PR #1, opened 2018-11-16
Slack: Import send.php and users.php from http://meta.svn.wordpress.o…

vedanshujain:
Slack: Import send.php and users.php from http://meta.svn.wordpress.o…rg/sites/trunk/common/includes/slack/

Copying these paths instead of setting these as externals as we ideally would want to keep externals to only plugins folder, this will make it easier later on if we want to move to git.
Also modifying wcorg-misc.php to use this file.

### DIFF
--- a/public_html/wp-content/mu-plugins/wcorg-misc.php
+++ b/public_html/wp-content/mu-plugins/wcorg-misc.php
@@ -459,7 +459,21 @@ function send_error_to_slack( $err_no, $err_msg, $file, $line ) {
 		return false;
 	}
 
-	if ( $err_no !== E_ERROR && $err_no !== E_USER_ERROR && $err_no !== E_CORE_ERROR && $err_no !== E_COMPILE_ERROR ) {
+	$error_whitelist = array(
+		E_ERROR,
+		E_USER_ERROR,
+		E_CORE_ERROR,
+		E_COMPILE_ERROR,
+		E_PARSE,
+		E_WARNING,
+		E_DEPRECATED,
+
+		// TODO: Enable notices once we get rid of all noisy E_WARNING.
+		// E_NOTICE,
+
+	);
+
+	if ( ! in_array( $err_no, $error_whitelist ) ) {
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


public_html/wp-content/mu-plugins/send.php
```diff
@@ -0,0 +1,133 @@
...
+		}
+		return 'Bot';
+	}
+
+	function get_text() {
+		return $this->text;
+	}
+
+	function set_text( $text ) {
+		echo 'hello from set text';
```
iandunn: It looks like this was added as a temporary debugging step after copying the file, so it should probably be removed now.

vedanshujain:
```suggestion
```
nice catch. will remove.

iandunn:
I made some changes in `r14577-dotorg` and c01a5d1d254ca6c1bba0c204a1fe58920656e044. It seemed easier to show how the `svn:external` could work rather than trying to explain it with words.

I don't think this'll introduce any problems for the eventual Git migration, since we can use the `Git and SVN side-by-side` approach. If we don't end up with that approach, then we'll need to figure something else out for the other externals anyway, so I don't think this'll introduce any new challenges. It feels a lot cleaner than having a duplicate copy that we need to keep in sync.

The rate limiting worked pretty well in my testing, so I think we're getting close here.

One big problem I noticed was that it's returning early for parse errors, instead of sending notifications. e.g.,

```php
<?php

this_should_not_be_here

namespace Foo;

// ...
```

I think we also want warnings and notices, since those can be helpful clues, and I tend to think of them as errors, even if they're not fatal errors.

That will produce some noise in the channel initially, but it's good to know that they're happening, and we can easily fix most of them. They're technically bugs, so it's better to fix them than ignore them.

iandunn:
Oh, you'll need to look at the SVN commit to see the external part:

https://meta.trac.wordpress.org/changeset/7855/sites/trunk/wordcamp.org/public_html

vedanshujain:
@iandunn If we also want warnings and notices, then can you please create another temporary public channel in dotorg slack with you, me and @coreymckrill in it.  Because other wise its at least 2 notifications every 2 minutes. We will follow that channel but also mute it. That way we can fix those notification urgently, but also not disrupt any of other work. We will archive this channel once notifications are under control.

Also I did not used externals here because I wanted to keep externals only in plugins folder for now, since we are not sure what approach we are going to take for git. But maybe I am overthinking this, so lets go ahead with externals for now. 

Please let me know after creating another channel. I will push the changes for `wcorg-misc.php`.

iandunn:
I'm not a Slack admin, so I can't create new private channels. I don't think we really need one, though. Since the current channel isn't being used, it won't really disrupt anything. If the people who were invited automatically are disturbed, they can always leave the channel, since it's probably not relevant to them anyway.

iandunn:
Looks like we still need to loosen the error handling to include warnings, notices, parse errors, etc. Other than that I think we're good to merge this.

PR #2, opened 2018-11-16
Add repo meta files

coreymckrill:
Add GitHub's recognized meta files

- [ ] Readme
- [ ] License
- [ ] Contributing
- [ ] Code of Conduct
- [ ] Issue template

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,1 @@
+Thanks for contributing to WordCamp.org! This page will explain all the ways you can contribute.
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,357 @@
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation; either version 2 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+
+===================================
+
+
+GNU GENERAL PUBLIC LICENSE
+   Version 2, June 1991
+
+Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+Everyone is permitted to copy and distribute verbatim copies
+of this license document, but changing it is not allowed.
+
+		Preamble
+
+The licenses for most software are designed to take away your
+freedom to share and change it.  By contrast, the GNU General Public
+License is intended to guarantee your freedom to share and change free
+software--to make sure the software is free for all its users.  This
+General Public License applies to most of the Free Software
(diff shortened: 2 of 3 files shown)


iandunn:
We'd need to add these to `meta.svn` in order for them to be synced, so I wonder if it'd be better to wait until we switch to Git natively?

Alternatively, we could add them to `meta.svn`, but make them generic enough to cover GitHub and Trac.

If we do that, it might be nice to use `.github/CONTRIBUTING.md` to keep things organized, since there are probably other GitHub templates we might want to add. e.g., https://github.com/WordPress/gutenberg/tree/480016dd1833728f9028a0235dc4ebe28393f303/.github


I see this repo more as a temporary things for early-stage collaboration on large new features like the blocks, rather than something for contributors, which should still use instead of Trac for the time being. Maybe we could setup some automation etc that would make it useful for all contributions, though? Let me know what y'all think.

vedanshujain:
> I see this repo more as a temporary things for early-stage collaboration on large new features like the blocks, rather than something for contributors, which should still use instead of Trac for the time being. 

While we still use Trac for time being, lets also move (slowly) towards making this repo our main hub for contributors as well. To that end, its probably a good idea to add git specific repo meta files in `meta.svn` as well (maybe mentioning that this is intended for our GitHub repo with link).

Also, specifically what automatons you have in mind that we can setup.

CONTRIBUTING.md
```diff
@@ -0,0 +1 @@
+Thanks for contributing to WordCamp.org! This page will explain all the ways you can contribute.
```
iandunn, line 1: `SHOULD` - We can maybe just link to the existing Handbook page, to keep things DRY.

LICENSE.txt
```diff
@@ -0,0 +1,357 @@
+This program is free software; you can redistribute it and/or modify
```
iandunn, line 1: `MIGHT` - IIRC, we're _technically_ supposed to have the full file, but it feels much cleaner to just make the `License` section of the `README` link to https://www.gnu.org/licenses/old-licenses/gpl-2.0.html? That feels much cleaner.

IIRC, GitHub also has something where you can mark the license, and it'll show up in GH's interface somewhere.

readme.md
```diff
@@ -0,0 +1,21 @@
+# WordCamp.org
+
+This is a mirror of the [WordCamp directory](https://meta.svn.wordpress.org/sites/trunk/wordcamp.org/) from the WordPress Meta SVN repository, though it does not include externals.
```
iandunn, line 3: `SHOULD` - maybe link to the Meta Environment provision for details on getting those? Or maybe #113 will include instructions for Docker instead.

readme.md
```diff
@@ -0,0 +1,21 @@
...
+
+WordCamp.org is a WordPress multisite network that supports WordCamp events around the world.
```
iandunn, line 5:
```suggestion
[WordCamp.org](https://central.wordcamp.org) is a WordPress multisite network that supports WordCamp events around the world.
```

readme.md
iandunn, line 3: `SHOULD` - We'll need to update this soon to reflect that it won't be a mirror.

CONTRIBUTING.md
iandunn, line 1: `SHOULD` - I think this should live in `.github/` to avoid cluttering the root, and since we'll probably want to add issue templates, etc.

iandunn (changes requested):
Some changes needed, and I think we'll need to revamp this in light of #113 too

iandunn:
d339420 covers a lot of what we'll want here, but doesn't add contributing templates, etc. We can open new PRs for anything else we still want to add.

PR #3, opened 2018-11-23
Speaker block

vedanshujain:
Add support for speaker group in speakers shortcode. This is needed for upcoming speaker block.


coreymckrill:
Per another discussion elsewhere, we're going to build a new rendering function for the Speakers block, rather than modifying the shortcode callback.

PR #9, opened 2019-01-01
Discuss blocks review

iandunn:
This PR shouldn't be merged, it just seemed like the easiest way to use GitHub's review tools and have a discussion on individual lines of code.

I read through all the code and played around with things, and overall I love it. Great work @coreymckrill and Vedanshu!

I've added comments with a few suggestions, and lots of questions I don't have the answers to :)

I'm happy to push commits for those changes, but want to discuss them first so that:

1. We all agree on what should happen
1. We all learn from the process

@vedanshujain, if you haven't already, it'd be good for you to read through all the code that you didn't write, and add your own comments to this branch as well.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar-size/index.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar-size/index.js
@@ -51,6 +51,7 @@ function AvatarSizeControl( {
 	onChange,
 	initialPosition,
 	...props
+	// i may just be misunderstanding the spread operator, but it seems like a bad idea to have arbitrary inputs mapped to fixed function params. wouldn't we want the params to be explicitly and unchanging?
 } ) {
 	const id = `wordcamp-inspector-avatar-size-control-${ instanceId }`;
 
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar/index.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar/index.js
@@ -14,6 +14,13 @@ function AvatarImage( {
 	name,
 	size,
 	url,
+	// this feels like a weird way to declare default arguments, mostly from a readability standpoint. i wonder if there's a better way, but i'm not really sure.
+	// i don't think i've seen it done this way in gutenberg, but i could be wrong
+	// maybe this should extend Component so it could simply accept a `props` object and then have a `defaultProps` property?
+		// or is this some es6 way of picking specific things out of the props that are already being passed in?
+	// or maybe have a statement where it merges the passed props with the defined defaults or something?
+	// or maybe just put it all on 1 line?
+	// if we change it here, we should change all the other instances as well
 } ) {
 	const getSizedURL = ( avatar_url, avatar_size ) => {
 		return addQueryArgs( avatar_url, { s: avatar_size } );
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/shared/image-alignment/index.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/shared/image-alignment/index.js
@@ -38,6 +38,7 @@ function ImageAlignmentControl( {
 				{ options.align.map( ( alignment ) => {
 					const optLabel = alignment.label;
 					const optValue = alignment.value;
+					// should we use `const { label, value } = alignment;` here for consistency?
 					const isCurrent = value === optValue;
 					const iconSlug = `align-${ optValue }`;
 
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/shared/versatile-select/index.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/shared/versatile-select/index.js
@@ -19,6 +19,10 @@ import './style.scss';
 class VersatileSelect extends Component {
 	constructor( props ) {
 		super( props );
+		// do we need to pass props here, or is just calling super() enough? i think it should be if all we want is to access `this`
+		// passing `props` is only needed when we want to reference `this.props` inside the constructor
+		// https://stackoverflow.com/questions/30571875/whats-the-difference-between-super-and-superprops-in-react-when-using-e
+		// if we change it here, we should change the other instances as well
 
 		this.state = {
 			selectedOptions: null,
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
@@ -19,10 +19,24 @@ import AvatarImage from '../shared/avatar';
 import './block-content.scss';
 
 class SpeakersBlockContent extends Component {
+	// it's not immediately obvious why `add` is here, so jsdoc would really help explain that. we might not want to keep it, though (see below)
 	static maybeAddMoreLink( content, add ) {
 		if ( add ) {
+			/*
+			 this feels a bit odd. is it here because JSX can only contain expressions, so we cant do something like:
+
+			if ( speaker_link ) { addMoreLink() )
+
+			what if instead we used a ternary?
+
+			{ ( speaker_link ? this.constructor.addMoreLink( post.excerpt.rendered ) : post.excerpt.rendered ).trim() }
+
+			then this function would be more straightforward, and the calling function read more naturally and be self-explanatory
+			*/
+
 			const more = sprintf(
 				'<a href="#" class="wordcamp-speaker-more-link">%s</a>',
+				// should the href have the actual value? blocks like Recent Posts have real links, but make them open in a new window if clicked on inside gutenberg
 				__( 'Read more', 'wordcamporg' )
 			);
 
@@ -30,12 +44,14 @@ class SpeakersBlockContent extends Component {
 
 			if ( Array.isArray( content.match( pattern ) ) ) {
 				content = content.replace( pattern, ' ' + more + '</p>' );
+				// could post.content.rendered contain divs or other things we need to worry about here?
 			} else {
 				content += ' ' + more;
 			}
 		}
 
 		return content;
+		// if we return content.trim() then we wouldn't have to worry about using it in the caller statements
 	}
 
 	render() {
@@ -57,11 +73,15 @@ class SpeakersBlockContent extends Component {
 		}
 
 		return (
+			// this feels a bit overwhelming. i think this would be easier to grok and maintain if it were modularized more
+			// maybe pull out the `wordcamp-speaker-content` and session parts into functions that just get called from here?
+
 			<ul className={ classnames( containerClasses ) }>
 				{ speakerPosts.map( ( post, i ) =>
 					<li
 						key={ i }
 						className={ classnames( 'wordcamp-speaker', 'wordcamp-speaker-' + post.slug ) }
+						// should we use decodeEntities here too?
 					>
 						<h3 className={ 'wordcamp-speaker-name-heading' }>
 							{ decodeEntities( post.title.rendered.trim() ) || __( '(Untitled)', 'wordcamporg' ) }
@@ -80,8 +100,13 @@ class SpeakersBlockContent extends Component {
 							<div className={ 'wordcamp-speaker-content' }>
 								{ 'full' === content &&
 									<Disabled>
+										{ /* why is Disabled used here? */ }
 										<RawHTML>
 											{ this.constructor.maybeAddMoreLink( post.content.rendered.trim(), speaker_link ).trim() }
+											{ // why do we need a read more link if we're showing the full bio?
+												// is it just to link to the full session post? if so, maybe the anchor text should say something like "view speaker page" instead?
+												// or just let their name be a link?
+											}
 										</RawHTML>
 									</Disabled>
 								}
@@ -89,6 +114,12 @@ class SpeakersBlockContent extends Component {
 									<Disabled>
 										<RawHTML>
 											{ this.constructor.maybeAddMoreLink( post.excerpt.rendered.trim(), speaker_link ).trim() }
+											{ /* i think calling it like `SpeakersBlockContent.maybeAddMoreLink()` seems a bit nicer than `this.constructor. obviously not a big deal either way though.  */ }
+											{ /*
+											 excerpts shouldn't have HTML, should they? it seems like we can remove the RawHTML wrapper.
+											or maybe it's only needed around the speaker link, but not the excerpt?
+											maybe `maybeAddMoreLink` should return an Element instead of a string, so that we wouldn't need rawhtml? i'm worried about using rawhtml when we don't absolutely need to, since it can be an attack vector
+											*/ }
 										</RawHTML>
 									</Disabled>
 								}
@@ -96,6 +127,7 @@ class SpeakersBlockContent extends Component {
 									<Disabled>
 										<RawHTML>
 											{ this.constructor.maybeAddMoreLink( '', speaker_link ).trim() }
+											{ /* similar question to above about avoiding rawhtml */ }
 										</RawHTML>
 									</Disabled>
 								}
@@ -118,18 +150,29 @@ class SpeakersBlockContent extends Component {
 												<a
 													className={ 'wordcamp-speaker-session-link' }
 													href={ session.link }
+													// is there an esc_url() equivalent in g? or do we not need it because we're returning an Element, which gets escaped automatically?
 												>
 													{ decodeEntities( session.title.rendered.trim() ) || __( '(Untitled)', 'wordcamporg' ) }
 												</a>
 												<br />
 												<span className={ 'wordcamp-speaker-session-info' }>
 													{ session.session_track.length &&
+														// will this fatal similar to https://meta.trac.wordpress.org/changeset/8025/ ?
+														// if so, it'd be good to check for other instances of that problem throughout the plugin
+
 														sprintf(
 															/* translators: 1: A date; 2: A time; 3: A location; */
 															__( '%1$s at %2$s in %3$s', 'wordcamporg' ),
 															session.session_date_time.date,
 															session.session_date_time.time,
 															get( find( tracks, ( value ) => { return parseInt( value.id ) === head( session.session_track ) } ), 'name' )
+															// this is pretty difficult for a human to parse when it's all written as a single line, can it be broken up into more digestable chunks?
+															// i don't think it needs to be a named function, but just spreading it across multiple lines would increase the readability a lot
+
+															// why use `head` instead of `session.session_track[0]` ?
+															// it's one extra thing to learn, but the benefit isn't obvious. is it useful because it does some extra error checking or something?
+															// if so, could that be avoided by an extra condition in the `{ session.session_track.length &&` bit above? maybe that's overkill and head() is easier, though.
+															// seems like it's unnecessary in es6 though? https://www.sitepoint.com/lodash-features-replace-es6/
 														)
 													}
 													{ ! session.session_track.length &&
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.scss
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.scss
@@ -19,6 +19,7 @@
 	}
 
 	.wordcamp-speaker:after {
+		/* why not use the existing wp-clearfix class? */
 		content: "";
 		display: table;
 		clear: both;
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-controls.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-controls.js
@@ -26,6 +26,34 @@ class SpeakersBlockControls extends Component {
 
 		const hasPosts = Array.isArray( speakerPosts ) && speakerPosts.length;
 
+		/*
+		 what do you think about modularizing all these different returns into named functions?
+		 it seems like it'd be easier to scan the structure and also focus on individual parts that way
+		 that could also simplify the structure into a single if..then..else block, rather than an if, a switch, and an implicit else
+
+		```
+		let content;
+
+		if ( mode && ! hasPosts ) {
+			content = noPostsPlaceholder();
+		} elseif ( 'all' === mode ) {
+			content = ( <SpeakersBlockContent { ...this.props } /> );
+		} elseif ( 'specific_posts' === mode ) {
+			content = specificPostsView();
+		} elseif ( 'specific_terms' === mode ) {
+			content = specificTermsView();
+		} else {
+			content = newBlockPlaceholder();
+		}
+
+		return content;
+		```
+
+		that'd also resolve the CS violation about returning from switch statements
+		https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#switch-statements
+
+		 */
+
 		if ( mode && ! hasPosts ) {
 			return (
 				<Placeholder
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/edit.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/edit.js
@@ -24,6 +24,14 @@ class SpeakersEdit extends Component {
 		const { mode } = this.props.attributes;
 
 		return (
+			/*
+			the naming of speakerblockcontrols isn't clear to me. it seems like it's wrapper for block content in its various states
+		    (nothing selected, no posts available, all speakers, specific speakers, specific terms)
+		    so why is it called a "control", when it doesn't seem to have any kind of interactive switches?
+		    maybe my understanding of "control" is off?
+			*/
+
+
 			<Fragment>
 				<SpeakersBlockControls { ...this.props } />
 				{ mode &&
@@ -59,6 +67,8 @@ const speakersSelect = ( select, props ) => {
 	}
 
 	const speakersQuery = pickBy( args, ( value ) => ! isUndefined( value ) );
+	// it sounds like `pick` isn't needed in es6, does that also apply to pickBy?
+	// https://www.sitepoint.com/lodash-features-replace-es6/
 
 	return {
 		speakerPosts : getEntityRecords( 'postType', 'wcb_speaker', speakersQuery ),
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/index.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/index.js
@@ -18,5 +18,9 @@ export const settings = {
 	edit,
 	save        : function() {
 		return null;
+		// should we return a cached copy of the markup here, as a failsafe in case we ever want to disable this block or make some big changes to it?
+		// disabling it entirely seems unlikely, but we might eventually want to build a v2 from scratch or something
+		// not having to worry about back-compat might give us some flexibility in the future
+		// https://wordpress.org/gutenberg/handbook/designers-developers/developers/block-api/block-edit-save/#save
 	},
 };
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
@@ -18,6 +18,10 @@ const { __ } = wp.i18n;
 import AvatarImage from '../shared/avatar';
 import VersatileSelect from '../shared/versatile-select';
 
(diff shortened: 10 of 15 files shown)


public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar-size/index.js
```diff
@@ -51,6 +51,7 @@ function AvatarSizeControl( {
 	onChange,
 	initialPosition,
 	...props
+	// i may just be misunderstanding the spread operator, but it seems like a bad idea to have arbitrary inputs mapped to fixed function params. wouldn't we want the params to be explicitly and unchanging?
```
coreymckrill: I think I was sort of using `...props` for anything that is optional, while parameters that are explicitly named in the signature are "required".

I'll have to think about this some more, but my initial thought is that there's no harm in including the spread of additional inputs because any values that aren't explicitly used are ignored anyway.

One potential advantage of using `...props` that I can think of is that in a case like this one where a component contains other components, which have their own signatures, is that if any of those components change or are replaced, we don't necessarily have to update the parent component's signature.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar/index.js
```diff
@@ -14,6 +14,13 @@ function AvatarImage( {
 	name,
 	size,
 	url,
+	// this feels like a weird way to declare default arguments, mostly from a readability standpoint. i wonder if there's a better way, but i'm not really sure.
+	// i don't think i've seen it done this way in gutenberg, but i could be wrong
+	// maybe this should extend Component so it could simply accept a `props` object and then have a `defaultProps` property?
+		// or is this some es6 way of picking specific things out of the props that are already being passed in?
+	// or maybe have a statement where it merges the passed props with the defined defaults or something?
+	// or maybe just put it all on 1 line?
+	// if we change it here, we should change all the other instances as well
```
coreymckrill: I got this pattern from some other component in Gutenberg, but I'm not sure which one. Unfortunately, as I discovered over the course of building this, Gutenberg does not have a consistent way of defining components or blocks, probably due to both varying contributor styles and the fact that the codebase was developed over more than a year, during which a lot of best practices in React and es6 probably changed.

I agree that we should standardize, though, and I think extending `Component` is a good way to do that.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/image-alignment/index.js
```diff
@@ -38,6 +38,7 @@ function ImageAlignmentControl( {
 				{ options.align.map( ( alignment ) => {
 					const optLabel = alignment.label;
 					const optValue = alignment.value;
+					// should we use `const { label, value } = alignment;` here for consistency?
```
coreymckrill: I didn't do that in this case because `label` and `value` are already used in the component signature.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/versatile-select/index.js
```diff
@@ -19,6 +19,10 @@ import './style.scss';
 class VersatileSelect extends Component {
 	constructor( props ) {
 		super( props );
+		// do we need to pass props here, or is just calling super() enough? i think it should be if all we want is to access `this`
+		// passing `props` is only needed when we want to reference `this.props` inside the constructor
+		// https://stackoverflow.com/questions/30571875/whats-the-difference-between-super-and-superprops-in-react-when-using-e
+		// if we change it here, we should change the other instances as well
```
coreymckrill: > do we need to pass props here, or is just calling super() enough?

I don't know. It's worth testing.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -19,23 +19,39 @@ import AvatarImage from '../shared/avatar';
...
 			if ( Array.isArray( content.match( pattern ) ) ) {
 				content = content.replace( pattern, ' ' + more + '</p>' );
+				// could post.content.rendered contain divs or other things we need to worry about here?
 			} else {
 				content += ' ' + more;
 			}
 		}
 
 		return content;
+		// if we return content.trim() then we wouldn't have to worry about using it in the caller statements
```
coreymckrill: All of the above are good points, and I agree that the whole `maybeAddMoreLink` method feels a bit hacky (along with the PHP version). The problem is that I couldn't find a good way to consistently get the Read More link inside the rendered content, on the same line as the trimmed excerpt (instead of in a new, separate paragraph). For the record, I have strong negative feelings about the way excerpts are implemented in WP ;P

I would love to get some collaboration on this to figure out how to do this in a way in which we don't need this method at all :D  At the very least, I agree with your suggestions, and we should implement them.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -57,11 +73,15 @@ class SpeakersBlockContent extends Component {
 		}
 
 		return (
+			// this feels a bit overwhelming. i think this would be easier to grok and maintain if it were modularized more
+			// maybe pull out the `wordcamp-speaker-content` and session parts into functions that just get called from here?
+
```
coreymckrill: > maybe pull out the `wordcamp-speaker-content` and session parts into functions that just get called from here?

Sure, works for me.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -57,11 +73,15 @@ class SpeakersBlockContent extends Component {
...
 			<ul className={ classnames( containerClasses ) }>
 				{ speakerPosts.map( ( post, i ) =>
 					<li
 						key={ i }
 						className={ classnames( 'wordcamp-speaker', 'wordcamp-speaker-' + post.slug ) }
+						// should we use decodeEntities here too?
```
coreymckrill: Maybe? Unless the `classnames` function takes care of that. I'll have to doublecheck the source...

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -80,22 +100,34 @@ class SpeakersBlockContent extends Component {
 							<div className={ 'wordcamp-speaker-content' }>
 								{ 'full' === content &&
 									<Disabled>
+										{ /* why is Disabled used here? */ }
```
coreymckrill: So that links in the content don't take the user off the page if they get clicked.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -118,18 +150,29 @@ class SpeakersBlockContent extends Component {
 												<a
 													className={ 'wordcamp-speaker-session-link' }
 													href={ session.link }
+													// is there an esc_url() equivalent in g? or do we not need it because we're returning an Element, which gets escaped automatically?
```
coreymckrill: I'm not aware of a G equivalent of `esc_url`. I'm also not sure if the escaping gets handled by something else or not. I know I saw a similar implementation without escaping in some component in Gutenberg, but that doesn't necessarily mean it's safe ;P

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -118,18 +150,29 @@ class SpeakersBlockContent extends Component {
...
 												>
 													{ decodeEntities( session.title.rendered.trim() ) || __( '(Untitled)', 'wordcamporg' ) }
 												</a>
 												<br />
 												<span className={ 'wordcamp-speaker-session-info' }>
 													{ session.session_track.length &&
+														// will this fatal similar to https://meta.trac.wordpress.org/changeset/8025/ ?
+														// if so, it'd be good to check for other instances of that problem throughout the plugin
+
```
coreymckrill: Good question. I'll have to test, as I'm not sure if the REST API strips out the key if the value is empty.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -118,18 +150,29 @@ class SpeakersBlockContent extends Component {
...
 															session.session_date_time.date,
 															session.session_date_time.time,
 															get( find( tracks, ( value ) => { return parseInt( value.id ) === head( session.session_track ) } ), 'name' )
+															// this is pretty difficult for a human to parse when it's all written as a single line, can it be broken up into more digestable chunks?
+															// i don't think it needs to be a named function, but just spreading it across multiple lines would increase the readability a lot
+
+															// why use `head` instead of `session.session_track[0]` ?
+															// it's one extra thing to learn, but the benefit isn't obvious. is it useful because it does some extra error checking or something?
+															// if so, could that be avoided by an extra condition in the `{ session.session_track.length &&` bit above? maybe that's overkill and head() is easier, though.
+															// seems like it's unnecessary in es6 though? https://www.sitepoint.com/lodash-features-replace-es6/
```
coreymckrill: > why use `head` instead of `session.session_track[0]` ?

I think I did this because `session.session_track[0]` doesn't always exist. `head` always grabs the first item, regardless of the index.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -118,18 +150,29 @@ class SpeakersBlockContent extends Component {
...
+
 														sprintf(
 															/* translators: 1: A date; 2: A time; 3: A location; */
 															__( '%1$s at %2$s in %3$s', 'wordcamporg' ),
 															session.session_date_time.date,
 															session.session_date_time.time,
 															get( find( tracks, ( value ) => { return parseInt( value.id ) === head( session.session_track ) } ), 'name' )
+															// this is pretty difficult for a human to parse when it's all written as a single line, can it be broken up into more digestable chunks?
+															// i don't think it needs to be a named function, but just spreading it across multiple lines would increase the readability a lot
+
```
coreymckrill: > can it be broken up into more digestable chunks?

Worth trying. I'm not sure why I left it in one line, because I usually like lots of whitespace :D

It might have been an esLint issue.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.scss
```diff
@@ -19,6 +19,7 @@
 	}
 
 	.wordcamp-speaker:after {
+		/* why not use the existing wp-clearfix class? */
```
coreymckrill: `wp-clearfix` isn't available on the front end. Currently we're loading this stylesheet on both the WP Admin and the front end so that the layout options work correctly regardless of theme.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-controls.js
```diff
@@ -26,6 +26,34 @@ class SpeakersBlockControls extends Component {
...
+		}
+
+		return content;
+		```
+
+		that'd also resolve the CS violation about returning from switch statements
+		https://make.wordpress.org/core/handbook/best-practices/coding-standards/javascript/#switch-statements
+
+		 */
+
```
coreymckrill: Sure, works for me. Good catch with the coding standards, I wasn't aware of that one.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/edit.js
```diff
@@ -24,6 +24,14 @@ class SpeakersEdit extends Component {
 		const { mode } = this.props.attributes;
 
 		return (
+			/*
+			the naming of speakerblockcontrols isn't clear to me. it seems like it's wrapper for block content in its various states
+		    (nothing selected, no posts available, all speakers, specific speakers, specific terms)
+		    so why is it called a "control", when it doesn't seem to have any kind of interactive switches?
+		    maybe my understanding of "control" is off?
+			*/
+
```
coreymckrill: `SpeakersBlockControls` does contain interactive controls. When no "mode" is selected, it displays the buttons to choose whether to show all speakers or specific ones. When specific speakers or terms are selected, it shows the picker beneath the content that allows the selection to be modified.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/edit.js
```diff
@@ -59,6 +67,8 @@ const speakersSelect = ( select, props ) => {
 	}
 
 	const speakersQuery = pickBy( args, ( value ) => ! isUndefined( value ) );
+	// it sounds like `pick` isn't needed in es6, does that also apply to pickBy?
+	// https://www.sitepoint.com/lodash-features-replace-es6/
```
coreymckrill: Good question. `pick` and `pickBy` don't do exactly the same thing, but it certainly would be worth exploring whether we could eliminate this and maybe other lodash dependencies.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/index.js
```diff
@@ -18,5 +18,9 @@ export const settings = {
 	edit,
 	save        : function() {
 		return null;
+		// should we return a cached copy of the markup here, as a failsafe in case we ever want to disable this block or make some big changes to it?
+		// disabling it entirely seems unlikely, but we might eventually want to build a v2 from scratch or something
+		// not having to worry about back-compat might give us some flexibility in the future
+		// https://wordpress.org/gutenberg/handbook/designers-developers/developers/block-api/block-edit-save/#save
```
coreymckrill: Maybe? I haven't tested this, but my assumption is that the markup would get saved to `post_content` and then when the post is loaded on the front end, we would end up with the cached version, followed by a dynamically generated version. But I could very well be wrong about that.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
```diff
@@ -18,6 +18,10 @@ const { __ } = wp.i18n;
 import AvatarImage from '../shared/avatar';
 import VersatileSelect from '../shared/versatile-select';
 
+// lots of stuff here, can/should some of it be black-boxed into VersitileSelect, so it's DRY when we start using this component in other blocks?
+
+// this is pretty big, could it be broken reasonably down into smaller components?
+
```
coreymckrill: It is kind of big, but as far as I can tell, all of it is fairly specific to the Speakers block. Happy to take suggestions for stuff that could be black-boxed, though.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
```diff
@@ -39,6 +43,8 @@ class SpeakersSelect extends Component {
 						name={ optionData.label }
 						size={ 24 }
 						url={ optionData.avatar }
+
+						// formatting is messed up when no gravatar exists for user
```
coreymckrill: This might be a problem with our custom AvatarImage component not handling default avatars, rather than something with this select component. I'll need to dig further to figure this out.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
```diff
@@ -125,6 +132,7 @@ class SpeakersSelect extends Component {
 
 		if ( ids ) {
 			currentValue = filter( selectOptions[ 0 ].options, ( o ) => {
+				// what is `o`? an option? the name should be descriptive and obvious
```
coreymckrill: Good point.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
```diff
@@ -61,6 +67,7 @@ class SpeakersSelect extends Component {
 
 	static optionLabel( optionData ) {
 		const { type } = optionData;
+		// would it be better to include label and count in ^, and then name the return value something like `labelMarkup` ?
```
coreymckrill: We could. I think I split out the image and the label into separate methods to keep things shorter, but it might make more sense for them to be combined...

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
```diff
@@ -151,9 +159,12 @@ class SpeakersSelect extends Component {
 					);
 				} }
 				onChange={ ( selectedOptions ) => {
+					// this seems like it's big enough that it'd be more readable if it were modularized into a named function
+					// even the smaller ones above feel like they could be too. if it's not a 1-liner then it makes the element harder to scan quickly, and mixes logic with markup
```
coreymckrill: > if it's not a 1-liner then it makes the element harder to scan quickly, and mixes logic with markup

Good point about mixing logic with markup, although that's the whole problem with React in general, right? :D

I'm not sure about scanability, though. If you have to scroll back and forth to different parts of the file to figure out what the named callback does, is it really more readable?

(I'm just being pedantic. I think I have done it both ways, even within the codebase of this block. We should probably pick one pattern and stick with it)

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
```diff
@@ -151,9 +159,12 @@ class SpeakersSelect extends Component {
...
 					const value = map( selectedOptions, 'value' );
 
 					if ( ! value.length ) {
+						// i think this would flow better if it were `if ( value.length ) { stuff } else { other stuff }`. `if ( ! true } {} else {}` is a bit awkward to read
```
coreymckrill: Fair enough.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
```diff
@@ -192,6 +203,8 @@ const optionsSelect = ( select, props ) => {
 	let options = [];
 
 	if ( ! mode || 'specific_terms' === mode ) {
+		// should there be some kind of UI to let the user know they can only select groups at the start, and can't mix speakers/groups?
+
```
coreymckrill: To clarify, you _can_ switch from speakers to groups, or vice versa, if you remove all of the selected items first. Good question, though. I think this another thing that would be good to discuss with the designer.

public_html/wp-content/mu-plugins/blocks/includes/speakers.php
```diff
@@ -64,6 +64,9 @@ function add_script_data( array $data ) {
 			'mode'    => get_options( 'mode' ),
 			'sort'    => get_options( 'sort' ),
 		),
+		// would ^ be easier if the get_options() had a `default` case that just returned all of them?
+		// that way this would just be `options => get_options()` instead of listing each one manually
+		// get_options would need to be refactored a bit, so maybe not worth it
```
coreymckrill: Good point. I don't think the refactor would be very complicated.

public_html/wp-content/mu-plugins/blocks/webpack.config.js
```diff
@@ -24,6 +24,10 @@ const webpackConfig = {
 	output: {
 		filename: '[name].min.js',
 		path: path.resolve( __dirname, 'assets' ),
+		// should we leave the js un-minified in dev mode, so that it's easier to debug?
+
+		// the browser debugger isn't usable b/c it can only see the minified file, i'm guessing we need source maps enabled
+		// er, well, it seems like it works sometimes, and doesn't others
```
coreymckrill: Yeah, I didn't take the time to figure out the source mapping stuff, but I definitely think it would help for dev mode.

public_html/wp-content/mu-plugins/blocks/blocks.php
```diff
@@ -103,6 +109,8 @@ function register_assets() {
  * Fix a CSS bug in Gutenberg when PanelRows are used with RangeControls.
  *
  * @todo Remove this when https://github.com/WordPress/gutenberg/pull/4564 is fixed.
+ *       it doesn't sound like they're going to accept any PRs for this, so we can probably remove ^ comment -- https://github.com/WordPress/gutenberg/issues/4402
+ *       it might also be good to refactor this to a more permenant solution, with a link to the ^ issue for documentation
```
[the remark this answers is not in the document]
iandunn: That's the impression I got from the comments on the issue/PR, but it does seem odd to me, since lots of plugins will have to add the same styles to avoid the problem. If you think they might accept a new PR then that sounds good to me.

public_html/wp-content/mu-plugins/blocks/blocks.php
```diff
@@ -14,13 +14,17 @@
...
 		return;
 	}
 
 	require_once PLUGIN_DIR . 'includes/speakers.php';
 }
 
 add_action( 'plugins_loaded', __NAMESPACE__ . '\load' );
+// what do you think about moving all of the add_action|filter to the top of the file?
+// i personally like that because it serves as a "table of contents" for what's in the file
+// and then the rest of the file just has function declerations, rather than a mix of declerations and executing code
```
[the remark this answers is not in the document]
iandunn: Yeah, I think you're right that duplicating the info in a comment would lead to it getting out of sync. Sounds like it's just personal preference then, so i'm fine leaving it like this.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/speakers-select.js
iandunn: haha, that's true :)

it's seems like there are cases where it helps to keep clear boundaries, though, and to me this feels like one of them. i don't feel strongly, though.

having to scroll to find out what the callback does is a valid point. hopefully that would be somewhat mitigated by a descriptive function name, and `cmd+b` and `cmd+alt+left` in PhpStorm, but it would still add that extra work in some cases.

it might just come down to personal preference for where the lines get drawn, and if that's the case i'm fine leaving it as a case-by-case judgement call.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.scss
iandunn: Ah, good point. I wonder if we should have our own `wordcamp-clearfix`, or maybe a duplicate `wp-clearfix`?

At the very least, it seems like it'd be good to have a rule with `clearfix` in the name, so its purpose would be obvious, and we could reuse it across blocks.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
iandunn: Ah, good point about `[0]` being empty but `[n]` not being. IIRC, the ES6 replacement for `head()` solves that, so it might be the best option regardless.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
iandunn: Ah, that'd make sense. Feel free to modify `.eslintrc.js` as you see fit, it's definitely a rough draft. 

My take is that if a rule will create false positives more than ~5% of the time, it's worth turning it off and just relying on our judgement to stick to the standard. Too many false positives and we won't be able to trust the tool, and it'll become more of a nuisance than a helper.

If we ever run into a situation where one of us commits a rule change that others disagree with, we can start a discussion and come to a consensus. I don't think that'll happen too often, though, so better to just go ahead and commit what you think is best, rather than getting bogged down in discussion first.

cc @vedanshujain, same goes for you for modifying the rules.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.scss
coreymckrill: Good call, I like the idea of adding a rule for `wordcamp-clearfix`.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
iandunn: Ah, ok. I read the `README` for the component but you explained it much better than they did :)

public_html/wp-content/mu-plugins/blocks/assets/src/shared/versatile-select/index.js
iandunn: It worked in my limited testing, but I wanted to check if there was a reason that I was missing. It sounds like it'll be ok to remove it though.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/image-alignment/index.js
iandunn: Ah, ok, that makes sense.

I wonder if we need the extra variable, though? Would it be slightly more straight-forward to  just `alignment.label`?

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
coreymckrill: Nope, doesn't look like `classnames` does any sanitization. So yes, we should probably use `decodeEntities` here.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar/index.js
coreymckrill: @iandunn I dug into this a bit more, and I think I have a better explanation now. I guess I'm not entirely sure which parts of this feel "weird" to you, but here's what I've got:

* This is actually a fairly common way to "destructure" function parameters in es6. They aren't really defaults in this case, because we're not assigning default values to them, just declaring them. [Here's one post that discusses this approach.](https://simonsmith.io/destructuring-objects-as-function-parameters-in-es6/)
* Declaring the component as a "pure function" instead of a class extending `Component` is considered by some to be a best practice in the case that the component doesn't have any modifiable state. If the component did have state and/or need to use lifecycle methods, then it would make sense to extend the `Component` class. Some discussion [here](https://stackoverflow.com/questions/36097965/react-when-to-use-es6-class-based-components-vs-functional-es6-components) and [here](https://code.tutsplus.com/tutorials/stateful-vs-stateless-functional-components-in-react--cms-29541).

Given these points, most of our components could probably actually be functions instead of classes...

coreymckrill: There might be some minor performance efficiencies when using stateless function components, but I agree, it's probably not worth refactoring existing stuff. Gutenberg's coding guidelines actually mention transitioning function components to classes when state needs to be introduced, so maybe we could take a similar approach:

https://github.com/WordPress/gutenberg/blob/master/docs/contributors/coding-guidelines.md#class-definition

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -19,23 +19,39 @@ import AvatarImage from '../shared/avatar';
 import './block-content.scss';
 
 class SpeakersBlockContent extends Component {
+	// it's not immediately obvious why `add` is here, so jsdoc would really help explain that. we might not want to keep it, though (see below)
 	static maybeAddMoreLink( content, add ) {
```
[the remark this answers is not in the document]
coreymckrill: My IDE told me it should be static ;P  I think because it's a pure function? I don't remember if it was an esLint thing or some other inspection.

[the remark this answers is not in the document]
iandunn: It seems reasonable to me for it to be static, since it doesn't need to be aware of the class' state.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/index.js
iandunn: I think that's only true for static blocks; for dynamic blocks, it only saves a comment in `post_content`

```html
<!-- wp:heading -->
<h2>"Heading" is a static block</h2>
<!-- /wp:heading -->

<!-- wp:wordcamp/speakers {"mode":"all","sort":"date_desc","layout":"grid","grid_columns":4,"content":"none","speaker_link":true} /-->
```

public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar/index.js
iandunn: Ah, thanks for those links. I remove my objections :)

For cases like this where the list of properties is short, and we're not assigning values, I think it still might be nice to put it all on one line, though. If either of those is false, though, then I agree it's more readable to use separate lines. I'm also fine using separate lines if you think the consistency is better. Not a big deal either way.

I also like the idea of using pure functions in cases where we don't need state/lifecycle.

PR #14, opened 2019-01-11
Add a trigger for Organizer Reminders that allows sending notifications to global sponsors when meetups are added to the chapter

coreymckrill:


iandunn:

Sorry for the hassle :(

@coreymckrill  , can you re-submit this?

PR #18, opened 2019-01-17
Add an ignore list mechanism for errors we don't need to send to Slack

coreymckrill:
Two things to note here:

* This uses the `ABSPATH` constant to obscure the full file paths that come with the error messages, since this ignore list is in a public repo. This might be overkill though. Or we could just put the ignore list in a non-public file. Thoughts?
* This looks for substring matches in the error messages instead of exact matches. This way we can still catch and ignore error messages that contain unique parts, like `wp-content/temp-write-test-1543906354`

### DIFF
--- a/public_html/wp-content/mu-plugins/wcorg-misc.php
+++ b/public_html/wp-content/mu-plugins/wcorg-misc.php
@@ -525,7 +525,7 @@ function send_error_to_slack( $err_no, $err_msg, $file, $line ) {
 
 	$domain    = get_site_url();
 	$page_slug = esc_html( trim( $_SERVER['REQUEST_URI'], '/' ) );
-	$text      = $text . "Message : \"$err_msg\" occured on \"$file:$line\" \n Domain: $domain \n Page: $page_slug \n Error type: $err_no";
+	$text      = $text . "Message : \"$err_msg\" occurred on \"$file:$line\" \n Domain: $domain \n Page: $page_slug \n Error type: $err_no";
 
 	$message = array(
 		'fallback'    => $text,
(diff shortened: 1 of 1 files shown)


public_html/wp-content/mu-plugins/wcorg-misc.php
```diff
@@ -475,9 +475,19 @@ function send_error_to_slack( $err_no, $err_msg, $file, $line ) {
...
 		E_DEPRECATED,
 		E_WARNING,
-	);
+	];
+
+	if ( ! in_array( $err_no, $error_safelist ) ) {
+		return false;
+	}
+
+	// Always use the ABSPATH constant in the keys here to avoid path disclosure.
```
vedanshujain: 👍

vedanshujain (approved):
looks good to me

vedanshujain:
one request, not a blocker for this PR though, since you are working on this. Can you also add a feature to send error notifications to personal slack if triggered through sandbox? You suggested in one of our recent chats and would be really cool to have.

iandunn (approved):
LGTM :)

coreymckrill:
> Can you also add a feature to send error notifications to personal slack if triggered through sandbox?

This'll have to wait for a different PR. Sending to a username instead of a channel exists in WCPT, but not in this **wcorg-misc.php** file. It might be worth adding a file to **mu-plugins/utilities** for Slack notifications that can be used across multiple plugins...

coreymckrill:
Committed to SVN: https://meta.trac.wordpress.org/changeset/8093

PR #19, opened 2019-01-18
Automate request approval

iandunn:
if the request is within the approved budget, there's no need for manual approval

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments-network/includes/wordcamp-budgets-dashboard.php
+++ b/public_html/wp-content/plugins/wordcamp-payments-network/includes/wordcamp-budgets-dashboard.php
@@ -15,6 +15,8 @@ define( 'REDACT_PAID_REQUESTS_CRON_ID', 'wcb_redact_paid_requests' );
 
 /*
  * Core functionality and helper functions shared between modules
+ *
+ * @todo the import/export stuff should be moved to its own file for better organization.
  */
 
 add_action( 'network_admin_menu', __NAMESPACE__ . '\register_budgets_menu' );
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wordcamp-payments-network/includes/wordcamp-budgets-dashboard.php
```diff
@@ -655,9 +663,85 @@ function process_action_approve() {
 	\WordCamp_Budgets::log( $post->ID, get_current_user_id(), 'Request approved via Network Admin', array(
 		'action' => 'approved',
 	) );
+	// edit this to say 'manually approved by {name}' or 'automatically approved' ?
```
coreymckrill: I think making this more specific is a good idea.

public_html/wp-content/plugins/wordcamp-payments-network/includes/wordcamp-budgets-dashboard.php
```diff
@@ -627,6 +631,10 @@ function convert_currency( $from, $to, $amount ) {
  * Approve a payment or reimbursement request.
  */
 function process_action_approve() {
+	// call this when user submits, will need way to bypass nonce, maybe just $force or $auto param
+		// or maybe modularize so can call model parts and this remains as controller. yeah, that's probably better
+		// that would be a separate commit
+
```
coreymckrill: > maybe modularize so can call model parts and this remains as controller

+1

public_html/wp-content/plugins/wordcamp-payments-network/includes/wordcamp-budgets-dashboard.php
```diff
@@ -655,9 +663,85 @@ function process_action_approve() {
...
+		//process_action_approve(); // pass params, maybe need to modularize it so it can be called
+		// if do that, then need to add current_user_can, and maybe verify_nonce to this?
+
+		// watch out for recursive loop when calling this
+	}
+}
+
+// phpdoc
+function request_can_be_automatically_approved( $request ) {
+	$meets_criteria = true;
```
coreymckrill: Seems like it would be safer if this defaulted to `false`?

iandunn:

I'll re-submit this.

PR #21, opened 2019-01-22
Prevent submission of payment requests that have no documentation

coreymckrill:
For both Payment Requests and Reimbursement Requests:

* Add a method to check and see if the request post can be submitted. Currently this just checks to see if the post has one or more files attached to it.
* When the post fails the check, show a note in the Status metabox that documentation must be attached first, and don't render the _Submit for Review_ button.

Screenshots:

Payment request with no attachments:
payment-request_no-attachment

Reimbursement request with no attachments:
reimbursement-request_no-attachment

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
@@ -464,6 +464,21 @@ class WCP_Payment_Request {
 	}
 
 	/**
+	 * Check if a post meets the requirements to be submitted for review.
+	 *
+	 * @param WP_Post $post
+	 */
+	protected function can_submit_request( $post ) {
+		$files = WordCamp_Budgets::get_attached_files( $post );
+
+		if ( empty( $files ) ) {
+			return false;
+		}
+
+		return true;
+	}
+
+	/**
 	 * Get the value of a given field.
 	 *
 	 * @param string $name
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
@@ -249,10 +249,10 @@ class WCP_Payment_Request {
 			$current_user_can_edit_request = true;
 		} elseif ( in_array( $post->post_status, $editable_statuses ) ) {
 			if ( WordCamp_Budgets::can_submit_request( $post ) ) {
-				$submit_text = esc_html__( 'Submit for Review', 'wordcamporg' );
-				$submit_note = esc_html__( 'Once submitted for review, this request cannot be edited.', 'wordcamporg' );
+				$submit_text = __( 'Submit for Review', 'wordcamporg' );
+				$submit_note = __( 'Once submitted for review, this request cannot be edited.', 'wordcamporg' );
 			} else {
-				$submit_note = esc_html__( 'Please add an invoice or other supporting documentation in the Files section and save the draft.', 'wordcamporg' );
+				$submit_note = __( 'Please add an invoice or other supporting documentation in the Files section and save the draft.', 'wordcamporg' );
 				$submit_note_class = 'error';
 			}
 
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
@@ -111,6 +111,15 @@ class WCP_Payment_Request {
 	}
 
 	/**
+	 * Get a list of statuses for which posts can be edited by non-admins.
+	 *
+	 * @return array
+	 */
+	protected static function get_editable_statuses() {
+		return [ 'auto-draft', 'draft', 'wcb-incomplete' ];
+	}
+
+	/**
 	 * Register meta boxes
 	 */
 	public function init_meta_boxes() {
@@ -230,7 +239,7 @@ class WCP_Payment_Request {
 			$post->post_status = $back_compat_statuses[ $post->post_status ];
 		}
 
-		$editable_statuses = array( 'auto-draft', 'draft', 'wcb-incomplete' );
+		$editable_statuses = self::get_editable_statuses();
 		$current_user_can_edit_request = false;
 		$submit_text = esc_html_x( 'Update', 'payment request', 'wordcamporg' );
 		$submit_note = '';
@@ -583,7 +592,7 @@ class WCP_Payment_Request {
 
 		// Submit for Review button was clicked.
 		if ( ! current_user_can( 'manage_network' ) ) {
-			$editable_statuses = array( 'auto-draft', 'draft', 'wcb-incomplete' );
+			$editable_statuses = self::get_editable_statuses();
 			if ( ! empty( $post_data_raw['wcb-update'] ) && in_array( $post_data['post_status'], $editable_statuses ) ) {
 				$post_data['post_status'] = 'wcb-pending-approval';
 			}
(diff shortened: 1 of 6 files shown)


vedanshujain (approved):
LGTM

hlashbrooke:
Looks good.

public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
```diff
@@ -230,16 +239,21 @@ public function render_status_metabox( $post ) {
 			$post->post_status = $back_compat_statuses[ $post->post_status ];
 		}
 
-		$editable_statuses = array( 'auto-draft', 'draft', 'wcb-incomplete' );
+		$editable_statuses = self::get_editable_statuses();
```
iandunn: Good job making this DRY.

public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
```diff
@@ -454,6 +468,21 @@ protected function render_files_input( $post, $label, $name, $description = '' )
...
+	 * @param WP_Post $post
+	 */
+	protected function can_submit_request( $post ) {
+		$files = WordCamp_Budgets::get_attached_files( $post );
+
+		if ( empty( $files ) ) {
+			return false;
+		}
+
+		return true;
```
iandunn: Could this be simplified to `return ! empty( $files );` ?

public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
```diff
@@ -289,8 +291,13 @@ function render_status_metabox( $post ) {
...
 	} elseif ( in_array( $post->post_status, $editable_statuses ) ) {
-		$submit_text = esc_html__( 'Submit for Review', 'wordcamporg' );
-		$submit_note = esc_html__( 'Once submitted for review, this request can not be edited.', 'wordcamporg' );
+		if ( can_submit_request( $post ) ) {
+			$submit_text = esc_html__( 'Submit for Review', 'wordcamporg' );
+			$submit_note = esc_html__( 'Once submitted for review, this request can not be edited.', 'wordcamporg' );
+		} else {
+			$submit_note = esc_html__( 'Documentation must be attached and saved to the draft before this request can be submitted.', 'wordcamporg' );
+		}
+
```
iandunn: Do we want to make this DRY also, since it seems identical?

public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
```diff
@@ -404,6 +410,21 @@ function display_post_states( $states ) {
...
+function can_submit_request( $post ) {
+	$files = WordCamp_Budgets::get_attached_files( $post );
+
+	if ( empty( $files ) ) {
+		return false;
+	}
+
+	return true;
+}
+
```
iandunn: seems like this could also be DRY

public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
```diff
@@ -250,7 +252,7 @@ function enqueue_assets() {
  * requests and to submit notes, but they should only be able to change the form fields if the post hasn't been
  * submitted yet, or if we've asked for more information.
  *
- * @param \WP_Post $post
+ * @param WP_Post $post
```
iandunn: Since there's a lot of PHPCS fixes, I think it'd be ideal for them to be in a separate commit, to avoid diff noise, but it's probably not worth breaking the commits apart now that they're already pushed.

public_html/wp-content/plugins/wordcamp-payments/views/payment-request/metabox-status.php
```diff
@@ -1,3 +1,13 @@
+<?php
+/** @var WP_Post $post */
+/** @var WCP_Payment_Request $this */
+/** @var bool $current_user_can_edit_request */
+/** @var string $submit_text */
+/** @var string $submit_note */
+/** @var bool $date_vendor_paid_readonly */
+/** @var string $incomplete_notes */
+/** @var bool $incomplete_readonly */
+?>
```
iandunn: This would be a bit more readable if the `$` were aligned.

public_html/wp-content/plugins/wordcamp-payments/views/payment-request/metabox-status.php
```diff
@@ -86,23 +96,24 @@ class="large-text"
...
-				<?php if ( current_user_can( 'delete_post', $post->ID ) ) : ?>
+			<?php if ( current_user_can( 'delete_post', $post->ID ) ) : ?>
+				<div id="delete-action">
 					<a class="submitdelete deletion" href="<?php echo get_delete_post_link( $post->ID ); ?>">
 						<?php _e( 'Delete', 'wordcamporg' ); ?>
 					</a>
-				<?php endif; ?>
-			</div>
+				</div>
+			<?php endif; ?>
```
iandunn: I wonder if the `delete-action` element was intentionally left in, even if it's blank, maybe to preserve the layout or something? It might be affected by cases where `#publishing-action` is shown/not-shown.

If it tests well for you though, then it's probably safe to change this. It does make more sense to not create it when the button won't be created, and we can always revert this part if we find a problem in the future.

iandunn (approved):
Left some minor thoughts but overall LGTM

sarahmonster:
Some thoughts here from a user experience perspective:

- Change the wording of the notice here to a more active voice—something along the lines of "Please attach an invoice, contracts, or other vendor correspondence to submit your payment request. Your request will be saved as a draft until files are attached."
- Highlight the files field to indicate where this documentation should be attached.
- Use a standard error notice styling and placement.
- It may make sense to nix the notes field and the bit about "If no supporting documentation is available, please indicate the reason in the notes below." as well, unless submitting a request without an attachment, but with text in this field, would be allowed through.

I can put a quick mockup of this together if that's helpful!

public_html/wp-content/plugins/wordcamp-payments/views/payment-request/metabox-status.php
coreymckrill: I haven't noticed any issues in testing. I'd prefer to leave it until someone reports a problem.

iandunn: 👍

public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
iandunn: `wordcamp-payments/wordcamp-budgets.php` has some shared stuff already, so that might be good.

public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
coreymckrill: Good point

public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
```diff
@@ -287,15 +287,17 @@ function render_status_metabox( $post ) {
...
 		$current_user_can_edit_request = true;
 	} elseif ( in_array( $post->post_status, $editable_statuses ) ) {
-		if ( can_submit_request( $post ) ) {
+		if ( WordCamp_Budgets::can_submit_request( $post ) ) {
 			$submit_text = esc_html__( 'Submit for Review', 'wordcamporg' );
-			$submit_note = esc_html__( 'Once submitted for review, this request can not be edited.', 'wordcamporg' );
+			$submit_note = esc_html__( 'Once submitted for review, this request cannot be edited.', 'wordcamporg' );
 		} else {
-			$submit_note = esc_html__( 'Documentation must be attached and saved to the draft before this request can be submitted.', 'wordcamporg' );
+			$submit_note = esc_html__( 'Please add an invoice or other supporting documentation in the Files section and save the draft.', 'wordcamporg' );
```
iandunn: Should this be escaped when it's output rather than here?

coreymckrill:
Committed to SVN: https://meta.trac.wordpress.org/changeset/8136

PR #22, opened 2019-01-22
WCPT: Move post type check to before nonce check.

vedanshujain:
This was giving an error sometimes while submitting a post, looks like because `metabox_save` is getting enqueued some other place as well.

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
@@ -412,17 +412,17 @@ abstract class Event_Admin {
 			return;
 		}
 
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -412,7 +412,11 @@ public function metabox_save( $post_id, $post ) {
 			return;
 		}
 
-		// Make sure the requset came from the edit post screen.
+		if ( $this->get_event_type() !== get_post_type() ) {
```
iandunn: Using `$post` might be a bit safer than `get_post_type()`. I can't think of a case where they'd be different, but there might be some.

PR #25, opened 2019-02-13
WordCamp Payment: Add status "Paid" and "Cancelled" to list of status…

vedanshujain:
…es where we do not enforce required fields.

We periodically delete bank details data for privacy reasons. Because of this, it was not possible to update any other field in the Reimbursement request or Vendor Payment request without filling junk data into deleted fields. This patch allows updating these posts after the bank data is deleted.


iandunn (approved):
👍

coreymckrill (approved):

PR #26, opened 2019-02-13
Docker setup

vedanshujain:


vedanshujain:
Creating this PR just for record keeping. No need to merge

nginx.conf
```diff
@@ -0,0 +1,44 @@
...
+server {
+		listen 443 ssl http2;
+        listen [::]:443 ssl http2;
+
+        server_name wordcamp.test;
+		ssl_certificate /etc/ssl/certs/wordcamp.crt;
+		ssl_certificate_key /etc/ssl/private/wordcamp.key;
+		ssl_protocols TLSv1.2 TLSv1.1 TLSv1;
+
+        root /usr/src/pulic_html;
```
vedanshujain:
```suggestion
        root /usr/src/public_html;
```

iandunn:
If we circle back to this:

* We should also match the PHP version to what production is running
* I think it'd be good to place all the new files inside a `provision` folder, to avoid cluttering the root folder. 
* Could the certs be ignored by Git, since it looks like they're created automatically?

iandunn:
Closing since there's nothing actionable to do right now. I added the `Maybe Later` tag so it'll be easy to find in the future. I'll add a link to Trello as well.

coreymckrill:
It looks like there are a bunch of changes in here that are unrelated to Docker. @vedanshujain did these get committed already elsewhere?

iandunn:
I think that's just because of the problems we had earlier w/ commit authorship/etc. If this were rebased then those other commits would go away.

PR #28, opened 2019-02-19
Error handling improvements

coreymckrill:
* Rename 0-slack.php to 0-error-handling.php, since that file's only
  purpose is to load the Slack library for error handling purposes
  anyway.
* Move all of the error handling stuff from wcorg-misc.php to
  0-error-handling.php. There's enough of it to warrant a separate file,
  and this way it all gets loaded before the rest of the mu-plugins
  so it can potentially catch errors there as well.
* Run the custom error handler and shutdown function in other
  environments besides production, as long as the SANDBOX_SLACK_USERNAME
  constant is defined. This way the errors will be sent to the sandbox
  owner's Slack account, rather than the log channel.
* In the shutdown function, only send the error to Slack if it is one
  of the types of errors that can't be handled in the custom error
  handler.


iandunn:
I like the idea of moving this all the that file, but I think we might want it to load earlier than mu-plugins, like during `define-wordcamp.php`, which is the equivalent of where w.org loads theirs. That way it'll catch things that happen earlier in the process.

iandunn:
Maybe we can store the file somewhere in `wp-content` though, so it can be in the public repo, and just include it from `define`?

vedanshujain (approved):
LGTM. Other than Ian's comment.

coreymckrill:
I'm not sure that the `add_action` and `wp_schedule_event` functions would be available when *define-wordcamp* is loading. I'll do some testing.

public_html/wp-content/mu-plugins/0-error-handling.php
```diff
@@ -0,0 +1,169 @@
...
+	$err_key = substr( base64_encode("$file-$line-$err_no" ), -254 );
+
+	$error_file = ERROR_RATE_LIMITING_DIR . "/$err_key";
+
+	$text = '';
+
+	$data = array(
+		'last_reported_at' => time(),
+		'error_count'      => 0, // since last reported.
+	);
```
iandunn: low priority, but it'd be nice to remove those blank lines and align the `=` operator.

public_html/wp-content/mu-plugins/0-error-handling.php
```diff
@@ -0,0 +1,169 @@
...
+			$data['error_count']      = 0;
+
+			file_put_contents( $error_file, wp_json_encode( $data ) );
+		} else {
+			file_put_contents( $error_file, wp_json_encode( $data ) );
+			return false;
+		}
+	}
+
+	$domain    = get_site_url();
```
iandunn: If we move the file earlier, this (and similar functions) might not exist yet either.

public_html/wp-content/mu-plugins/0-error-handling.php
```diff
@@ -0,0 +1,169 @@
...
+	$unhandled_error_types = [ E_ERROR, E_PARSE, E_CORE_ERROR, E_CORE_WARNING, E_COMPILE_ERROR, E_COMPILE_WARNING ];
+
+	if ( ! empty( $error ) && in_array( $error['type'], $unhandled_error_types, true ) ) {
+		send_error_to_slack( $error['type'], $error['message'], $error['file'], $error['line'] );
+	}
+}
+
+/**
+ * Check and create filesystem dirs to manage rate limiting in error handling.
+ * For legacy bugs we are doing rate limiting via filesystem. We would be investigating to see if we can instead use memcache to rate limit sometime in the future.
```
iandunn: Nitpicky, but there should be an empty line here.

public_html/wp-content/mu-plugins/0-error-handling.php
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * @return bool Return true if file permissions etc are present
+ */
+function init_error_handling() {
+	if ( ! file_exists( ERROR_RATE_LIMITING_DIR ) ) {
+		mkdir( ERROR_RATE_LIMITING_DIR );
```
iandunn: `wp_mkdir_p()` would be a best practice, but probably not helpful in this case, since `/tmp` should always exist and we wouldn't have permission to create it if it didn't.

public_html/wp-content/mu-plugins/0-error-handling.php
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	return is_dir( ERROR_RATE_LIMITING_DIR ) && is_writeable( ERROR_RATE_LIMITING_DIR );
+}
+
+/**
+ * Function `send_error_to_slack` above also creates a bunch of files in /tmp/error_limiting folder in order to rate limit the notification.
```
iandunn: This is missing a short doc.

iandunn (approved):
Mostly nitpicky comments, feel free to ignore :)

It'd be really nice to get the handler setup as early in the execution path as possible, though. Maybe this could be split into two files? One that loads super early and just logs the errors to disk, and a "phase 2" that runs at shutdown and sends to Slack, removes files, etc?

That's not essential if it's gonna be a ton of time, though. In that case maybe just leave a comment at the top that it's a future interation. We'll want to remind ourselves that it's possible for errors to occur early and not be logged, in case we're trying to debug something and can't figure out what's going on.

coreymckrill:
Committed in https://meta.trac.wordpress.org/changeset/8287

PR #29, opened 2019-02-21
Travis: Run unit test suite

iandunn:
I started working on this, but didn't get very far. @vedanshujain, how did you test the original config before committing it? I couldn't find any good info on that.

Also, do you have any feedback on this?

In addition to running the tests, I pulled in some stuff from Core's config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,43 @@
-# Travis CI configuration file.
-# @link https://travis-ci.org/
+# travis-ci.org continuous integration testing.
 
-# Declare project language and PHP versions to test against.
-# @link http://about.travis-ci.org/docs/user/languages/php/
+dist: xenial
+
 language: php
 php:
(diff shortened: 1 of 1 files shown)


iandunn:
We should switch to phpunit 7.x, since that's the latest version that Core's framework supports.

iandunn:
> how did you test the original config before committing it? I couldn't find any good info on that.

looks like you can [run it locally in a Docker instance](https://stackoverflow.com/a/43625282/450127)

it sounds like it should also work by sending a PR, but it doesn't look like this PR was tested.

vedanshujain:
I think I tested it by committing changes to PR, but clearly tests are not running for this one. Let me know if you want to pair up for this. 
BTW kudos for adding unit tests infra , looks like a well needed addition.

iandunn:

I'll re-submit this.

PR #30, opened 2019-02-25
WCPT: Do not sent slack notifications for same events again.

vedanshujain:


vedanshujain:
- [ ] Backfill notification sent flags for WordCamp and Meetup that are already declined / scheduled / added to chapter.

coreymckrill:
👍

public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-admin.php
```diff
@@ -526,8 +526,8 @@ public static function notify_new_meetup_group_in_slack( $meetup ) {
 
 			$message = sprintf(
 				"Let's welcome the new WordPress meetup group%s%s, to the chapter! :tada: :community: :wordpress:\n%s",
-				isset( $city ) ? " in $city," : '',
-				isset( $organizer_slack ) ? " organized by @$organizer_slack" : '',
+				! empty( $city ) ? " in $city," : '',
+				! empty( $organizer_slack ) ? " organized by @$organizer_slack" : '',
```
iandunn: I think it's a bit easier to read like this:

`empty( $city ) ? '' : " in $city,"`

That way you don't have to mentally apply the `!` and it all flows smoother.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -455,6 +456,7 @@ public function metabox_save( $post_id, $post, $verify_nonce = true ) {
 		}
 
 		// Don't add/remove meta on trash, untrash, restore, etc.
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing
```
iandunn: That sniff has so many false positives, I tempted to just say we should disable it.

It is a security one, though, which makes me leery of that. I hate adding cluttering the code with `phpcs:ignore`, but maybe in this case that's a reasonable compromise?

What do y'all think? @vedanshujain , @coreymckrill

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -258,6 +258,7 @@ public function metabox_status( $post ) {
 	 *
 	 * @return string
 	 */
+	// phpcs:ignore WordPress.NamingConventions.ValidFunctionName.MethodNameInvalid
```
iandunn: I think it's ok to say `set_locale_to_en_us` here instead of adding the `phpcs:ignore`. Not a big deal either way though.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -330,16 +330,29 @@ public static function schedule_decline_notification( $event, $label, $location
 	 * @param int    $event_id
 	 * @param string $label
 	 * @param string $location
+	 *
+	 * @return null|bool|string
 	 */
 	public static function send_decline_notification( $event_id, $label, $location ) {
+		$declined_notification_key = 'Sent declined notification';
```
iandunn: It's possible to have spaces in keys, but conventionally they're `underscore_case`.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -330,16 +330,29 @@ public static function schedule_decline_notification( $event, $label, $location
...
+		$already_notified          = get_post_meta( $event_id, $declined_notification_key, true );
+		if ( $already_notified ) {
+			return null;
+		}
```
iandunn: This could be shortened to:

```php
if ( get_post_meta( $event_id, $declined_notification_key, true ) ) {
    return null;
}
```

...since `$already_notified` isn't being used.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -330,16 +330,29 @@ public static function schedule_decline_notification( $event, $label, $location
...
 		);
 
 		$attachment = create_event_status_attachment( $message, $event_id, '' );
-		wcpt_slack_notify( COMMUNITY_TEAM_SLACK, $attachment );
+
+		$notification_sent = wcpt_slack_notify( COMMUNITY_TEAM_SLACK, $attachment );
+		if ( $notification_sent ) {
+			update_post_meta( $event_id, $declined_notification_key, true );
+		}
+		return $notification_sent;
```
iandunn: I'm wondering if this should return anything, since it's a cron callback. The return value won't ever be used, and it feels a bit odd for a function to return `null` or a `bool` or a `string`.

What's the goal/benefit of tracking that? If we want to know about failures at some point, it seems like adding logging to `wcpt_slack_notify()` might be a better place for them. If it's about unit-testability, we'd probably need to mock `wcpt_slack_notify()` anyway, so we could do that in a way that lets us test for success/failure (similar to how PHPMailer is mocked by Core's test suite).

public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-admin.php
```diff
@@ -529,7 +533,11 @@ public static function notify_new_meetup_group_in_slack( $meetup ) {
 
 			$attachment = create_event_status_attachment( $message, $meetup->ID, $title );
 
-			return wcpt_slack_notify( COMMUNITY_EVENTS_SLACK, $attachment );
+			$notification_sent = wcpt_slack_notify( COMMUNITY_EVENTS_SLACK, $attachment );
+			if ( $notification_sent ) {
+				update_post_meta( $meetup->ID, $new_group_notification_key, true );
+			}
+			return $notification_sent;
```
iandunn: Same comments as above regarding the `return`.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
```diff
@@ -761,7 +767,11 @@ public static function notify_new_wordcamp_in_slack( $wordcamp ) {
 
 			$attachment = create_event_status_attachment( $message, $wordcamp->ID, $title );
 
-			return wcpt_slack_notify( COMMUNITY_EVENTS_SLACK, $attachment );
+			$notification_sent = wcpt_slack_notify( COMMUNITY_EVENTS_SLACK, $attachment );
+			if ( $notification_sent ) {
+				update_post_meta( $wordcamp->ID, $scheduled_notification_key, true );
+			}
+			return $notification_sent;
```
iandunn: Same comments as above regarding the `return`, and removing `$already_notified`.

There's a lot of duplicating in these 3 functions. I wonder if we should refactor them to be more DRY? It may not be worth it, though, since it's just a few lines in each.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
vedanshujain: lets keep it, because as you say, it sound like a reasonable compromise given that this affects security. we can revisit this discussion later if this gets too annoying.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
vedanshujain: no good reason, so it returns null|bool|string because it returns whatever the response is from the web request, which could be anything really ( and null in case of connection error ) . i like keeping return values documented if a function can return something in case we may need that in future, but right we are not using it.
lets keep it for now, because there is no overhead in keeping it, but you are correct that its very unlikely that we will use it.

vedanshujain:
this is merged

PR #32, opened 2019-03-01
Make forms-to-drafts compatible w/ the block editor

coreymckrill:
* Changes all the post and page stubs for new sites to use block editor grammar. This includes the stub with the Call For Speakers form, which is what is currently used by the Forms To Drafts plugin.
* This also moves all the post and page stub content to separate files so that they are easier to manage and modify.
* Modify the callbacks that create the speaker and session drafts to use block editor grammar for the post content.

Question:

* Does this fully address [this conversation](https://wordpress.slack.com/archives/C08M59V3P/p1546874499054900)?

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
@@ -488,50 +488,31 @@ class WordCamp_New_Site {
 	 * @return array
 	 */
 	protected function get_stub_pages( $wordcamp, $meta ) {
-		// todo remove the to field from all contact forms and notes, just let it default to the admin email
-
 		$pages = array(
 			array(
 				'title'   => __( 'Schedule', 'wordcamporg' ),
-				'content' =>
-					'<p>'  . __( '<em>Organizers note:</em> You can enter content for this page in the Sessions menu item in the sidebar.', 'wordcamporg' ) . '</p> ' .
-					'<h1>' . __( 'Saturday, January 1st', 'wordcamporg' ) . '</h1> ' .
-					'<p>[schedule date="YYYY-MM-DD" tracks="example-track,another-example-track,yet-another-example-track"]</p>',
+				'content' => $this->get_stub_content( 'page', 'schedule' ),
 				'status'  => 'publish',
 				'type'    => 'page',
 			),
 
 			array(
 				'title'   => __( 'Speakers', 'wordcamporg' ),
-				'content' =>
-					'<p>' . __( '<em>Organizers note:</em> You can enter content for this page in the Speakers menu item in the sidebar.', 'wordcamporg' ) . '</p> ' .
-					'<p>[speakers]</p>',
+				'content' => $this->get_stub_content( 'page', 'speakers' ),
 				'status'  => 'publish',
 				'type'    => 'page',
 			),
 
 			array(
 				'title'   => __( 'Sessions', 'wordcamporg' ),
-				'content' =>
-					'<p>' . __( '<em>Organizers note:</em> You can enter content for this page in the Sessions menu item in the sidebar.', 'wordcamporg' ) . '</p> ' .
-					'<p>[sessions orderby="session_time" order="asc"]</p>',
+				'content' => $this->get_stub_content( 'page', 'sessions' ),
 				'status'  => 'publish',
 				'type'    => 'page',
 			),
 
 			array(
 				'title'   => __( 'Sponsors', 'wordcamporg' ),
-				'content' =>
-					'<p>' . sprintf(
-						/* translators: %s: Global Community Sponsorship page URL */
-						__( "<em>Organizers note:</em> Multi-event sponsors have been automatically created in the Sponsors menu, but you'll need to remove the ones that don't apply to your specific event. To find out which ones apply, please visit the <a href=\"%s\">Global Community Sponsorship</a> handbook page. After that, you should add the sponsors that are specific to your event. For non-English sites, make sure the URL below matches the Call for Sponsors page.", 'wordcamporg' ),
-						'https://make.wordpress.org/community/handbook/wordcamp-organizer/planning-details/fundraising/global-community-sponsorship-for-event-organizers/'
-					) . '</p> ' .
-					'<h3>' . __( 'Our Sponsors', 'wordcamporg' ) . '</h3> ' .
-					'<p>'  . __( 'Blurb thanking sponsors', 'wordcamporg' ) . '</p> ' .
-					'<p>[sponsors]</p> ' .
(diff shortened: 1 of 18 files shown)


public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
```diff
@@ -488,50 +488,31 @@ protected function create_post_stubs( $wordcamp, $meta, $lead_organizer ) {
...
-		// todo remove the to field from all contact forms and notes, just let it default to the admin email
-
 		$pages = array(
 			array(
 				'title'   => __( 'Schedule', 'wordcamporg' ),
-				'content' =>
-					'<p>'  . __( '<em>Organizers note:</em> You can enter content for this page in the Sessions menu item in the sidebar.', 'wordcamporg' ) . '</p> ' .
-					'<h1>' . __( 'Saturday, January 1st', 'wordcamporg' ) . '</h1> ' .
-					'<p>[schedule date="YYYY-MM-DD" tracks="example-track,another-example-track,yet-another-example-track"]</p>',
+				'content' => $this->get_stub_content( 'page', 'schedule' ),
```
iandunn: I like this a lot better than inlining everything, kudos!

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
```diff
@@ -762,6 +631,27 @@ protected function get_stub_posts( $wordcamp, $meta ) {
...
+	 * @param string $post_type
+	 * @param string $stub_name
+	 *
+	 * @return string
+	 */
+	protected function get_stub_content( $post_type, $stub_name ) {
+		$content   = '';
+		$stub_file = WCPT_DIR . "stubs/$post_type/$stub_name.php";
+
+		if ( is_readable( $stub_file ) ) {
```
iandunn: I'm not sure we need this check here, since it's our own file and inside this same plugin. If one of us changes one of the files and forgot to update calls to it, we'd probably want to see a fatal in our local environments to notify us before we deploy that change.

To me, having this check feels similar to `WordCamp_New_Site::configure_new_site()` doing this:

```php
if ( is_callable( array( $this, 'get_user_or_current_user' ) ) {
	$lead_organizer = $this->get_user_or_current_user( $meta['WordPress.org Username'][0] );
}
```

Having said that, I don't feel strongly about it.

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
```diff
@@ -269,6 +269,8 @@ protected function simulate_post_type( $post_type ) {
 	 *   taxonomy and the selected term is applied to the drafted post. Maybe need to send PR to add filter to
 	 *   insert custom fields programmatically.
 	 * - Sideload the logo from submitted URL and set it as the featured image.
+	 * - Add a Company Description field to the form stub, or populate content from something else?
```
iandunn: I'm assuming this is for the `call for sponsors` stub, but it's probably worth making that explicit.

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
```diff
@@ -378,11 +380,19 @@ protected function get_speaker_from_user_id( $user_id ) {
 	 * @return int | WP_Error
 	 */
 	protected function create_draft_speaker( $speaker ) {
+		$content = ( ! empty( $speaker['Your Bio'] ) ) ? $speaker['Your Bio'] : '';
```
iandunn: Could this be shortened to `$speaker['Your Bio'] ?? ''` ?

If not, I don't think the parens are needed around the condition, and to me it feels like it makes it harder to read. Not a big deal either way though.

Same applies to below code too.

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
```diff
@@ -378,11 +380,19 @@ protected function get_speaker_from_user_id( $user_id ) {
...
+
+		if ( $content ) {
+			$content = wpautop( $content );
+			$content = str_replace( '<p>', "<!-- wp:paragraph -->\n<p>", $content );
+			$content = str_replace( '</p>', "</p>\n<!-- /wp:paragraph -->", $content );
```
iandunn: Is it worth doing that conversion here? Won't G do it anyway if they save in the block editor? Or do we need to do it to avoid it being set as a Classic block?

Same applies to below code too.

iandunn (approved):
Left a few suggestions, but it'd be fine to commit as-is too. Great job :)

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
coreymckrill: I see what you're saying. If we did change something that cause this to fatal, though, it would halt the whole process of setting up a new site, which I think is a lot worse than having a stubbed page with missing content. Maybe there could be an `else` that logs an error if the file isn't readable?

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
coreymckrill: In my testing, content without the G grammar loads a Classic block, so I think this is preferable.

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
coreymckrill: Ah good point. I always get `??` and `?:` confused, and `?:` wasn't working (see previous commit) so I just switched to the longform ternary. This `??` seems to work, though.

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
coreymckrill: Good call.

coreymckrill:
Committed: https://meta.trac.wordpress.org/changeset/8384

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
iandunn: Yeah, that's a fair compromise :)

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
iandunn: 👍

PR #31, opened 2019-02-26
Sponsor block

vedanshujain:

PR #27, opened 2019-02-18
Sessions block

coreymckrill:
A block version of the sessions shortcode.

Discussion

* https://make.wordpress.org/community/2018/10/30/wordcamp-block-sessions/

Design

* https://wp.invisionapp.com/share/7DOUB92NTEZ#/screens/328383331
* https://wp.invisionapp.com/share/JUOUBYORWBZ#/screens/328389037
* https://wp.invisionapp.com/share/QWOUA278STJ#/screens/328372426

PR #33, opened 2019-03-01
Add Sessions and Sponsor blocks, and required shared components

coreymckrill:
This expands the work on the Speakers block to include several more blocks:

- [x] Sessions
- [x] Sponsors

This is a combination of two previous pull requests, #27 and #31, because we realized that we needed to collaborate on the shared components, which would be easier in a single branch.


public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,183 @@
...
+	 */
+	constructor( props ) {
+		super( props );
+	}
+
+	componentWillMount() {
+		let availableSizes;
+		const { wpMediaDetails } = this.props;
+		if ( wpMediaDetails ) {
+			availableSizes = FeaturedImage.parseWPMediaDetails( wpMediaDetails );
```
coreymckrill: Can `this` be used here as well, instead of `FeaturedImage`?

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,183 @@
...
+		}
+		return availableSizes;
+	}
+
+	/**
+	 * Pre-calculate and store aspect ratio of all images. This will help us in picking the best image available for the size that we have.
+	 */
+	parseAvailableSizes( availableSizes ) {
+		let largestSizeValue = 0;
+		let largestSize = {};
```
coreymckrill: It looks like these two variables aren't used...

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,183 @@
...
+				} else {
+					return 1;
+				}
+			}
+		);
+
+		this.setState(
+			{
+				sizes,
+				...this.state
```
coreymckrill: I don't think adding the spread of the whole state object is necessary here. You can just include the items that are being added or updated.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,183 @@
...
+
+		// Lets loop through all possible images, see if we have an image that matches requested aspect ration. if we have, we will return it. Also keep track of image which has the most closest aspect ration.
+		let smallestImage;
+		let selectedImage;
+		const requiredAspectRatio = width / height;
+
+		for ( const size of sizes ) {
+
+			// Reject all the images with less height or width then 20px.
+			if (size.height < height - 20 || size.width < width - 20) {
```
coreymckrill: Will this be sufficient to catch all images that aren't the right size? What if the height is within 20px but the width is way less?

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,183 @@
...
+			}
+		);
+	}
+
+
+	/**
+	 * Returns URL of appropriate image. We find appropriate by looking for an image with similar aspect ratio, and similar or greater area.
+	 * If there is no image with similar aspect ratio or area, we will return strictly bigger image then currently requested.
+	 * If there is no bigger image then currently request one, we will return the max size image URL.
+	 * Note: Core Image block allows user to select image size by name, but we cannot really do that because we will be controlling multiple images at once, and one or more images may not have all the available sizes. So an automated solution is preferred here.
```
coreymckrill: Kudos, I think this sounds like a good approach. 

> we will be controlling multiple images at once, and one or more images may not have all the available sizes.

I wonder, though, if it would be worth still letting the user pick a size, and then use these algorithms as a backup for the case when the correct size isn't available. That way, when the size _is_ available, we could skip all these calculations...

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/block-content.js
```diff
@@ -0,0 +1,50 @@
+import { get } from 'lodash';
+import RawHTML from '@wordpress/element/build-module/raw-html';
```
coreymckrill:
```suggestion
const { RawHTML } = wp.element;
```

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: 🤔 thats a good point. So in this case this component will need two inputs, one would be of size name, and other would be an object of `{ height, width }`. We will use height, width and this algo if size is not available.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: So if width is less then 20px, then the image will be rejected.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
coreymckrill: 👍

vedanshujain: So I ended with not having pre-defined sizes, but having height and width pass via props. Algo will automatically pick exact size if its available, otherwise will make an effort to find appropriately sized image.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: fixed

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: fixed

vedanshujain:
Pending stuff in Sponsor block:
- [ ] Add option to display `excerpt` instead of just `None` and `Full`.
- [ ] Add option for linking the logos as well.
...

vedanshujain:
@coreymckrill Can you review the code for sponsor block. There are some minor stuff pending as stated in above comment, but I think its now in a state where a review would be helpful.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: I still have not looked at this, but its on task list. I should be able to, but it was not working as is, so have to take a detailed look.

coreymckrill (commented):
A few inline comments below, but here are some general comments as well:

* I think the list/grid layout implementation should be consistent across all of our blocks that use it. Right now it looks like you've simplified from what is being done in the Speakers block by just having a range control to choose from 1 to 4 columns, and omitting the block toolbar buttons for switching between list view and grid view. This is simpler, but it doesn't match the design mockups, and personally I think it's not as intuitive for the end user when you don't differentiate between list and grid. If you feel strongly, though, we should have a discussion and include some design perspective, maybe in a separate GitHub issue.
* I see you've got a todo list that includes adding an excerpt option for the sponsor post content. Another thing missing is sorting. In the design spec the default sorting should be first by sponsor level and second by post title.
* The `FeaturedImage` component seems to work well, but I think the image sizing options in the Inspector sidebar still have a ways to go before they're ready. First of all, I think they should be extracted out into a shared component, because the Sessions block needs to have featured image size options as well. But also, they aren't really usable at the moment. The Small/Medium/Large buttons assume a 1:1 aspect ratio, so a lot of images are skewed when you use them, and the manual width/height inputs are hard to use because you have to calculate the correct aspect ratio yourself. I think we should try to emulate the sidebar sizing options found in the core Image block, that focus on scaling rather than actual pixel dimensions.
* The HTML markup of the content isn't consistent with the other blocks. If you take a look at the latest in the Sessions block, I just added some shared components for outputting a post title and post content, that I think would be good to use here too.

vedanshujain:
> Right now it looks like you've simplified from what is being done in the Speakers block by just having a range control to choose from 1 to 4 columns, and omitting the block toolbar buttons for switching between list view and grid view. This is simpler, but it doesn't match the design mockups, and personally I think it's not as intuitive for the end user when you don't differentiate between list and grid

Why do you think that its not intuitive when you don't differentiate between list and a grid? It feels like this is an implementation detail, because, in our case a grid with column = 1 is basically a list. 
Do you think its more intuitive because its in a toolbar? In that case we can have column buttons in toolbar itself.
I do agree that we should have consistent UI across all blocks though. I will separate out grid logic in a shared component so that we can use it across. 

> Another thing missing is sorting. In the design spec the default sorting should be first by sponsor level and second by post title.

So I did not implemented sorting because anyone could use different names for sponsor level, and I don't think we enforce a strict naming convention so we won't know the correct sort order. In this case, having a sort option for A->Z and Z->A should be good enough? And if someone wants to sort by levels, they would have to add different block for each sponsor level. I could be missing context information here, so please correct me if I am wrong.

>  The Small/Medium/Large buttons assume a 1:1 aspect ratio, so a lot of images are skewed when you use them, and the manual width/height inputs are hard to use because you have to calculate the correct aspect ratio yourself. I think we should try to emulate the sidebar sizing options found in the core Image block, that focus on scaling rather than actual pixel dimensions.

That's a good point. Will go ahead and add options for alignment as well. I think instead of Small/Medium/Large button, we can have a drop down with a bunch of different sizes, like the core image block would have for an image with different sizes actually present. Although I'd still keep the height and width boxes, not just for customization, but because they are also in core image block.

> The HTML markup of the content isn't consistent with the other blocks. If you take a look at the latest in the Sessions block, I just added some shared components for outputting a post title and post content, that I think would be good to use here too.

Sounds good.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,185 @@
...
+			if ( typeof( onChange ) === 'function' ) {
+				onChange( imageURL );
+			}
+		}
+		return(
+			<img
+				className={ classnames( 'featured-image', className ) }
+				src = { imageURL }
+				alt = { alt }
+				style = { { height: height + "px"} }
```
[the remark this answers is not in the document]
vedanshujain: IIRC height attribute directly was not working because of an overzealous CSS rule.

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/inspector-controls.js
```diff
@@ -0,0 +1,151 @@
...
+];
+
+/**
+ * Class for defining Inspector control in sponsor block.
+ */
+class SponsorInspectorControls extends Component {
+
+	setFeaturedImageSize( value ) {
+		const { setAttributes } = this.props;
+		const { height, width } = SponsorInspectorControls.getSizeChart()[ value ];
```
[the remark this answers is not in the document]
vedanshujain: this was a stranded function, removed now

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/edit.js
```diff
@@ -0,0 +1,82 @@
...
+	const sponsorLevelQuery = {
+		orderby : 'id',
+		order: 'asc',
+		per_page: MAX_PAGE,
+		_embed: true
+	};
+
+	return {
+		sponsorPosts: apiFetch( { path: addQueryArgs( '/wp/v2/sponsors', sponsorQuery ) } ),
+		sponsorLevels: apiFetch( { path: addQueryArgs('/wp/v2/sponsor_level', sponsorLevelQuery ) } )
```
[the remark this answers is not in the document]
vedanshujain: Ah, this is a regression, thanks for pointing it out. I have fixed with a cleaner solution, so basically I am fetching all the posts in one API call ( since we need them anyway ) for sponsor and one for sponsor level. Then doing the filtering on client side. 
This approach prevents at least 1 API call per resource and have less code.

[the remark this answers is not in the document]
coreymckrill: Ah cool. I will take a look at that. It might be worth doing that for all the blocks. My original reasoning for using `withSelect` was that the query parameters could result in different sets of speakers, with different sort orders. But since we're currently retrieving all the posts anyway for the Select, you're right that we could just pull them all once, and then do the filtering and sorting in-browser.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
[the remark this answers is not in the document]
coreymckrill: Ugh, I hate it when that happens. I do think, however, the HTML tag and attributes should be the same on the front and back end, and that we should use `get_the_post_thumbnail` for the PHP front end version, because it handles all the stuff around sizing and srcset.

There are some other ways around overly-broad CSS rules. Here's what I did for one case in Speakers: https://github.com/WordPress/wordcamp.org/blob/master/public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.scss#L8-L14

coreymckrill:
> Why do you think that its not intuitive when you don't differentiate between list and a grid? It feels like this is an implementation detail, because, in our case a grid with column = 1 is basically a list.

Yeah, I agree that when considered from an implementation perspective, this is simpler. But I'm not sure that end users would think about it the same way. Additionally:

* Having the layout mode button in the toolbar makes it more discoverable, but having a dropdown for the number of columns or a separate button in the toolbar for each column option might be overkill. 
* Have separate `layout-list` and `layout-grid` classes gives themers more flexibility, especially if you're including more than one instance of a block on a post/page.
* Keeping a layout mode option separate from the number of columns makes it easier if we wanted to add other layout options in the future.

> So I did not implemented sorting because anyone could use different names for sponsor level, and I don't think we enforce a strict naming convention so we won't know the correct sort order.

True, sponsor level names are arbitrary. But there's actually a tool in WP-Admin for putting the levels in a certain order, so I think we can leverage that. Check out `get_sponsor_levels()` in **wc-post-types**. However...

> And if someone wants to sort by levels, they would have to add different block for each sponsor level.

This is a good point. This might be worth adding to our list of followup questions when we ask for review.

> I think instead of Small/Medium/Large button, we can have a drop down with a bunch of different sizes, like the core image block would have for an image with different sizes actually present.

Yeah, I think this would be a good approach. I don't remember the function name off hand, but there's a way to get a list of all of the officially registered images sizes (maybe the Core block does this already?), so we could use that to populate the dropdown. Then the logic that you've already put into the `FeaturedImage` component would handle the case where the correct size isn't available for a particular featured image.

> Although I'd still keep the height and width boxes, not just for customization, but because they are also in core image block.

Agreed. I also think it would be good to link the two fields so if you change one, it updates the other to maintain the aspect ratio. I surprised that the core Image block doesn't do this, actually.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,187 @@
...
+		const imageSize = this.props.size || { height: 150, width: 150 };
+
+		const { height, width } = imageSize;
+
+		const imageURL = this.getSizedUrl( height, width );
+		if ( imageURL !== this.currentImageURL ) {
+			this.currentImageURL = imageURL;
+			if ( typeof( onChange ) === 'function' ) {
+				onChange( imageURL );
+			}
```
coreymckrill: I'm not sure I understand the purpose of this. Won't just changing the value of `src` in the `img` element cause it to update?

Also, would it make sense to store this in state instead of a custom property? Several things I've read suggest it's an anti-pattern to modify properties directly, although I couldn't tell you exactly why that is.

coreymckrill:
@vedanshujain Here are some notes before I head out:

* I created a new shared component that combines stuff from `VersatileSelect` and `CustomPostTypeSelect`. I'm hoping we can use this instead of having two separate components that do very similar things. I refactored Sessions and Speakers to use it. More info in the [commit message](https://github.com/WordPress/wordcamp.org/pull/33/commits/bad2798ad8f78bbb921898ad90cacdf801e02fdb).
* If you run `npm run lint:js`, you will see that there's a lot of linting cleanup to do.
* The Sessions block is pretty close. It still needs to implement the `FeaturedImage` component and add Inspector controls for image-related attributes. If you get to a place next week where those are in good shape in Sponsors, feel free to take a crack at adding them to Sessions as well.
* Sessions also might need the toolbar with buttons for list and grid layout modes. I wasn't sure multiple columns actually made sense for Sessions, but I guess there might be a use case, and that would keep it consistent with the other blocks.
* For places where our blocks deviate from the original design mockups, I think we should draft a post seeking review and feedback from the designer. I've got a few things in mind for the Sessions block, but it would be good to start thinking about Sponsors as well.

iandunn:
Since this is a very large PR, it'd be nice to break it into reasonably atomic commits for SVN. e.g.,

* Adding shared block content functions, another for block controls, etc
* Adding shared featured image component, another for grid-layout, another for i18n, etc
* Updating speakers block to use new shared components
* Adding sessions block
* Adding sponsors block
* Adding organizers block

vedanshujain:
I guess we can do that once we are ready to merge. Otherwise we would have too do all that again.

public_html/wp-content/mu-plugins/blocks/view/sessions.php
```diff
@@ -5,135 +5,132 @@
 use function WordCamp\Blocks\Shared\{ get_all_the_content, array_to_human_readable_list };
```
vedanshujain: this file mostly has alignment changes because of removing a loop.

iandunn:
Yeah, it might be easiest to just do it during the commit to SVN, rather than trying to do it in Git.

public_html/wp-content/mu-plugins/blocks/view/shared/featured-image.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @param array    $class_names        Additional classes to add inside <img> tag.
+ * @param \WP_Post $post               Current post object. This will be used to calculate srcset attribute.
+ * @param string   $selected_image_url URL for selected sized image.
+ * @param int      $height             Height of the image
+ * @param int      $width              Width of the image
+ *
+ * @return string Output markup for featured image.
+ */
+function render_featured_image( $class_names, $post, $height, $width ) {
```
iandunn: @vedanshujain , this doesn't feel like a "view" to me, since it doesn't have any actual HTML. I wonder if it'd be more logical to live in `blocks/includes/shared.php`?

public_html/wp-content/mu-plugins/blocks/view/shared/grid-layout.php
```diff
@@ -0,0 +1,41 @@
...
+ * Provides rendering for grid-layout component.
+ *
+ * @param string $layout            Whether the layout is `grid` or `list`.
+ * @param int    $columns           Number of columns if layout is `grid`. Assumed to be 1 if layout is grid.
+ * @param array  $children          Array of posts to rendered inside layout. Should be output markup.
+ * @param array  $container_classes Array of classes that will be added to container.
+ *
+ * @return string Markup of output layout.
+ */
+function render_grid_layout( $layout, $columns, $children, $container_classes ) {
```
iandunn: @vedanshujain Typically the way we've done these in other parts of the codebase is to have the function that creates the variables live inside the normal `includes`/`classes`/etc folder, and then only the HTML itself in the `view/` folder. That keeps the logic and presentation separated. I wonder if we should do that here too, for consistency, and more of a pure MVC-ish layout?

public_html/wp-content/mu-plugins/blocks/view/sessions.php
```diff
@@ -0,0 +1,126 @@
...
+/** @var string $container_classes */
+
+setup_postdata( $session );
+
+?>
+
+<div class="wordcamp-block-post-list-item wordcamp-session wordcamp-session-<?php echo sanitize_html_class( $session->post_name ); ?> wordcamp-clearfix">
+	<h3 class="wordcamp-item-title wordcamp-session-title">
+		<a href="<?php echo esc_url( get_permalink( $session ) ); ?>">
+			<?php echo get_the_title( $session ); // phpcs:ignore WordPress.Security.EscapeOutput ?>
```
iandunn: In this particular situation, rather than ignoring this, I think we should add `wp_kses_post()` or similar. 

Related: https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1538

----

In general, I think adding `phpcs:ignore` comments to the codebase should be rare. Seeing a lot of these comments in the codebase is a "smell" that our config file needs to be improved, and it has the effect of making the code harder to read.

I think we should only really use them in situations like [overriding `$post` in `get_contact_form_shortcode()`](https://github.com/WordPress/wordcamp.org/pull/34/files#diff-eed50fd407482588ef91f414ac7259fcR300), where it's necessary because of a true exception to the rule.

I think most of the time these comments tend to be used because a particular rule has false positives. In some cases, we can tweak the options in `phpcs.xml.dist` to allow the desired formatting. If that's not possible, though, then I'd rather see us just `<exclude>` the rule from the config file and rely on our own judgement to keep the code formatted properly, rather than litter the codebase with metadata that is essentially irrelevant at that layer.

@vedanshujain , @coreymckrill

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,186 @@
...
+	 * ]
+	 * @param props.height Height in pixels for image.
+	 * @param props.width Width in pixels for image.
+	 * @param props.className Classname for image element
+	 * @param props.alt Alt text for image
+	 * @param props.onChange Function callback for when Image URL is changed. Current URL will be passed as a parameter.
+	 */
+	constructor( props ) {
+		super( props );
+	}
```
iandunn: @vedanshujain , why is the `constructor()` needed when it's not doing anything? Is it just there to document the props? Would that be better done w/ `propTypes` ?

iandunn:
@vedanshujain it looks like there's some CSS issues w/ the Sponsors block in FireFox (not sure if you're aware or not, so posting just in case)

Screen Shot 2019-03-15 at 3 49 03 PM

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
```diff
@@ -0,0 +1,142 @@
...
+		super();
+
+		this.state = {
+			allSessionPosts      : null,
+			allSessionTracks     : null,
+			allSessionCategories : null,
+		};
+	}
+
+	componentWillMount() {
```
iandunn: @coreymckrill , @vedanshujain : `componentWillMount()`, `componentWillRecieveProps()`, and `componentWillUpdate()` are being deprecated in React 16.3 and removed in 17; would it be best to migrate away from them now?

https://blog.northcoders.com/react-componentwillmount-to-be-deprecated
https://github.com/WordPress/gutenberg/issues/7897#issuecomment-428242442
https://github.com/facebook/react/issues/7671

iandunn: ( note that there are lots of places where we use them, so if we change one we should change them all )

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
```diff
@@ -0,0 +1,142 @@
...
+			path: addQueryArgs( `/wp/v2/sessions`, ALL_POSTS_QUERY ),
+		} );
+		const allSessionTracks = apiFetch( {
+			path: addQueryArgs( `/wp/v2/session_track`, ALL_TERMS_QUERY ),
+		} );
+		const allSessionCategories = apiFetch( {
+			path: addQueryArgs( `/wp/v2/session_category`, ALL_TERMS_QUERY ),
+		} );
+
+		if ( this.isStillMounted ) {
```
iandunn: @coreymckrill , I'm probably missing something, but I don't see how this would ever be `false` at this point, so it seems like the condition is unnecessary.

Is it something to do with `apiFetch()` being asynchronous or a promise or something?

( note that there are a few other places where we do this, so if we change one we should change them all )

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
iandunn: er, Gutenberg is up to 16.8.4, so they're already deprecated.

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
```diff
@@ -0,0 +1,142 @@
...
+		const { mode } = this.props.attributes;
+
+		return (
+			<Fragment>
+				<SessionsBlockControls
+					icon={ SESSIONS_ICON }
+					{ ...this.props }
+					{ ...this.state }
+				/>
+				{ mode &&
```
iandunn: This feels like of unintuitive to me, like from the context I expect `mode` to be a boolean, but then it's name doesn't describe what it is, so I'm left confused.

I had to look through other parts of the code to realize it's a string that has an empty value until a mode like "all speakers" or "group of speakers" is selected.

I wonder if the code would be more descriptive/self-documenting/obvious if we gave `mode` an initial value of `placeholder`, and then changed things like this to read `'placeholder' !== mode` ?

@coreymckrill , @vedanshujain

iandunn:
I wonder if, for the sake of clarity, we should establish a convention for explicitly setting expectations for review comments?

[Brandon Savage suggests using `MIGHT`, `SHOULD`, and `MUST`](https://www.brandonsavage.net/the-pitfalls-of-code-review-and-how-to-fix-them/), which feels familiar to RFC conventions.

e.g.,

* `[MIGHT]` This might be easier to read if it were written like `{foo}`, but I can see it this way too. Just use your judgement which you think is best.
* `[SHOULD]` This would be more performant if we `{bar}`, but it's not a deal-breaker. If you have time go for it, but if we need to move on to higher priority things, it's not the end of the world.
* `[MUST]` This could be an XSS vector with a payload like `{baz}`. We can't merge this until that's mitigated.

What do y'all think, @vedanshujain , @coreymckrill ?

public_html/wp-content/mu-plugins/blocks/view/shared/featured-image.php
vedanshujain: Hmm, yeah thats a good idea. Will move it there.

public_html/wp-content/mu-plugins/blocks/view/shared/grid-layout.php
vedanshujain: this file does not have too much of a view, so we can move to `/include` folder. I will do that, meanwhile feel free to use these files.

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/edit.js
```diff
@@ -0,0 +1,124 @@
...
+const { addQueryArgs }        = wp.url;
+
+/**
+ * Internal dependencies
+ */
+import OrganizersBlockControls     from './block-controls';
+import OrganizersInspectorControls from './inspector-controls';
+import OrganizersToolbar           from './toolbar';
+
+export const ORGANIZERS_ICON = 'groups';
```
iandunn: @coreymckrill , @vedanshujain : what do you think about this?

It feel odd to me that in the other blocks the icon is duplicated in `index` and `edit`, instead of being DRY, so this is my attempt to fix that.

I first tried to define it in `index` and pass it to `edit`, since that feels more logical, but didn't see a good way. This works, but feels a bit odd that it's being passed from `edit` to `index`.

Here's the usage in `index`:

https://github.com/WordPress/wordcamp.org/pull/33/commits/6560472e4740813d7bbf66613a625d417220432f#diff-bb376325754a536e359c629d2a98226fR10

https://github.com/WordPress/wordcamp.org/pull/33/commits/6560472e4740813d7bbf66613a625d417220432f#diff-bb376325754a536e359c629d2a98226fR17

coreymckrill: I like this. We could do this for `LABEL` as well. And I agree it would be better in index than in edit.

I'm not sure I understand why putting it in `index` didn't work. Couldn't you just export it there and then do `import { ORGANIZERS_ICON } from './index.js';` in `edit`?

public_html/wp-content/mu-plugins/blocks/view/sessions.php
coreymckrill: I agree that we should minimize exceptions. In this particular case, I assumed that since `the_title` doesn't trigger the sniff, `get_the_title` shouldn't either. But it looks like that's not [the](https://developer.wordpress.org/reference/functions/get_the_title/#comment-2150) [case](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards/issues/1538).

So, I agree, `wp_kses_post` or `wp_kses_data` would be appropriate here.

This makes me wonder when it would ever make sense to use `the_title`, since it echoes the output before it can be modified further. Unless you added an escaping function to the `the_title` filter hook...

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
coreymckrill: Blerg.

coreymckrill: Yes, we should refactor now.

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
coreymckrill: I got the `this.isStillMounted` pattern from the Latest Posts block in Gute Core. I suspect the reasoning is that since it's working with promises, there's a chance that between the time that the fetch happens and when the promise is resolved, the block could have been deleted, and then it would be trying to manipulate variables that no longer exist.

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
coreymckrill: That would be fine with me.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/grid-layout/inspector-control.js
```diff
@@ -0,0 +1,52 @@
...
+class GridInspectorControl extends Component {
+
+	render() {
+		const { attributes, setAttributes } = this.props;
+		const { layout, grid_columns } = attributes;
+
+		if ( 'grid' !== layout ) {
+			return null;
+		}
+		const schema = DEFAULT_SCHEMA;
```
coreymckrill: [SHOULD] Since this is a shared component, this should look for values in `props` before using the static defaults, so that different values can be passed in.

iandunn:
`[SHOULD]` @vedanshujain , it'd be good to run `eslint` on the new code before committing parts of this to SVN, since it'd be cleaner fix the violations now than after it's merged.

public_html/wp-content/mu-plugins/blocks/package.json
```diff
@@ -43,7 +43,7 @@
 	],
 	"scripts": {
 		"start": "cross-env BABEL_ENV=development NODE_ENV=development webpack --watch",
-		"build": "cross-env BABEL_ENV=development NODE_ENV=development webpack",
+		"build": "cross-env BABEL_ENV=production NODE_ENV=production webpack",
```
iandunn: @coreymckrill , @vedanshujain : Is there any reason this was set to `development`? This is a huge performance increase.

public_html/wp-content/mu-plugins/blocks/webpack.config.js
```diff
@@ -17,6 +17,8 @@ const webpackConfig = {
 		minimize: true,
 	},
 
+	devtool: 'production' === NODE_ENV ? 'none' : 'cheap-module-eval-source-map',
```
iandunn: @vedanshujain , @coreymckrill : FYI.

This adds about `500ms` to the `watch` process, but seems worth it. [The only faster options](https://webpack.js.org/configuration/devtool/) would result in more difficult debugging experiences, but we can experience with them if the slower rebuilds have negative impacts on our workflow.

iandunn: It feels odd that we have lines 5, 14, and 20 all doing essentially the same thing w/r/t setting the environment var. I'd be ideal to have it set once (in 5), and then be able to use that in `14` and `20`, e.g.:

```js
mode: NODE_ENV,
devtool: 'production' === mode ? 'none' : 'cheap-module-eval-source-map',
```

Does that seem right to y'all as well?

I'm not sure that line 20 can reference `mode` though, since it won't be fully defined yet? Or maybe it needs to use `this.mode` or something?

coreymckrill:
@vedanshujain I still have a lot of concerns about the `FeaturedImage` component. I don't think it's usable in its current state. If you upload the latest **mu-plugins/blocks** files from this PR to your sandbox and then take a look at [this post on 2017.testing](https://2017.testing.wordcamp.org/wp-admin/post.php?post=11868&action=edit), I think you'll see what I mean.

* All of the logos have skewed aspect ratios
* None of the statically-defined size presets match the images that are currently being used
* Setting a width/height manually doesn't work because many of the logos have different aspect ratios

There are probably a lot of different approaches to making this work, but here's what I would suggest:

* I don't think we can try to control the exact dimensions of the images, since they might not all have the same aspect ratios. Instead, I think we should let the user pick from the list of WP-generated image size names, and then allow them to scale downward to a specific percentage or pixel width, while the heights are automatically calculated to maintain correct aspect ratios.
* As such, instead of a whole bunch of static predefs of various specific image sizes and aspect ratios, just populate the dropdown with the standard size names that all WP sites have: Thumbnail, Medium, Large, Full Size. Don't worry about the dimensions or aspect ratios of these--they might be customized in WP settings anyway.
* When a size is selected, the image URL for each displayed post can be determined based on what the REST API tells us is actually available (i.e. `_embedded.wp:featuredmedia[0].media_details.sizes`).
* My preference for scaling controls would be to have buttons for `25%`, `50%`, `75%`, and `100%` (similar to the Image block), plus an input for the actual width in pixels. We could show height too, but it would be readonly, since it would be calculated based on the set width. When a size is selected, the scale input would reset to `100%`, so that it would show the maximum width using that particular size.
* I think we should avoid setting image dimensions via inline styles. Otherwise it makes it harder to customize via the stylesheet with things like `width: 100%; height: auto;`. Seems like we should be able to feed the set width and calculated height into the `width` and `height` attributes of the img tag directly.
* The srcset issue is tricky. Maybe always us the Full size for that, but if it's not the same aspect ratio as the chosen size (if, for example, Thumbnail is chosen, which sometimes crops the image), maybe skip srcset altogether?

Let me know your thoughts.

vedanshujain:
@coreymckrill So I spent a lot (all) of time on this today and ended up implementing a lot of feedback that you mentioned, but also deviated from some stuff. Summarizing the points below:

1. I removed the height customization attribute, so from now on height will always be dependent on width to preserve the aspect ratio of the "full" sized image. Because of this, if you disable name and content, and only have logos, then different rows can have different height. I tried several approaches, including adding a config to specify max - height, but added complexity did not seem worth it.

2. I am planning to not have srcset attribute in the editor, instead we will always render "full" sized image. This is what Gallery block in core does. However, we will have srcset attribute rendered in actual post render, to optimize page and image load. All this behavior is still WYSIWYG as long as images are not cropped.

>we should be able to feed the set width and calculated height into the width and height attributes of the img tag directly.

Done for width attribute. I do not think we need to do it for height attribute as well.

> My preference for scaling controls would be to have buttons for 25%, 50%, 75%, and 100% (similar to the Image block), plus an input for the actual width in pixels

It seemed like there will be too many controls if we have % buttons as well as absolute image size. Also % options are sort of redundant because you can anyway set the image size. I have added `max-width` attribute to 100% though so it shouldn't cause issues on small screen sizes.

All these changes are in https://github.com/WordPress/wordcamp.org/pull/33/commits/1461fb1521a849fae52a3126d2f70ab69857f727https://github.com/WordPress/wordcamp.org/pull/33/commits/1461fb1521a849fae52a3126d2f70ab69857f727 . I have not fully tested the rendered post side yet, but feel free to do a preliminary review.

public_html/wp-content/mu-plugins/blocks/webpack.config.js
vedanshujain: Ideally we should not be minimizing files on local, instead of using source map.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -9,150 +9,56 @@ import classnames from 'classnames';
...
+	 * @param props.wpMediaDetails Available sizes of images in the format as returned by WP API. This is the `sizes` object inside `media_details` inside `wp:featuredMedia` object.
 	 * @param props.height Height in pixels for image.
 	 * @param props.width Width in pixels for image.
 	 * @param props.className Classname for image element
 	 * @param props.alt Alt text for image
-	 * @param props.onChange Function callback for when Image URL is changed. Current URL will be passed as a parameter.
 	 */
 	constructor( props ) {
-		super();
+		super( props );
```
iandunn: I intentionally removed `props` in 2e720880c751a70df2d48b6293e46dad77574e24. Is there a reason to add it back?

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/inspector-control.js
```diff
@@ -2,48 +2,33 @@
...
-	},
-	{
-		label : __( 'Medium - 16:9', 'wordcamporg' ),
-		value : '720x405',
-	},
-	{
-		label : __( 'Large - 16:9', 'wordcamporg' ),
-		value : '1280x720'
+		label : __( 'Large', 'wordcamporg' ),
+		value : '1024',
```
iandunn: Should we grab these dynamically from Core rather than hardcoding them?

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/inspector-control.js
```diff
@@ -2,48 +2,33 @@
  * WordPress dependencies.
  */
 const { Component } = wp.element;
-const { PanelBody, PanelRow, TextControl, BaseControl, SelectControl } = wp.components;
+const { PanelBody, PanelRow, TextControl, BaseControl, SelectControl, ToggleControl } = wp.components;
 const { __ } = wp.i18n;
+import './style.scss';
 
 /**
  * Not sure if these sizes are actually useful.
```
iandunn: `[SHOULD]` We should be sure before merging this, and update the comment to describe _what_ the var is (or just remove if unnecessary).

public_html/wp-content/mu-plugins/blocks/webpack.config.js
iandunn: 🤔 

I love the idea of not needing to add the extra time to the rebuild process, but I'm not sure if debugging would be too hard. Currently the code gets minified, but at least in FireFox it shows it formatted, so that isn't a problem for me. The bigger problem is that the code you see in the browser is the _generated_ code, rather than the original _source_, which has a few problems:

1) It's full of verbose/obscure var names, polyfills, transpiled alternatives, etc, so it's difficult to read and understand
2) It's inconvenient to manually trace line in the generated source back to the line in the original source.

There are advantages to that, though, in the case where a bug _results from_ the transpilation, but that's probably fairly rare.

I could go either way. @vedanshujain , @coreymckrill , what has your debugging experience been like so far? Have you found it difficult to track down errors? If not, then the source maps may be unnecessary. They seemed helpful to me, but that could just be because I'm newer to this.

If we do remove them, then maybe an alternative idea would be to add a 3rd build mode for debugging, and in that mode the source maps would be generated. e.g,

```diff
"scripts": {
	"start": "cross-env BABEL_ENV=development NODE_ENV=development webpack --watch",
	"build": "cross-env BABEL_ENV=production NODE_ENV=production webpack",
+	"debug": "cross-env BABEL_ENV=development NODE_ENV=debug webpack",
	"lint:js": "wp-scripts lint-js ."
}
```

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/block-content.js
```diff
@@ -0,0 +1,178 @@
...
+				wpMediaDetails={featuredImageSizes}
+				alt={sponsorPost.title.rendered}
+				attributes={attributes}
+			/>
+			}
+			{ ( show_desc || show_desc === undefined ) &&
+			<ItemHTMLContent
+				className={ classnames( 'wordcamp-sponsor-content' ) }
+				content={ sponsorPost.content.rendered.trim() }
+			/>
```
iandunn: `[SHOULD]` It looks like the design calls for excerpts support, similar to Speakers. Did you leave that out because the post type doesn't support it? It seems like we should enable support in the post type instead.

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/edit.js
```diff
@@ -0,0 +1,90 @@
...
+			per_page: MAX_PAGE,
+			_embed: true
+		};
+
+		this.setState(
+			{
+				sponsorPosts: apiFetch( { path: addQueryArgs( '/wp/v2/sponsors', sponsorQuery ) } ),
+				sponsorLevels: apiFetch( { path: addQueryArgs( '/wp/v2/sponsor_level', sponsorLevelQuery ) } ),
+				siteSettings: apiFetch( { path: addQueryArgs( '/wp/v2/settings', {} ) } ),
+			}
```
iandunn: `[MIGHT]` @vedanshujain why is this done differently than the Speaker's block? I'm not implying it's wrong, I just want to understand the reason why it's different. If this is better, then should be do it in the other blocks as well, for consistency?

cc @coreymckrill

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/edit.js
```diff
@@ -0,0 +1,124 @@
...
+
+	componentWillMount() {
+		this.isStillMounted = true;
+
+		const allSpeakerPosts = apiFetch( {
+			path: addQueryArgs( `/wp/v2/organizers`, ALL_POSTS_QUERY ),
+		} );
+		const allSpeakerTerms = apiFetch( {
+			path: addQueryArgs( `/wp/v2/organizer_team`, ALL_TERMS_QUERY ),
+		} );
```
iandunn: `[MIGHT]` I'm getting an `Unhandled promise rejection` error here. It's probably just because I haven't finished setting up the API endpoint the way that the block expects, but the underlying issue is that it feels like a code smell that we're not catching rejections.

https://stackoverflow.com/questions/40500490/what-is-an-unhandled-promise-rejection

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

It seems equivalent to doing a `try` without a `catch` in PHP. Should we add `catch` everywhere we use promises?

@coreymckrill , @vedanshujain

public_html/wp-content/mu-plugins/blocks/webpack.config.js
coreymckrill: With FF Developer's script debug console, I've usually been able to decipher the generated code and figure out which line of the source is causing the bug. I'm not opposed to source maps though. I just never took the time to figure out how to set them up and get them to work right...

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
coreymckrill: I don't think it works correctly if you don't include `props` as a parameter in `super()`...

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/edit.js
iandunn: Huh, I guess you can. For some reason that didn't work the first time I tried it, but I must have done it wrong. Thanks!

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/edit.js
coreymckrill: In the Sessions block I added `catch` in the `SessionsSelect` component, because that's where the promises ultimately get resolved.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,86 @@
...
+		return widestImage;
+	}
+
+	/**
+	 * Renders FeaturedImage component.
+	 *
+	 * @returns {*}
+	 */
+	render() {
+		const { className, alt, attributes } = this.props;
```
iandunn: @vedanshujain, @coreymckrill  it looks like Sponsors is passing in the attributes, but Sessions isn't, so sessions has a fatal error.

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/edit.js
coreymckrill: +1 for consistency, one way or the other.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/inspector-control.js
coreymckrill: > Should we grab these dynamically

That was my suggestion above. Is there a blocker to doing this, or some other reason it's still statically defined?

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/inspector-controls.js
```diff
@@ -0,0 +1,90 @@
...
+							onChange={ ( value ) => setAttributes( { show_desc: value } ) }
+						/>
+					</PanelRow>
+					<PanelRow>
+						<SelectControl
+							label = { __( 'Sort by', 'wordcamporg' ) }
+							options = { sortOptions }
+							value = { sort_by || 'name_asc' }
+							onChange={ ( value ) => setAttributes( { sort_by: value } ) }
+							help = { __( 'Select whether to sort by name or sponsor level. Order of sponsor level can be configure by going to Sponsor -> Order Sponsor Levels admin menu.') }
```
iandunn: `MIGHT`

> Select whether to sort by name or sponsor level.

This part seems redundant, since it's obvious from the options in the dropdown.

>  Order of sponsor level can be configure by going to Sponsor -> Order Sponsor Levels admin menu.') }

Can we provide a direct link here, or does Gutenberg only allow text? If it's the latter, it might be good to open an issue suggesting they allow links for this type of use case.

We could probably also make this more succinct in order to make it easier to skim, and clutter the UI less. Maybe something like,

> [Configure sponsor levels](#).

...if a link is possible. It should probably open in a new tab, as much as i hate that.

Or,

> Configure sponsor levels from the Sponsor -> Order Sponsor Levels page.

...if a link isn't if possible.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
iandunn: It seems to work fine for me, and everything I read suggests that it's not necessary. Do you see anywhere that it doesn't work?

public_html/wp-content/mu-plugins/blocks/webpack.config.js
vedanshujain: There have been some instances where instead of where the error message came from, I have been presented with a link which make me go to a webpage to see the error message.
Other than this though debugging experience has been good.

iandunn: hm, ok, I'll switch it to add the `debug` mode instead, then, so that we can get good source maps when we need them, but that it won't slow down the regular rebuild process.

vedanshujain: I would keep that as a low priority though. This does not cause much pain.

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/block-content.js
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	return image;
+}
+
+function SessionDetails( { session, show_meta, show_category } ) {
+	let meta, metaContent, category;
+	const terms = get( session, '_embedded[\'wp:term\']', [] ).flat();
+
+	if ( show_meta ) {
```
iandunn: `[SHOULD]` This condition feels like it's big enough that it should be modularized into its own function.

The same goes for `if ( show_category && session.session_category.length ) {` below.

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
```diff
@@ -0,0 +1,142 @@
...
+		per_page           : MAX_POSTS, // -1 is not allowed for per_page.
+		_embed             : true,
+		context            : 'view',
+		_wcpt_session_type : 'session',
+	};
+
+	if ( 'session_time' !== sort ) {
+		const [ orderby, order ] = split( sort, '_', 2 );
+		args.orderby = orderby;
+		args.order = order;
```
iandunn: `[SHOULD]` I mentioned this before in a regular comment, but I should have put that in a code comment so that it can be tracked as unresolved.

We should run ESLint and PHPCS before merging any commits from this PR.

iandunn: We should probably run it twice, actually. Once for new code, and once for existing code. That way we can keep the two commits separate during the merge to SVN. The commit on new code could be squashed into the others, but the commit on existing code should be kept separate, so that it doesn't add a ton of diff noise.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
coreymckrill: My read is that if the parent constructor needs any of the parameters being fed to the child, then `super` should include it.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super

There are probably some situations where that isn't the case, so it doesn't affect anything to omit it. But I would argue that for consistency's sake we should just always include it, unless there's a situation where we specifically _don't_ want the parent constructor to get the same parameters, in which case we would want to explain that in a code comment.

public_html/wp-content/mu-plugins/blocks/view/sessions.php
coreymckrill: Fixed.

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/block-controls.js
```diff
@@ -0,0 +1,345 @@
...
+
+				<SponsorBlockContent
+					selectedPosts={ selectedPosts }
+					sponsorTermOrder = { sponsorTermOrder }
+					{...this.props}
+				/>
+
+				{'all' !== mode &&
+					<Placeholder
+						icon='heart'
```
iandunn: `[MUST]` - @vedanshujain , what's the reason that this is done differently than Speakers? Here it's hardcoded instead of pulling in the `icon` from `this.props`.

It seems like it'd be better to have it DRY like Speakers, but the larger issue is the inconsistency. I think in these cases we need to discuss the two approaches, come to a consensus on the best one, and update all the blocks to use it, so that everything is consistent.

In cases where something is clearly better and it's unlikely that there'd be any disagreement, it's also totally fine to just go ahead and adopt it (since we can always revert if there is a disagreement and we change our minds), but it's still important to update the other components to be consistent, and then give everyone a heads up.

In this case, I'm working on updating the icons per https://github.com/WordPress/wordcamp.org/pull/33#discussion_r267058114, so I'm gonna go ahead and change this to be consistent w/ the others. If there's a reason to change them all to this approach, though, we can definitely start a discussion about that.

public_html/wp-content/mu-plugins/blocks/webpack.config.js
vedanshujain: Opened https://github.com/WordPress/wordcamp.org/issues/37 for this

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: Resolving this comment. Please re-open if any further discussion is needed.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/inspector-control.js
vedanshujain: Added as part of https://github.com/WordPress/wordcamp.org/issues/36

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/block-content.js
vedanshujain: Opened new issue for this https://github.com/WordPress/wordcamp.org/issues/38

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/edit.js
vedanshujain: @iandunn you will probably get this when request fails, I am not sure if we can do anything inside `catch` statement which will fix it and is not complicated. So user would have to reload their page in case this happens. 
Please feel free too add a new issue for our standard behavior in case network call fails (maybe display a notification on top of editor?)

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: this is fixed

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/inspector-controls.js
vedanshujain: Added in https://github.com/WordPress/wordcamp.org/issues/41

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
vedanshujain: Added PR for this https://github.com/WordPress/wordcamp.org/pull/42

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/block-controls.js
vedanshujain: Opened issue for this https://github.com/WordPress/wordcamp.org/issues/43

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/edit.js
vedanshujain: So this is so that I can use results of same API call for rendering options in drop down, as well as rendering posts. Also, `withSelect` feels a bit opaque to me (it doesn't have good documentation as of now) so I wanted to avoid it if I can.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/inspector-control.js
vedanshujain: I was not able to find a good place in front-end where I can fetch these from. Image block in core seems to generate these from the image itself (which can have different sizes per image). So we cannot use any of the image data to render options for all images.

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
```diff
@@ -47,24 +47,28 @@ function SpeakerSessions( { speaker, tracks } ) {
 									{ decodeEntities( session.title.rendered.trim() ) || __( '(Untitled)', 'wordcamporg' ) }
 								</a>
 								<span className="wordcamp-speaker-session-info">
-									{ session.session_track.length &&
+									{ ( session.session_track.length && Array.isArray( tracks ) ) &&
```
iandunn: `[MIGHT]` 🤔 , this feels like kind of a sideways way of handling this. Would it be better to have something more explicit, like this?

```js
function render {
    if ( promise not resolved ) {
        return;
    }
}
```

( or maybe even do it earlier in the lifecycle? )

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/edit.js
iandunn: Good idea, moved to https://github.com/WordPress/wordcamp.org/issues/46

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
iandunn: Marking as unresolved because right now it's inconsistent, most of the constructors don't have it, but this one does, so we should decide one way or the other and make them all consistent.

It seems like, so far, it's not needed in 100% of the situations, so it feels more logical to leave it off by default, and then pass it in when necessary.

It seems like Andrew [felt like it was unnecessary](https://github.com/WordPress/wordcamp.org/pull/13#discussion_r249476093), so this might make us more consistent w/ how Gutenberg does it. But actually, it looks like [G typically calls it with `...arguments`](https://github.com/WordPress/gutenberg/search?l=JavaScript&q=super%28), so maybe we should do that always instead of `props`?

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
coreymckrill: As far as I can tell, when using `withSelect` there isn't a promise exposed anywhere that you can resolve, unfortunately. You just have a prop (tracks in this case) that is `null` until the underlying promise is resolved, at which point it is an array of data.

This might end up being moot if we [move away](https://github.com/WordPress/wordcamp.org/pull/33#discussion_r268608584) from using `withSelect`.

iandunn:
All code that hasn't been merged to `master`, and all unresolved conversations have been added to #44, so closing this.

iandunn:
( Note that we should not delete this branch until #44 is resolved, just in case that would delete the comments here that we still need to refer to in order to resolve all the remaining conversations )

public_html/wp-content/mu-plugins/blocks/webpack.config.js
iandunn: Done in 508984cce4e5a65bc65f907ad369ee7e64928f9c. Run `npm run debug` to enable the source maps.

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/edit.js
coreymckrill: Related: #39

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
coreymckrill: Andrew's comment was actually about binding the render function, further down, not about the `super` call.

But, I found this: https://stackoverflow.com/questions/30571875/whats-the-difference-between-super-and-superprops-in-react-when-using-e

...which would indicate that including `props` in the `super` call is only necessary if you need to refer to `this.props` elsewhere within the constructor method. **But**, it also points out that in the React docs, it says [this](https://reactjs.org/docs/state-and-lifecycle.html#adding-local-state-to-a-class):

> Class components should always call the base constructor with `props`.

iandunn: haha, yeah, that's the thing I read that made me feel like we should remove it :)

So it seems like it's pretty intangible. For the sake of not bikeshedding it, I'll just revert 2e72088 and we can move on to more important things :)

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,185 @@
...
+			}
+		}
+
+		if ( selectedImage ) {
+			return selectedImage.source_url;
+		} else if ( smallestImage ) {
+			return smallestImage.source_url;
+		} else {
+			// All images are small, lets just return the largest image url.
+			return sizes[ sizes.length - 1 ].source_url;
```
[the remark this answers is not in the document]
vedanshujain: this was fixed

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/edit.js
iandunn: Possibly related: https://github.com/WordPress/gutenberg/pull/11300#issuecomment-437925427

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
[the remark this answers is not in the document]
iandunn: When saying something is fixed, can you leave the commit number where it was fixed? Otherwise I have to dig through the logs to find it.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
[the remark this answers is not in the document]
iandunn: The corresponding checkbox for this is marked off in #44, but this is still marked as unresolved. If it's resolved, can you leave a comment w/ the commit that resolved it, and then mark as resolved?

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/block-content.js
iandunn: The corresponding checkbox for this is marked off in #44, but this is still marked as unresolved. If it's resolved, can you leave a comment w/ the commit that resolved it, and then mark as resolved?

public_html/wp-content/mu-plugins/blocks/assets/src/speakers/block-content.js
coreymckrill: We are rethinking the way fetched data is handled in #39, so I'm going to resolve this discussion in favor of that.

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
coreymckrill: Opened #63 to address this.

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
coreymckrill: This approach may change when we address #63. Let's continue the conversation there.

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/block-content.js
coreymckrill: https://github.com/WordPress/wordcamp.org/pull/44/commits/15e377aa893c986810bd40a1b784cb975bb476c9

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: This code was removed. Resolving

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: This code was removed. Resolving.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: Tracking this in https://github.com/WordPress/wordcamp.org/issues/36. Resolving this conversation.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
```diff
@@ -0,0 +1,185 @@
...
+		if ( imageURL !== this.currentImageURL ) {
+			this.currentImageURL = imageURL;
+			if ( typeof( onChange ) === 'function' ) {
+				onChange( imageURL );
+			}
+		}
+		return(
+			<img
+				className={ classnames( 'featured-image', className ) }
+				src = { imageURL }
```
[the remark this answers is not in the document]
vedanshujain: Tracking this in https://github.com/WordPress/wordcamp.org/issues/36 . Resolving this conversation.

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/edit.js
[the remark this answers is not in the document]
vedanshujain: We decided that we will use `withSelect` instead of `apiFetch`, so resolving this conversation. Tracking for implementing `withSelect` is here https://github.com/WordPress/wordcamp.org/issues/65.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: This code was removed because we changed our approach. Resolving this conversation.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/inspector-control.js
vedanshujain: Tracking this in https://github.com/WordPress/wordcamp.org/issues/36. Resolving this conversation.

public_html/wp-content/mu-plugins/blocks/assets/src/sponsors/edit.js
vedanshujain: We have decided to use `withSelect` because it can cache request across blocks. Its being tracked in https://github.com/WordPress/wordcamp.org/issues/65 . Resolving this conversation.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/grid-layout/inspector-control.js
vedanshujain: I did not added this because we only support minimum 2 and maximum 4, so it makes sense to not make this configurable.
I am marking this is as resolved #44 to make everything more manageable, but please feel free to uncheck it there if you think we should discuss this more.

public_html/wp-content/mu-plugins/blocks/view/sessions.php
vedanshujain: Added this comment to help any reviewer parsing the changes. Resolving now.

public_html/wp-content/mu-plugins/blocks/package.json
vedanshujain: It was set to development because IIRC it produced more readable digested that way. Now that we have a separate `debug` setting, its not relevant anymore and I am resolving this conversation.

public_html/wp-content/mu-plugins/blocks/view/shared/featured-image.php
vedanshujain: Being fixed in https://github.com/WordPress/wordcamp.org/pull/66 .

public_html/wp-content/mu-plugins/blocks/view/shared/grid-layout.php
vedanshujain: This is being fixed in #66

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/edit.js
iandunn: Done in 59cdcb645cd6dd62397ae527f2eaab5e53fdf104

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
vedanshujain: This constructor is not empty anymore and initializes `state`. Resolving this conversation.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/grid-layout/inspector-control.js
coreymckrill: This kind of ties in with #57. If the block schema gets defined in PHP and loaded via `add_script_data`, like in the other blocks, it would be simple to use the schema values instead of a static variable. That would potentially make it DRYer as well. I'll add a comment about that over in the issue so we can close up this thread.

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
iandunn: Thanks!

public_html/wp-content/mu-plugins/blocks/assets/src/shared/featured-image/index.js
iandunn: Done in 494f7f32c4b2ec095b26cd184f5479c1a3ad1249

PR #34, opened 2019-03-14
Coming soon mode: Also support jetpack/contact-form block.

vedanshujain:
We already support `contact-form` shortcode which is used to rendered contact us form when Coming Soon mode is enabled for a site.
This patch also adds support for jetpack/contact-form block which can also be used to create a contact-us form.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
+++ b/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
@@ -263,7 +263,7 @@ class WordCamp_Coming_Soon_Page {
 	}
 
 	/**
-	 * Retrieve the contact form shortcode string
+	 * Loop through all pages and renders first contact-us form or contact-us block
 	 *
 	 * We can't just create an arbitrary shortcode because of https://github.com/Automattic/jetpack/issues/102. Instead we have to use a form that's tied to a page.
 	 * This is somewhat fragile, though. It should work in most cases because the first $page that contains [contact-form] will be the one we automatically create
@@ -274,8 +274,8 @@ class WordCamp_Coming_Soon_Page {
 	 * @return string|false
 	 */
 	public function get_contact_form_shortcode() {
-		$contact_form_shortcode = false;
-		$shortcode_regex        = get_shortcode_regex();
+		$contact_form_content = false;
+		$shortcode_regex      = get_shortcode_regex();
 
 		$all_pages = get_posts( array(
 			'post_type'      => 'page',
@@ -285,6 +285,10 @@ class WordCamp_Coming_Soon_Page {
 		foreach ( $all_pages as $page ) {
 			preg_match_all( '/' . $shortcode_regex . '/s', $page->post_content, $matches, PREG_SET_ORDER );
 
+			if ( $contact_form_content ) {
+				break;
+			}
+
 			foreach ( $matches as $shortcode ) {
 				if ( 'contact-form' === $shortcode[2] ) {
 					global $post;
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
```diff
@@ -263,7 +263,7 @@ public function get_dates() {
 	}
 
 	/**
-	 * Retrieve the contact form shortcode string
+	 * Loop through all pages and renders first contact-us form or contact-us block
```
iandunn: I think the short description is meant to say _what_ a function does, rather than _how_ it does it, so I'd leave out the bit about looping through the pages, since that's just an implementation detail. Maybe something more along the lines of "Retrieve the contact form shortcode/block output" ?

public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
```diff
@@ -293,15 +297,31 @@ public function get_contact_form_shortcode() {
...
+			foreach( $blocks as $block ) {
+				if ( 'jetpack/contact-form' === $block['blockName'] ) {
+					global $post;
+					$post = $page;
+					setup_postdata( $post );
+					$contact_form_content = render_block( $block );
+					wp_reset_postdata();
+					break;
+				}
+			}
```
iandunn: It might be a bit more readable to structure it like this:

```php
foreach ( $all_pages as $page ) {
	if ( has_block( 'jetpack/contact-form' ) {
		// get content from block
	} elseif ( has_shortcode( 'contact_form' ) {
		// get content from shortcode
	}

	if ( $contact_form_content ) {
		break;
	}
}
```

I think the original code may have been written before `has_shortcode()` existed, so it'd be nice to update it to take advantage of the function, since it'll make the structure of the code more obvious and descriptive.

Placing the `break` after checking the block/shortcode prevents an extra unnecessary iteration of the loop. That'd be a useless micro-optimization from a performance perspective, but it feels a bit more logical to me than, since it happens right after the condition might be met.

public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
```diff
@@ -92,6 +92,7 @@ public function render_dynamic_styles() {
 			return;
 		}
 
+		// phpcs:ignore WordPress.PHP.DontExtract.extract_extract -- Not sure whats the alternative to this could be.
```
iandunn: 🤔 

That's a good question.

I'd be tempted to remove the use of `extract`, since it's a violation of the guidelines. https://core.trac.wordpress.org/ticket/22400 has more discussion and a link to an article.

We don't have to do that now, though. For now I'd just remove the `phpcs:ignore`, and let the warning remain so we can fix it in the future.

On the other hand, this does kind of feel like an appropriate usage of `extract()`, since it keeps the logic DRY (it's used by multiple templates). One way to keep it and avoid the problems with it would be to add phpdoc to declare the variables at the point where they're `extract()`ed, but then those docs wouldn't be DRY.

🤔 

I don't have a good answer. 🤷‍♂️ 

Maybe it's fine to leave the `phpcs:ignore`? Whatever you think is best works for me.

iandunn (approved):
Left a few suggestions, but they're not blockers or anything. Good work!

public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
vedanshujain: I guess in this case I would keep the Loop through pages bit as well, because this methodology is a but unusual and unexpected. Just saying `Retrieve the contact form shortcode/block output` implies that there is a dedicated shortcode or block somewhere just for this contact form. Explaining that we loop through pages could clearly help someone in debugging issues related to this function.

public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
vedanshujain: I am going to keep this for now. But lets re visit when we have some time. Adding a todo here.

vedanshujain:
This is merged now https://meta.trac.wordpress.org/changeset/8463

public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
iandunn: 👍

PR #45, opened 2019-03-25
WordCamp: Change sponsor order to reflect their donation status

vedanshujain:
In acknowledgement message to sponsor on purchase successful page and ticket receipt email while purchasing a ticket, we were showing sponsors who were sponsoring in that region in random order each time.
This patch changes that bahavior to:
1. Recognise sponsor irrespective of the region that they sponsor in.
2. Order sponsor names w.r.t. their donation level (Gold, Silver and then Bronze) and sponsor region.


vedanshujain:
This is merged https://meta.trac.wordpress.org/changeset/8529

PR #48, opened 2019-03-26
WIP: WordCamp: Add docker setup to ease development flow

vedanshujain:
This is adds initial docker configuration for running WordCamp meta environment. Right now expectation is to have a MySQL running on host and configured in wp-config. Eventually MySQL will be added as part of docker configuration.

These are pending tasks before its ready to be merged
- [ ] Add MySQL configuration with updated database
- [ ] Move files to new `provisioning` folder
- [ ] Add inbuilt wp-config file
- [ ] Add linting support

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,11 @@ RUN cd ~ && openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout wordcamp
 RUN cp ~/wordcamp.crt /etc/ssl/certs/wordcamp.crt
 RUN cp ~/wordcamp.key /etc/ssl/private/wordcamp.key
 
+# Install wkhtmltopdf for camptix invoices. See https://stackoverflow.com/a/38336153/1845153
+RUN curl -L https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz > wkhtmltox.tar.xz
+RUN tar xvf wkhtmltox.tar.xz
+RUN mv wkhtmltox/bin/wkhtmlto* /usr/bin/
+
 # Install cli
 RUN curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
 RUN chmod +x wp-cli.phar
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,43 @@
+FROM wordpress:5.0.3-fpm
+
+RUN mv /usr/src/wordpress /usr/src/public_html
+WORKDIR /usr/src/public_html
+
+RUN apt-get update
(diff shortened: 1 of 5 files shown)


vedanshujain:
Eh, this is still a draft PR, I accidentally marked it as regular and do not see any option to convert it now.

Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM wordpress:5.0.3-fpm
```
iandunn: Is there a use the latest branch here, instead of pinning to a specific tag?

vedanshujain: So the idea is to match this with whatever we have in production as close as possible, so that we can reduce the number of things that we have to test on sandbox. I will update this, but lets keep it manual. Whenever we update production version, we will push a commit here as well.

Dockerfile
```diff
@@ -13,6 +13,11 @@ RUN cd ~ && openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout wordcamp
 RUN cp ~/wordcamp.crt /etc/ssl/certs/wordcamp.crt
 RUN cp ~/wordcamp.key /etc/ssl/private/wordcamp.key
 
+# Install wkhtmltopdf for camptix invoices. See https://stackoverflow.com/a/38336153/1845153
+RUN curl -L https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz > wkhtmltox.tar.xz
+RUN tar xvf wkhtmltox.tar.xz
+RUN mv wkhtmltox/bin/wkhtmlto* /usr/bin/
+
```
iandunn: The Meta Environment does it a little different:

https://github.com/WordPress/meta-environment/blob/844bef6/wordcamp.test/provision/vvv-init.sh#L23-L29

It'd probably be good to install that latest version via `apt`, since that's probably closer to production. I don't know if the `xvfb-run` stuff is really relevant or not, but it might be.

vedanshujain: I would rather keep this version locked with whatever we have on production. Also installing with `apt` unfortunately does not comes with few libraries that are needed for this to be run inside docker (qt). 
Also, I'd really avoid a display server just for this if we can, because wkhtmltopdf have its own rendering engine. I think it will be fine without it, but lets add `xvfb` if we see any problems.

iandunn: 👍

tomjn:
I'd recommend avoiding mailcatcher in favour of mailhog, it's much easier to setup and to install, and less prone to breakage. The UI is also much nicer and has push notification support

iandunn:
Yeah, that's what I use locally. The one downside I've run into is that I've never been able to get it configured to properly release messages, for situations when you want them delivered.

iandunn:

Sorry for the hassle :(

@vedanshujain , can you re-submit this?

Issue #41, opened 2019-03-25
Sponsor Block: Improve help text for order by control


vedanshujain:
> Select whether to sort by name or sponsor level.

This part seems redundant, since it's obvious from the options in the dropdown.

>  Order of sponsor level can be configure by going to Sponsor -> Order Sponsor Levels admin menu.') }

Can we provide a direct link here, or does Gutenberg only allow text? If it's the latter, it might be good to open an issue suggesting they allow links for this type of use case.

We could probably also make this more succinct in order to make it easier to skim, and clutter the UI less. Maybe something like,

> [Configure sponsor levels](#).

...if a link is possible. It should probably open in a new tab, as much as i hate that.

Or,

> Configure sponsor levels from the Sponsor -> Order Sponsor Levels page.

...if a link isn't if possible.

_Originally posted by @iandunn in https://github.com/WordPress/wordcamp.org/pull/33_

PR #49, opened 2019-03-27
WordCamp Sponsor Blocks: Change help text to make it more clearer.

vedanshujain:
In `Order By` sponsor level control, help text has a bit of redundant instruction which is already implied by the SelectControl label.
This fixes https://github.com/WordPress/wordcamp.org/issues/41


vedanshujain:
this is merged

PR #52, opened 2019-03-28
Add Organizers block

iandunn:


iandunn:
I haven't pushed the Organizers block itself yet, but I did debounce `AvatarSizeControl`. I had to drop the `wait` down to `10` milliseconds in order to preserve the UX, though, so it may not be all that useful. It also feeds a little janky (see comments in 91e2217). What do y'all think?

public_html/wp-content/mu-plugins/blocks/assets/src/shared/avatar-size/index.js
```diff
@@ -41,64 +43,82 @@ const sizePresets = [
...
-					isLarge
-					isDefault
-					onClick={ () => onChange( Number( initialPosition ) ) }
-				>
-					{ __( 'Reset', 'wordcamporg' ) }
-				</Button>
-			</div>
+			// converting from props to state just to debounce feels like maybe it's the wrong way?
+			// maybe related to https://github.com/vasanthk/react-bits/blob/master/anti-patterns/01.props-in-initial-state.md
+			// ^ might also mean that tracking value in state is also wrong?
```
coreymckrill: Would it work to just put the `debounce` call in the RangeControl? Like:

```
<RangeControl
	...
	onChange={ debounce( this.onChange, 10 ) }
	...
</RangeControl>
```

Tracking the value in state seems ok (we're doing it in `ItemSelect` as well), but this does seem a bit overkill...

iandunn: I tried that first, but then the slider element doesn't update, which would be a really bad UX.

iandunn:
The CSS in be495d8 may not be the best way to fix the problem. It feels a bit wrong for a child component like this to know about the parent structure, but it needs to be shared across all blocks (at least the ones we have so far), so maybe this should move to a new "common" CSS file that gets included by all the blocks?

I'm guessing y'all didn't run into this problem, but I was seeing it in Firefox and Chrome, for Organizers and Speakers blocks.

iandunn:
@coreymckrill , did your 👍 mean that you agree that it should move to a "common" CSS file?

coreymckrill:
Sorry for the ambiguous emoji usage. I think the approach in this commit is fine. I actually don't think we need to create a new "common" CSS file, at least not yet. This style is already part of a shared component, and it all gets compiled into one stylesheet ultimately anyway.

iandunn:
Thanks :)

iandunn:
I went ahead and merged this in cc534712ec28380b34df98da1a561c79f5fc6c90 so that we can get design/user feedback, but the code still needs review.

Since this code is already merged, you'll need to [compare 245d05a to db2db05](https://github.com/WordPress/wordcamp.org/compare/db2db0522d052336a76a2f678dfe1a3b6d32176c...245d05a) to see the changes that need review.

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/organizers-select.js
```diff
@@ -0,0 +1,199 @@
...
+	buildSelectOptions( mode ) {
+		const options = [];
+
+		const labels = {
+			wcb_organizer      : __( 'Organizers', 'wordcamporg' ),
+			wcb_organizer_team : __( 'Teams',      'wordcamporg' ),
+		};
+
+		for ( const type in this.state ) {
+			if ( ! this.state[ type ].length ) {
```
coreymckrill: PhpStorm gives me an inspection warning here because there's not a `hasOwnProperty` check. The purpose of that check is to make sure you're only iterating over properties in `this.state` that you have explicitly set, and not ones that are inherited from the prototype. I'm not sure that's actually a big deal here, but you can see how I mitigated the inspection warning in **sessions-select.js**.

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/block-content.js
```diff
@@ -0,0 +1,74 @@
...
+								</Disabled>
+							</div>
+						}
+
+						{ ( 'none' !== content ) &&
+							<ItemHTMLContent
+								className={ classnames( 'wordcamp-organizer-content-' + decodeEntities( content ) ) }
+								content={  'full' === content ? post.content.rendered.trim() : post.excerpt.rendered.trim() }
+								link={ (   'full' === content || excerpt_more ) ? post.link : null }
+								linkText={ 'full' === content ? __( 'Visit session page', 'wordcamporg' ) : __( 'Read more', 'wordcamporg' ) }
```
coreymckrill: Forgot to change `session` here.

coreymckrill (approved):

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/organizers-select.js
iandunn: Yeah, IIRC I intentionally left that out, because it felt like it added clutter without a practical benefit, since in this case there shouldn't be any enumerable properties coming from the prototype (and `for...in` ignores non-enumerable ones).

I tried using `for ( const type in Object.getOwnPropertyNames( this.state ) )` for a cleaner approach, but it didn't work for some reason. It might if I play with it a bit more, though, I might have just being making a basic mistake.

I kinda feel like that inspection is over-zealous, **but** I also feel like consistency is important, and triggering inspections hurts devex by distracting us from the code, so in this case it's probably better to add the `hasOwnProperty()` check if `getOwnPropertyNames()` won't work.

_Side note: It might be nice to share some basic PhpStorm config for things like inspections, but it seems like [PhpStorm makes that way more trouble than it's worth](https://www.reddit.com/r/PHP/comments/al87zr/how_do_you_guys_handle_versioning_of_the_phpstorm/)._

iandunn: Ah, yeah, I just need to use `Object.getOwnPropertyDescriptors()` instead :)

It doesn't look like IE11 supports that. Babel might transpile it into something more primitive, or provide a polyfill or something. I'll have to setup a VM to test.

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/index.js
```diff
@@ -6,10 +6,11 @@ const { __ } = wp.i18n;
 /**
  * Internal dependencies
  */
-import { edit }            from './edit.js';
-import { ORGANIZERS_ICON } from './edit.js';
+import { edit }            from './edit';
 
 export const name = 'wordcamp/organizers';
+export const LABEL           = __( 'Organizers', 'wordcamporg' );
+export const ORGANIZERS_ICON = 'groups';
```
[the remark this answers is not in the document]
iandunn: Fixed in c2784116387aab1a8269d455d9b1d47fc2cda786

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/organizers-select.js
iandunn: Yup, works in IE11. Fixed in 2809c1e4b080c8e8798f952f012d736b4ee5aa3a.

public_html/wp-content/mu-plugins/blocks/assets/src/organizers/block-content.js
iandunn: Fixed in 5f405708f3a6e6a8717f53f9b2539b9841bd6366

iandunn:
Resolved all the feedback, so closing this out.

PR #53, opened 2019-03-28
Travis js lint

vedanshujain:


vedanshujain:
For whatever reason, Travis is taking too much time to run a node build that its making debugging really ineffective. Punting this for, may take a look again after some time.

iandunn:

Sorry for the hassle :(

@vedanshujain , can you re-submit this?

PR #72, opened 2019-04-09
Blocks: Add shared definitions

coreymckrill:
Many of our blocks share several attribute definitions in their schemas, as well as several sets of option label/value pairs. This provides a centralized way to load those definitions to make our codebase DRYer, while still allowing the flexibility for blocks to define their own attributes/options, and even modify the shared definitions after loading them.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/includes/shared/definitions.php
+++ b/public_html/wp-content/mu-plugins/blocks/includes/shared/definitions.php
@@ -175,6 +175,3 @@ function get_shared_definition( $key, $type ) {
 
 	return $result;
 }
-
-
-
--- a/public_html/wp-content/mu-plugins/blocks/includes/shared/definitions.php
+++ b/public_html/wp-content/mu-plugins/blocks/includes/shared/definitions.php
@@ -1,0 +1,169 @@
+<?php
+namespace WordCamp\Blocks\Shared\Definitions;
+defined( 'WPINC' ) || die();
+
+/**
+ * Retrieve an array of definitions for use within a block.
+ *
+ * Currently there are two types of definitions: attributes, which make up a block's schema, and options, which
+ * are label/value pairs used in situations where the value of an attribute must come from an enumerated list.
+ *
+ * @param array|string $keys The keys of the definitions to retrieve. An array of strings, or 'all'.
+ * @param string       $type The type of definition to retrieve. 'attribute' or 'option'.
+ *
+ * @return array
+ */
+function get_shared_definitions( $keys, $type ) {
+	switch ( $type ) {
+		case 'attribute':
+			$definitions = [
+				'align'        => [
+					'type'    => 'string',
+					'enum'    => wp_list_pluck( get_shared_definition( 'align_block', 'option' ), 'value' ),
+					'default' => '',
+				],
+				'className'    => [
+					'type'    => 'string',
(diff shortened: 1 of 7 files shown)


public_html/wp-content/mu-plugins/blocks/includes/shared/definitions.php
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * @param array|string $keys The keys of the definitions to retrieve. An array of strings, or 'all'.
+ * @param string       $type The type of definition to retrieve. 'attribute' or 'option'.
+ *
+ * @return array
+ */
+function get_shared_definitions( $keys, $type ) {
+	switch ( $type ) {
+		case 'attribute':
+			$definitions = [
```
vedanshujain: do you think if it would be better if there is a way for a shared component to hook and add the definitions which are supported by that component.

coreymckrill: We _could_ do that... not all of the shared attributes have a correlated shared component. And, I think it would make it harder to understand what shared definitions are available, if you have to go and look in multiple files to see them all...

So, I'm leaning towards "no", but I'm still willing to be convinced.

vedanshujain: I was imagining that to use a shared component, we will just write a line somewhere and it will hook all the files, attributes, options that we would need. but I am not sure if there is a way to do that without complicating the code or if it would be worth it.

that said, it seems a bit more logical to me to have attributes and options defined in the component itself, so that we don't forget why are we using an attribute, or when using a component in a new block we won't have to search through the code to figure out what attributes it needs. but I guess this can be handled by a doc comment, if its not trivial to do it via code.

iandunn: 🤔 

I can see Vedanshu's point about it being more logical to define the attributes in the component itself, but I also see the benefits of making it DRY.

If there's a simple way to get the best of both worlds, then I feel like that'd be worth exploring (if it wouldn't take too much time), but otherwise I feel like this approach is fine.

With this approach, if you're looking at the code for a component, you can see that `get_shared_definitions()` is being called, which makes it obvious enough what's going on. If you're looking at the data in the API response or browser console, then you'll still see the full list.

iandunn (approved):

public_html/wp-content/mu-plugins/blocks/includes/shared/definitions.php
coreymckrill: > I was imagining that to use a shared component, we will just write a line somewhere and it will hook all the files, attributes, options that we would need.

I could see something like this working if we basically created a PHP version of a React Component, and extended that for each of our shared components. Then each would be a class that has a public `render` method, and it could also have a method that defines its attribute schema.

That might not work for `option` definitions, though, since those don't necessarily have a 1-to-1 relationship with a component.

coreymckrill: In the interest of keeping this moving forward, I'm going to go with this approach for now. We could refactor later, in a future iteration, without too much pain, I think.

coreymckrill:
Committed: https://meta.trac.wordpress.org/changeset/8624

PR #73, opened 2019-04-10
WordCamp Camptix: Throttle ticket purchase request.

vedanshujain:
This patch adds throttling support to ticket purchase request to 39 per hour. This is not intended to prevent any kind of sophiscated attacks, but is just there so that we don't have to delete lots of failed tickets when a security runs any automated tests on us.

`Form_Spam_Prevention` provides honeypot feature as well, which we are intentionally not using here  because we want don't want to be aggressive in blocking ticket purchase form.

Also fixes a minor bug in `Payment_Options` where output was directly getting rendered instead of being returned.

Depends on https://github.com/Automattic/camptix/pull/238

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
@@ -69,6 +69,7 @@ class Payment_Options extends CampTix_Addon {
 	 * @param string $selected_payment_method Already selected payment method.
 	 */
 	function generate_payment_options( $payment_output, $total, $payment_methods, $selected_payment_method ) {
+		ob_start();
 		?>
 		<div class="tix-submit">
 			<?php if ( $total > 0 ) : ?>
@@ -93,6 +94,7 @@ class Payment_Options extends CampTix_Addon {
 			<br class="tix-clear" />
 		</div>
 		<?php
+		return ob_get_clean();
 	}
 
 	/**
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
@@ -3,6 +3,7 @@
 namespace WordCamp\CampTix_Tweaks;
 use CampTix_Plugin;
 use WP_Post;
+use WordCamp\Utilities\Form_Spam_Prevention;
 
 defined( 'WPINC' ) or die();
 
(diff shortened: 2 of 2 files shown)


public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
```diff
@@ -93,6 +94,7 @@ function generate_payment_options( $payment_output, $total, $payment_methods, $s
 			<br class="tix-clear" />
 		</div>
 		<?php
+		return ob_get_clean();
```
iandunn: `MUST` - Nice catch! This should be a separate commit when merged to SVN.

public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
```diff
@@ -919,6 +924,50 @@ function modify_shortcode_contents( $shortcode_contents, $tix_action ) {
...
+ * Show error message if IP Address has been throttled by `Form_Spam_Prevention`.
+ */
+function show_throttle_notice() {
+	global $camptix;
+
+	$fsp = new Form_Spam_Prevention(
+		[
+			'prefix' => WC_CAMPTIX_FSP_PREFIX,
+		]
+	);
```
iandunn: `MIGHT` - Since there's only 1 array item, and it's relatively short, it feels like this would be more readable if it were all on a single line. It'd only be about `75` chars, which is far under the `115` limit we typically use for wrapping.

If there were multiple non-trivial array items, though, then it probably would be better to wrap regardless of length.

Same applies to the other function as well.

public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
```diff
@@ -919,6 +924,50 @@ function modify_shortcode_contents( $shortcode_contents, $tix_action ) {
...
+		// Therefore its important that we disable payment methods tab if we are going to block the checkout request.
+		add_filter( 'tix_render_payment_options', '__return_empty_string', 20 );
+	}
+
+}
+
+/**
+ * Checks if IP is throttled. If not then increments the score by 0.1. This does not handle any sophisticated attack, but is just there so that we do not have to delete junk tickets if a security researcher runs a test on site.
+ *
+ * Maximum score threshold in Form_Spam_Prevention is 4, so using 0.1 implies an IP address will be able to make 39 purchase request before getting throttled.
```
iandunn: `MIGHT` - How did you arrive at `39` as the limit? That feels pretty high to me. I was thinking somewhere closer to `10`; maybe `15` or even `20` if we wanna be real cautious.

I'm open to a higher number, though, but am curious about the reasoning.

coreymckrill: I was thinking it felt high, too, but then I realized that the way this is implemented, the score increases every time the `form_checkout` callback runs, which I think may be multiple times during a successful ticket purchase.

[MIGHT] Related, since this is a fairly unconventional implementation of `Form_Spam_Prevention` (not using any of the form fields or validation, only the throttle transient), it might be worth adding more explanation to the code comments about how/why it's set up this way so it's easier to grok later on.

iandunn: Oh, that's a good catch. It doesn't seem like we want it to trigger each time, only once per checkout attempt, right?

public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
vedanshujain: :+1:

public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
vedanshujain: As Corey said, someone might try several times before actually successfully purchasing a ticket, maybe because of payment gateway misbehaving or some other reason. And since this is the ticket purchase page, IMO we should really give benefit of doubt as much as possible.
So my intention is to only block when we are extremely sure that its a bad actor, and even if it is a bot, let it go if its not annoying.

iandunn: 👍 

We can always lower it if it becomes a problem in practice.

vedanshujain:
This is merged

Issue #95, opened 2019-04-12
Blocks: No featured image has been set: should we add a placeholder?


coreymckrill:
In #87 we discovered that for blocks using the `FeaturedImage` component (currently Sessions and Sponsors), when no featured image is set, it still renders an `img` tag with an empty `src` attribute. This is not ideal behavior. Two better options would be:

* Render a placeholder instead. Either a generic image or a `div` container with an icon or some kind of background. Perhaps this could be defined outside of the component so different placeholders could be used in different contexts.
* Render nothing.

In the case of Sessions, it seems like we should render a placeholder so that in a grid layout there is consistency in having some kind of image for each session. But maybe we should also have a setting where the placeholder could be turned off? Or maybe we just let organizers hide the placeholder with CSS if they want?

coreymckrill:
@melchoyce Any thoughts on this? If we were to have a placeholder for Session or Sponsor posts that don't have a featured image set, what would you want it to look like?

melchoyce:
I think we shouldn't have a placeholder — especially for sessions, including a featured image is pretty rare (we only add them to Boston for social sharing), and so adding in a placeholder feels more unexpected to me than having a wonky grid.

coreymckrill:
Works for me!

PR #97, opened 2019-04-19
Blocks: Add fallback when no featured image is available

coreymckrill:
~Modifies both JS and PHP to render a fallback container with icon when a post does not have a featured image.~

Ensures that an empty string is output instead of a broken image tag when a post does not have a featured image available.

This also makes some improvements to the featured image output when there _is_ a featured image both on the JS and the PHP side.

These blocks also need to be updated to work with this:

- [x] Sessions
- [x] Sponsors

Refs #95

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/assets/src/sessions/block-content.js
+++ b/public_html/wp-content/mu-plugins/blocks/assets/src/sessions/block-content.js
@@ -166,10 +166,9 @@ class SessionsBlockContent extends Component {
 
 						{ show_images &&
 							<FeaturedImage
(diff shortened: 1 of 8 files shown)


iandunn:
I started reviewing this on Friday, but now it sounds like [Mel doesn't think we should have a fallback](https://github.com/WordPress/wordcamp.org/issues/95#issuecomment-485530517), so I'll hold off doing anything else until.

coreymckrill:
Sounds good. I've changed the description here to reflect that this only ensures that _nothing_ is output when a featured image isn't available, instead of outputting a new fallback element.

coreymckrill:
Committed here: https://meta.trac.wordpress.org/changeset/8668

PR #98, opened 2019-04-23
Add the CampTix Invoices plugin to the WordCamp codebase

coreymckrill:
* Changes the text domain to `wordcamporg` and removes files and functions related to having a separate text domain. The strings will instead be added to the [WordCamp project](https://translate.wordpress.org/projects/meta/wordcamp/) on translate.wordpress.org.
* Removes unnecessary meta files (composer.json, .gitignore, etc.)
* Uses **camptix-invoices** as the plugin directory/slug instead of **invoices-camptix** so there won't be any collision with the version in the plugin repository.


coreymckrill:
Committed here: https://meta.trac.wordpress.org/changeset/8657

PR #99, opened 2019-04-24
Add schedule block

iandunn:
So far it's just the static interface. Asked mark/mel for feedback:

https://make.wordpress.org/community/2018/10/26/wordcamp-block-schedule/#comment-26814


public_html/wp-content/mu-plugins/blocks/assets/src/schedule/index.js
```diff
@@ -0,0 +1,50 @@
...
+	 * Using the standard mechanism for full-width blocks would also add a toolbar icon which could be used to turn
+	 * off the wider view in the editor, which could cause confusion and clutter, without add benefit.
+	 *
+	 * Another potential solution would have been to manually add CSS rules to make the block full-width in the editor,
+	 * but that would have meant duplicating the existing rules from Core, and tightly coupling the editing experience,
+	 * which wouldn't be future-proof.
+	 *
+	 * Adding the alignment through `getEditWrapperProps()` allows us to achieve the desired result without any side-
+	 * effects.
+	 */
```
coreymckrill: Doesn't this break the concept of What-You-See-Is-What-You-Get? It seems misleading to me to have the schedule be inexplicably wider than all the rest of the content in the block editor, but the same width on the front end.

I would lean towards actually adding official support for wide and full alignments, like we just added for the other blocks. It seems like the schedule would have a similar use case, but also, the organizer could temporarily toggle it on if they need to get a better look at their schedule in the editor while also getting a more accurate WYSIWYG.

iandunn: 🤔 

You make a good point about the inconsistency between the front and back ends.

I was thinking that was a there wasn't really a use case for turning off the full width, so it didn't make sense to use those options. Thinking about it again, though, I guess for single- and double-track camps, the default width would actually be fine, so maybe it would be better to use those, and let camps select wide/full when they have more tracks. I'll play around with it, thanks!

iandunn:
I rebased this against `vedanshu-store` to pull in the new data stuff. We'll need to merge #69 to master before this one.

Side note: I was pleasantly surprised that GitHub no longer destroys comments like https://github.com/WordPress/wordcamp.org/pull/99#pullrequestreview-230803419 when a branch is rebased 😄

iandunn:

I'll re-submit this.

PR #103, opened 2019-05-01
Automate request approval 

iandunn:
if the request is within the approved budget, there's no need for manual approval

Supercedes #19


timiwahalahti:
Not sure if we want this anymore, actually. During the budget review deputy discuss essential topics with the organisers, and even if the budget is within approved limits, there might be single lines the reviewer asks to change (like an imbalance between some costs).

What do you think @kcristiano, @harmonyromo, @dorsvenabili, @angelasjin?

kcristiano:
@timiwahalahti I agree I don't think automatic approval is warranted any longer.

harmonyromo:
I would agree

dorsvenabili:
I agree with not automating the approval.

timiwahalahti:
Going to close this PR per voiced opinions of not doing automated budget approvals.

PR #104, opened 2019-05-01
Add/schedule block

iandunn:
_WIP_

So far it's just the static interface. Asked mark/mel for feedback:

https://make.wordpress.org/community/2018/10/26/wordcamp-block-schedule/#comment-26814

Supercedes #99 

### Testing - not ready yet

1. import tests/sample-sessions.xml
1. try to think of cases not covered by ^, so that you're not just repeating the same tests i've already done many times


public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/index.js
```diff
@@ -0,0 +1,59 @@
...
+	 */
+	//getEditWrapperProps( props ) {
+	//	return {
+	//		'data-align': 'full',
+	//		// maybe apply this conditionally based on screen width? show responsive layout on smaller screen?
+	//	};
+    //},
+	// probably remove ^ in favor of `supports align wide full` above, but test with 1 and 2 column schedules once get live data integrated
+	// if do that, then maybe turn `full` on by default if the camp has 3+ tracks, or maybe just always, and let them turn it off if they want
+	// probably safe to always make it the default, as long as grid is center-aligned to content area, instead of left-aligned?
```
ryelle: This all seems like a great thing to put into an issue and address in a following PR, once the base schedule is merged

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
```diff
@@ -0,0 +1,131 @@
...
+		return (
+			<CoreInspectorControlsContainer>
+				<PanelBody title={ __( 'Display Settings', 'wordcamporg' ) } initialOpen={ true }>
+					<ToggleControl
+						label={ __( 'Show categories', 'wordcamporg' ) }
+						checked={ show_categories }
+						onChange={ ( value ) => setAttributes( { show_category: value } ) }
+					/>
+
+					<PanelRow className="wordcamp-schedule__choose-specific-days">    {/* todo not sure if this is correct */}
```
ryelle: Why are these in a PanelRow? They aren't all meant to be in one row, so maybe this should just be a `div`? Then you won't need to worry about the conditional wrapping note below.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,331 @@
...
+
+
+
+/* todo
+ *
+ *
+ * IMPORTANT: refer to https://codepen.io/mrwweb/pen/ZaONLW for lots of critical notes and info, also https://meta.trac.wordpress.org/ticket/3117
+ *
+ * markup/css needs to have fallback for unsupported browsers? doesn't have to look good, but has to be legible/accessible
+ *      will babel add prefix for ie11?
```
ryelle: babel won't help with CSS, but we can add autoprefixer. the IE11 grid support is not great though (might collapse everything into the first cell), and I'm not sure how much just checking `@supports` will help us here. Testing/supporting IE would be another good issue I think :)

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,331 @@
...
+ *      if so, still need to do anything special? if so, enough to just wrap stuff in @supports?
+ *
+ * test and make compatible w/ personalized schedule builder - maybe by making that add support for this rather than this be back-compat w/ that
+ * test all the scenarios listed at https://meta.trac.wordpress.org/ticket/3117#comment:2. lightning talks too, if not already covered there.
+ *
+ * test other camps, like seattle 2017, berlin 2017, 2018.montreal, wcus, new york, europe, boston, kansas city, miama, tokyo, etc across various years
+ * 	    create live posts that mimics all the situations mark covered in the static mockup
+ * 	    what will this look lke with 15 tracks ala 2012.nyc? or a slightly-less-unreasonable 8 tracks ala 2009.newyork ? probably don't need to handle those edge cases
```
ryelle: Europe & US both have 6 tracks this year, so it's not unreasonable and maybe going to be more common on the flagship events.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,331 @@
...
+ * @param {string} date In a format acceptable by `wp.date.format()`.
+ * @param {Array} sessions
+ * @param {Array} tracks
+ *
+ * @return {Element}
+ */
+function ScheduleDay( { date, sessions, tracks } ) {
+	return (
+		<Fragment>
+			<h2 className="wordcamp-schedule-date">
```
ryelle: It would be nice to [use `InnerBlocks` for the headings,](https://github.com/WordPress/gutenberg/tree/master/packages/block-editor/src/components/inner-blocks) so that organizers can edit this content, change the heading level if they need to, etc. See the template prop, where you can pass in `[ 'core/heading', { content: format( 'l, F jS', date ) } ]` so they can have the right initial content (see an example [in my recipe block plugin](https://github.com/ryelle/rmb-recipe-block/blob/0c6535d1086be232e6b2941f420e6e2f2676fb14/assets/js/blocks/recipe-ingredients.js#L60-L71)).

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,331 @@
...
+					aria-hidden="true"
+					style={ { gridColumn: `wordcamp-schedule-track-${ track.id }` } }
+				>
+					{ track.name }
+				</span>
+			) ) }
+
+			{/* todo test w/ 3+ tracks, make sure column headers match assigned sessions */}
+			{/* maybe different name because needs to be unique to avoid conflicting styles? */}
+			{/* need js to toggle aria-hidden when css changes display:none? is aria-hidden needed here? */}
```
ryelle: Why are these `aria-hidden`?

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,331 @@
...
+ * workaround should be removed.
+ *
+ * https://github.com/WordPress/gutenberg/issues/16218
+ * https://github.com/WordPress/gutenberg/issues/15221
+ *
+ * @param {Array} sessions
+ *
+ * @return {Array}
+ */
+function deriveSessionStartEndTimes( sessions ) {
```
ryelle: Could these `derive*` functions be pulled out of this component file & run instead in the select query? that way you know that the data coming from the select is already correctly processed.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,331 @@
...
+ * Get the tracks that the given Sessions are assigned to.
+ *
+ * In general, it's more useful for this block to know which tracks are assigned, rather than which ones exist,
+ * because e.g., you don't want to print all tracks in `GridColumnHeaders`, only the ones being used.
+ *
+ * @param {Array} sessions
+ *
+ * @return {Array}
+ */
+function getPopulatedTracks( sessions ) {
```
ryelle: Maybe this function should get some tests, since it seems like there can be a lot of edge cases from the notes here.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,331 @@
...
+	tracks.sort( ( a, b ) => {
+		return a.slug > b.slug;
+	} );
+
+	return tracks;
+}
+
+// todo
+export class ScheduleGrid extends Component {
+	// does this need to be a class component? only if it has state? if so, are there other blocks which don't need to be Components?
```
ryelle: > does this need to be a class component?

If all you're doing is a render function, it can be a functional component 👍

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/sessions.js
```diff
@@ -0,0 +1,168 @@
...
+	const timeGroups            = [];
+
+	for ( const timestamp of Object.keys( sessionsGroupedByTime ) ) {
+		const gridRow = `
+			time-${ format( 'Hi', sessionsGroupedByTime[ timestamp ][0].derived.startTime ) } /
+			time-${ format( 'Hi', sessionsGroupedByTime[ timestamp ][0].derived.endTime ) }
+		`;
+
+		timeGroups.push(
+			<h3
```
ryelle: Is this the start of a row? I wouldn't use `h3` here, probably just a div or span (it's not totally clear to me what this ends up rendering since i can't get the schedule to load on my test site). Otherwise you're being opinionated on what the heading hierarchy can be on the page, and it'll end up being a lot of headings for screen reader users to read.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/sessions.js
```diff
@@ -0,0 +1,168 @@
...
+		<div
+			className={ classes }
+			data-wordcamp-schedule-session-id={ id }
+			data-wordcamp-schedule-session-slug={ slug }
+			data-wordcamp-schedule-session-tracks={ tracks.map( track => track.slug ) }
+				// test w/ 0, 1, and 2 tracks
+				// need space or comma or something inbetween? think about how they'll be targeted w/ css
+			style={ { gridColumn, gridRow } }
+		>
+			<h4 className="wordcamp-schedule-session-title">
```
ryelle: Same point as above, we shouldn't be this opinionated about heading levels, and I don't think that these need to be headings. I'll take a look at what seems to be best practices around this for a11y.

ryelle (commented):
I had some time this morning to check this out, so I left a few comments. I wasn't able to get the schedule to render locally at all, I kept getting console errors, but I walked through the code a bit here on github 🙂

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
iandunn: That comment was more shorthand for "are we going to need any state or lifecycle methods in the future? if not, then come back at the end and make this a functional component".

I could go ahead and make it functional now, and then switch back to stateful in the future if there's a need; there's nothing wrong with that approach, the only reason I haven't done is is because I've been focused on the more important work of getting the basic features built, and didn't want to get distracted by tangential minor issues.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/index.js
iandunn: It seems to me like something that should work before the block is merged. The switch from `getEditWrapperProps()` to `supports` has already been done, it just needs to be tested more before I'm confident that it's working in all cases.

I'm not planning to leave a bunch of `todo` comments in the code after the PR is merged, 99% of those are reminders for myself to fix/complete things while the branch is a WIP.

The reason I left those comments in the PWA files was because that feature was rushed to get deployed long before it was ready for production, and I wanted to make sure the code was brought up to standards before being activated for all sites. Having comments in the code means they're a lot less likely to get lost/ignored than if they're buried in some external issue tracker.

ryelle: I still think adding wide/full alignment support can be done in a following PR – get a basic schedule working in one PR, and then iterate on it. This isn't a live feature, it doesn't need to be perfect before the first PR is merged, and smaller, single-feature changes are easier for your teammates to review.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
iandunn: They're hidden on the mobile view, but appear on the desktop/Grid view. It seemed like it needed `aria-hidden` since it can change dynamically, but maybe that's not really an issue in real life, except maybe tablets switching from landscape to portrait? What do you think?

ryelle: `display: none` will already hide these from screen reader users, so I think it's redundant here. The only reason you'd use `aria-hidden` would be to hide visible content from screen readers (like if it's already on the page in `screen-reader-text`, for example).

iandunn: Ah, that's a good point, now I'm not sure why; maybe because it's not _essential_ information, and can be implied, so it'd be more convenient for screen reader users?

Do you remember, @mrwweb?

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
mrwweb: I'd specifically recommend _against_ IE11 grid. They should just receive the fallback/mobile version. Because of that `@supports( display: grid ) {}` is perfect because IE11 "supports" grid but doesn't support `@supports`. Reference: https://css-tricks.com/building-a-conference-schedule-with-css-grid/#article-header-id-2

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
mrwweb: Here's the rationale:
- The track column "headings" are neither headings nor table headers. They are semantically unmoored both in markup and source order. They're essentially a nice-to-have feature for people who can see the layout. Therefore, screen readers neither need them nor could make sense of them.
- The track information is included with each session so it's accessible there. (It should never be `display:none`'d.)

The column "headings" are `display: none` on mobile because there's no easy or obvious place to put them, and, again, the information is provided directly on each session.

Reference: https://css-tricks.com/building-a-conference-schedule-with-css-grid/#article-header-id-6

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/sessions.js
mrwweb: Reference: https://css-tricks.com/building-a-conference-schedule-with-css-grid/#article-header-id-1

I think they are very much necessary.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
iandunn: Ah, that makes sense to me. Thanks! 

What do you think, Kelly?

iandunn:
I've been thinking about how the tracks should be sorted in the grid view. The simplest thing is to just sort them alphabetically, but ideally they should be sorted in a way that minimizes the number of times when a session spans non-contiguous tracks.

e.g., if session Foo is assigned to tracks Bar and Quix, those tracks should be adjacent on the grid layout:

```
| Bar | Quix | Baz |
--------------------
| Foo        | Cat |
--------------------
```

...rather than being sorted alphabetically, and making the session un-renderable:

```
| Bar | Baz | Quix |
--------------------
| Foo |     |      |
--------------------
```

Developing an algorithm for that seems pretty complex at first glance, though, and spanning sessions is an edge case anyway (except for cases where _all_ sessions are spanned, like lunch, etc, which isn't relevant to this question).

So, I think for now it's probably fine to just sort them alphabetically, and if needed organizers can rename the tracks to get them in the desired order. I don't anticipate that being needed frequently, though, and maybe not ever. 

If it does become a common need, we could always iterate on the method in the future, using a feature flag to preserve back-compat.

...Another solution entirely might be to duplicate the markup, so that the same session appears in both tracks:

```
| Bar | Baz | Quix |
--------------------
| Foo |     | Foo  |
--------------------
```

...but I think that's probably also premature for something that seems like an edge case right now.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
ryelle: Ok, I see now– since I couldn't get the schedule to load, I had to just rely on reading the code & that wasn't clear. As long as this info does appear in the session itself, using `aria-hidden` is fine.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/sessions.js
ryelle: Ah, I see. I was thinking each session would have a rowheader/columnheader for this kind of information (which is also why I thought the tracks shouldn't be hidden). I think this works, but we might want to check that it's not overly verbose for screen reader users.

ryelle: For example, having each individual session in a heading makes navigating by headings very time-consuming, since you need to move through each track to get past the schedule.

iandunn:
@mrwweb , what were your thoughts around using `span` for tracks/categories/speakers? My instinct would be to use `ul`, since there can be multiple.

```html
<span class="session-track">Track: 1</span>
<span class="session-presenter">Presenter</span>
```

vs

```html
<ul class="wordcamp-schedule__session-tracks">
	<li>Tahoma 1</li>
	<li>Tahama 2</li>
</ul>

<ul class="wordcamp-schedule__session-speakers">
	<li>Jane Doe</li>
	<li>Fred Smith</li>
</ul>
```

_( The mockup didn't call for `Track:` / `Speaker:` labels, but we can consider those if you think it'd be better to have them )_

mrwweb:
Good point. As you might have already inferred, that's one area where I put less thought into things. First reactions:

1. My primary goal with markup in the session is to keep things super granular and CSS-class-laded to enable themers to easily customize the presentation.
2. My gut says the labels (e.g. "Tracks") are necessary for screen reader accessibility given that tracks often have abstract names that the track column headings are not available to screen readers.
3. Lists often require a bit extra un-theming because they come with more styles baked in and folks often style them generically in the content area (e.g. `.entry-content ul`). Spans feel a bit more flexible and the flatter structure might be slightly more useful, but I think it's not a huge deal either way. I'd get feedback from a few other people, ideally ones who have themed a WC site somewhat recently.

iandunn:
> primary goal... keep things super granular ... easily customize the presentation.

Agreed.

> labels... are necessary ... given that tracks often have abstract names ... track column headings are not available to screen readers.

Do you feel like they're only needed for screen readers, or also for sighted users? If it's the former, then we could use `screen-reader-text` to add them for readers, without deviating from the mockup.

> Lists often require a bit extra un-theming

Yeah, I think the block's default styles would remove most of what browsers and Core adds, though, so that organizers wouldn't have to. Does that seem like it'd be an easy solution, or would it miss some things?

> Spans ... more flexible and the flatter structure might be slightly more useful

My main concern was just being semantically correct, but I agree that it's important to consider practical concerns as well. What are some examples of how spans would be more flexible/useful?

mrwweb:
re: visible labels - I tend to believe that kind of stuff is more valuable for users than people often think, but I often get outvoted on stuff like that :) If they were visible, I worry people would `display:none` them frequently, so I'm kind of of the mind that accessibly hiding them is the best compromise.

re: spans - The extra layer of nesting means some grid layouts might be harder to do, but maybe that grouping would actually be valuable! For some reason, I'm struggling to come down one way or another on this one. If we're going to get super picky about semantics, I would think a definition list with the labels is the most appropriate structure.

iandunn:
@ryelle, @coreymckrill: one of the remaining blockers for the static interface is how to handle hierarchical tracks, if you feel like diving in and giving advice.

* https://github.com/WordPress/wordcamp.org/pull/104/commits/45e8e17c2a82db4fdc6564d6581f82169a5a03bf
* https://github.com/WordPress/wordcamp.org/pull/104/commits/2613ebc90b06bb0cd9054575a2f4390481e9bae2

iandunn:
@ryelle, @coreymckrill: Now that we have the Live Schedule block, it seems like we should rename this to something like "Full Schedule", to avoid confusion.

Does "Full" sound good, or is there something more descriptive?

ryelle:
IMO "Schedule" is still fine, since it _is_ the default layout people think of when thinking of a schedule.

coreymckrill:
I agree, "Schedule" still seems best. When both of them are available, I would guess that people would most likely choose the one with the simpler name, unless they really are looking for the live updating one.

mrwweb:
Maybe consider renaming "Live Schedule" to "Day-of Schedule" or something else similarly descriptive?

iandunn:
The big open question I still have here is around how the session data flows through the components and is modified.

### PHP side

* `get_session_posts()` in `controller.php` is the source on the PHP side. It's called by `render()` and `print_dynmaic_styles()`.
* `render()` might be the PHP view in the future, but will probably be removed and we'll just render everything in JS. If we do keep it, it'd probably need the data grouped and derived similar to `print_dynamic_styles()` and `ScheduleGrid()`.
* `print_dynamic_styles()` - Prints styles for specifics days/tracks that can't be in static `.scss` files. It  groups the sessions by their date, and calls functions to derive the assigned tracks and start/end times from each session.


### JavaScript

* Session data enters the system from the `withSelect()` call in `edit.js`.
* `ScheduleGrid()` derives the sessions' terms and start/end times, then groups them by the date (similar to `print_dynamic_styles()` on the PHP side).
* `Sessions()` determines if any of the sessions overlap, and groups them by their time slot.
* `Session()` determines if the session spans non-contiguous tracks (b/c Grid doesn't support that)
* Inspector controls aren't using live data yet, but will need to know the terms and dates that have been assigned to all sessions.

Maybe all of the derivation of data and grouping by date/time should be done as part of the `withSelect()` call, and/or inside `ScheduleEdit()`?

Maybe we should make it DRY w/ PHP side, like with a custom API endpoint that PHP could also call internally?

Or maybe it's more straight-forward the way it is?

@coreymckrill , @ryelle

iandunn:
You can search the `schedule/` folder for `todo-data` to see specific places where the data-handling raises questions.

public_html/wp-content/mu-plugins/blocks/source/utils/render-frontend.js
```diff
@@ -92,6 +92,10 @@ export default ( selector, Block, getProps = getPropsFallback ) => {
 				/>,
 				element
 			);
+
+			// todo-front why pulling render() from react-dom instead of @wordpress/element ?
+			// make sure that react-dom, lodash, etc aren't getting bundled into the build, should be using from core versions as external instead
+				// seems lke they're being bundled live-schedule.min.js
```
ryelle: Happened to see this in my git app — it's meant to save from enqueuing all of `wp-element` on the frontend, it should not affect build since since both `react-dom` and `wp-element` are externals (the only change is to the dependency list in the asset.php file).

however, the block in the live-schedule uses wp-element, so it's added to the deps list anyway, so this is kind of a pointless "optimization" 🤷🏼‍♀️  Since I see you're already importing from `wp-element` in this file, you can just combine them.

iandunn:
Closing this in favor of #502, since so much has changed.

Issue #108, opened 2019-05-01
Non-parity of HTML markup between PHP and JS renderers


coreymckrill:
There are still some places where the markup is not the same on the front and back end. It looks like this is due to some abstractions that we implemented such as `grid-layout`.

coreymckrill:
This is fixed with https://meta.trac.wordpress.org/changeset/8737

Issue #96, opened 2019-04-16
Class names don't follow convention


iandunn:
If I understand [the guidelines](https://wordpress.org/gutenberg/handbook/contributors/develop/coding-guidelines/) correctly, it seems like all of our class names are wrong, and should be renamed to the BEM-esque convention 😱

e.g.:

* `wordcamp-speaker-sessions-list` should be `wordcamp-speaker__sessions-list`.
* `wordcamp-image-container` should be `wordcamp-avatar` (or `wordcamp-avatar__container` if we want to keep the redundant descriptor)
* `wordcamp-item-permalink` should be `wordcamp-block-content__item-html-content-permalink`, or maybe `wordcamp-block-content__item-html-content__permalink`? Those would both be awful, so maybe (hopefully) I'm reading the guidelines wrong? Maybe `wordcamp-item-html-content__permalink` if we want to break the guidelines for sanity's sake?
* `wordcamp-image-link` should be `wordcamp-featured-image__link`
* `wordcamp-clearfix` should be `wordcamp__clearfix`

Those are just a few examples, but it seems like probably every single class needs to be renamed 😢. I hope I'm wrong.

If not, we should probably fix this before launch since fixing it in the future would break BC.

@coreymckrill , @vedanshujain : what do you think?

iandunn:
The BEM-esque style is nice, because it matches the component structure, but using it would also create inconsistency across our codebase. That would be especially inconvenient for organizers, when they're trying to style the front-end.

However, that inconsistency already exists, since the Core blocks use the BEM style, while everything else output by Core follows [the current standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).

It turns out that the inconsistency has been discussed a few times:

* https://github.com/WordPress/gutenberg/issues/11689
* https://wordpress.slack.com/archives/C02QB2JS7/p1496251618220866

Unfortunately, that wasn't resolved before the 5.0 merge, so now we're left dealing with the fallout.

It's unclear whether the Gutenberg standards will be merged into Core at some point in the future. Obviously there'd be some back-compat hurdles to overcome if they are; I'm guessing the only way to do it would be to add duplicate BEM-style classes everywhere that currently has WP-style classes (for lack of a better term). That would be awkward, cluttered, and somewhat ineffective :(

[...]

PR #110, opened 2019-05-02
Bring HTML markup closer to parity between PHP and JS renderers

coreymckrill:
* Renames the `grid-layout` components to `post-list` because it more accurately describes what the components handle.
* Refactors all blocks to use the post list components. Some were not using the `grid-layout` components yet either.
* Removes toolbar components from individual blocks in favor of the shared `layout-toolbar` (formerly `grid-toolbar`).
* Adds PHP helper functions for rendering post list items in a consistent way, similar to Item* components on the JS side.
* Tries to ensure that our current HTML class naming conventions are applied consistently, and that all the same classes appear on both the front end and the back end. (These might change again depending on what happens with #96)

Refs #108


coreymckrill:
Committed: https://meta.trac.wordpress.org/changeset/8737

PR #111, opened 2019-05-03
Miscellanous patches from April bug scrub

vedanshujain:


public_html/wp-content/themes/campsite-2017/inc/jetpack.php
```diff
@@ -27,6 +28,19 @@ function jetpack_setup() {
 		)
 	);
 
+	$featured_image_enabled = ! ( function_exists( 'wcorg_skip_feature' ) && wcorg_skip_feature( 'cs17_display_featured_image' ) );
```
vedanshujain: In https://meta.trac.wordpress.org/ticket/4339#comment:5 we discussed using 2 separate feature flag such that:
1. Change default values for content option so that its disabled by default for old sites, and enabled for new sites.
2. Always enable for new sites

While implementing I realized that 2nd feature flag is probably redundant, because in first one, if we disable by default for old sites it will also automatically enables for new sites, which was the purpose of 2nd feature flag.

What do you think?

coreymckrill: The reason for two separate flags is that the first one applies to the default settings for all of the places that Jetpack's content options apply (archives, posts, pages), while the second one is specifically for custom post types other than posts and pages (speakers, sponsors, sessions, etc.)

In the notes on that Trac ticket, it questions whether Jetpack's content options for "posts" applies to custom post types as well, or just to the `post` type itself. I think we need to find the answer to that question before this can be resolved.

public_html/wp-content/themes/campsite-2017/inc/jetpack.php
```diff
@@ -27,6 +28,19 @@ function jetpack_setup() {
...
+	add_theme_support( 'jetpack-content-options', array(
+		'featured-images'    => array(
+			'archive'           => true,
+			'archive-default'   => false,
```
coreymckrill: Why not apply the `$featured_image_enabled` variable here as well?

public_html/wp-content/themes/campsite-2017/inc/jetpack.php
coreymckrill: On another note, I think you can omit the `function_exists( 'wcorg_skip_feature' )` part. If that function isn't available, we've got bigger problems ;P

public_html/wp-content/themes/campsite-2017/template-parts/content.php
```diff
@@ -15,6 +15,12 @@
 	<header class="entry-header">
 		<?php
 
+		if ( ( ! is_single() || is_singular( 'post' ) ) && has_post_thumbnail() ): ?>
```
coreymckrill: This is where the second feature flag would need to go, assuming that Jetpack content options do apply to custom post types as well. I think it would be something like

```suggestion
		if ( wcorg_skip_feature( 'cs17_display_cpt_featured_image' ) ) {
			$show_featured_image = ( ! is_single() || is_singular( 'post' ) ) && has_post_thumbnail();
		} else {
			$show_featured_image = has_post_thumbnail();
		}

		if ( $show_featured_image ) : ?>
```

coreymckrill (changes requested):
I left a longer inline comment about this, but the short version is that I think we do need both feature flags, but first we need to clarify Jetpack's functionality.

public_html/wp-content/themes/campsite-2017/template-parts/content.php
```diff
@@ -15,6 +15,12 @@
 	<header class="entry-header">
 		<?php
 
+		if ( ( ! is_single() || is_singular( 'post' ) ) && has_post_thumbnail() ) : ?>
```
vedanshujain:
```suggestion
		if ( ( ! is_single() || is_singular() ) && has_post_thumbnail() ) : ?>
```

public_html/wp-content/themes/campsite-2017/inc/jetpack.php
vedanshujain: Ah got it, I was sure I was missing something. Whether Jetpack's content option will apply to all the post types or not is actually configurable. Right now its onIy enabled for `posts` which means this code actually has a bug and will not fix 3645. 

By configurable, its actually not a Jetpack option, but can be configured in theme like so: https://github.com/WordPress/wordcamp.org/pull/111/files#r280874548 . Right now, CampSite 2017 do not display featured images irrespective of the post type, even when its `post`. 
Which probably also means we may not need different featured flags for different post types, because currently featured image are disabled for all posts? What do you think?

coreymckrill: Oh, I think I see what you're saying. There's no need to differentiate between regular posts and custom post types because for new sites the same Jetpack content option would apply to them all anyway. Now I'm not sure why we thought we needed two separate feature flags. Any thoughts on this, @iandunn ?

public_html/wp-content/themes/campsite-2017/template-parts/content.php
coreymckrill: Assuming we don't need a separate feature flag here, couldn't this be simplified further to

```suggestion
		if ( has_post_thumbnail() ) : ?>
```

public_html/wp-content/themes/campsite-2017/inc/jetpack.php
vedanshujain: `function_exists( 'wcorg_skip_feature' )` removed in https://github.com/WordPress/wordcamp.org/pull/111/commits/e539b28421d0f44e708b3fd945c0bf76df97b08b

public_html/wp-content/themes/campsite-2017/inc/jetpack.php
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/111/commits/e539b28421d0f44e708b3fd945c0bf76df97b08b

public_html/wp-content/themes/campsite-2017/inc/jetpack.php
iandunn: 🤔 I've re-read things a few times, and am still a bit confused, but I think it'd be fine to just have one flag.

It seems like the following is a comprehensive description of what we want to achieve and avoid with the patch:

* New sites should display featured images for all post types, unless the organizer disables that via JP's content options
* Old sites should continue hiding featured images for all post types, unless the organizer enables them via JP's content options

If that's correct, then yeah, I think the current PR achieves that, as long as Corey's suggestion to remove the `! is_single() || is_singular( 'post' ) )` condition is integrated.

iandunn (changes requested):

public_html/wp-content/themes/campsite-2017/template-parts/content.php
coreymckrill: @vedanshujain I think this is the only thing left to figure out what then I'd say it's ready to commit.

vedanshujain: I think this comment answers https://github.com/WordPress/wordcamp.org/pull/111#discussion_r281394415 . Going to go ahead and implement discussed changes and deploy .

public_html/wp-content/themes/campsite-2017/inc/jetpack.php
vedanshujain: Feedback addressed in https://github.com/WordPress/wordcamp.org/pull/111/commits/80ec42ce295c21af1f56a52a6715f45fbfb2bf5b . Resolving this conversation. Please unresolve if we should discuss anything further.

public_html/wp-content/themes/campsite-2017/template-parts/content.php
vedanshujain: We won;t need this as per https://github.com/WordPress/wordcamp.org/pull/111#discussion_r281394415. Resolving

public_html/wp-content/themes/campsite-2017/template-parts/content.php
vedanshujain: Addressed in https://github.com/WordPress/wordcamp.org/pull/111/commits/80ec42ce295c21af1f56a52a6715f45fbfb2bf5b. Resolving

public_html/wp-content/themes/campsite-2017/inc/jetpack.php
iandunn: 👍 

FYI @ryelle, in case you notice anything unexpected once we deploy this.

iandunn (approved):

vedanshujain:
This was merged in https://meta.trac.wordpress.org/changeset/8748 and https://meta.trac.wordpress.org/changeset/8747

PR #113, opened 2019-05-06
Wordcamp docker

vedanshujain:
Adds docker for provisioning WordCamp. This will create `.data` directory to persists MySQL. 

See added `readme.md` for more details.


.gitignore
```diff
@@ -1 +1,3 @@
 ./public_html/wp-content/mu-plugins/blocks/assets/blocks.min*
+.data/
+.idea
```
iandunn: `SHOULD` - I think `.idea` is similar to other things, like `.DS_Store`, `node_modules`, etc. It should be in people's `~/.gitignore_global`, instead of individual repos. It's specific to a dev's local system -- some devs may use PhpStorm while others use Sublime Text, Netbeans, etc; we wouldn't want to have to add all possible IDEs/editors. 

It's also in every repo that person works on, so it's better to just ignore it once in `.gitignore_global` instead of every individual repo.

.gitignore
```diff
@@ -1 +1,3 @@
 ./public_html/wp-content/mu-plugins/blocks/assets/blocks.min*
+.data/
```
iandunn: `MIGHT` - I would if that directory could be in `.docker/` instead, since it's a consequence of Docker, rather than being relevant to all local dev environments?

That'd keep the folder structure and this file less cluttered, and make it obvious where it comes from.

`SHOULD` - I think `mysql` or `database` would be a more descriptive/meaningful name.

.docker/Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM wordpress:5.0.3-fpm
+
+# Move WordPress files to Working Directory
+RUN mv /usr/src/wordpress /usr/src/public_html
```
iandunn: `SHOULD` - It seems like it should be in `public_html/mu` to match production.

readme.md
```diff
@@ -0,0 +1,63 @@
...
+
+This repository mirrors https://meta.svn.wordpress.org/sites/trunk/wordcamp.org/ and can be used to contribute used Git instead of SVN.
+
+###### Setup 
+
+There are two primary ways to setup this repo for local development.
+
+1. Using VVV. WordCamp.org is available as part of WordPress.org's meta network, and can be setup by following meta-environment setup guide: https://github.com/WordPress/meta-environment/blob/master/docs/install.md
+
+1. If you'd like to only setup WordCamp.org and would like to use Docker, you can use inbuilt docker provisioning. Please follow these steps
```
iandunn: `SHOULD` - I think the docker-specific stuff should be modularize the docker into `.docker/README.md`, and then this item should just link to that. That keeps this file more digestable, and focused on being an overview, and avoids overwhelming people who aren't using Docker with irrelevant info.

Rather than adding this file now, though, I think it'd be better to do that as part of #2, since that has a lot of other stuff we probably want to include.

.docker/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+COPY nginx.conf /etc/nginx/sites-enabled/wordcamp.conf
+COPY php-fpm.conf /usr/local/etc/php-fpm.d/zz-www.conf
+
+# Install and configure SSL
+COPY wordcamp.ssl.conf /var/wordcamp.ssl.conf
+RUN cd ~ && openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout wordcamp.key -out wordcamp.crt -config /var/wordcamp.ssl.conf
+RUN cp ~/wordcamp.crt /etc/ssl/certs/wordcamp.crt
+RUN cp ~/wordcamp.key /etc/ssl/private/wordcamp.key
+
+# IAdd script to install plugin and themes
```
iandunn:
```suggestion
# Add script to install plugin and themes
```

readme.md
```diff
@@ -0,0 +1,63 @@
...
+        127.0.0.1	central.wordcamp.test
+        127.0.0.1	wordcamp.test
+        127.0.0.1	2014.content.wordcamp.test
+        127.0.0.1	2014.misc.wordcamp.test
+        127.0.0.1	2014.new-site.wordcamp.test
+        ```
+        
+        `/wp-admin` pages for these sites should now be accessible. Use `admin` as username and `password` as password to login.
+        
+        **Note:** `https` URL scheme must be used to visit these sites. Security exception will be required in first time run.
```
iandunn: `MIGHT` - It'd be nice to add instructions for adding the cert to the browser, so you get a green link instead of yellow. That's not necessary, just a nice-to-have

.docker/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+ADD install-plugin-theme.sh /usr/src/public_html/install-plugin-theme.sh
+
+# Unfortunately themes and plugins cannot be installed as part of docker setup and have to be run manually.
+# This is because everything in wp-content folder will be overwritten when we mount this directory from volume.
+# RUN sh install-plugin-theme.sh
+
+
+# Install wkhtmltopdf for camptix invoices. See https://stackoverflow.com/a/38336153/1845153
```
iandunn:
```suggestion
# Install `wkhtmltopdf` for WordCamp Docs, CampTix Invoices, etc. See https://stackoverflow.com/a/38336153/1845153.
```

.docker/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+RUN curl -L https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.4/wkhtmltox-0.12.4_linux-generic-amd64.tar.xz > wkhtmltox.tar.xz
+RUN tar xvf wkhtmltox.tar.xz
+RUN mv wkhtmltox/bin/wkhtmlto* /usr/bin/
+
+# Install cli
```
iandunn:
```suggestion
# Install WP-CLI.
```

.docker/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+RUN curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
+RUN chmod +x wp-cli.phar
+RUN mv wp-cli.phar /usr/local/bin/wp
+RUN apt-get install less -y
+
+# Install Mail catcher
```
iandunn:
```suggestion
# Install MailCatcher.
```

.docker/install-plugin-theme.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Install plugins
+echo "Installing plugins"
+wp --allow-root --url=wordcamp.test plugin install akismet buddypress bbpress jetpack wp-multibyte-patch wordpress-importer polldaddy liveblog wp-super-cache custom-content-width camptix-network-tools email-post-changes tagregator supportflow camptix-pagseguro camptix-payfast-gateway camptix-trustpay camptix-trustcard camptix-mercadopago camptix-kdcpay-gateway campt-indian-payment-gateway camptix
```
iandunn: `MUST` - CampTix should probably be cloned from GitHub, since we'll want the development version.

IIRC the Meta Environment provision script does that with a few others too.

.docker/install-plugin-theme.sh
```diff
@@ -0,0 +1,9 @@
...
+
+# Install themes
+echo "Installing themes"
+wp --allow-root --url=wordcamp.test theme install twentyten twentyeleven twentytwelve twentythirteen twentyfourteen
```
iandunn: `MUST` - I think we'll need the rest of the Twenty* themes too. We can't rely on Core to bundle them in the future, because that changes over time.

iandunn: `MUST` - Also `p2`

.docker/nginx.conf
```diff
@@ -0,0 +1,51 @@
+# Upstream to abstract backend connection(s) for php
```
iandunn:
```suggestion
# Upstream to abstract backend connection(s) for PHP.
```

.docker/wp-config.php
```diff
@@ -0,0 +1,129 @@
+<?php
+/**
+ * The base configuration for WordPress
```
iandunn, line 3: `SHOULD` - I think all the comments can be removed, they're intended for new users and just clutter the file. Anything non-obvious should be documented, though.

.docker/wp-config.php
```diff
@@ -0,0 +1,129 @@
+<?php
```
iandunn, line 1: `SHOULD` - The Meta Environment version might be good to look at to see if it has anything that's missing from here.

.docker/wp-config.php
```diff
@@ -0,0 +1,129 @@
...
+/** Database Charset to use in creating database tables. */
+define('DB_CHARSET', 'utf8');
+
+/** The Database Collate type. Don't change this if in doubt. */
+define('DB_COLLATE', '');
+
+define( 'JETPACK_DEV_DEBUG', true);
+define( 'SAVEQUERIES', true );
+define( 'SCRIPT_DEBUG', true );
+define( 'MULTISITE', true );
```
iandunn: `MUST` - Run the linter to align these, and catch other deviations from the coding standards.

.docker/wp-config.php
```diff
@@ -0,0 +1,129 @@
...
+ * in their development environments.
+ *
+ * For information on other constants that can be used for debugging,
+ * visit the Codex.
+ *
+ * @link https://codex.wordpress.org/Debugging_in_WordPress
+ */
+define('WP_DEBUG', true);
+
+/* That's all, stop editing! Happy blogging. */
```
iandunn: `SHOULD` - The only stuff past this line should be defining `ABSPATH` and including `wp-settings`

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: "3.2"
```
iandunn: `SHOULD` - Can this live in `.docker/` ?

readme.md
iandunn: `MUST` - Do `HTTP` URLs redirect to `HTTPS`?

readme.md
```diff
@@ -0,0 +1,63 @@
...
+        
+    1. (Optional) Basic installation comes without themes and with only very basic plugins. Add plugins and themes which are installed in wordcamp.org if needed like so:
```
iandunn: `MUST` - Can this be automated as part of installation?

readme.md
```diff
@@ -0,0 +1,63 @@
...
+            ```
+        
+        1. Run predefined command for installing plugins and themes. This could take some time depending on your internet connection:
+            ```bash
+               sh install-plugin-theme.sh
+            ```
+            
+        1. Activate plugin or apply theme as needed.
+        
+    1. (Optional) By default, docker will start with data defined in `.docker/wordcamp_dev.sql` and changes to data will be persisted across runs. To start with different database, delete `.data` directory and replace the `.docker/wordcamp_dev.sql` file and run `docker-compose up --build` again.
```
iandunn:
```suggestion
    1. (Optional) By default, docker will start with data defined in `.docker/wordcamp_dev.sql` and changes to data will be persisted across runs in `.data/`. To start with different database, delete `.data` directory and replace the `.docker/wordcamp_dev.sql` file and run `docker-compose up --build` again.
```

readme.md
```diff
@@ -0,0 +1,63 @@
...
+    
+    After first time provisioning, docker can be started by using `docker-compose up` command from inside the directory. 
+        
+To contribute, you can send pull requests to this repo, or add patches to https://meta.trac.wordpress.org/.
```
iandunn:
```suggestion
To contribute, you can send [pull requests on GitHub](https://github.com/WordPress/wordcamp.org/), or add [patches on Meta Trac](https://make.wordpress.org/meta/handbook/about/get-involved/learn-how-to-contribute-code/#share-your-new-code).
```

iandunn (changes requested):
Looks good overall. I left a few suggestions.

I didn't spend any time testing it, since it sounds like @coreymckrill is eager to try an alternative to VVV.

.gitignore
vedanshujain: Agree. Removed `.idea` in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.gitignore
vedanshujain: Moved `.data` to `.docker/database` and updated refs in readme.md in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.docker/Dockerfile
vedanshujain: I am not in favor of doing this because it would require more non obvoius configuration, which I am not sure if we should do (atleast at this point). If you feel strongly about it though then I will open a Github issue to track it.

readme.md
vedanshujain: I would think Docker and VVV instructions should reside within same parent folder. Maybe same level pages on make/meta or inside a docs folder in git repo with readme linking to those. What do you think?

.docker/Dockerfile
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

readme.md
vedanshujain: Punting this for later. Because the process will actually be other way around. You would generate a cert on host first and then add it inside Docker. 
Keeping this conversation un-resolved for now, and will open a issue when we merge this.

.docker/Dockerfile
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.docker/Dockerfile
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.docker/install-plugin-theme.sh
vedanshujain: I thought about this, but I guess responsibility of VCS for anything inside `wp-content` folder should be on host instead of Docker 🤔 . So if someone wants to develop on CampTix, they should ideally replace the directory inside `wp-content/plugin` folder. What do you think?

.docker/install-plugin-theme.sh
vedanshujain: This would add some non-neglible time to setup procedure. I added this because they were also getting added by meta environment, and TBH I wanted to install not more than 2 3 themes.

User can always add whatever themes they like on their own. What do you think?

vedanshujain: Same as https://github.com/WordPress/wordcamp.org/pull/113#discussion_r281663564

.docker/nginx.conf
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.docker/wp-config.php
vedanshujain: Removed redundant comments in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.docker/wp-config.php
vedanshujain: Synced and added some missing configs in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.docker/wp-config.php
vedanshujain: Aligned now. Although interestingly this does not get flagged by the linter setup to run on PRs. Is this flagged by your linter?

.docker/wp-config.php
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

docker-compose.yml
vedanshujain: It can but then folks would have to go inside a folder starting with `.` to start a docker. This might be annoying for several windows programs like Windows. IMO its fine to clutter the root with this file.

readme.md
vedanshujain: Yes they do. See https://github.com/WordPress/wordcamp.org/pull/113/files#diff-821cdce799b2a90c846089e20e997ff2R8

readme.md
vedanshujain: Unfortunately I was not able to find a non complicated solution to do this. Main reason being that we mount the `wp-content` directory from host filesystem, which is going to overwrite Docker image's `wp-content` folder.

readme.md
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

readme.md
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.docker/Dockerfile
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/1e3bc2e5e0ebcc4643db21c0d98a69b94a51b6c8

.docker/Dockerfile
iandunn: What's the extra config that's needed?

On the WP side, it's just defining `WP_CONTENT_DIR` and `WP_CONTENT_URL`. It doesn't seem like it'd need any custom config in Docker, unless `wordpress:5.0.3-fpm` is some kind of special package that doesn't let you config the path or something.

iandunn: I also personally find it really annoying to have a 3rd-party external like WP mixed into the root of the site, and it can cause sometimes cause problems (like the recurring `update-root-files` fiasco on w.org).

I'm open to it if the alternatives are worse, though.

readme.md
iandunn: 🤔 Having them inside the repo is probably best, so that people don't have to switch back and forth between a browser and IDE/terminal. Having it in the repo also gives us a rendered version on GitHub.com that we can link to from the handbooks.

I think it'd be good for `/README.md` to _mention_ the VVV and Docker instructions, and link to them, but having the full instructions in the main readme feels overwhelming and cluttered to me. I think it'll make it difficult for people to navigate that document and find what they need.

The VVV instructions should probably remain inside the Meta Environment repo, though, since that's where the provisioning code/database/etc is. It'd be good to mention/link to it from `/README.md`, though.

I like the idea of a `docs/` folder, but it doesn't seem like we have enough for that yet, since right now it's just this file. Maybe as part of #2 we could discuss moving some of our internal devdocs there, though? Even if we did move it there, I still think the best location for the Docker readme would be inside `.docker/`, but it'd definitely be good to link (or just symlink) to that from `docs/`.

I'm open to other ideas, though. Do you have any thoughts, @coreymckrill ?

readme.md
iandunn: 👍

.docker/install-plugin-theme.sh
iandunn: 🤔 , I thought that Docker shared the folder between the host and container, so that they're essentially the same? Given that, I'm not sure why the responsibility would shift?

It seems like we'd just be adding an extra step that has to be done manually, rather than automating it.

.docker/Dockerfile
iandunn: Er, I was thinking of configuring custom content folders, not installing WP in a subfolder. What's the config that's needed? I always install WP in a subfolder, so I might have Nginx config you can just copy/paste, etc.

.docker/install-plugin-theme.sh
iandunn: I think the fundamental principle of development environments is that they should match production as closely as possible. Having a fast setup time is nice, but I think accurately representing production is far more important.

.docker/wp-config.php
iandunn: Ah, you're right, the linter doesn't flag that yet. I'll add a `todo` comment to it.

docker-compose.yml
iandunn: That's a good point.

🤔 

It sounds like we could get around that by [setting the `COMPOSE_FILE` environment variable](https://docs.docker.com/compose/reference/envvars/#compose_file):

`COMPOSE_FILE=.docker/docker-compose.yml:other/paths/the/user/wants`

We'd have to add that to the README, though, so it does add a bit of extra work for the user, and potential confusion.

I'm probably just being too pedantic, so the root folder is fine with me unless @coreymckrill would also prefer it in a subdir.

readme.md
iandunn: 👍

readme.md
iandunn: 🤔 I'm confused, I would think they'd be the same folder, so any changes within the container would be synced to the host. Is that not the case?

.docker/Dockerfile
vedanshujain: Ah I see. Yeah we can move WordPress files to `/public_html/mu`. Will take care of this

.docker/install-plugin-theme.sh
vedanshujain: Docker does shared folder, however VCS tools like Git or SVN are not guaranteed to be installed inside it.
I visualize Docker setup like a run time environment, so everything related to running the site should be inside the Docker, but source control should be outside it. For egs, someone can download the source code zip and be able to run Docker, without having to setup or log in to Git, SVN or any other tool which is not required to run it.
Since CampTix is also open source, there is no blocker to actually clone it from GitHub although I feel the setup is much cleaner because of this distinction, but that's just a preference.

readme.md
vedanshujain: Yup, so during the first time docker build process, a container image will be build. But while starting up the container, we will mount the host's `wp-content` directory inside that image (overwriting image's wp-content folder) so that files are synced. That's why if during the build process we install the plugins and themes, they will get overwritten. 

We can solve this by not mounting `wp-content` but instead sub-folders one by one, but that will make the config very big and hard to keep updated. Let me also take a look to see if there are other solutions, but so far this looks like to be the case.

.docker/Dockerfile
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/113/commits/9b89964fb5e1eaf28b8792cd54d3f9f0a937f79b

.docker/install-plugin-theme.sh
vedanshujain: Added themes in https://github.com/WordPress/wordcamp.org/pull/113/commits/db214d69ebe62cfa4be3af4e8432092c90f72061

vedanshujain: Added themes in https://github.com/WordPress/wordcamp.org/pull/113/commits/db214d69ebe62cfa4be3af4e8432092c90f72061

readme.md
vedanshujain: I splitted readme.md in https://github.com/WordPress/wordcamp.org/pull/113/commits/db5db5a1d8c4db4cd3f90b0c97aa60795cfa729a . Please resolve this conversation if this is good enough for now.

iandunn: 👍

.docker/install-plugin-theme.sh
iandunn: 🤔 yeah, I can see that point of view. I personally still lean towards the view that the practical goal is more important than maintaining purity.

Another way to archive the goal would be to make CampTix a Git submodule, but Git submodules are pretty awful. 

If there are any other steps like this, then maybe another solution would be to provide a script that the user could run post-installation in order to replace the downloaded code w/ a cloned repo, etc. Maybe not worth it if this is the only thing, though.

🤷‍♂ , not a big deal either way. Go with whatever feels best to you.

readme.md
iandunn: 🤔 

huh. Maybe this is another case where we'd want [a script that the user runs on the host after installation](https://github.com/WordPress/wordcamp.org/pull/113#discussion_r284009060)?

Then it seems like we couldn't need to install anything during the initial build.

.docker/install-plugin-theme.sh
vedanshujain: For now lets keep these separate. If people complain about this, then we will revisit and make the change. Resolving the conversation for now

.docker/wp-config.php
vedanshujain: Created an issue to track it. Resolving the conversation now. https://github.com/WordPress/wordcamp.org/issues/120

docker-compose.yml
vedanshujain: I still lean towards not having to create an environment variable. I have not tested it out, but it looks like `COMPOSE_FILE` is used to specify multiple compose files, instead of defining a path. Path can simply be specified while executing `docker-compose` command. 

I am resolving this conversation now, but please re-open this if you feel we should rethink.

readme.md
vedanshujain: We already have a script for installing plugins and themes in `install-pluign-theme.sh`. Perhaps we can mark this step as mandatory instead of optional. What do you think?

iandunn: That sounds good to me, since they won't really have a completely functioning environment until they do that.

vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/113/commits/8e679936b8e51e534ca63ea91ea26639384690de

.docker/Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+RUN cd ~ && openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout wordcamp.key -out wordcamp.crt -config /var/wordcamp.ssl.conf
+RUN cp ~/wordcamp.crt /etc/ssl/certs/wordcamp.crt
+RUN cp ~/wordcamp.key /etc/ssl/private/wordcamp.key
+
+# Add script to install plugin and themes
+ADD install-plugin-theme.sh /usr/src/public_html/install-plugin-theme.sh
+
+# Unfortunately themes and plugins cannot be installed as part of docker setup and have to be run manually.
+# This is because everything in wp-content folder will be overwritten when we mount this directory from volume.
+# RUN sh install-plugin-theme.sh
```
iandunn: `SHOULD` - It doesn't seem like we need this comment here anymore, and maybe not the `ADD` command above, since they're artifacts of when it was run automatically.

Now that it's run manually, it seems like it makes more sense for the script to just be mentioned in the `README`, and then add a comment inside the script itself that documents why it can't be run automatically during `compose`.

.docker/Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+RUN sed -i -e "s/#FromLineOverride=YES/FromLineOverride=YES/" /etc/ssmtp/ssmtp.conf
+RUN sed -i -e "s/mailhub=mail/mailhub=127.0.0.1:1025/" /etc/ssmtp/ssmtp.conf
+RUN cp /usr/local/etc/php/php.ini-development /usr/local/etc/php/php.ini
+
+# Utils
+RUN apt-get install vim -y
+
+COPY wp-config.php /usr/src/public_html/wp-config.php
+
+# Replace nginx default conf because it conflicts with wordcamp.cong
```
iandunn:
```suggestion
# Replace nginx default conf because it conflicts with wordcamp.conf
```

`MUST`

.docker/Dockerfile
```diff
@@ -0,0 +1,120 @@
...
+RUN mv /etc/nginx/sites-enabled/wordcamp.conf /etc/nginx/sites-enabled/default
+
+CMD tail -f /dev/null
+
```
iandunn, line 115:
```suggestion

RUN echo "\n\nSuccess: The WordCamp container is now running.\n"
RUN echo "You can visit https://2014.content.wordcamp.org/ to view a sample WordCamp site. You can log in with username 'admin' and password 'password'.\n"
RUN echo "You can also visit http://localhost:1080/ to open MailCatcher."

```

`MUST` - At first I got tripped up because I didn't realize that `docker-compose` up was supposed to run indefinitely. I expected it to finish and return me to the shell, similar to a `vagrant up`, but it just sat there after starting MySQL for several minutes before I got it. This will help avoid that, and also give the user some necessary information.

Alternatively, it'd be cool if instead of showing the Nginx access log, if the terminal instead automatically entered a shell inside the container, so that you don't have to do that manually. If we don't do that, we'll probably want a quick way to enter it without having to lookup the current name and type a command. Maybe just `ssh wordcamp.test` via `~/.ssh/config` or something?

.docker/install-plugin-theme.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Install plugins
+echo "Installing plugins"
+wp --allow-root plugin install akismet buddypress bbpress jetpack wp-multibyte-patch wordpress-importer polldaddy liveblog wp-super-cache custom-content-width camptix-network-tools email-post-changes tagregator supportflow camptix-pagseguro camptix-payfast-gateway camptix-trustpay camptix-trustcard camptix-mercadopago camptix-kdcpay-gateway campt-indian-payment-gateway camptix
+
+# Install themes
+echo "Installing themes"
+wp --allow-root theme install twentyten twentyeleven twentytwelve twentythirteen twentyfourteen twentyfifteen twentysixteen twentyseventeen twentyeighteen p2
```
iandunn:
```suggestion
wp --allow-root theme install twentyten twentyeleven twentytwelve twentythirteen twentyfourteen twentyfifteen twentysixteen twentyseventeen twentynineteen p2
```

`MUST` - There wasn't a `twentyeighteen` theme, we skipped that year, but there is a `twentynineteen`

```sh
Installing Twenty Seventeen (2.2)
Downloading installation package from https://downloads.wordpress.org/theme/twentyseventeen.2.2.zip...
The authenticity of twentyseventeen.2.2.zip could not be verified as no signature was found.
Unpacking the package...
Installing the theme...
Theme installed successfully.
Warning: twentyeighteen: Theme not found
Warning: The 'twentyeighteen' theme could not be found.
***************** ^
Installing P2 (1.5.8)
Downloading installation package from https://downloads.wordpress.org/theme/p2.1.5.8.zip...
The authenticity of p2.1.5.8.zip could not be verified as no signature was found.
Unpacking the package...
Installing the theme...
Theme installed successfully.
Error: Only installed 9 of 10 themes.
***************** ^
```

.docker/nginx.conf
```diff
@@ -0,0 +1,57 @@
+# Upstream to abstract backend connection(s) for PHP
+upstream php {
+        server unix:/tmp/php-cgi.socket;
+        server 127.0.0.1:9000;
+}
+
+server {
+    listen 80 default_server;
```
iandunn, line 8: `MIGHT` - I wonder if it'd be better to use non-standard ports for this and `443`? I have a local Nginx install that I leave running 24/7, and it has a bunch of other sites in addition to WordCamp.test. It'd be really inconvenient to have to stop this container, and restart the local Nginx/MySQL/PHP servers every time I want to change to another site.

It seems like that might be common, but if I'm the only one, though, maybe I can work out another solution. It might be normal for people to only run one container at a time?

.docker/php-fpm.conf
```diff
@@ -0,0 +1,6 @@
+[global]
+
+[www]
+php_flag[display_errors] = on
+php_admin_value[error_log] = /var/log/fpm-php.www.log
+php_admin_flag[log_errors] = on
```
iandunn, line 6:
```suggestion
php_admin_flag[log_errors] = on
php_admin_flag[post_max_size] = 100M
php_admin_flag[upload_max_filesize] = 100M
```

`SHOULD` - Now that this is in a separate file, the `if you'd like too...` seems like an artifact leftover from when this sentence was next to one referring to the Meta Environment. If they're loading this file, we can assume they just want the Docker process.

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+**Note:** This will create `.docker/database` directory which will contain MySQL files to persist data across docker restarts.
+
+1. Make sure you have Docker installed and Docker daemon running on your system.
+
+1. Clone the repo: 
+    ```
+    git clone git@github.com:WordPress/wordcamp.org.git
+    ```
+
+1. Change into the directory, and run docker compose command. This could take some time depending upon the internet speed.
```
iandunn:
```suggestion
1. Change into the directory, and run Docker's `compose` command. This could take some time depending upon the speed of your Internet connection.
```

`SHOULD` (or something similar)

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+    ```
+    cd wordcamp.org && docker-compose up --build
+    ```
+     This will also start the docker environment for WordCamp.org development.
```
iandunn:
```suggestion
     This will also start the docker environment for WordCamp.org development. This command will continue running as long as the environment is running.
```

`SHOULD` - At first I got tripped up because I didn't realize that that command was supposed to run indefinitely. I expected it to finish and return me to the shell, similar to a `vagrant up`, but it just sat there after starting MySQL for several minutes before I got it.

`MIGHT` - It'd also be nice to provide a way to just load the container when the user logins in to their OS account, rather than having to manually start it and leave it running in a terminal which then can't be used for anything else.

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+1. Add the following resolutions to your host file:
+    ```
+    127.0.0.1	central.wordcamp.test
+    127.0.0.1	wordcamp.test
+    127.0.0.1	2014.content.wordcamp.test
+    127.0.0.1	2014.misc.wordcamp.test
+    127.0.0.1	2014.new-site.wordcamp.test
+    ```
+    
+    `/wp-admin` pages for these sites should now be accessible. Use `admin` as username and `password` as password to login.
```
iandunn:
```suggestion
    `/wp-admin` pages for these sites should now be accessible. Use `admin` as username and `password` as password to login. Front end pages will not be accessible until you complete the remaining steps.
```

`SHOULD`

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+    
+    **Note:** `https` URL scheme must be used to visit these sites. Security exception will be required in first time run.
+    
+1. Basic installation comes without themes and with only very basic plugins. Add plugins and themes which are installed in wordcamp.org if needed like so:
```
iandunn:
```suggestion
The installation doesn't have any 3rd-party plugins or themes yet, you must add them like so:
```

`SHOULD` - it seems like the previous explanation is an artifact from before this step became required.

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+
+    1. Find the docker which is running WordCamp. It will start with `wordcamporg_wordcamp.test` and will most likely be `wordcamporg_wordcamp.test_1`
```
iandunn:
```suggestion
     1. Find the WordCamp container by running `docker ps`. It will be named similar to `wordcamp-docker_wordcamp.test_1`
```

`SHOULD` - this will help them find the exact name quickly if they're new to docker

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+    
+    1. Go inside this docker:
+    
+       **Note:** You might need to change `wordcamporg_wordcamp.test_1` to whatever is docker name in your setup.
```
iandunn:
```suggestion
```

`SHOULD` - This probably isn't needed if we provide them `docker ps` above

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+        ```bash
+           docker exec -it wordcamporg_wordcamp.test_1 bash
+        ```
+    
+    1. Run predefined command for installing plugins and themes. This could take some time depending on your internet connection:
+        ```bash
+           sh install-plugin-theme.sh
```
iandunn, line 44: `MIGHT` - I'm getting some warnings:

```sh
> docker exec -it wordcamp-docker_wordcamp.test_1 bash
root@878d69d889f9:/usr/src/public_html# sh install-plugin-theme.sh
Installing plugins
Installing Akismet Anti-Spam (4.1.2)
Downloading installation package from https://downloads.wordpress.org/plugin/akismet.4.1.2.zip...
The authenticity of akismet.4.1.2.zip could not be verified as no signature was found.
************ ^
Unpacking the package...
Installing the plugin...
Plugin installed successfully.
```

I've never seen that before. It could be related to the 5.2 package signing stuff? Not sure.

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+        ```
+        
+    1. Activate plugin or apply theme as needed.
```
iandunn: `MUST` - Shouldn't the default DB already have them activated?

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+    
+1. By default, docker will start with data defined in `.docker/wordcamp_dev.sql` and changes to data will be persisted across runs in `.docker/database`. To start with different database, delete `.docker/database` directory and replace the `.docker/wordcamp_dev.sql` file and run `docker-compose up --build` again.
+
+1. Note that if you want to work on WordCamp blocks, [you would have to install all the node dependencies](../public_html/wp-content/mu-plugins/blocks/readme.md). This can be done either inside, or even from outside the Docker.
+
+After first time provisioning, docker can be started by using `docker-compose up` command from inside the directory. 
+
+Once the Docker instance has started, you can visit [2014.content.wordcamp.org](2014.content.wordcamp.org) to view a sample WordCamp site. WordCamp central would be [central.wordcamp.test](central.wordcamp.test). You can also visit [localhost:1080](localhost:1080) to view the MailCatcher dashboard.
```
iandunn, line 111:
```suggestion
Once the Docker instance has started, you can visit [2014.content.wordcamp.org](2014.content.wordcamp.org) to view a sample WordCamp site. WordCamp central would be [central.wordcamp.test](central.wordcamp.test). You can also visit [localhost:1080](localhost:1080) to view the MailCatcher dashboard.

To stop the WordCamp container, press `ctrl + c` in the terminal window that's running the `docker-compose` command.
```

`MUST` - Otherwise new Docker users won't know how to close it. I'm not sure if `ctrl-c` is the recommended way, or if there's something better that shuts it down gracefully.

.docker/install-plugin-theme.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Install plugins
+echo "Installing plugins"
+wp --allow-root plugin install akismet buddypress bbpress jetpack wp-multibyte-patch wordpress-importer polldaddy liveblog wp-super-cache custom-content-width camptix-network-tools email-post-changes tagregator supportflow camptix-pagseguro camptix-payfast-gateway camptix-trustpay camptix-trustcard camptix-mercadopago camptix-kdcpay-gateway campt-indian-payment-gateway camptix
```
iandunn:
```suggestion
wp --allow-root plugin install akismet buddypress bbpress  classic-editor jetpack wp-multibyte-patch wordpress-importer polldaddy liveblog wp-super-cache custom-content-width camptix-network-tools email-post-changes tagregator supportflow camptix-pagseguro camptix-payfast-gateway camptix-trustpay camptix-trustcard camptix-mercadopago camptix-kdcpay-gateway campt-indian-payment-gateway camptix
```

`MUST` - I'm seeing this at https://wordcamp.test/wp-admin/network/upgrade.php

```
The following plugins are missing:

    classic-editor/classic-editor.php
    wporg-profiles-wp-activity-notifier/wporg-profiles-wp-activity-notifier.php
    json-rest-api/plugin.php

Please let a WordCamp.org developer know about this message.
```

`wporg-profiles-wp-activity-notifier` comes from the Meta repo. `json-rest-api` can be left out since we'll remove it soon anyway.

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
+If you'd like to only setup WordCamp.org and would like to use Docker, you can use inbuilt docker provisioning. Please follow these steps
```
iandunn:
```suggestion
Follow these steps to setup WordCamp.org using [Docker](https://www.docker.com/).
```

`SHOULD` - Now that this is in a separate file, the if you'd like too... seems like an artifact leftover from when this sentence was next to one referring to the Meta Environment. If they're loading this file, we can assume they just want the Docker process.

.docker/readme.md
```diff
@@ -0,0 +1,55 @@
...
+    1. Run predefined command for installing plugins and themes. This could take some time depending on your internet connection:
+        ```bash
+           sh install-plugin-theme.sh
+        ```
+        
+    1. Activate plugin or apply theme as needed.
+    
+1. By default, docker will start with data defined in `.docker/wordcamp_dev.sql` and changes to data will be persisted across runs in `.docker/database`. To start with different database, delete `.docker/database` directory and replace the `.docker/wordcamp_dev.sql` file and run `docker-compose up --build` again.
+
+1. Note that if you want to work on WordCamp blocks, [you would have to install all the node dependencies](../public_html/wp-content/mu-plugins/blocks/readme.md). This can be done either inside, or even from outside the Docker.
```
iandunn, line 49: `SHOULD` - can this be automated in the install plugin script? If so that should maybe be renamed to `post-compose.sh` or something.

There's probably some other stuff that needs to be built, like the Application Tracker.

.docker/wp-cli.yml
```diff
@@ -0,0 +1,2 @@
+path: /usr/src/public_html/mu
+url: wordcamp.test
```
iandunn, line 2:
```suggestion
```

`MIGHT` - We can't use that on sandboxes/production, so it might be best to just leave it out. Alternatively, we could `gitignore` this file? But it would be nice to have one on production too, so maybe the script that pushes to the SVN repo could swap `.test` for `.org` or something?

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+services:
+  wordcamp.test:
```
iandunn: `MIGHT` - Can we use tabs for indentation instead of spaces, to be consistent? IIRC YAML allows that.

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    volumes:
+      - "./public_html/wp-content:/usr/src/public_html/wp-content"
+      - ".docker/wp-config.php:/usr/src/public_html/wp-config.php"
+    ports:
+      - "80:80"
+      - "443:443"
+      - "1080:1080"
+    command: >
+      bash -c "touch /usr/src/public_html/wp-content/mu-plugins/blocks/build/blocks.min.js &&
+               touch /usr/src/public_html/wp-content/mu-plugins/blocks/build/blocks.min.css &&
```
iandunn: `MIGHT`  - This won't be necessary if we automate the building of this and other things like the Application Tracker

public_html/wp-content/mu-plugins/wcorg-json-api.php
```diff
@@ -361,9 +361,11 @@ function wcorg_json_v2_compat( $request ) {
 	}
 
 	// Route v2 requests to Core handler
-	$request->query_vars['rest_route'] = $request->query_vars['json_route'];
-	unset( $request->query_vars['json_route'] );
-	$request->matched_query = preg_replace( '#^json_route=(.+)$#', 'rest_route=$1', $request->matched_query );
+	if ( isset( $request->query_vars['json_route'] ) ) {
```
iandunn, line 364: `MUST` - Why is this change necessary? Because the `json-rest-api` plugin isn't available in the w.org repo anymore? Do the v1 and v2 endpoints still work on your sandbox with this change?

iandunn (changes requested):
Great job, kudos!

There's some smaller things I think should change, but nothing big.

public_html/wp-content/mu-plugins/wcorg-json-api.php
vedanshujain: Yup, thats because `json-rest-api` plugin isn't available in w.org repo. I tested out few `v2` endpoints, and some `json_route` endpoints like `?json_route=/posts/types/page` and they seem to work.

Is there anything specific you have in mind that we should test?

.docker/Dockerfile
vedanshujain: I moved the comment over to `readme.md` and removed the `RUN` command. Although, we still need `ADD` command to actually push the file inside docker environment.

vedanshujain: ^ this was done in https://github.com/WordPress/wordcamp.org/pull/113/commits/41867e7168934f7be3093a8ade78a9c1d3604520

.docker/Dockerfile
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/41867e7168934f7be3093a8ade78a9c1d3604520

.docker/Dockerfile
vedanshujain: I have edited Docker's `readme.md` and changed some of the provisioning command.
`docker-compose up --build -d` will exit when provisioning is done and docker instance will keep on running in background. 

Also `docker-compose exec wordcamp.test bash` can be used to go inside the shell. See https://github.com/WordPress/wordcamp.org/pull/113/commits/ac11b763700e157d761baedda4253bf3722d6e60

Please resolve the conversation if you feel that this addressed your issues.

.docker/install-plugin-theme.sh
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/113/commits/41867e7168934f7be3093a8ade78a9c1d3604520

.docker/nginx.conf
vedanshujain: So this configuration will actually go into `docker-compose.yml:11` where you can choose a different port on your host machine to map to port 80 of Docker's webserver. For egs, we map host's 3307 to MySQL Docker's 3306 to avoid conflict. 

That said, using any port other than 80 might not work because lots of image path are hardcoded without port number in the database. So we would have to clean that first, which should ideally be a different PR.

You can comment out the line in `docker-compose.yml` in your local setup if docker-setup conflicts with your local one.

.docker/php-fpm.conf
vedanshujain: Added in https://github.com/WordPress/wordcamp.org/pull/113/commits/936ea89eb0aaaedcc90dda6925af03fdac01ad41

.docker/readme.md
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/113/commits/41867e7168934f7be3093a8ade78a9c1d3604520

.docker/readme.md
vedanshujain: Default provisioning command are now changed in https://github.com/WordPress/wordcamp.org/pull/113/commits/ac11b763700e157d761baedda4253bf3722d6e60 to exit when its complete, instances will keep on running in background. 

That command can also be now added to host's startup script to keep the Docker service running.

.docker/readme.md
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/ac11b763700e157d761baedda4253bf3722d6e60

.docker/readme.md
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/ac11b763700e157d761baedda4253bf3722d6e60

.docker/readme.md
vedanshujain: Changed the default scripts in https://github.com/WordPress/wordcamp.org/pull/113/commits/ac11b763700e157d761baedda4253bf3722d6e60. 

This can be done by using `docker-compose exec wordcamp.test bash`

.docker/readme.md
vedanshujain: Changed in https://github.com/WordPress/wordcamp.org/pull/113/commits/ac11b763700e157d761baedda4253bf3722d6e60

.docker/readme.md
vedanshujain: 🤔 Looks like it.

.docker/readme.md
vedanshujain: It is. removed this line. https://github.com/WordPress/wordcamp.org/pull/113/commits/ac11b763700e157d761baedda4253bf3722d6e60

.docker/readme.md
vedanshujain: Instructions are now changed, so this is not a problem anymore because `compose` command will automatically exit using `-d` flag. To stop the container use `docker-compose stop`. This is documented in https://github.com/WordPress/wordcamp.org/pull/113/commits/ac11b763700e157d761baedda4253bf3722d6e60

.docker/install-plugin-theme.sh
vedanshujain: Added `classic-editor` in https://github.com/WordPress/wordcamp.org/pull/113/commits/41867e7168934f7be3093a8ade78a9c1d3604520

.docker/readme.md
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/41867e7168934f7be3093a8ade78a9c1d3604520

.docker/readme.md
vedanshujain: Having a separate that automatically does all this is a good idea, but perhaps it should not be part of this PR, because it can also be used in VVV (or any other) setup. I will open an issue to track it before merging. What do you think?

.docker/wp-cli.yml
vedanshujain: Its added to `public_html` during Docker compose, but otherwise its always inside `.docker` folder. So it should not cause an issue on sandboxes/production. What do you think?

docker-compose.yml
vedanshujain: Unfortunately, [YAML does not allow tabs](https://yaml.org/faq.html).

docker-compose.yml
vedanshujain: Depends upon resolution of https://github.com/WordPress/wordcamp.org/pull/113#discussion_r285101877

public_html/wp-content/mu-plugins/wcorg-json-api.php
iandunn: Nope, that sounds good to me. I think this should be a separate commit from the other Docker stuff, and the commit message should document that it's because the plugin isn't in the repo (publicly) anymore. Looks good though :)

.docker/Dockerfile
iandunn: 👍

.docker/Dockerfile
iandunn: That commit looks great!

.docker/nginx.conf
iandunn: 🤔 

Yeah, let's just go with this for now, and I can play around with it later and maybe iterate.

I wonder if there's a way to do an "local" config that overrides parts of `docker-compose.yml`, similar to how you can have `phpcs.xml.dist` committed to version control, and then an ignored `phpcs.xml` file that takes precedence.

.docker/readme.md
iandunn: 👍

iandunn: > That command can also be now added to host's startup script to keep the Docker service running.

I'll look into that when I make the switch, and then add instructions to the README.

.docker/readme.md
iandunn: Nice! That's much better.

.docker/readme.md
iandunn: 👍 Probably nothing for us to worry about here, then.

.docker/readme.md
iandunn: 👍

.docker/readme.md
iandunn: 👍

.docker/wp-cli.yml
iandunn: 👍

docker-compose.yml
iandunn: huh, ok :)

New issue: from that FAQ it sounds like maybe we should use `.yaml` instead of `.yml`, but not a big deal.

iandunn (approved):
Great work!

There're a few unresolved threads but they're not blockers.

iandunn:
`MUST` - Oh, one more thing I just noticed:

```sh
> git stat
?? .docker/database/
?? public_html/wp-content/plugins/akismet/
?? public_html/wp-content/plugins/bbpress/
?? public_html/wp-content/plugins/buddypress/
?? public_html/wp-content/plugins/campt-indian-payment-gateway/
?? public_html/wp-content/plugins/camptix-kdcpay-gateway/
?? public_html/wp-content/plugins/camptix-mercadopago/
?? public_html/wp-content/plugins/camptix-pagseguro/
?? public_html/wp-content/plugins/camptix-payfast-gateway/
?? public_html/wp-content/plugins/camptix-trustcard/
?? public_html/wp-content/plugins/camptix-trustpay/
?? public_html/wp-content/plugins/camptix/
?? public_html/wp-content/plugins/custom-content-width/
?? public_html/wp-content/plugins/email-post-changes/
?? public_html/wp-content/plugins/jetpack/
?? public_html/wp-content/plugins/liveblog/
?? public_html/wp-content/plugins/polldaddy/
?? public_html/wp-content/plugins/supportflow/
?? public_html/wp-content/plugins/tagregator/
?? public_html/wp-content/plugins/wordpress-importer/
?? public_html/wp-content/plugins/wp-multibyte-patch/
?? public_html/wp-content/plugins/wp-super-cache/
?? public_html/wp-content/themes/p2/
?? public_html/wp-content/themes/twentyeleven/
?? public_html/wp-content/themes/twentyfifteen/
?? public_html/wp-content/themes/twentyfourteen/
?? public_html/wp-content/themes/twentyseventeen/
?? public_html/wp-content/themes/twentysixteen/
?? public_html/wp-content/themes/twentyten/
?? public_html/wp-content/themes/twentythirteen/
?? public_html/wp-content/themes/twentytwelve/
```

Our "svn to git migration" doc pushes the 3rd-party plugin/themes to `.git/info/exclude`, but I feel like we can probably go ahead and just add them to `.gitignore` now.

Seems like `.docker/database/` can be `.gitignore`'d too.

docker-compose.yml
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/113/commits/0a485e31005fb66c47a0c5c1bfa51ad6a6ca680e

vedanshujain:
@iandunn Added plugins and themes to .gitignore

vedanshujain:
Merged in r8808-r8836

readme.md
vedanshujain: Raised https://github.com/WordPress/wordcamp.org/issues/121

.docker/readme.md
vedanshujain: Raised https://github.com/WordPress/wordcamp.org/issues/122

PR #114, opened 2019-05-07
Enhance Sessions endpoint to allow for querying by specific meta keys

### LINKED ISSUE #109: Sessions Block: Should non-"Regular" sessions be included?
There are currently two types of sessions: "Regular" and "Break, Lunch, etc.". In the old `[sessions]` shortcode, the non-regular session posts were excluded from the query.

Right now in the Sessions block, the PHP renderer excludes these sessions as well, but the backend JS UI does not. We need to decide which way to go, and/or create a setting for the block that toggles these non-"Regular" sessions on/off.

My inclination would be to omit them for now, and if people ask for a way to include them, we can add the toggle, with the default value still being "off".

coreymckrill:
This fixes #109 by allowing the `meta_key` and `meta_value` parameters to be used when requesting sessions posts from the REST API, and thus, being able to specify only sessions that have a value of `session` for the `_wcpt_session_type` post meta field. These request parameters are not available for any post type endpoints out of the box, presumably for security reasons. Here I attempt to address security concerns by only safelisting meta keys that have already been registered with `show_in_rest` set to `true`.

This also changes how we're registering meta fields to use Core's `register_post_meta` function instead of our custom, hacky `wcorg_register_meta_only_on_endpoint`, which is deleted. 🎉


public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -6,51 +6,59 @@
...
+			),
+		],
+	];
 
-	$public_sponsor_fields = array(
-		'_wcpt_sponsor_website' => array(
-			'single' => true,
-		),
-	);
+	foreach ( $public_meta_fields as $post_type => $fields ) {
```
iandunn: I think defining everything in an array made since in [the original commit](https://github.com/WordPress/wordcamp.org/commit/5f3d839075dbeaad9e3f30526e9aa32a2aa3bc54), but now it makes things really hard to read. I had to slowly go through each line and reference other lines to make sure I understand what each var was.

`SHOULD` - I think this is a great example of where [being "clever" is bad](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#clever-code), and refactoring to a more straight-forward approach would greatly benefit readability/maintainability.

```php
$args = array(
	'single'       => true,
	'show_in_rest' => true,
);

register_post_meta( 'wcb_session', _wcpt_session_time',   array_merge( $args, array( 'type' => 'number' ) ) );
register_post_meta( 'wcb_session', _wcpt_session_type',   $args );
register_post_meta( 'wcb_session', _wcpt_session_slides', $args );
register_post_meta( 'wcb_session', _wcpt_session_video',  $args );
// etc
```

public_html/wp-content/mu-plugins/blocks/assets/src/sessions/edit.js
```diff
@@ -47,8 +47,14 @@ class SessionsEdit extends Component {
 const sessionsSelect = ( select ) => {
 	const { getEntities } = select( WC_BLOCKS_STORE );
 
+	const sessionArgs = {
+		_embed     : true,
+		meta_key   : '_wcpt_session_type',
+		meta_value : 'session',
```
iandunn:
```suggestion
		meta_value : 'session', // Regular sessions only, no breaks/lunch/etc sessions.
```

`'_wcpt_session_type' => 'session'` 

😱 

"session" is a really undescriptive name for that :(

`SHOULD` - It'd be too much of a hassle to go back through all the legacy code and data and fix it, but it might be worth adding a comment here indicating that "session" means "regular", as opposed to "custom" -- also poorly named -- which are breaks.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -147,8 +155,60 @@ function register_additional_rest_fields() {
...
+function prepare_meta_query_args( $args, $request ) {
+	if ( isset( $request['meta_key'], $request['meta_value'] ) ) {
+		$args['meta_key']   = $request['meta_key'];
+		$args['meta_value'] = $request['meta_value'];
+	}
+
+	return $args;
+}
+
+add_filter( 'rest_wcb_session_query', __NAMESPACE__ . '\prepare_meta_query_args', 10, 2 );
```
iandunn, line 153: > These request parameters are not available for any post type endpoints out of the box, presumably for security reasons.

🤔 

The *fields* aren't included in the response by default, [because of security concerns](https://github.com/WP-API/WP-API/pull/135), but it does seem a bit odd that the `meta_key`/`meta_value` *parameters* aren't enabled by default. I would have thought they'd been enabled, but would ignore any data that wasn't registered with `show_in_rest => true`.

I wonder if there's some reason we shouldn't enable them, or need to be more cautious when using them? 

It feels weird that we need to enable those params, rather than just registering the data, since exposing that data is a common use case, and it'd be really weird to require all devs to enable those fields any time they want to do something simple like this. 

- [x] `SHOULD` - This feels like it might need a bit more research, unless you've already done that and found something to indicate this is the recommended way. If it is necessary, I imagine there's an existing ticket complaining about it :)

- [x] `SHOULD` - At the very least, we should include a link to that ticket as a `@todo` note so we can remove this extra code when Core provides a better way to do it.

- [x] `SHOULD` - If that ticket doesn't exist, it seems like it'd be good to create it :)

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -147,8 +155,60 @@ function register_additional_rest_fields() {
...
+ * safelisted by filtering for ones that have `show_in_rest` set to `true`.
+ *
+ * @param array        $query_params
+ * @param WP_Post_Type $post_type
+ *
+ * @return array
+ */
+function add_meta_collection_params( $query_params, $post_type ) {
+	$public_meta_fields = array_filter( get_registered_meta_keys( 'post', $post_type->name ), function( $registered ) {
+		return $registered['show_in_rest'];
```
iandunn: > Here I attempt to address security concerns by only safelisting meta keys that have already been registered with show_in_rest set to true.

That feels like a good solution, if it turns out that we need to keep this and `prepare_meta_query_args()`.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -6,51 +6,59 @@
...
+		foreach ( $fields as $field_key => $field_args ) {
+			$args = array_merge(
+				$field_args,
+				[
+					'show_in_rest' => true,
+				]
+			);
 
-	wcorg_register_meta_only_on_endpoint( 'post', $public_sponsor_fields, '/wp-json/wp/v2/sponsors/' );
+			register_post_meta( $post_type, $field_key, $args );
```
iandunn: When you view `wp-json/wp/v2/session/N`, does this place the safelisted meta fields at `$session->meta->_wcpt_session_type`, or at `$session->_wcpt_session_type`?

I'm assuming (and hoping) that it's the former. If it's the latter, that would really annoy me, but would probably be a decision that Core made, so probably wouldn't be worth trying to work around anymore.

`MUST` - We might need to consider back-compat concerns, though, since we'd be breaking any clients that rely on that field being under `$session->meta` like it currently is.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -147,8 +155,60 @@ function register_additional_rest_fields() {
...
+ * This enables and validates simple meta query parameters for the Sessions endpoint. Specific meta keys are
+ * safelisted by filtering for ones that have `show_in_rest` set to `true`.
+ *
+ * @param array        $query_params
+ * @param WP_Post_Type $post_type
+ *
+ * @return array
+ */
+function add_meta_collection_params( $query_params, $post_type ) {
+	$public_meta_fields = array_filter( get_registered_meta_keys( 'post', $post_type->name ), function( $registered ) {
```
iandunn:
```suggestion
	// Avoid exposing potentially sensitive data.
	$public_meta_fields = array_filter( get_registered_meta_keys( 'post', $post_type->name ), function( $registered ) {
```

`SHOULD` - I think the security issues are implied by the docblock comment and the name of this var, but it's worth being extra safe and declaring it explicitly.

iandunn (changes requested):
Overall it looks good, but it might need more research to confirm this is the correct way to approach it, and there are a few small changes to make.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
coreymckrill: It's the former, `$session->meta->_wcpt_session_type`. There shouldn't be any backcompat issues.

iandunn: 👍

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
coreymckrill: Works for me. Changed in 2a0df824e1540bfb55c52232e4ea031868c465f2

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
coreymckrill: Before the API was merged into Core, meta queries were possible via a `filter` parameter:

```
?filter[meta_key]=foo&filter[meta_value]=bar
```

This was [removed](https://core.trac.wordpress.org/ticket/38378) when the API was merged. There ~is~ was [a plugin](https://github.com/WP-API/rest-filter) that restored this functionality, but that doesn't seem like a good long term solution. All discussions I have found so far, though, just recommend installing that plugin. My Trac-fu might be failing me, but I can't find any Core tickets about adding the ability to do meta queries.

I'm waiting to hear back in the `core-restapi` channel to see if anyone there has insight.

coreymckrill: Here's a relevant discussion:
https://wordpress.slack.com/archives/C02RQC26G/p1532356391000058

coreymckrill: Here's a conversation thread in `core-restapi` about this with Kadam:
https://wordpress.slack.com/archives/C02RQC26G/p1557261445107000

The gist:

> It's possible to use meta queries to construct a very expensive DB query, even with the hard limit of 100 posts_per_page, so in an abundance of caution we restricted passing meta queries via the REST API query for the API beta and subsequent initial core release of the plugin
> It has since become appealing to add it back in, with the improvements to meta registration you mention and the introduction of register_post_meta and friends, but it hasn't been a widespread enough request yet to prioritize

iandunn: 👍 

Sounds like we can move forward with the current approach then, but it'd be nice to add a `todo` with a link to the ticket, once you create it.

coreymckrill: Ok, here's a Trac ticket: https://core.trac.wordpress.org/ticket/47194

I'm thinking we should namespace our custom collection params since we don't know what, if anything, a Core implementation would look like.

coreymckrill: c024005856199dbdbb19ba5ab850d1a35eb5eb5d

coreymckrill:
Committed:

* https://meta.trac.wordpress.org/changeset/8755
* https://meta.trac.wordpress.org/changeset/8756
* https://meta.trac.wordpress.org/changeset/8757

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
iandunn: 👍

PR #115, opened 2019-05-09
WC Campsite17: Adds featured image support with feature flag.

vedanshujain:
In an earlier patch r8748, Jetpack's content options was used to add featured image support. For backwards compatibility, this was disabled by default using a feature flag on all existing sites.

This caused an issue because content options were removing featured images on *all* renders including in blocks and shortcode. Therefore r8748 was eventually reverted in r8753.

This patch adds featured image support natively instead of using Jetpack Content Support so as to not disable images in blocks and shortcodes. This also means that if any exisiting camp needs featured image supported, they would have to request to have the backwards compatibility flag removed for them.

See #3645 #4339


vedanshujain:
This was merged  https://meta.trac.wordpress.org/changeset/8762

PR #116, opened 2019-05-09
Consolidate and normalize image settings for all blocks

### LINKED ISSUE #91: Sessions: Image Settings
- [ ] "Image size" should be called "Image Settings"
- [ ] Image Settings should match regular core image settings UI (see Image block)
- [ ] Should be able to hide images

coreymckrill:
This consolidates several related shared components into one `image` folder. Avatars, featured images, and inspector panels and controls related to each can now all be imported from `shared/image`.

It also normalizes the image settings available in each block so that for each you can now toggle images on/off, set an image "size" (which is really width, and the height scales to maintain aspect ratio) and set an image alignment. Speakers and Organizers already had these settings, but Sponsors and Sessions were inconsistent. This fixes #91.

This also fixes some minor outstanding issues with the Sponsors block.


coreymckrill:
Committed in https://meta.trac.wordpress.org/changeset/8768

Issue #47, opened 2019-03-25
Blocks: Simplify folder structure


iandunn:
It feels like the `assets` folder adds an unnecessary lever of depth to the tree structure, since there's nothing else inside it except `src` and the build files. If anything, it feels like the build files should be in a subfolder (or separate folder), since we don't need to access them frequently, but there's only 2 of them, so that also feels like overkill. Putting the `src` files deeper makes them more inconvenient to access.

`assets` also isn't very descriptive, and doesn't seem to make sense in a component architecture where CSS, images, etc would go into the folders of their respective components, rather than being separated into folders based on file type.

The `src` folder name isn't very descriptive, and I personally hate abbreviations in most cases. What about a root-level `components` or `javascript` folder instead?

`view/` should probably be plural since it contains multiple views.

What do y'all think about something like this:

```
blocks/
      components/
            organizers/
            shared/
            ...
            blocks.js
            blocks.min.js
            blocks.min.css
      includes/
      views/
      blocks.php
      ...
```

coreymckrill:
I agree this should be simplified somehow. It seems like there are some conventions that have been widely adopted in other projects that we might look to. For example, what about putting the built files in a separate root-level folder called `dist` or `build`? That might make it easier to have VCS's ignore those files in certain environments, especially if we eventually have more than the two current files.

iandunn:
That WFM. I'd lean towards `build` since it's not abbreviated 😄

iandunn:
We should wait to do this until #44, etc are done, to minimize the merge conflicts.

iandunn:
We might want to use `src/index.js` and `build/index.js`, to watch the convetions that `@wordpress/scripts` uses.

https://make.wordpress.org/core/2019/03/25/building-javascript/

That'd make sense if we want to switch to that instead of our custom webpack/babel configs too.

iandunn:
I wonder if it'd be good to have separate folders for blocks and components, the way G does.

```
blocks/
      src/
            components/
                  avatar/
                  grid-layout/
                  ...            
            blocks/
                  organizers/
                  speakers/
                  ...
            index.js
      build/
            index.min.js
            index.min.css
      includes/
      views/
      blocks.php
      ...
```

coreymckrill:
Yeah, that works for me. We could also include an **index.js** file in each of those folders that just re-exports all of the components that are in the subfolders. That way we could do something like:

```
import { AvatarImage, GridContentLayout } from '../components';
```

[...]

PR #117, opened 2019-05-10
Blocks: Reorganize files and directories

coreymckrill:
This reorganizes the JS files for blocks into a structure that is a bit more logical and more closely matches the organization found in the Gutenberg plugin. It attempts to address the issues discussed in #47 and pretty much follows the pattern suggested [in this comment](https://github.com/WordPress/wordcamp.org/issues/47#issuecomment-485598737)

```
blocks
    build
        blocks.min.css
        blocks.min.js
    source
        block
            organizers
            sessions
            speakers
            sponsors
        component
            block-content
            block-controls
            image
            item-select
            post-list
        data
        i18n
        blocks.js
```


public_html/wp-content/mu-plugins/blocks/source/block/index.js
```diff
@@ -0,0 +1,11 @@
+import * as organizers from './organizers';
```
iandunn: `SHOULD` - Should the parent folder be named `blocks/` (plural), since it contains multiple, and for consistency with G? Same for `components/`, `views/`, the `BLOCK` constant, etc.

iandunn: `SHOULD` - Normally I'd definitely prefer `source` to `src`, but in this case I'm wondering if the grandparent folder be  `src`, to match what `wp-scripts` expects? That'll make it easier to eventually migrate to that instead of our custom configs (assuming that's possible).

public_html/wp-content/mu-plugins/blocks/blocks.php
```diff
@@ -52,14 +52,14 @@ function register_block_categories( $default_categories ) {
 function register_assets() {
```
iandunn, line 62: `SHOULD` - Should we rename `includes/shared/` to `includes/components` too, so it's consistent w/ the JS? Maybe move the block PHP files into `includes/blocks` for the same reason?

public_html/wp-content/mu-plugins/blocks/source/block/sessions/block-content.js
```diff
@@ -13,11 +13,11 @@ const { __ }        = wp.i18n;
...
-import { ItemTitle, ItemHTMLContent, ItemPermalink, BlockNoContent } from '../shared/block-content';
-import { tokenSplit, arrayTokenReplace, intersperse, listify }       from '../shared/i18n';
-import PostList                                                      from '../shared/post-list';
-import { FeaturedImage }                                             from '../shared/image';
-import { filterEntities }                                            from '../blocks-store';
+import { ItemTitle, ItemHTMLContent, ItemPermalink, BlockNoContent } from '../../component/block-content';
+import { FeaturedImage }                                             from '../../component/image';
+import { PostList }                                                  from '../../component/post-list';
+import { filterEntities }                                            from '../../data';
+import { tokenSplit, arrayTokenReplace, intersperse, listify }       from '../../i18n';
```
iandunn: `SHOULD` - I wonder if `helpers/i18n.js` would be more future-proof? We probably won't have more custom i18n stuff, but it seems somewhat likely that we'll have more generic helpers.

I'm open to other terms besides "helpers" too, that was just the first that came to mind.

public_html/wp-content/mu-plugins/blocks/blocks.php
```diff
@@ -52,14 +52,14 @@ function register_block_categories( $default_categories ) {
...
+		PLUGIN_URL . 'build/blocks.min.css',
 		[],
-		filemtime( PLUGIN_DIR . 'assets/blocks.min.css' )
+		filemtime( PLUGIN_DIR . 'build/blocks.min.css' )
 	);
 
 	wp_register_script(
 		'wordcamp-blocks',
-		PLUGIN_URL . 'assets/blocks.min.js',
+		PLUGIN_URL . 'build/blocks.min.js',
```
iandunn, line 72: I like `blocks.min.js` better than `index.min.js` 👍

public_html/wp-content/mu-plugins/blocks/source/block/organizers/index.js
```diff
@@ -6,22 +6,22 @@ const { __ } = wp.i18n;
...
 
 const supports = {
 	align: [ 'wide', 'full' ],
 };
 
-export const settings = {
-	title       : __( 'Organizers',                'wordcamporg' ),
+export const SETTINGS = {
+	title       : __( 'Organizers', 'wordcamporg' ),
 	description : __( 'Add a list of organizers.', 'wordcamporg' ),
```
iandunn: `SHOULD` - Why remove the alignment here?

public_html/wp-content/mu-plugins/blocks/source/block/sessions/block-content.js
```diff
@@ -13,11 +13,11 @@ const { __ }        = wp.i18n;
...
  */
-import { ItemTitle, ItemHTMLContent, ItemPermalink, BlockNoContent } from '../shared/block-content';
-import { tokenSplit, arrayTokenReplace, intersperse, listify }       from '../shared/i18n';
-import PostList                                                      from '../shared/post-list';
-import { FeaturedImage }                                             from '../shared/image';
-import { filterEntities }                                            from '../blocks-store';
+import { ItemTitle, ItemHTMLContent, ItemPermalink, BlockNoContent } from '../../component/block-content';
+import { FeaturedImage }                                             from '../../component/image';
+import { PostList }                                                  from '../../component/post-list';
+import { filterEntities }                                            from '../../data';
```
iandunn: I like having a separate folder for `data` 👍 

`MIGHT` - I wonder if `stores` or `data-stores` would be a more descriptive name, since `data` is very generic. I can see it fitting in this context, though.

public_html/wp-content/mu-plugins/blocks/source/block/sessions/index.js
```diff
@@ -6,22 +6,22 @@ const { __ } = wp.i18n;
...
 
-export const settings = {
+export const SETTINGS = {
 	title       : __( 'Sessions', 'wordcamporg' ),
 	description : __( 'Add a list of sessions.', 'wordcamporg' ),
 	icon        : ICON,
 	category    : 'wordcamp',
 	supports    : supports,
-	edit        : edit,
+	edit        : Edit,
```
iandunn: Nice work making things like this consistent ❤️

public_html/wp-content/mu-plugins/blocks/source/block/speakers/block-content.js
```diff
@@ -15,11 +15,11 @@ const { escapeAttribute }     = wp.escapeHtml;
...
-import { AvatarImage }                                               from '../shared/image';
-import { ItemTitle, ItemHTMLContent, ItemPermalink, BlockNoContent } from '../shared/block-content';
-import { tokenSplit, arrayTokenReplace }                             from '../shared/i18n';
-import PostList                                                      from '../shared/post-list';
-import { filterEntities }                                            from '../blocks-store';
+import { ItemTitle, ItemHTMLContent, ItemPermalink, BlockNoContent } from '../../component/block-content';
+import { AvatarImage }                                               from '../../component/image';
+import { PostList }                                                  from '../../component/post-list';
+import { filterEntities }                                            from '../../data';
+import { tokenSplit, arrayTokenReplace }                             from '../../i18n';
```
iandunn: ❤️ alphabetizing :)

public_html/wp-content/mu-plugins/blocks/source/blocks.js
```diff
@@ -0,0 +1,13 @@
...
+ * WordPress dependencies
+ */
+const { registerBlockType } = wp.blocks;
+
+/**
+ * Internal dependencies
+ */
+import { BLOCKS } from './block';
+
+BLOCKS.forEach( ( { NAME, SETTINGS } ) => {
```
iandunn, line 11: `SHOULD` - When these are passed to the closure, it seems like they're no longer constants, but instead local vars that would be camelCase.

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-content.js
```diff
@@ -19,7 +19,7 @@ import { filterEntities }                             from '../../data';
 /**
  * Component for displaying the block content.
  */
-class OrganizersBlockContent extends Component {
+export class BlockContent extends Component {
```
iandunn: ❤️ 

I'm guessing you're planning to do this for the rest of the blocks/components too?

public_html/wp-content/mu-plugins/blocks/source/block/organizers/inspector-controls.js
```diff
@@ -2,7 +2,7 @@
  * WordPress dependencies
  */
 const { PanelBody, PanelRow, SelectControl, ToggleControl } = wp.components;
-const { InspectorControls }                                 = wp.editor;
+const { InspectorControls: Inspector }                      = wp.editor;
```
iandunn: 🤔 

`SHOULD` - This feels kinda wrong to me. If there's a conflict, it seems like our custom components should be the ones be named more descriptively, rather than essential G components.

I'm not entirely sure either way, though.

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-controls.js
```diff
@@ -7,28 +7,31 @@ import classnames from 'classnames';
  * WordPress dependencies
```
iandunn: ❤️

iandunn (changes requested):
Looks great overall. Left some comments that we'll want to resolve before merging.

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-controls.js
coreymckrill: I can't tell what you're heart-ing here. I'll just take it as a general sign of encouragement 😁

public_html/wp-content/mu-plugins/blocks/source/block/organizers/inspector-controls.js
coreymckrill: I thought about this too. I could flip it so ours is the one named just `Inspector`, but then I'd probably want to rename the file to just `inspector.js` as well.

On the other hand, the Gutenberg element really is the whole Inspector Sidebar, while ours is just a component that adds some additional inspector controls, so semantically this makes more sense to me, even if we're renaming the Core comonent (which, yes, is due to a naming conflict)

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-content.js
coreymckrill: Yep, just committed the other blocks as well.

public_html/wp-content/mu-plugins/blocks/source/block/index.js
vedanshujain: I'd actually prefer `src`, because what I see more commonplace. 
I'm curious, can `wp-scripts` could be configured to use with `source`? If so then might as well keep it this way.

public_html/wp-content/mu-plugins/blocks/source/block/sessions/block-content.js
coreymckrill: `data` is what the data stores package is called in Gutenberg, so 🤷‍♂

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-controls.js
```diff
@@ -7,49 +7,52 @@ import classnames from 'classnames';
...
  */
-import { BlockControls, PlaceholderSpecificMode } from '../shared/block-controls';
-import OrganizersBlockContent                     from './block-content';
-import OrganizersSelect                           from './organizers-select';
-import { LABEL }                                  from './index';
+import { PlaceholderSpecificMode } from '../../component/block-controls';
+import { getOptionLabel }          from '../../component/item-select';
+import { BlockContent }            from './block-content';
+import { Select }                  from './select';
+import { LABEL }                   from './index';
```
vedanshujain: Generic names here looks way cleaner, but it could cause problem down the line if need to grep for stuff, or if we have conflict with another library. Although I don't feel strongly about this so feel free to keep however :) .

public_html/wp-content/mu-plugins/blocks/source/block/sessions/block-content.js
coreymckrill: This follows the convention of Gutenberg, which also has a `i18n` package. I don't think it's a problem to have this separate directory with only a few functions in it. In fact, I think that having more, explicitly named folders will make it easier for people to browse the code and see what kinds of "helper" functions are available.

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-controls.js
vedanshujain: That said, perhaps we can think about renaming `Select` to something else. Its way to generic IMO.

public_html/wp-content/mu-plugins/blocks/source/block/index.js
coreymckrill: Aw man, I picked `source` because I remembered that you preferred spelling things out! 😆 

It's unclear to me right now if/how we would migrate to a pure usage of `wp-scripts` instead of our own config, but if we did, I think it would be easy to change it to `src` at that time if necessary. There are very few places in the code that need to reference that part of the path.

vedanshujain (approved):
Left some pretty minor comments but this looks awesome 👍

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-controls.js
coreymckrill: Fair point. What about `ContentSelect`?

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-controls.js
iandunn: Er, yeah, it makes more sense if you [view it in the context of the commit](https://github.com/WordPress/wordcamp.org/pull/117/commits/e9bcd2897aff6b1682a68496592134049d1d0e40), which is what I was looking at when I wrote it, but even then it's not entirely clear.

I was referring to the commit message (or, really, to the intent behind the commit, not the literal message, but that was fine too). It doesn't really apply to a specific line, but GitHub doesn't always let me comment on the commit as a whole (haven't noticed the pattern to when it does and doesn't, maybe it's only enabled for ones that are merged?)

Anyway, I should have quoted the commit message before my reply to make that clear :)

public_html/wp-content/mu-plugins/blocks/source/block/organizers/inspector-controls.js
iandunn: 🤔 

Well, to further complicate things, to me `Inspector` makes it sound like it's the entire `BlockInspector`, rather than just the controls for it, so I'm not sure that that name works well for either `wp.editor.InspectorControls` or our `OrganizerInspectorControls`.

iandunn: If it'd be more practical/semanic to rename `wp.editor.InspectorControls` rather than our custom components, then maybe it should be named `BlockInspectorControls` ?

That still feels odd to me, though, because I perceive the platform's components as (conceptually) immutable, whereas the custom stuff we build on top of that is malleable, and created in light of that platform's prior existence. 

To draw an analogy to PHP, I don't think we'd ever to this:

```php
<?php

namespace WordCamp\Foo;
use function wp_insert_post as insert_post;

function wp_insert_post( $arr ) {
    // do some custom wrapper processing on $arr
    insert_post( $arr );
}
```

...instead, we'd just

```php
<?php

namespace WordCamp\Foo;

function insert_post( $arr ) {
    // do some custom wrapper processing on $arr
    wp_insert_post( $arr );
}
```

iandunn: <a href="https://www.commitstrip.com/en/2015/10/27/one-of-the-coders-hardest-problems/">There are only 2 hard things in computer science...</a>

😉

public_html/wp-content/mu-plugins/blocks/source/block/index.js
iandunn: >  can wp-scripts could be configured to use with source?

I think so, but it'd require a custom webpack config, which would default a lot of the purpose of using it. There may be a way to do it with simple npm script parameters or something, though.

> It's unclear to me right now if/how we would migrate to a pure usage of wp-scripts instead of our own config, but if we did, I think it would be easy to change it to src at that time if necessary. 

Good point, I withdrawal my objection :)

coreymckrill: JS folders renamed here: 116c34c9a2b120c045d89964db1e6af8c6bc9465

public_html/wp-content/mu-plugins/blocks/source/block/sessions/block-content.js
iandunn: 👍

public_html/wp-content/mu-plugins/blocks/source/block/sessions/block-content.js
iandunn: 🤷‍♂

public_html/wp-content/mu-plugins/blocks/source/block/index.js
coreymckrill: PHP folders renamed here: dc2a6650bb6b7d8e8b87abdaba951f644ec64ee6

public_html/wp-content/mu-plugins/blocks/blocks.php
coreymckrill: PHP stuff reorganized here: dc2a6650bb6b7d8e8b87abdaba951f644ec64ee6

public_html/wp-content/mu-plugins/blocks/source/block/organizers/inspector-controls.js
coreymckrill: 💯

public_html/wp-content/mu-plugins/blocks/source/blocks.js
coreymckrill: Not sure what you're suggesting here. They are defined as constants, which is why I made them all caps (though we are not completely consistent in the codebase with this). When they are imported and deconstructed, the case of the variable names still applies, similar to how we're importing `LABEL` and `ICON` in various places. Are you suggesting we rename them during deconstruction, i.e. `{ NAME: name, SETTINGS: settings }`? That seems unnecessary to me...

public_html/wp-content/mu-plugins/blocks/source/block/organizers/index.js
coreymckrill: This made it consistent with the other blocks 🤷‍♂

iandunn: 👍 We can fix them all later

public_html/wp-content/mu-plugins/blocks/source/blocks.js
iandunn: er, yeah, good point, that would be kind of annoying. it's fine to leave it.

iandunn: Although, I don't think we should uppercase local vars that are declared with `const` instead of `let`, only "global" ones, like these.

public_html/wp-content/mu-plugins/blocks/source/block/organizers/block-controls.js
coreymckrill: Renamed to `ContentSelect`: c89a50f75be17fca4e69e78f98867e03ccd523c5

public_html/wp-content/mu-plugins/blocks/source/block/organizers/inspector-controls.js
coreymckrill: Good point that platform component names should probably be immutable. What about eschewing "Inspector" altogether and going with something like `BlockSidebar`? Same idea, and it still follows the convention we're using with `BlockControls` and `BlockContent`...

iandunn: 🤔 I don't love it, since "inspector" feels more specific and meaningful.

🤔, maybe `InspectorControlContents` ? It's a bit of a mouthful, but kinda fits. It's not specific to Organizers, but I guess the intent is to make it more generic inside the component anyway. We could even shorter it to `ControlContents` since the "inspector" part is mostly obvious from it being wrapped in an `<InspectorControl>`. 

🤔 I don't love that either, and it could also conflict if we have other types of controls inside the parent component.

🤷‍♂Oh well, it's a crapshoot, pick the one you like best and :shipit:  😄

public_html/wp-content/mu-plugins/blocks/source/blocks.js
coreymckrill: Ah, that's a good distinction.

public_html/wp-content/mu-plugins/blocks/source/block/organizers/inspector-controls.js
coreymckrill: Maybe not a popular decision, but a decision nonetheless: 8ec11263bd5c149d52dc241d398cc6cecde5e2c1

coreymckrill:
Committed: https://meta.trac.wordpress.org/changeset/8779

PR #118, opened 2019-05-11
WIP: Try using React Context to pass values to child components within a block

### LINKED ISSUE #100: Passing `...this.props` to child components causes unnecessary re-renders?
We do stuff like this a lot:

```jsx
<SessionsBlockControls
	icon={ ICON }
	{ ...this.props }
	{ ...this.state }
/>
```

I wonder if that causes `SessionsBlockControls` to re-render unnecessarily whenever something in `props` or `state` changes? It looks like the Core blocks just pass in the exact values that are needed by the child, rather than everything.

`wp.compose.pure` might avoid that, but I don't see any Core blocks using it, which makes me wonder if it's not intended to be used as often as I would think from reading about pure components. It seems like maybe just not passing in extra stuff would avoid re-renders without the need for extra memoization?

coreymckrill:
@iandunn brought up with example of using React Context:
https://lucasmreis.github.io/blog/simple-react-patterns/

Which got me thinking about ways to fix #100. This is an experiment to see if contexts will work for that problem.

Some (not that great) documentation about React Context:
https://reactjs.org/docs/context.html

One problem I ran into was that the child components weren't receiving any data into `this.context`. It turned out it was because I was defining and exporting the context component in the **edit.js** file. Since **edit** was also where the child components were getting instantiated, it was creating [a circular dependency](https://github.com/facebook/react/issues/13969#issuecomment-433253469). When I moved the context definition and export to its own file, and imported it from there into all the other files that needed it, it started working.

I'm not sure about including `attributes` in the context, since that's the thing that will actually keep changing (everything else I'm putting in it should stay static after it has loaded). Every time the context changes, everything that is a child of it will update. It might be better to just feed each child component with the attributes it needs as individual props.

A couple of other things we could add into the context: `LABEL` and `ICON`. That way we aren't importing them from **index.js** and potentially creating another circular dependency.

Feel free to add to this PR if you have ideas. @iandunn @vedanshujain

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-content.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-content.js
@@ -15,6 +15,7 @@ import { AvatarImage }                                from '../../components/ima
 import { BlockNoContent, ItemTitle, ItemHTMLContent } from '../../components/block-content';
 import { PostList }                                   from '../../components/post-list';
 import { filterEntities }                             from '../../data';
+import { BlockContext }                               from './block-context';
 
 /**
  * Component for displaying the block content.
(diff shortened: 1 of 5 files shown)


pkevan:

Issue #56, opened 2019-03-29
Blocks: Abstract into parent/child classes to make code more DRY?


iandunn:
It feels like Speakers/Sessions/Sponsors/Organizers share _a lot_ of code in common, with minor differences. Here's some examples of where things are very close, with with minor differences in variable names, function parameters, etc

* `includes/{type}.php`
* `view/{type}.php`
* `src/{type}/edit.js`
* `src/{type}/block-content.scss`
* `src/{type}/block-controls.js`
* `src/{type}/inspector-controls.js`
* `src/{type}/speakers-select.js`
* `src/{type}/toolbar.js`

I know we want to be careful about over-engineering things, because having a maze of abstraction layers can be worse than code duplication, but if feels like we may be erring too far to the side of duplicating right now.

What do y'all think about the above? Are there some places where it'd make sense to add an abstraction layer, and other where it makes sense to keep it the way it is?

coreymckrill:
I'm open to seeing what you would propose specifically in terms of abstraction, but I would hesitate to abstract away any of the files you list that exist for each of the blocks. Having those files keeps the blocks flexible for future development and maintenance, where their current commonalities might diverge.

Rather than abstracting entire chunks of the file structure, I think I would prefer to abstract more pieces into shared components. That maintains a bit more flexibility, because then we have repeatable chunks that don't necessarily need to appear in the same context in every block, or the same order.

vedanshujain:
@iandunn I also thought this while working on Sponsors block, and considered adding a component for`CustomPostType`, which will basically be a abstraction over these different blocks. However, these blocks are different enough, that it could result into abstraction maze instead of simplifying stuff. 
I agree with @coreymckrill here, I think we can put more thought into creating better shared components to keep our code DRY.

iandunn:
👍 

That's the direction I was thinking too: more shared components that can accept props to account for minor differences, but not replacing entire files.

I definitely agree that a maze of abstractions should be avoided.

I'll work on a PR once some of the other things have settled down, in order to minimize merge conflicts.

iandunn:
Good food for thought:

> "Duplication is far cheaper than the wrong abstraction." ([>>](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction))

hat-tip @coreymckrill

ryelle:
At this point I think we've settled on the architecture we have, so I'm going to close this out – feel free to re-open if there's still something to discuss here.

PR #123, opened 2019-05-20
Apply Bem-esque convention for Blocks CSS

vedanshujain:
- [x] Organizers
- [x] Sessions
- [x] Speakers
- [x] Sponsors
- [x] Shared Components

- [x] Testing


vedanshujain:
@iandunn This is now ready for review. Some noteworthy changes:

1. I have used convention of adding `s` for block classifier, because our blocks name is plural, even if the element is singular. For egs `wordcamp-organizers__post` as opposed to `wordcamp-organizer__post`. What do you think about this?

2. Added a file `_z-index.scss` to store z-index attributes in one place to all files. This sort of not comply with BEM standard, but [there is some prior art for rare cases.](https://github.com/WordPress/gutenberg/blob/master/assets/stylesheets/_z-index.scss)

3. Added components `EditModeOption` and `ItemMeta`. Only purpose of these components is to apply common styles, so that we won't have to duplicate those in all blocks. 

I am still testing the whole PR, but it should be ready for review.

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
```diff
@@ -28,7 +28,7 @@ function render_featured_image( $post, $width, $class_names = [], $image_link =
 	$size         = array( $width, $height );
 
 	$container_classes = array_merge(
-		[ 'wordcamp-image-container', 'wordcamp-featured-image-container' ],
+		[ 'wordcamp-image__featured-image-container' ],
```
iandunn, line 31: `MUST` - If I understand the convention, then I think this should just be `wordcamp-featured-image`, since 
 root elements don't need a descriptor.

> Any descendent of the component’s root element must append a dash-delimited descriptor, separated from the base by two consecutive underscores __. 

Related: WordPress/gutenberg#15764

Also see the comment on line `42` for why I think it should be `wordcamp-featured-image` instead of `wordcamp-image`, `wordcamp-image-featured`, etc.

There's probably other places where a similar change is needed.

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
```diff
@@ -39,7 +39,7 @@ function render_featured_image( $post, $width, $class_names = [], $image_link =
 	?>
 		<div class="<?php echo esc_attr( $container_classes ); ?>">
 			<?php if ( '' !== $image_link ) { ?>
-				<a href="<?php echo esc_html( $image_link ); ?>" class="wordcamp-image-link wordcamp-featured-image-link">
+				<a href="<?php echo esc_html( $image_link ); ?>" class="wordcamp-image__featured-image-link">
```
iandunn, line 42: `MUST` - If I understand the convention, then I think this should be `wordcamp-featured-image__link`. The "directory" portion of the name should match the folder, which is `featured-image` rather than `image`, and the "modifier" portion should describe the child element _relative to the root element_, so the "featured" part would be redundant there (assuming it's already been moved to the "directory" portion).

Related: https://github.com/WordPress/gutenberg/pull/15764

There's probably other places where a similar change is needed.

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
```diff
@@ -43,7 +43,7 @@ function render_post_list( array $rendered_items, $layout = 'list', $columns = 1
 	?>
 		<ul class="<?php echo esc_attr( $container_classes ); ?>">
```
iandunn, line 44: `MUST` - it looks like these haven't been updated yet, e.g. `organizers.php::render()` should pass in `wordcamp-organizers` rather than `wordcamp-organizers-block`.

There's probably other places that need a similar change.

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
```diff
@@ -43,7 +43,7 @@ function render_post_list( array $rendered_items, $layout = 'list', $columns = 1
...
 			<?php foreach ( $rendered_items as $item ) : ?>
-				<li class="wordcamp-post-list-item wordcamp-clearfix">
+				<li class="wordcamp-post-list__post wordcamp-clearfix">
```
iandunn, line 46: `MIGHT` - This is a bit weird, because the root element is `wordcamp-organizers-block` (which should be renamed to `wordcamp-organizers`, see above), but the child elements are `wordcamp-post-list__post`.

It makes sense that the child elements would be marked as under the `post-list` "directory", but it's odd that the root and child elements don't have consistent "directory" parts to their name.

I wonder if the root element should be `wordcamp-post-list wordcamp-organizers`, and the children `wordcamp-post-list__post`? Or the root should be `wordcamp-organizers` and the children `wordcamp-organizers__post`? Or even something else?

🤔 Not sure, what do you think?

public_html/wp-content/mu-plugins/blocks/source/_z-index.scss
```diff
@@ -0,0 +1,8 @@
+.wordcamp__edit-placeholder .components-placeholder__fieldset {
```
iandunn, line 6:
```suggestion
/*
 * When adding elements here, make sure you leave a comment fully explaining which other element they're conflicting with.
 */

.wordcamp__edit-placeholder .components-placeholder__fieldset {
```

`SHOULD` - It'd be good to add a note to the top of this file, similar to the one in the G file that you linked to, asking that any additions here be explained with comments, since comments are as normal in CSS as they should be.

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-content.js
```diff
@@ -79,15 +79,15 @@ export class BlockContent extends Component {
...
 			<PostList
 				{ ...this.props }
-				className="wordcamp-organizers-block"
+				className="wordcamp-organizers__posts"
 			>
 				{ posts.map( ( post ) => /* Note that organizer posts are not 'public', so there are no permalinks. */
 					<div
 						key={ post.slug }
-						className={ classnames( 'wordcamp-organizer', 'wordcamp-organizer-' + post.slug.trim() ) }
+						className={ classnames( 'wordcamp-organizers__post', 'has-slug-' + post.slug.trim() ) }
```
iandunn: `MIGHT` - I wonder if this should be `'has-slug', 'slug-{slug}'`?

Maybe that's premature though, since we don't have a use case for targeting elements that `has-slug` rather than individual ones? But maybe site designers would have a use case in their custom CSS? And if we do eventually need that, we'll have to plan for it now to avoid back-compat complications.

`SHOULD` - I used single quotes in the example above b/c GitHub/markdown won't let me escape backticks, but using JS string interpolation rather than `+` seems more readable.

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
```diff
@@ -61,28 +61,34 @@ export class BlockControls extends Component {
 			default :
 				output = (
 					<Placeholder
-						className={ classnames( 'wordcamp-block-edit-placeholder', 'wordcamp-block-edit-placeholder-no-mode' ) }
+						className={ classnames( 'wordcamp__edit-placeholder', 'has-no-mode' ) }
```
iandunn, line 64: `MUST` - I think this should be `wordcamp-organizers__block-contents wordcamp-placeholder` for similar reasons as earlier comments.

> Notably, if your folder contains multiple files, each with their own default exported component, only the element rendered by that of index.js can be considered the root. All others should be treated as descendents.

There's probably other places that need a similar change.

iandunn (changes requested):
I've only looked through Organizers so far, but it seems like we have some fundamentally different interpretations of the guidelines, so we'll need to figure out that first.

After that, I can come back and review the rest.

iandunn:
> I have used convention of adding s for block classifier, because our blocks name is plural, even if the element is singular.

👍 

> Added a file _z-index.scss to store z-index attributes in one place to all files. 

👍 

> Added components EditModeOption and ItemMeta. Only purpose of these components is to apply common styles, so that we won't have to duplicate those in all blocks.

🤔 

I think this may have been the type of DRYness that @coreymckrill was leery of in #56, since the added complexity from the extra abstraction layer might be greater than the complexity of the duplication.

I'd lean towards reverting back to the previous way, but need to sleep on it.

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
vedanshujain: 🤔 In this case, `featured-image` component is actually part of `image` folder, which has its own `index.js`. Guidelines also states that:

> Notably, if your folder contains multiple files, each with their own default exported component, only the element rendered by that of `index.js` can be considered the root.

So that's why I am using `image` as root descriptor and `featured-image` as component. What do you think?

vedanshujain: Ah I see, so in `php` we have `featured-image` in as the folder name, but in `js` we have `image` as the folder name. What do you think we should give preference to? Or should we rearrange the folders in `php` to reflect `js`?

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
vedanshujain: Depends upon resolution of https://github.com/WordPress/wordcamp.org/pull/123#discussion_r286231604

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
vedanshujain: 🤔 Where do you `wordcamp-organizers-block`? Looks like its actually passing `wordcamp-organizers__posts`.

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
vedanshujain: Hmm, so right now, root has actually `wordcamp-block wordcamp-post-list layout-list wordcamp-organizers__posts` and then children `li` has `wordcamp-post-list__post`, and whose children has `wordcamp-organizers__post has-slug-organizer-1`, which sound reasonable to me, because `li` is rendered by the `post-list` component, and then `wordcamp-organizer wordcamp-organizers__post has-slug-{organizer-slug}` which is a `div` is rendered by the block itself.

Are you suggesting that we merge the classes in `li` and `div`? Looks like we can do that by passing the class name while rendering `post-list`. 
For the root element, I agree that `wordcamp-organizers` is cleaner that `wordcamp-organizers__posts`, but then because of the way `js` is structured, it won't follow the guideline of only treating `index.js` export as root and rest of all as descendants. What do you think?

public_html/wp-content/mu-plugins/blocks/source/_z-index.scss
vedanshujain: 👍

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-content.js
vedanshujain: > using JS string interpolation rather than + seems more readable.

👍 Will do that.

>I wonder if this should be 'has-slug', 'slug-{slug}'?

Why not just use `slug-{slug}`, only purpose of `has` adjective here is to improve readability. 
Also,

> Because a modifier class name is not contextualized to a specific component, it should 
always be written in stylesheets as accompanying the component being modified (.components-panel.is-opened).

Given this line, I would think that you are not supposed use modifier class name on its own. If some would want to target `has-slug` they would hopefully use `wordcamp-organizers__post` instead. What do you think?

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
iandunn: Oh, you're right, good point.

🤔 , wonder if we should really have all those inside an image folder? It makes sense in a way, but also feels kind of awkward to have the extra level there, especially with the way it affects the class names. This might be simpler:

```
components
 |-- avatar
 |-- featured-image
 |-- image-inspector-controls
```

Although, [this comment](https://github.com/WordPress/wordcamp.org/issues/96#issuecomment-491443218) might be relevant here too:

> For us the block/ and component/ folders are just transparent organizational devices for the filesystem, rather than part of the actual component hierarchy in the DOM. Maybe that's a useful way to think about it?

Even if we keep the `image` container folder, I'm not sure it necessarily needs to be taken into account in the class names.

We could look at what G does for some conventions, IIRC there are some nested components in a few places too, but it also feels like generally the hierarchy is kept mostly flat.

----

If we were to make the PHP consistent w/ the JS, I guess we'd want it structured more like this:

* `includes/components/featured-image/controller.php`
* `includes/components/featured-image/view.php`

...rather than being split between `/includes` and `/views. That feels kinda nice, because it reflects the component-ized nature of the system.

Another discrepancy is that the views for the PHP components are embedded in the controller files, while the views for the PHP blocks are separated into the `views` folder. I'm guessing that was just done because they're fairly small, but it might be nice to separate them for consistency, readability, and separation-of-concerns.

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
iandunn: Oh, I was looking at https://github.com/WordPress/wordcamp.org/blob/a7eeea1e0330c8ec4f7c3ddcd0ce1956b950551a/public_html/wp-content/mu-plugins/blocks/includes/blocks/organizers.php#L51, but I didn't actually test it, so maybe I'm just reading the code wrong.

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
iandunn: 🤔 

Or, I wonder if we should instead just move the PHP files into `source`, instead of keeping them in a separate folder? That way we'd have:

```
source
|-- blocks
|-- |-- organizers
|-- |-- |-- controller.php
|-- |-- |-- edit.js
|-- |-- |-- index.js
|-- |-- |-- view.php
|-- |-- |-- ...etc
```

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
iandunn: > Hmm, so right now, root has actually wordcamp-block wordcamp-post-list layout-list wordcamp-organizers__posts and then children li has wordcamp-post-list__post, and whose children has wordcamp-organizers__post has-slug-organizer-1, which sound reasonable to me, because li is rendered by the post-list component, and then wordcamp-organizer wordcamp-organizers__post has-slug-{organizer-slug} which is a div is rendered by the block itself.

Ah, good point 👍 

> Are you suggesting that we merge the classes in li and div? Looks like we can do that by passing the class name while rendering post-list.

🤔 yeah, I think so. It does feel weird that we have the `li` and then the `div`; the `div` feels unnecessary.

It seems like we should replace the `<div>` with a `Fragment`, and pass the class names for the `li` like you said.

I don't feel too strongly, though, so let me know if you see any problems with that.

> For the root element, I agree that wordcamp-organizers is cleaner that wordcamp-organizers__posts, but then because of the way js is structured, it won't follow the guideline of only treating index.js export as root and rest of all as descendants. What do you think?

Yeah, I see your point. It feels like the convention kind of breaks down here, and is complicated by the fact that a block is more than a simple component, and it has the "index" which is really just a controller, and the "edit" which is the view.

It seems like maybe `edit.js` should have a `<div class="wordcamp-organizers">` instead of a `<Fragment>`, but 🤷‍♂ . Trying to follow this convention is getting pretty awful :(

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
```diff
@@ -43,7 +43,7 @@ function render_post_list( array $rendered_items, $layout = 'list', $columns = 1
 	?>
```
iandunn, line 43: `MUST` - I think this should this be `has-layout-{layout}` instead, since it'd be considered a state.

https://github.com/WordPress/wordcamp.org/pull/123/files#diff-edfb05ceecdc904d955b60b5e64046bcL27

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-content.js
iandunn: 🤔 

Yeah, I guess the slug is really a property of the post, rather than a state, so it probably shouldn't have the `is/has` prefix. `slug-{slug}` makes sense to me.

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
vedanshujain: We agreed on suggestion by @iandunn that it makes more logical sense in this case to move `php` files to source folder as well.

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
vedanshujain: Noting here that in separate discussion, we agreed to
1. Make root class name as `wordcamp-organizer` i.e. to treat `block-content.js` as the root file instead of `index.js`
2. Keep both `li` and `div` as it is now.

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/123/commits/909a6ad397da18e573c1584e746cf459a3eff99d

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
vedanshujain: We decided to use `block-content` as root, which is done in https://github.com/WordPress/wordcamp.org/pull/123/commits/af6f25b4a21ebb99b8077bddf8fe6e8ead8a0087.  Resolving conversation.

public_html/wp-content/mu-plugins/blocks/includes/components/featured-image.php
vedanshujain: In a separate chat, we decided to keep it as it is for now, and moved the controller to inside the source folder in https://github.com/WordPress/wordcamp.org/pull/123/commits/909a6ad397da18e573c1584e746cf459a3eff99d

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/123/commits/af6f25b4a21ebb99b8077bddf8fe6e8ead8a0087

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-content.js
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/123/commits/be41c3ddc196715dc6a4b2a237f9a40bc62ec458

public_html/wp-content/mu-plugins/blocks/includes/components/post-list.php
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/123/commits/47f250ec5776657736fcb324c48d36530322c92f

public_html/wp-content/mu-plugins/blocks/source/_z-index.scss
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/123/commits/b03b01398481daf879cd95966a202885ac6b14a7

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
vedanshujain: So `wordcamp__edit-placeholder` actually applies a shared style defined in `z-index.scss`, so it would make sense to keep it as it is. What do you think?

vedanshujain:
@iandunn Since we decided to use shared styles, I have also removed both `EditModeOption` and `ItemMeta`

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-content.js
```diff
@@ -80,7 +80,7 @@ export class BlockContent extends Component {
 		return (
 			<PostList
 				{ ...this.props }
-				className="wordcamp-organizers__posts"
+				className="wordcamp-organizerss"
```
iandunn: `MUST` - Looks like there's an extra `s`.

public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/block-content.js
```diff
@@ -289,16 +286,15 @@ export class BlockContent extends Component {
...
-								content={
-									show_meta &&
-									<SessionMeta session={ post }/>
-									+
-									show_category &&
-									<SessionCategory session={ post } />
-								}
-							/>
+						<div
+							className="wordcamp__post-meta wordcamp-sessions__details">
```
iandunn: `SHOULD` - it seems like the class should be on the same line as the div, since it's the only attribute, and it's not very long.

public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/block-content.js
```diff
@@ -289,16 +286,15 @@ export class BlockContent extends Component {
...
+							{ show_meta &&
+							<SessionMeta session={ post }/>
+							}
+							{ show_category &&
+							<SessionCategory session={ post }/>
```
iandunn: `SHOULD` - I think `<SessionCategory` and `<SessionMeta` should be indented.

`SHOULD` - There should be an empty space before the `/>`.

The linter might catch other things in the PR, so it'd be a good idea to run it.

public_html/wp-content/mu-plugins/blocks/blocks.php
```diff
@@ -12,23 +12,25 @@
...
 function load_includes() {
-	$includes_dir = PLUGIN_DIR . 'includes/';
+	$includes_dir   = PLUGIN_DIR . 'includes/';
+	$blocks_dir     = PLUGIN_DIR . 'source/blocks/';
+	$components_dir = PLUGIN_DIR . 'source/components/';
 
 	require_once $includes_dir . 'definitions.php';
 
 	// Utilities.
 	require_once $includes_dir . 'utilities/content.php';
```
iandunn: `MIGHT`  - It seems like the `utilities` folder is no longer needed, since there's only 1 thing in it. I think it'd be good to move `content.php` up into `includes` directly, and remove `utilities`.

Low-ish priority and not directly related to this PR though.

public_html/wp-content/mu-plugins/blocks/source/_z-index.scss
```diff
@@ -1,3 +1,7 @@
+/*
+ * When adding elements here, make sure you leave a comment fully explaining which other element they're conflicting with.
+*/
+
```
iandunn, line 5: `MIGHT` - It feels like this commit can be squashed into 71903933b5823bcb179dc632b897132b30cc733e, since they're directly related, and the note here reflects the similar one in the Gutenberg file that's being referenced in 71903933b5823bcb179dc632b897132b30cc733e's commit message.

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
```diff
@@ -61,28 +61,34 @@ export class BlockControls extends Component {
...
-								isDefault
-								isLarge
-								onClick={ () => {
-									setAttributes( { mode: 'all' } );
-								} }
-							>
-								{ getOptionLabel( 'all', options.mode ) }
-							</Button>
-						</div>
+						<EditModeOption
```
iandunn: `MUST` - I thought we were removing the elements for these? It looks like they one of the commits in this PR removed them, but a later one added them back.

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
iandunn: 🤷‍♂ , that works for me.

public_html/wp-content/mu-plugins/blocks/source/components/block-content/index.js
```diff
@@ -26,7 +26,7 @@ import './style.scss';
  */
 export function BlockNoContent( { loading } ) {
 	return (
-		<div className="wordcamp-block-content-none">
+		<div className="wordcamp-block_posts has-no-content">
```
iandunn: `MUST` - Looks like this was intended to be a `__`

public_html/wp-content/mu-plugins/blocks/source/components/block-content/style.scss
```diff
@@ -1,4 +1,4 @@
-.wordcamp-block-content-none {
+.wordcamp-block_posts.has-no-content {
```
iandunn: `MUST` - Also `__`. There may be others too.

public_html/wp-content/mu-plugins/blocks/source/components/image/style.scss
```diff
@@ -1,4 +1,4 @@
-.wordcamp-image-container {
+.wordcamp-image__avatar-container,.wordcamp-image__featured-image-container {
```
iandunn: `SHOULD` - The second selector should be on a new line.

iandunn (changes requested):
Needs a few minor changes, but after that it should be good to merge.

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
vedanshujain: Where are you seeing this? Its already removed, I force pushed the branch today, so you might need to force pull it if you are testing on local

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-content.js
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/123/commits/92ab54753af3ccb3e8248f03065b2e340521a261

public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/block-content.js
vedanshujain: Right, I was configuring my editor for standards today, looks like I missed a few spots. Did a linter check and fixed stuff in https://github.com/WordPress/wordcamp.org/pull/123/commits/92ab54753af3ccb3e8248f03065b2e340521a261

public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/block-content.js
vedanshujain: Fixed in https://github.com/WordPress/wordcamp.org/pull/123/commits/92ab54753af3ccb3e8248f03065b2e340521a261

public_html/wp-content/mu-plugins/blocks/blocks.php
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/123/commits/92ab54753af3ccb3e8248f03065b2e340521a261

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
vedanshujain: Resolving this, please re-open if action is needed here

public_html/wp-content/mu-plugins/blocks/source/components/block-content/index.js
vedanshujain: Yup, thanks for the catch. Fixed in https://github.com/WordPress/wordcamp.org/pull/123/commits/92ab54753af3ccb3e8248f03065b2e340521a261

public_html/wp-content/mu-plugins/blocks/source/components/block-content/style.scss
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/123/commits/92ab54753af3ccb3e8248f03065b2e340521a261

public_html/wp-content/mu-plugins/blocks/source/components/image/style.scss
vedanshujain: Fixed along with lint stuff in https://github.com/WordPress/wordcamp.org/pull/123/commits/92ab54753af3ccb3e8248f03065b2e340521a261

public_html/wp-content/mu-plugins/blocks/source/_z-index.scss
vedanshujain: Done in https://github.com/WordPress/wordcamp.org/pull/123/commits/a30b2a1e7950c8c975863f0c18b1b928185228b9

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
iandunn: Huh, that's weird, I'm not seeing `EditModeOption` in `production` either, and it's not shown as being removed [in the current diff](https://github.com/WordPress/wordcamp.org/pull/123/files). I'm confused.

vedanshujain: Noting here that confusion was because `EditModeOption` was never merged, so it was not showing as removed.

PR #139, opened 2019-07-09
Git: Ignore the `node_modules` directory

ryelle:
Rather than assuming the developer has a global gitignore set up, the project should explicitly ignore any untracked files it creates.

This is a friendlier developer experience, and prevents accidentally committing all of `node_modules` 🙂

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 #
 wp-cli.yml
 script-backups/
+node_modules/
 public_html/.well-known/
 public_html/mu/
 public_html/bin/


iandunn:
Is there a compelling reason why someone wouldn't want to add `node_modules` to their global ignore file? That seems much easier and more efficient than manually adding it to dozens of projects individually.

There are also a dozen other common files that could be present, like `.DS_STORE`, `sass-cache`, `.svn`, `.idea`, `npm-debug.log`, `*.map`, `*.diff`, `*.patch`, and `debug.log`. It feels tedius to add them all individually, and it clutters the local ignore file with things that aren't unique to the project. Is there a reason that we should add `node_modules`, but not all of those others too?

Having said that, this is really just a minor annoyance of mine, and not a big deal either way.

coreymckrill (approved):
Since this is public, I don't think we can assume that everyone has a global gitignore. Especially since so many projects add it to their local gitignore files, I imagine some people take it for granted. Some other meta files like `.DS_STORE` and `.idea` aren't guaranteed to be there because they are added by the end user's environment, but `node_modules` is added by our own build script, so I think it's especially relevant here. And accidentally committing it would really suck.

iandunn:
👍 

I'm still personally apprehensive, since I think the correct way to do achieve the goal is with `.gitignore_global`, we should always check `git stat` before committing anything, and nothing within `node_modules` should be executable over unauthed HTTP requests.

But more importantly, I can see where y'all are coming from, and it's not important enough to spend more time on, so I'm happy to defer to y'all.

ryelle:
I don't have project-specific files/folders in my global gitignore, but I do have system-created things like `.DS_STORE` there. My general opinion is that if the project creates it, we should ignore it. So if we create a `.sass-cache`, it should also get added.

Ideally everyone is using some process that checks files before they're committed, but accidents happen and since this is an easy way to automatically prevent that, I'd rather have it.

iandunn (approved):
There's solid reasoning behind it.

:shipit:

Issue #86, opened 2019-04-11
Organizers: "Biography length" setting is somehow half-width


melchoyce:
![image](https://user-images.githubusercontent.com/2846578/55988601-4844b400-5c72-11e9-9ce1-fa02be1bd43c.png)

Should be full-width

iandunn:
This was fixed in 5175aff.

Issue #84, opened 2019-04-11
All: Sidebar selects should be full-width


melchoyce:
Dropdowns/select boxes within settings should all be full-width. 

This is how they look in core Gutenberg:
![image](https://user-images.githubusercontent.com/2846578/55988537-1df2f680-5c72-11e9-9e5f-0c903ae18b3f.png)

iandunn:
I think part of the reason for this is because we're using `PanelRow`, while the core blocks don't.

iandunn:
This was fixed in 5175aff.

PR #140, opened 2019-07-09
Blocks: Remove PanelRow wrapper from single-item controls

### LINKED ISSUE #79: All: Sidebar spacing
Sidebar setting margins don't match core Gutenberg margins, they're too big:

image

These should exactly match core's margins.

ryelle:
`PanelRow` applies flex styling, so the child elements are no longer full-width by default, the width is determined by the children elements. PanelRow is useful for aligning two controls or elements on one line, but isn't needed for single components. 

This fixes #79, #84, #86.

**To test**

- Add any of the blocks (this touches Organizers, Sessions, and Speakers)
- Check that the controls in the sidebar are displayed correctly, don't overlap or have extra space.


iandunn (approved):
LGTM!

In the commit message, I'd add an explanation of the negative side-effects of using `PanelRow` here, which explains why it's being removed. What you wrote in the PR description is great and would work for this purpose.

If a terse explanation of [the _why_](https://github.com/WordPress/wordcamp.org/blob/ffa15ac/.githooks/prepare-commit-msg#L8) will fit into the short description as well, that'd be useful. e.g., `Blocks: Remove unneeded 'PanelRow' for full width` (but with backticks instead of single quotes, I just used singles there b/c GitHub's Markdown doesn't let me escape nested backticks).

I think it's also good to add the `Fixes #79, #84, #86` to the commit itself, in addition to (or instead of) the PR, since commits are a more foundational part of the repository's history, while PRs are more optional and somewhat GitHub-specific.

coreymckrill (approved):
👍 

I think the reason those `PanelRow`s were there in the first place was because the layout was wonky for some components if they weren't wrapped in it. In particular, I recall that being the case with `Select`, `Toggle`, and `RangeControl`. Maybe this was fixed in a more recent version of Core or Gutenberg, as all the `Select`s and `Toggle`s look fine.

ryelle:
> What you wrote in the PR description is great and would work for this purpose.

Right, I was under the impression that the PR description gets merged into the commit but it's just the PR title. So used to everyone just being on github that it doesn't really matter what commits say as long as the PRs are descriptive 😅

PR #142, opened 2019-07-10
Docker: Update the test dB, and add useful database operations for maintaining it

coreymckrill:
**Changes to the test dB**

* Clean up the data on wordcamp.test and central.wordcamp.test. This includes switching all URLs to use `https` and removing references to Ian (who I think originally created the test data).
* Remove all of the mostly empty test sites, keeping only 2014.seattle.wordcamp.test, and cleaning it up a bit. That's the one with the most content.
* Add a user with each role for capability testing.
* Change central.wordcamp.test to have a blog_id of `5` to match production.
* Clean out post revisions, trashed posts, transients, and the camptix log. These are unnecessary for starting a new test, and make the file larger than it needs to be.

**New database.sh script**

This script is intended to standardize the process of updating the test database provision file, as well as give us some useful database tools for testing (think backing up and restoring states of the database while trying different code approaches). The details are in the readme.


.docker/bin/database.sh
```diff
@@ -0,0 +1,183 @@
+#!/bin/bash
```
iandunn, line 1: `FYI` - IMO we'll be better off writing stuff like this in PHP (see `docs/wordcamp-org.md#bin-scripts`), but I'm open to other ideas there, and we definitely should not spend time rewriting old Bash scripts to be in PHP. Just raising this as an issue for future.

.docker/bin/database.sh
```diff
@@ -0,0 +1,183 @@
...
+        help | * )
+            echo "Please enter a valid command."
+            echo
+            echo -e "  backup-current \n  clean \n  clean-export \n  import \n  export \n  reset \n  restore <file>"
+        ;;
+    esac
+
+    echo
+
+    cd $orig_dir
```
iandunn: `MIGHT` - I think you can just do `cd -` to switch back, rather than tracking this yourself. Not a big deal since this is already working though.

iandunn: ( Side note: you can do the same thing w/ Git branches; e.g., `git checkout temp && git checkout -` )

.docker/bin/database.sh
```diff
@@ -0,0 +1,183 @@
...
+
+    echo
+    echo -n "Importing database from provision file ${DATA_FILENAME}... "
+    $WP db import ${DATA_PATH}/${DATA_FILENAME} --quiet
+    echo "Done."
+}
+
+# Export the current state of the database as the new database provision file.
+#
+function export() {
```
iandunn, line 149: `MIGHT` - There's an `export` command for setting env vars. I don't think there's technically any conflict with naming a function the same, but ideally it might be better to avoid it, just for clarity, to avoid accidentally calling the command instead of the function, and maybe weird edge cases w/ esoteric Bash quirks.

Having said that, that's probably true of other functions like `import`, `clean`, etc, so it may not really be worth it.

.docker/bin/database.sh
```diff
@@ -0,0 +1,183 @@
...
+    esac
+
+    echo
+
+    cd $orig_dir
+}
+
+# Clean unnecessary data from the database.
+#
+# This currently includes post revisions and transients.
```
iandunn, line 60: `SHOULD` - It might be good to add a note here that this is only meant to be run in development environments where the data is already fake, and that it's *not* safe to run against production.

Actually, it might be even better to `echo` that to the console at the start of the script, rather than it being buried in a comment here.

.docker/readme.md
```diff
@@ -143,15 +143,27 @@ Note: All of these commands are meant to be executed from project directory.
...
+* `reset`: Delete the current state of the database and then re-import the provision file.
+* `clean-export`: Clean up the current state of the database and then export it to the provision file. See more on this below.
+* `help`: See the full list of subcommands available.
+
+### Updating the database provision file
+
+If the dev database needs to be updated to better reflect the state of production (e.g. a new version of the WP Core database, new network-activated plugins), or you've added data to your dev database that you think should be included for everyone (e.g. a new site with useful test cases), you can use the `clean-export` subcommand to update the file that is committed to version control. Before you do, please do these preflight checks:
+
+* Make sure WP is running the latest branch, and the database schema has been updated.
+* Make sure there isn't anything sensitive in the database. Scrub anything that is.
```
iandunn:
```suggestion
* Review each line of the diff to make sure there isn't anything sensitive in the database. Scrub anything that is.
```

`MIGHT` - It might be good to make this a bit more strict/explicit, just to make it clear how important that is. It might also help to provide some guidance on making that less inconvenient. [The Meta Environment has some tips there](https://github.com/WordPress/meta-environment/wiki/Reviewing-PRs-with-database-changes), but they're still pretty difficult to do. I'm not sure if there's a better way, though.

.docker/readme.md
```diff
@@ -143,15 +143,27 @@ Note: All of these commands are meant to be executed from project directory.
...
+
+* `backup-current`: Back up the current state of the database to a file in the **.docker/data** directory.
+* `restore <file>`: Import a specified file to the database after deleting the current state.
+* `reset`: Delete the current state of the database and then re-import the provision file.
+* `clean-export`: Clean up the current state of the database and then export it to the provision file. See more on this below.
+* `help`: See the full list of subcommands available.
+
+### Updating the database provision file
+
+If the dev database needs to be updated to better reflect the state of production (e.g. a new version of the WP Core database, new network-activated plugins), or you've added data to your dev database that you think should be included for everyone (e.g. a new site with useful test cases), you can use the `clean-export` subcommand to update the file that is committed to version control. Before you do, please do these preflight checks:
```
iandunn, line 166:
```suggestion
_WARNING_: Never export data directly from the production database, because it contains sensitive information. Instead, manually recreate the useful data locally with fake email addresses, etc.

If the dev database needs to be updated to better reflect the state of production (e.g. a new version of the WP Core database, new network-activated plugins), or you've added data to your dev database that you think should be included for everyone (e.g. a new site with useful test cases), you can use the `clean-export` subcommand to update the file that is committed to version control. Before you do, please do these preflight checks:
```

.docker/wp-config.php
```diff
@@ -34,7 +34,7 @@
 define( 'DOMAIN_CURRENT_SITE',  $wcorg_domain_current_site );
 define( 'PATH_CURRENT_SITE',    '/' );
 define( 'SITE_ID_CURRENT_SITE',  1 );
-define( 'BLOG_ID_CURRENT_SITE',  2 ); // central.wordcamp.test
+define( 'BLOG_ID_CURRENT_SITE',  5 ); // central.wordcamp.test
```
iandunn, line 37: Great catch!

iandunn (approved):
I didn't test locally, but the code LGTM.

.docker/bin/database.sh
coreymckrill: Good point.

.docker/bin/database.sh
coreymckrill: TIL. It appears this works as long as you don't switch directories more than once. This script doesn't, so it would be fine.

coreymckrill: c4134903

.docker/bin/database.sh
coreymckrill: Good idea. abd93d7e

This adds an environment variable to the docker config that the database script looks for, and it bails with an explanatory message if it doesn't find it.

PR #143, opened 2019-07-10
Attendance UI: Allow searching for a Full name, ie. 'John Smith'

dd32:
Currently the Attendance UI can only search on a First or Last Name.

If you enter a full name, ie. 'John Smith' it'll fail to find any results.
This attempts to work around it by matching on `FirstName LastName` in addition to them individually.


iandunn:
Thanks! Merged in c285868c09456d892104e93c133f89e3489bd02d and 31ae8232ecccf9cab25b9f38798aff57d8d6380b.

PR #144, opened 2019-07-10
Attendance UI: Add a shortcut to the Search view - Click the header

dd32:
Currently the only way to enter the Search view is to use Menu -> Search.

This adds a shortcut by allowing the user to click the header to open the search that's usually in that area.

Arguably the Search filter should always be visible, as it's an integral part of the UI, but this is better than changing the way the app works.


iandunn:
Thanks! I agree. I think we should probably just add the search input field permanently into the header, but this is a good improvement in the meantime.

Merged in 516a704.

PR #150, opened 2019-07-11
JS Code Standards: Enforce a line-length of 115 characters

ryelle:
_This is one of a few issues/PRs dealing with code standards, created after a discussion in slack._

Currently, we  wrap code at 115 characters, but have no enforcement or documentation of this for JS files (we do have it set in [the `stylelint` config](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/mu-plugins/blocks/.stylelintrc#L4)). We can set this up using [the `max-len` rule in eslint.](https://eslint.org/docs/rules/max-len) Unfortunately, `eslint --fix` won't fix these automatically, but there are only 8 violations in the code base right now (which are also fixed in this PR, in the second commit).

I've set up the rule to ignore line length in URLs, trailing comments, strings, and template literals. This prevents eslint from flagging "atomic" content, like a URL, that would be awkward to wrap.

**To test**

- Run `npm run lint:js`
- There should be no errors
- Open a file & add a line longer than 115 characters
- Run `npm run lint:js` again
- It should flag an error for that line

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/.eslintrc.js
+++ b/public_html/wp-content/mu-plugins/blocks/.eslintrc.js
@@ -81,6 +81,20 @@ module.exports = {
 		} ],
 
 		/*
+		 * Force a line-length of 115 characters.
+		 *
+		 * We ignore URLs, trailing comments, strings, and template literals to prevent awkward fragmenting of
+		 * meaningful content.
+		 */
+		'max-len': [ 'error', {
+			'code'                   : 115,
+			'ignoreUrls'             : true,
+			'ignoreTrailingComments' : true,
+			'ignoreStrings'          : true,
+			'ignoreTemplateLiterals' : true,
+		} ],
+
+		/*
 		 * Allow multiple spaces in a row.
 		 *
 		 * Ideally this should be on, because we don't want to allow things like `const foo  == bar;`, but the rule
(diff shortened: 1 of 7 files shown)


iandunn (approved):
Overall it LTGM :)

My first instinct is that I'm not a fan of a new line for every single property being deconstructed:

```js
const {
	BaseControl,
	Button,
	ButtonGroup,
	PanelBody,
	PanelRow,
	RangeControl,
	ToggleControl,
	Toolbar,
} = wp.components;
```

...because the properties themselves are only `93` chars, and could be wrapped if they were longer:

```js
const {
	BaseControl, Button, ButtonGroup, PanelBody, PanelRow, RangeControl, ToggleControl, Toolbar,
	SecondLine, Wrapped, WhenFirst, GetsTooLong,
} = wp.components;
```

...but that's a separate discussion, and I think it's fine to merge these commits as-is. We can potentially adjust those instances later if we decide to standardize on something different.

Before merging to `production`, the commit message should use the format described in the `prepare-commit-msg` hook (which is adapted from https://make.wordpress.org/core/handbook/best-practices/commit-messages/).

ryelle:
> My first instinct is that I'm not a fan of a new line for every single property being deconstructed

I was going to open another issue about that but it seems we're already putting each object on a new line in other destructuring assignments, probably due to this core standard:

> When an object declaration is too long to fit on one line, there must be one property per line.

coreymckrill (approved):

PR #152, opened 2019-07-12
Use Composer to manage third-party PHP packages

coreymckrill:
The impetus for doing this is so we can use third-party packages (such as [this oauth2 client](https://oauth2-client.thephpleague.com/)) as dependencies for some of our own custom modules and plugins. GitHub's recently added security features give us automatic notifications when a dependency has a vulnerability and a new version is available.

There are also a few other benefits to this:

* By adding all of the third-party WordPress plugins and themes we're using as dev dependencies, we can provision our docker container via Composer to ensure that it has the same versions of everything as we're running on production. Right now it just uses wp-cli to install the latest version of everything via a bash script.
* This wordcamp.org repo itself could eventually be provisioned in local dev environments via Composer. I'm not sure if that's possible right now since the composer.json file is not in the root.

**The setup**

As mentioned above, the composer.json file is in the public_html directory instead of the root of the repository, because of how Composer's installer for WordPress plugin packages works. It installs to `wp-content/plugins`, relative to the location of the composer file. I haven't found anything yet about being able to customize the path base.

All of the plugins are dev dependencies so that they can be skipped by running
`composer install --no-dev`.

For the non-dev dependencies, I picked `wp-content/mu-plugins/vendor` as the installation path. This way the files managed by Composer are kept distinct from our own mu-plugins code, but can still load and work similarly to other subdirectories we have in there. I kept the "vendor" name because it makes it glaringly obvious that it's Composer stuff, since that is the default directory name it uses. But I'm happy to discuss other names and locations.


public_html/composer.json
```diff
@@ -0,0 +1,129 @@
...
+			"type": "package",
+			"package": [
+				{
+					"name": "wordpress-plugin/bbpress",
+					"type": "wordpress-plugin",
+					"version": "2.6",
+					"source": {
+						"type": "svn",
+						"url": "https://plugins.svn.wordpress.org/bbpress/",
+						"reference": "branches/2.6/"
```
iandunn: > Right now it just uses wp-cli to install the latest version of everything via a bash script.

- [ ] `SHOULD` - Since committers will have the SVN externals checked out as well, it seems like everyone else using  the latest version is good enough. Prod usually runs the latest anyway, so if we can setup Composer to pull the latest -- IIRC there's a `latest.zip` redirect available -- then that'd save us a bunch of extra work when updating

- [ ] `SHOULD` - If we go with that approach, it'd be good to document the reason. Not sure where it'd be best, though, b/c JSON maddeningly doesn't support comments (b/c in theory it's just for data, but obviously that's not how it's used in reality). 
    - [ ] `MIGHT` - Maybe composer supports alternative formats like YAML (which has it's own annoyances, but at least allows comments)?

.gitignore
```diff
@@ -22,6 +22,7 @@ public_html/wp-content/cache/
 public_html/wp-content/languages/
 public_html/wp-content/mu-plugins/blocks/build
 public_html/wp-content/mu-plugins/includes/slack/
+public_html/wp-content/mu-plugins/vendor/
```
iandunn, line 25: What gets installed here? Just the non-WP dependencies like the oAuth client? I'd probably want to keep plugins/themes in their respective `wp-content` folders, but I'm fine w/ libraries and stuff going here.

public_html/composer.json
```diff
@@ -0,0 +1,129 @@
...
+	"config": {
+		"platform": {
+			"php": "7.2"
+		},
+		"vendor-dir": "wp-content/mu-plugins/vendor"
+	},
+	"repositories": [
+		{
+			"type": "composer",
+			"url": "https://wpackagist.org/"
```
iandunn: `MIGHT` - I'm a bit leery of this, mainly for [security reasons](https://github.com/outlandishideas/wpackagist/issues/169).

It may not be a big deal, though, since prod will never use it. Locally, committers will have the SVN externals locally,  so we'd spot any differences. Although, if we use the `latest.zip` here, like suggested below, then that might be an issue. I guess we'd just `svn revert` after `composer update`, though, so maybe not.

What's the advantage of using WPackagist over just specifying the URL directly? Is it that we don't have to also specify that they should be installed to `wp-content/plugins` ?

public_html/wp-content/mu-plugins/load-other-mu-plugins.php
```diff
@@ -24,6 +24,10 @@ function wcorg_include_individual_mu_plugins() {
 		require_once( __DIR__ . '/blocks/blocks.php' );
 	}
 
+	if ( is_readable( __DIR__ . '/vendor/autoload.php' ) ) {
```
iandunn, line 27: `MIGHT` - Are you checking `is_readable()` so that local installs won't fatal before running `composer install`?

it seems like it might be good to let it fatal, so that they can see something is wrong, and hopefully check the docs to see that they need to run it.

Or would the fatal interfere with some part of the automated provision process?

iandunn (approved):
Commented on a few things that might improve it, but LGTM at a high level, and the comments aren't blockers.

public_html/composer.json
```diff
@@ -0,0 +1,129 @@
+{
```
iandunn: `SHOULD` - If there's a practical way to install this in the root of the project, that'd be ideal. If we avoid WPackagist like suggested below, that might be easier.

public_html/composer.json
```diff
@@ -0,0 +1,129 @@
...
+						"reference": "wporg-profiles-wp-activity-notifier/"
+					}
+				}
+			]
+		}
+	],
+	"require": {
+		"league/oauth2-client": "2.4.1"
+	},
+	"require-dev": {
```
iandunn: `SHOULD` - Similar to above, it'd be good to document the reason why these are dev dependencies somewhere/somehow.

public_html/composer.json
iandunn: At first glance, this sounds like it'd work: https://github.com/composer/installers#custom-install-paths

iandunn: More: https://github.com/outlandishideas/wpackagist/issues/1

iandunn:
Related Core tickets:

* https://core.trac.wordpress.org/ticket/36335
* https://core.trac.wordpress.org/ticket/43449
* https://core.trac.wordpress.org/ticket/43558

public_html/composer.json
```diff
@@ -0,0 +1,129 @@
...
+		"wpackagist-theme/twentyten": "2.9",
+		"wpackagist-theme/twentyeleven": "3.3",
+		"wpackagist-theme/twentytwelve": "3.0",
+		"wpackagist-theme/twentythirteen": "2.9",
+		"wpackagist-theme/twentyfourteen": "2.7",
+		"wpackagist-theme/twentyfifteen": "2.5",
+		"wpackagist-theme/twentysixteen": "2.0",
+		"wpackagist-theme/twentyseventeen": "2.2",
+		"wpackagist-theme/twentynineteen": "1.4"
+	}
```
iandunn: `MIGHT` - Core has a `scripts` section too - https://github.com/WordPress/wordpress-develop/blob/e6b7205/composer.json

Maybe we should add phpunit/phpcs/phpmd here?

iandunn:
-[ ] `SHOULD` - If we do this, it seems like we'd want to remove `.docker/install-plugins-themes.sh`

-[ ] `SHOULD` - The docker readme should be updated to reflect ^ and the need to run `composer install` during provision, and `composer update` periodically

public_html/composer.json
iandunn: -[ ] SHOULD - If we do this, it seems like we'd want to remove .docker/install-plugins-themes.sh

-[ ] SHOULD - The docker readme should be updated to reflect ^ and the need to run composer install during provision, and composer update periodically

.gitignore
coreymckrill: Yeah, it's only libraries. Plugins and themes go where they're supposed to via installer scripts for `wordpress-plugin` and `wordpress-theme` package types.

public_html/composer.json
coreymckrill: Thanks for the leads. They helped me figure it out, so it's back in the root with a3d925eb

Sidenote, I also researched whether the **composer.json** file could live in a directory like **.config** or something, instead of the root. It appears that it's only possible by setting an environment variable, which would then apply everywhere you use composer locally, and could make things really confusing, in addition to it being an extra, unnecessary burden on contributors. So it's staying in the root for now.

public_html/wp-content/mu-plugins/load-other-mu-plugins.php
coreymckrill: 🤔 It shouldn't interfere with anything. But the fatal error message might not be informative enough for someone to understand that they need to run `composer install`.

We could add an `else` here and then throw a custom fatal error message. Something like:

```php
wp_die( 'Required files are missing. Please run <code>composer install</code> and try again.' );
```

public_html/composer.json
iandunn: 👍

public_html/wp-content/mu-plugins/load-other-mu-plugins.php
iandunn: That's fair.

The main benefit of a change, from my perspective , was just reducing the lines of code, so the current approach and adding `wp_die()` feel about the same to me. If you have a preference, that WFM.

public_html/composer.json
coreymckrill: Since this is only for local dev, and not for production, I think checking the lock file is sufficient, as [this comment explains](https://github.com/outlandishideas/wpackagist/issues/169#issuecomment-307415860).

The advantage is that Composer treats wpackagist as an actual Composer package repository, so the packages that come from there only need to be referenced once in the `require` or `require-dev` property, instead of also having to define a repository and therefore update the version number in three different places for a given package each time it upgrades.

Sidenote, checksums have been implemented for plugins and themes now, haven't they? It doesn't look like Outlandish has added anything related to that, despite [this comment](https://github.com/outlandishideas/wpackagist/issues/169#issuecomment-350897739).

iandunn: 👍 , that WFM, as long as we document that it should never be used on prod/remote sandboxes.

public_html/composer.json
coreymckrill: Done in 2e4a7122, except updating composer programatically. I'll open a separate issue for doing this, along with pulling the latest to the core branch.

public_html/composer.json
coreymckrill: Definitely worth exploring for an iteration. I don't think we should include it in this PR, but I'll open a separate issue.

public_html/wp-content/mu-plugins/load-other-mu-plugins.php
coreymckrill: Done in 99517c82

public_html/composer.json
coreymckrill: Noting that as far as I can tell, composer does not support alternative formats like yml for its package files.

coreymckrill: I'll document the reason for putting plugins/themes in dev dependencies in the merge commit. There doesn't seem like any other good place for it at the moment, given the json limitations you mention.

public_html/composer.json
coreymckrill: See https://github.com/WordPress/wordcamp.org/pull/152#discussion_r304109387

coreymckrill:
This was merged in e54cc82370881e9da8bb8fbf630aec9eda22168e, but I did something wrong so that GitHub isn't recognizing it :/

public_html/composer.json
iandunn: I worry that it'll get lost there. I wonder if the the `svn:externals` property in production's `wp-content/plugins` folder would be a good place? If someone ever mistakenly thought that the externals were no longer needed, then they hopefully see it when they went to delete the property?

Another good place might be in our private devdocs, similar to how we mention other security considerations around provisioning local and remove sandboxes?

public_html/composer.json
iandunn: Maybe in our devdocs then?

iandunn:
I don't see a `Closes #152` in the commit message, which IIRC is necessary when committing from the command line. If you did a squash merge from github.com, then I would think it'd close it automatically.

Reopening just so we don't lose track of the unresolved documentation comments.

public_html/composer.json
coreymckrill: Private devdocs updated.

public_html/composer.json
coreymckrill: Private devdocs updated to reflect this.

public_html/composer.json
coreymckrill: Done.

coreymckrill:
> I don't see a Closes #152 in the commit message, which IIRC is necessary when committing from the command line.

Ah, that must be what it was.

Documentation comments are now resolved.

public_html/composer.json
iandunn: What do you think about using `latest.zip` so that we don't have to manually update `composer.json` all the time? The only people using it for `plugins/themes` will be non-committers, and they don't need to worry about whether the version number is a few days/weeks out of date.

coreymckrill: If we're going to use **latest.zip**, it seems like we might as well switch back to the bash script. In my mind, the only advantage of putting these in composer.json is being able to specify version numbers, so if we're not doing that, the hassles and security concerns might not be worth it.

iandunn: 🤔 I thought that you wanted to define them in `composer.json` so that all 3rd-party dependencies were managed in one place?

It seems like the advantage over the Bash script is that regular contributors only need to run `composer install`, instead of `sh install-plugins-themes.sh && composer install`. If we already need to use composer for the Meetup client, we might as well use it for other stuff too? 

Maybe I misunderstood that part of it, though? I guess I don't feel strongly either way. If we do stick w/ composer, though, I'd much rather pin to the latest, rather than having to constantly update version numbers in 4 places instead of 1.

iandunn: Moved to https://github.com/WordPress/wordcamp.org/issues/159

PR #156, opened 2019-07-17
Composer: Document expected usage for security.

iandunn:
Previously:
* https://github.com/WordPress/wordcamp.org/pull/152#discussion_r303567529
* https://github.com/WordPress/wordcamp.org/pull/152#discussion_r303570561

The JSON format does not allow comments because, in theory, it's only a data-exchange format. In reality, it's widely used in situations like this where comments are extremely important.

To work around that, arbitrary properties are added. The keys must be unique, so they use BASIC-style line-numbering. Does anyone else remember BASIC? No? Ok, I must officially be old then.

`composer validate` will complain about those properties, but normal commands won't, and they  won't be automatically removed by `composer update`, etc. For our usage, they're practical, albeit hacky.

I also added some whitespace between the sections, because I find it much harder to visually parse a document when it's all crammed together.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,4 +1,13 @@
 {
+	"#10": "",
+	"#11": "This file is only intended for use in local environments, it should _NEVER_ be run on remote sandboxes",
+	"#12": "or on production, because we cannot trust/verify the package contents are what they should be.",
+	"#13": "An attacker could inject malware at wpackgist.org and achieve arbitrary remote code execution.",
+	"#14": "See https://github.com/outlandishideas/wpackagist/issues/169",
+	"#15": "See https://github.com/WordPress/wordcamp.org/pull/152#discussion_r303567529",
+	"#16": "",
+
+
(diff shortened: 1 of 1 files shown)


iandunn:
It didn't occur to me until this morning, but this seems like a better approach than the one we discussed in #152.

iandunn:
Related: https://github.com/Seldaek/jsonlint/issues/58

iandunn:
Ah, I found this [buried](https://github.com/composer/composer/issues/5364#issuecomment-221220653) in one of the many many many Composer issues complaining about this problem:

https://github.com/composer/composer/commit/98b0af1386f7ff730c3330e2525c1d66998c7f90

Looks like `_comment` is a valid, reusable key in `composer.json`, so we should we able to switch to that and not get any errors from `composer validate` etc. Haven't tested yet, though.

iandunn:
Tested `_comment` for 5ftF and it worked, so it should be a good path forward here.

PR #157, opened 2019-07-17
 Blocks: Use `@wordpress/scripts` for builds

### LINKED ISSUE #71: Blocks: Considering migrating to `@wordpress/scripts` build
https://make.wordpress.org/core/2019/03/25/building-javascript/

This would simplify things since we wouldn't need to maintain our own webpage/babel config.

Is there any reason we wouldn't want to do this?

ryelle:
Use `wp-scripts` commands to run builds and linting, and update the webpack config to merge in default WP configuration. This enables the `DependencyExtractionWebpackPlugin` plugin, so that `@wordpress/*` scripts are automatically extracted into a `*.deps.json` file, which we can read to automatically generate the correct dependencies for enqueuing.

To get the benefit of dependency extraction, we need to import the wordpress components as `import`s,  rather than const declarations - the wp-scripts webpack also takes care of converting those to the `wp.` globals, so we're not adding any extra libraries.

Post switch, file size effectively doesn't change:

- **blocks.min.js:**  before 163 KiB; after 163 KiB
- **blocks.min.css:**  before 3.82 KiB; after 3.92 KiB

It does add the `blocks.min.deps.json` file, but this is only read server-side.

The switch also brought in slightly updated eslint & stylelint config, so there were a few new issues. This PR also fixes those issues:

- in `ImageSizeControl` & `ImageAlignmentControl`, the BaseControl is used with label but without an id – this flagged [the `no-base-control-with-label-without-id`,](https://github.com/WordPress/gutenberg/blob/master/packages/eslint-plugin/docs/rules/no-base-control-with-label-without-id.md) because without an ID the label isn't attached to anything. Turns out we don't want to use a literal `<label>` in these contexts anyway, so I've moved the label to in-component `span`s.
- in the scss files, remove the unit from `line-height` and remove a redundant `&`.

Fixes #71

**To test**

- Run `npm run build` or `npm start` to build the files
- Check that each block still works
- Run `npm run lint:js`
- Run `npm run lint:css`
- There should be no errors


public_html/wp-content/mu-plugins/blocks/webpack.config.js
```diff
@@ -1,67 +1,38 @@
 const path = require( 'path' );
-const webpack = require( 'webpack' );
 const MiniCssExtractPlugin = require( 'mini-css-extract-plugin' );
+const defaultConfig = require( '@wordpress/scripts/config/webpack.config' );
 
-const NODE_ENV = process.env.NODE_ENV || 'development';
-const SOURCE_MAPS = process.env.SOURCE_MAPS ? true : false;
```
iandunn, line 6: How does `wp-scripts` handle source mapping?

public_html/wp-content/mu-plugins/blocks/webpack.config.js
```diff
@@ -1,67 +1,38 @@
...
 
+	// Override the default filename to keep `min` in the name.
 	output: {
-		filename : '[name].min.js',
-		path     : path.resolve( __dirname, 'build' ),
+		...defaultConfig.output,
+		filename: '[name].min.js',
 	},
 
+	// We need to extend the module.rules & plugins to add the scss build process.
```
iandunn, line 24: It looks like this fully take cares of https://github.com/WordPress/gutenberg/issues/14801 for our repo, so that we don't have to wait for upstream to implement it? Is that right?

iandunn (approved):
❤️

Great work!

iandunn:
Oh, this commit msg will need a `Fixes #71`.

public_html/wp-content/mu-plugins/blocks/webpack.config.js
ryelle: It doesn't, so I pushed a change to bring this back. Unfortunately it looks like `cheap-module-eval-source-map` isn't supported for CSS sourcemaps (at least, not using `MiniCssExtractPlugin`), so it only adds JS sourcemaps. But since that's probably always been the case I think it's fine.

public_html/wp-content/mu-plugins/blocks/webpack.config.js
ryelle: Right, since the core webpack config doesn't support CSS, we set it up ourselves. If that ends up being implemented, we'll probably want to remove our config- I'm not sure how webpack would handle duplicate loaders.

public_html/wp-content/mu-plugins/blocks/webpack.config.js
iandunn: Yup, that's great, thanks!

iandunn: I think this'll need to bring back the `npm run debug`  script too, right?

public_html/wp-content/mu-plugins/blocks/webpack.config.js
iandunn: Awesome! Thanks :)

public_html/wp-content/mu-plugins/blocks/webpack.config.js
ryelle: No, since it defaults to `development`– `npm start` builds the sourcemaps version. Then `npm run build`, which runs `wp-scripts build`, sets the `NODE_ENV` to `production` and no sourcemaps are built in that case.

I didn't notice any time difference between with or without the sourcemap step, so I don't think we need a separate command for it.

coreymckrill (approved):
I fully support using wp-scripts as much as possible and reducing our custom config. ❤️

public_html/wp-content/mu-plugins/blocks/webpack.config.js
iandunn: Perfect, thanks!

PR #158, opened 2019-07-18
Blocks: Update react-select package

ryelle:
_Note: this branch is based off the wordpress-scripts branch, once that's merged I'll rebase this on `production`_

`react-select` is now at v3.0.4, which comes with some bugfixes and improvements - notably some code splitting which shrinks our build 13% (to 141 KiB). This update needs almost no changes, except that the `onChange` callback now receives a null value when the last selected item is cleared, which causes `ContentSelect` to default back to the passed in `value`, rather than deleting the last item.

**To test**

- Run `npm install` to update packages
- `npm run build`
- Add any block to a post
- Select a few items, clear some of them, select more, clear the whole input… etc
- Select a few items, then click "Select" to display them
- Try to clear all the items from the input by clicking each item's X, you should be able to clear all of them, back to the initial placeholder screen.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/package-lock.json
+++ b/public_html/wp-content/mu-plugins/blocks/package-lock.json
@@ -567,61 +567,101 @@
 				}
 			}
 		},
-		"@emotion/babel-utils": {
-			"version": "0.6.10",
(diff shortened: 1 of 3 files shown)


iandunn:
> shrinks our build 13%

❤️ ❤️ ❤️

iandunn (approved):

PR #163, opened 2019-07-19
Blocks: Clean up the "add more" UI when some content is selected

### LINKED ISSUE #75: All: "Add More" placeholder design updates
- [ ] "Add more" placeholders should not be shown when block is not selected
- [ ] "Add more" placeholders should be simplified to just the "add additional content" view, per the mockups, rather than the default placeholder view (and shouldn't show content that's already selected; this should only be used to add additional content to the block, not edit what's already being shown)
- [ ] Wasn't in the mockups, but we should probably add an edit toggle in the toolbar to let people reset the content via the placeholder, like we do in embeds and whatnot

### LINKED ISSUE #161: Blocks: Dropdown falls under block toolbar on small screens
At <600px, the toolbar is always visible, and must have a higher z-index set:

Screen Shot 2019-07-19 at 11 58 39 AM

ryelle:
This PR does a few things to style the "placeholder" when adding content (I'm calling it an appender from here, though I did not change the component name).

- Only show the appender when the block is selected, which gives a cleaner preview of what the post will look like on the frontend
- Simplify the appender display to not use the full `Placeholder` setup
- Fix some bugs around z-index and class name consistency

Fixes #161, Fixes #75

Screen Shot 2019-07-19 at 4 25 55 PM

**To test**

- Add a block to a post, ex Speakers
- Select some content (a few speakers), hit select to show them
- You should see a smaller content selection UI now
- Click out of the block
- Only the selected items show, no selection UI
- Click back into the block, the selection UI is back and you can add/remove content.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
@@ -46,13 +46,13 @@ export class BlockControls extends Component {
 			case 'wcb_organizer_team' :
 				output = (
 					<PlaceholderSpecificMode
-						label={ getOptionLabel( mode, options.mode ) }
-						icon={ icon }
 						content={
 							<BlockContent { ...this.props } />
 						}
 						placeholderChildren={
-							isSelected && <ContentSelect { ...this.props } />
+							isSelected && (
+								<ContentSelect label={ getOptionLabel( mode, options.mode ) } { ...this.props } />
+							)
 						}
 					/>
 				);
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
@@ -29,7 +29,7 @@ export class BlockControls extends Component {
 	 * @return {Element}
 	 */
 	render() {
-		const { icon, attributes, setAttributes, blockData } = this.props;
+		const { attributes, blockData, icon, isSelected, setAttributes } = this.props;
 		const { mode } = attributes;
 		const { options } = blockData;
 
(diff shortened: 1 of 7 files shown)


public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
```diff
@@ -61,11 +61,11 @@ export class BlockControls extends Component {
 			default :
 				output = (
 					<Placeholder
-						className={ classnames( 'wordcamp__edit-placeholder', 'has-no-mode' ) }
+						className={ classnames( 'wordcamp-block__edit-placeholder', 'has-no-mode' ) }
```
iandunn: Corey and I have been using a `MUST`, `SHOULD`, `MAYBE` convention during reviews [to try and make expectations clear](https://www.brandonsavage.net/the-pitfalls-of-code-review-and-how-to-fix-them/).

"I think..."  is still implied, though, so even if I said `MUST`, that would just mean, I "I think this is a blocker", and you're always welcome to push back if you disagree.

----

`SHOULD` - I think this needs a wider discussion, and feels unrelated to this PR. I'd separate it out into a new PR so we can discuss it there.

See https://github.com/WordPress/wordcamp.org/issues/96 / #123

iandunn (changes requested):
I think changing the prefix needs to be a separate PR, but I think the rest is good to merge.

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/block-controls.js
ryelle: Makes sense. The standardization was more important before I slightly-refactored the PlaceholderSpecificMode component, since it was clearly supposed to be using the same prefix between the two. I'll push that to another PR so we can pick which prefix to use.

iandunn (approved):
Great job :)

PR #164, opened 2019-07-19
Meetup API Client: Use OAuth for authentication instead of an API key

coreymckrill:
Meetup.com [announced](https://groups.google.com/forum/#!topic/meetup-api/R__7mPzWJc0) that it's API key method of authentication would be going away on August 15th, 2019. That's what we've been using, so we need to switch to their OAuth2 method instead.

In order to avoid having to manually redirect to Meetup.com and login with credentials before each time we need to pull the data from the API, we are using the [Server Flow with User Credentials](https://www.meetup.com/meetup_api/auth/#oauth2servercredentials) approach. Unfortunately this requires a non-standard grant type, and a strange extra step of using an access token to request a separate, more special access token. As such, existing OAuth client libraries don't support it out of the box (that I'm aware of). Instead, this PR adds a very simple, custom OAuth2 client.

### DIFF
--- a/public_html/wp-content/mu-plugins/utilities/class-meetup-oauth2-client.php
+++ b/public_html/wp-content/mu-plugins/utilities/class-meetup-oauth2-client.php
@@ -1,0 +1,460 @@
+<?php
+namespace WordCamp\Utilities;
+
+use WP_Error;
+
+defined( 'WPINC' ) || die();
+
+/**
+ * Class Meetup_OAuth2_Client
+ *
+ * Implements the "Server Flow with User Credentials" for Meetup's OAuth2 authorization scheme.
+ *
+ * @see https://www.meetup.com/meetup_api/auth/#oauth2servercredentials
+ *
+ * @todo Maybe make some of the credential strings like consumer key, etc, class parameters that map to private
+ *       properties instead of constants. That way the client could be reused elsewhere in the Dotorg ecosystem with
+ *       different credentials, instead of being tightly coupled to WordCamp.
+ *
+ * @package WordCamp\Utilities
+ */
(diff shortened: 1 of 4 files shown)


public_html/wp-content/mu-plugins/utilities/class-meetup-oauth2-client.php
```diff
@@ -0,0 +1,460 @@
...
+		);
+
+		switch( $type ) {
+			case 'anonymous_code':
+				$request['code'] = $code;
+				break;
+			default:
+				$request[ $type ] = $code;
+				break;
+		}
```
iandunn, line 235: `SHOULD` - It's not obvious to me what's going on here. Is one of these for getting a new code, and the other for renewing an expired code? It'd be helpful to add inline comments for each saying what they're used for.

iandunn (approved):
LGTM, good job digging into their strange implementation on short notice so this critical part of our program can continue working!

public_html/wp-content/mu-plugins/utilities/class-meetup-oauth2-client.php
coreymckrill: Fixed in 0f24fca, but then I ended up doing a significant refactor in ea24b48. Might be worth taking another look to see if it's understandable now.

iandunn: Yeah, this looks good. Kudos for persevering through their inaccurate docs!

iandunn (approved):

coreymckrill:
Merged via aa9e9247f19036dddc64992082114cf23d4d3f63

PR #165, opened 2019-07-23
 Blocks: Remove eslint overrides for declaration spacing.

ryelle:
This PR removes the overrides to `no-multi-spaces` & `key-spacing`.

As a WordPress community project, we should be following the core coding standards with as few modifications as possible, so that anyone involved in a core project can jump into our codebase without needing to learn a new style. Additionally, by overriding the `no-multi-spaces` rule, we're also opening the possibility of incorrect spacing elsewhere in the code.

Fixes #148, fixes #149.

**To test**

1. Run `npm run lint:js`
2. There should be no errors

PR #166, opened 2019-07-23
Blocks: Add eslint rule to sort multiple member imports.

ryelle:
An attempt to add standards to our `import`s. We already break these into dependency groups per the [`dependency-group`](https://github.com/WordPress/gutenberg/blob/master/packages/eslint-plugin/docs/rules/dependency-group.md) rule, but we don't enforce any order inside each group. This proposes [`sort-imports`](https://eslint.org/docs/rules/sort-imports) to at least sort the multiple-item imports. By alphabetizing these imports, we can keep the order consistent across files.

For example, it will flag the following as incorrect:

	import { c, a, b } from 'foo';

Running `eslint --fix` will update this to

	import { a, b, c } from 'foo';

We could also look into codemods to sort all existing imports (see [this codemod from calypso](https://github.com/Automattic/calypso-codemods/blob/master/transforms/sort-imports.js)), 

**To test**

1. Run `npm run lint:js`
2. There should be no errors.

_(Note: this PR is built off #165, but can be judged as a stand-alone PR)_

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/.eslintrc.js
+++ b/public_html/wp-content/mu-plugins/blocks/.eslintrc.js
@@ -123,6 +123,25 @@ module.exports = {
 		} ],
 
 		/*
+		 * Sort imports alphabetically, at least inside multiple-member imports. Ignores declaration sorting since
+		 * this interfers with the External/WordPress/Internal groupings. For example, it will flag the following
+		 * as incorrect:
+		 *
+		 *   import { c, a, b } from 'foo';
+		 *
+		 * Running `eslint --fix` will update this to
+		 *
+		 *   import { a, b, c } from 'foo';
+		 *
+		 */
+		'sort-imports': [
+			'error',
+			{
+				ignoreDeclarationSort: true,
+			},
+		],
+
+		/*
 		 * Descriptions are often obvious from the variable and function names, so always requiring them would be
 		 * inconvenient. The developer should add one whenever it's not obvious, though.
 		 *


public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/content-select.js
```diff
@@ -12,7 +12,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
-import { buildOptions, ItemSelect, Option } from '../../components/item-select';
+import { ItemSelect, Option, buildOptions } from '../../components/item-select';
```
iandunn: `MAYBE` - It feels kind of odd to me that this is case-sensitive, but I don't feel too strongly.

iandunn (approved):

public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/content-select.js
ryelle: I went back & forth on that - we can ignore case if we want, but on the other hand if capitalization is an indicator of "type" (capital imports are components, while lowercase imports are helpers/settings), it does make sense to keep them separated somewhat.

coreymckrill: 👍 

Keeping it case-sensitive is fine with me.

coreymckrill (approved):

PR #167, opened 2019-07-23
Blocks: Update folder structure and component names, bring code up to new style standards.

ryelle:
Following conversations in #167, #183, #184, we've created specific code organization standards. This PR applies these standards to all existing blocks and components. Specifically, component names should reflect the content of the component, clarify our folder structure, and clean up the shared components folder.

----
Previous PR description:

Using the rule of one component per file makes it easier to find where a component lives, and makes individual components easier to test and document. We can also use `prop-types` to add some automatic prop validation (on development, will not add to production bundle). Additionally, this makes the CSS class names slightly easier to generate.


public_html/wp-content/mu-plugins/blocks/.eslintrc.js
```diff
@@ -137,6 +137,13 @@ module.exports = {
 			'destructuring': 'all',
 		} ],
 
+		/**
+		 * Disallow creating more than one component per file.
+		 *
+		 * Having one component per file makes components easier to test and easier to document.
+		 */
+		'react/no-multi-comp': [ 'error' ],
```
iandunn: Is this just for stateful components, or functional as well?

ryelle: This is for all components, but we could use `ignoreStateless` to allow functional components– as of right now I think even our functional components are complex enough that they could use the separate file + PropTypes approach, but I'm OK with allowing simple stateless components through.

iandunn: That would be my preference for a general rule, but I agree that there may be specific ones that warrant separating.

Things like `GridInspectorPanel` and `GridColumnsControl ` don't feel all that complex to me at first glance, but I'm open to it. I guess the thing I'm worried about is that having a ton of folders/files can be just as hard to navigate as having a component buried in a non-obvious file w/ other components, especially when everything is just named `index.js`.

ryelle: Ideally our folders would also have a better naming scheme where the folder name reflects the "top level" component (`index.js`), and each sibling component file name would match the component too - so `post-list/inspector-controls.js` would be `post-list/grid-inspector-panel.js` and `post-list/grid-columns-control.js`

But actually this one's not a great example, because I think `GridColumnsControl` could be merged into `GridInspectorPanel` directly since it's not used anywhere else.

ryelle: I mostly tried to leave file names the same so imports wouldn't change, but if we want to be more clear with files/folders I can also update the PRs to use that naming scheme too.

iandunn: I'm open to renaming stuff, but I agree it'd be more manageable if that were a separate issue.

ryelle: OK, once we've merged/decided on the sub-PRs here I can make another PR following that, but it doesn't make sense to start until this is settled.

coreymckrill: +1 for `ignoreStateless`

ryelle:
I've merged all the code-moving PRs into this PR, all of those have already been reviewed individually.

You can test this PR now by running `npm run lint:js`, and it should not pick up any eslint errors.

ryelle:
Ok, so, this PR is now a monster, but I've addressed all the component renaming we discussed in #183. I also got rid of every `{ ...this.props }` passed through components, since I was touching every file it was easy to track as I changed things.

I also added an `index.js` to the top-level `components` file, so we can just import all shared components from the one entry point (since we don't do any file chunking/tree shaking we don't loose anything by merging these all together, and it lets us rearrange, add & remove the components without needing to change every block import).

coreymckrill:
> since we don't do any file chunking/tree shaking we don't loose anything by merging these all together

Is that something that would be useful now or in the future for file size optimization or something? Would this change make that harder to implement?

ryelle:
> Is that something that would be useful now or in the future for file size optimization or something? Would this change make that harder to implement?

So AFAIK it's mainly used for apps with multiple pages/sections (like wp.com, for example) so that you don't need to load the bundle with all the post stuff if you're just looking at themes. Since we want all our JS on the one page (editor), we don't need to make this kind of optimization, and I don't see us needing to in the future. 

I suppose we could end up in a scenario where a component is imported into the index.js but then never used, and I'm not sure if webpack is smart enough to prune that out.

coreymckrill (approved):
`npm run lint:js` runs for me without any errors.

ryelle:
I've updated the PR title & description to more broadly describe what this PR does, and also added [a README](https://github.com/WordPress/wordcamp.org/blob/f501fbf1c7a0e83efddae90e138757908ecc64c8/public_html/wp-content/mu-plugins/blocks/source/README.md) to collect all the things we've talked about in the past week into one place.

PR #170, opened 2019-07-23
Blocks: Splitting block-content components

ryelle:
See #167 for background + merge plan

Split out the few components in `source/components/block-content/index.js` into separate files.

Previously `block-content/index.js` did a lot of work. The following components have moved:

- `DangerousItemHTMLContent` -> `./item-content.js`
- `ItemPermalink` -> `./item-permalink.js`
- `ItemTitle` -> `./item-title.js`
- `BlockNoContent` moved to a different folder -> `../post-list/no-content.js` since this component is more to do with an empty list than the others which deal with individual item content.

Each of the `item-*` files also have had their class names updated to reflect the file & component names (so they use `wordcamp-block__item-*` now, not `wordcamp-block__post-*`). `BlockNoContent` uses `wordcamp-post-list` like the post list that replaces it once loaded.

Lastly, 4d3324d updates the links to be more consistent across UI and interaction types.


coreymckrill (approved):

public_html/wp-content/mu-plugins/blocks/source/components/block-content/item-content.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+DangerousItemHTMLContent.propTypes = {
+	className : PropTypes.string,
+	content   : PropTypes.string,
+	link      : PropTypes.string,
+	linkText  : PropTypes.string,
+};
+
+export default DangerousItemHTMLContent;
```
iandunn, line 46: `SHOULD` - We've been using named exports only. Do you think there's a good reason to use `default`?

https://github.com/WordPress/wordcamp.org/commit/a7eeea1e0330c8ec4f7c3ddcd0ce1956b950551a

public_html/wp-content/mu-plugins/blocks/source/components/block-content/item-content.js
```diff
@@ -0,0 +1,46 @@
+/**
```
iandunn, line 1: `MAYBE` - I would have thought the file would be named `dangerous-item-html-content.js`; what's the reason for it to be different than the name of the component?

ryelle: It should be, but tbh I thought I'd get resistance for the long file name.

iandunn: No objection from me, I usually strongly prefer clarity over brevity.

public_html/wp-content/mu-plugins/blocks/source/components/block-content/item-content.js
ryelle: Mirroring gutenberg patterns. Each file has a default export which matches the file name, and is a good way of indicating which is the "top level" component in the case where you have some functional/stateless components in the same file.

iandunn: > Mirroring gutenberg patterns.

I don't think that is a compelling reason here, in-and-of itself, given [the problems with default exports](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad?gi=d96cb381d710).

> idicating which is the "top level" component

I can see the benefit of that, but IMO it's far outweighed by the downsides of named exports.

ryelle: I'm not really convinced by that article. I see the refactoring point about typos, but I don't think that's a reason alone to ban default. The better devex is really only a perk of named if you happen to use an IDE that does that, and the tree shaking point is an antipattern regardless of whether `default` is "good" or not.

Using default can also encourage single-component files, while named exports allows for exporting multiple components, and it might not always be clear from scanning a file which of the components are exported. With default exports, it's always just the one thing.

iandunn: > The better devex is really only a perk of named if you happen to use an IDE that does that

Does yours not do that? I'd assume that's a pretty universal feature, but could be wrong.

> Using default can also encourage single-component files

I don't think that's always a good thing. The are times when it's better to split, but also times when I find lots of little files harder to deal with.

> it might not always be clear from scanning a file which of the components are exported. 

I'm not sure what you mean, it seems like it'd be obvious if you include `export` in the function/class definition.

ryelle: > Does yours not do that?

Nope. I'm sure there's some extension I could add, but it's never seemed useful to me.

> The are times when it's better to split, but also times when I find lots of little files harder to deal with.

Small stateless components are not included in this because they're not exported. This kind of goes back to the file naming thing too, where the file name should match the default exported component. If we have multiple exports, that naming scheme falls apart.

> I'm not sure what you mean, it seems like it'd be obvious if you include export in the function/class definition.

Sure, but you still need to _see_ that is says `export function …` vs just `function …`. I missed that a few times in this project. With default exports, you know there's just the one thing, usually at the end of the file.

iandunn: > Small stateless components are not included in this because they're not exported. 

Doh, that's a great point.

> This kind of goes back to the file naming thing too, where the file name should match the default exported component. If we have multiple exports, that naming scheme falls apart.

I do agree that we probably shouldn't have multiple exported components in the same file.

The big thing that I like about named exports is that they're explicit, and force renaming to be explicit as well. For me that outweighs all the rest, but you make good points, so I'm happy to switch back to defaults if you and @coreymckrill prefer them.

public_html/wp-content/mu-plugins/blocks/source/components/block-content/item-content.js
coreymckrill: Works for me.

public_html/wp-content/mu-plugins/blocks/source/components/block-content/item-content.js
coreymckrill: Doing default exports would be a reversal of a decision we made earlier in the development of our blocks. That said...

After reading through this discussion, it makes sense to me to do default exports _as long as there is only one exported component per file_, which I think we all agree is a good idea. In our `components` directory, we're kind of doing both, right? The individual files have default exports, but then **index.js** exports them all explicitly (like [here](https://github.com/WordPress/wordcamp.org/pull/170/files#diff-07b48c2348be3259ed7a8b70cc89a8c1R4-R6))

ryelle: Yeah, that's basically right– any file with a component should have the component as a default export, while that root components file is more of a helper/convenience - so now you can `import { ItemTitle, ItemPermalink } from '../components';`, rather than having to do multiple imports drilling into different files.

PR #171, opened 2019-07-23
Blocks: Splitting block-content session components

ryelle:
See #167 for background + merge plan

Moves the extra session display components into a new `content` folder, adds PropTypes. Also updates the speaker name link to open in a new window, to prevent leaving the editor.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/block-content.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/block-content.js
@@ -1,111 +1,28 @@
 /**
  * External dependencies
  */
-import { get }    from 'lodash';
 import classnames from 'classnames';
+import { get }    from 'lodash';
 
 /**
  * WordPress dependencies
  */
-import { Component } from '@wordpress/element';
 import { __ }        from '@wordpress/i18n';
+import { Component } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
-import { ItemTitle, DangerousItemHTMLContent, ItemPermalink }  from '../../components/block-content';
-import { FeaturedImage }                                       from '../../components/image';
-import { BlockNoContent, PostList }                            from '../../components/post-list';
-import { filterEntities }                                      from '../../data';
-import { tokenSplit, arrayTokenReplace, intersperse, listify } from '../../i18n';
-
-/**
- * Component for the section of each session post that displays information about the session's speakers.
- *
- * @param {Object} session
- *
- * @return {Element}
- */
-function SessionSpeakers( { session } ) {
-	let speakerData = get( session, '_embedded.speakers', [] );
-
-	if ( speakerData.length === 0 ) {
-		return null;
-	}
-
-	speakerData = speakerData.map( ( speaker ) => {
-		if ( speaker.hasOwnProperty( 'code' ) ) {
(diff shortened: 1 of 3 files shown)


coreymckrill:
> Also updates the speaker name link to open in a new window, to prevent leaving the editor.

Ideally this link shouldn't go anywhere when it's rendered within the block editor, right? It just needs to appear as a link so it more closely represents what will be rendered on the front end. Elsewhere in our blocks we handle this by wrapping links in a `<Disabled />` component. Could we do that here as well instead of adding a target attribute?

ryelle:
> Ideally this link shouldn't go anywhere when it's rendered within the block editor, right? It just needs to appear as a link so it more closely represents what will be rendered on the front end. Elsewhere in our blocks we handle this by wrapping links in a <Disabled /> component. Could we do that here as well instead of adding a target attribute?

I swear I had a note about this somewhere, but I can't find it now 🤷‍♀  

The Disabled component doesn't really work like it says– it only "disables" inputs/textareas, and then overlays an element over top of the items to prevent mouse interaction. It still allows keyboard interactions on things like links, so in that case we have a different pattern if you're keyboarding or mousing. When [I raised the issue in core,](https://github.com/WordPress/gutenberg/issues/15799) it sounds like this is … if not expected behavior, not a fix they want to make. So rather than having links that do different behavior depending on interaction mode, I've been removing the `Disabled` wrappers and just having them open in new windows.

coreymckrill:
I may have just ran across the note you were thinking of:
https://github.com/WordPress/wordcamp.org/pull/176#discussion_r307877449

I'm not a fan of having live links within the content preview in the editor. If I'm reading the Core issue you raised correctly, it sounds like you aren't either :)

I'm not sure adding a target attribute is a sufficient solution, though, it still takes you away from the tab you are working on in a really distracting way. Since it sounds like the core `<Disabled>` component was not actually intended to be used for links, what if we created our own component that handles links consistently for all interaction modes? _(This might be a topic for a separate issue)_

ryelle:
No, I thought I'd explained it somewhere but I guess not :| 

I was actually convinced by the a11y issues– there isn't a good way to handle this for all interaction modes. If you're moving through a page and you land on a link, it should be announced as such (and removing the href might remove that). If we somehow short-circuit the link, it might not be clear for someone using a screen reader why they can't click the link. There might also be reasons to _want_ to click the link from the editor. So while opening in a new tab is a bit disruptive, I think it's the best solution.

coreymckrill:
Ok, I trust your assessment. If users complain vociferously about accidentally opening new tabs from the block content, we can revisit.

coreymckrill (approved):

iandunn:
I agree that opening in a new window is a good compromise.

iandunn (approved):
This feels like a grey area to me; I can how having multiple files might be just as difficult in some ways, but I don't feel strongly.

iandunn:
We should probably be consistent across blocks here, so it seems like the next step after merging this would be to add another PR to do this same thing for the others, is that right?

ryelle:
I just updated this PR, but I actually think I'm going to close this one instead, and wait on the outcome of #183 & #184 -- since that might also affect how we structure these files. Now that we've got the `ignoreStateless` rule, eslint no longer flags the `Session*` components.

PR #174, opened 2019-07-24
Blocks: Update selected items to match design

### LINKED ISSUE #80: All: Restyle "select" dropdown to match mockups
Restyle "select" dropdown to match mockups:
- [x] No divider between content and down chevron
- [x] Remove "clear all" x (previously: Don't need to have an X to remove both individuals and all contents within the dropdown) 
image

- [x] Tokens don't need featured images (like in the speakers block)
- [x] Tokens should be styled like core Gutenberg tokens (see the UI for tags)
- [x] Tokens don't need to display the number of items within them (it's okay in the search results, but too small to make out clearly within the token; we should just exclude) 
image
- [x] Select dropdown & button should be the same height

ryelle:
Fixes #80 – clean up the UI of the select input and selected tokens. `react-select` uses `formatOptionLabel` for both options in the list and selected tokens (or chips), but differentiates by passing in `context`. So I've updated `Option` to render a simplified view when context is `value`, while keeping more detail in the list (`menu`) view.

To style the tokens themselves, I needed to hook into react-select's use of [emotion,](https://emotion.sh/docs/introduction) which is a CSS-in-JS solution, so now we have some CSS-esque code in `ItemSelect` 🙃 but the things we needed to style are available, so it's all fine.

**Screenshots**

list

picked

remove-focus

**To test**

1. Add any WC block.
2. Select items.
3. Check that nothing is misaligned, that it looks like the screenshots above.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/content-select.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/content-select.js
@@ -112,11 +112,14 @@ export class ContentSelect extends Component {
 				selectProps={ {
 					options           : this.buildSelectOptions(),
 					isLoading         : this.isLoading(),
-					formatOptionLabel : ( optionData ) => {
-						return (
-							<Option { ...optionData } />
-						);
-					},
+					formatOptionLabel : ( optionData, { context } ) => (
+						<Option
+							context={ context }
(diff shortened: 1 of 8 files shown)


iandunn (approved):
LGTM

melchoyce (approved):
Screenshots look good!

PR #176, opened 2019-07-26
Blocks: Match speakers output to mockups

### LINKED ISSUE #94: Speakers: "Session Information" should match mockup
- [ ] "Sessions" header 18px and margin-bottom 8px
- [x] Title and meta on two separate lines
- [ ] Meta should be 14px and italic

image

ryelle:
Style tweaks and some minor component cleanup.

Fixes #94.

Screen Shot 2019-07-26 at 1 43 13 PM

To test:

1. Add a Speakers block to a post.
2. Toggle the content settings on & off.
3. Display should match the screenshot (given some flexibility for theme styles).


public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/index.js
```diff
@@ -6,6 +6,7 @@ import { __ } from '@wordpress/i18n';
 /**
  * Internal dependencies
  */
+import './style.scss';
```
iandunn, line 9: `SHOULD` - My first thought would be to move in the other direction, so `sessions/style.scss -> sessions/block-content.scss`. The Schedule block will have both `block-content.scss` and `inspector-controls.scss`, and I think it'll be harder to  work with if they're combined into a single file.

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/block-content.js
```diff
@@ -35,63 +33,59 @@ import './block-content.scss';
...
-										)
-									}
-								</span>
-							</Disabled>
-						</li>
-					) }
-				</ul>
-			</div>
-		);
+	if ( ! sessions.length ) {
```
iandunn, line 34: Returning early here is much nicer, kudos :)

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/block-content.js
```diff
@@ -157,16 +151,13 @@ export class BlockContent extends Component {
...
 				className="wordcamp-speakers"
 			>
 				{ posts.map( ( post ) =>
 					<div
 						key={ post.slug }
-						className={ classnames(
-							'wordcamp-speakers__post',
-							`slug-${ post.slug }`,
-						) }
+						className={ `wordcamp-speakers__post slug-${ post.slug }` }
```
iandunn, line 158: I like removing `classnames` 👍 It almost always seems silly to be to be dependent on 3rd party code just to concat some strings.

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/block-content.js
```diff
@@ -35,63 +33,59 @@ import './block-content.scss';
...
-					{ _n( 'Session', 'Sessions', sessions.length, 'wordcamporg' ) }
-				</h4>
-
-				<ul className="wordcamp-speakers__sessions-list">
-					{ sessions.map( ( session ) =>
-						<li
-							key={ session.id }
-							className="wordcamp-speakers__sessions-list-item"
-						>
-							<Disabled>
```
iandunn, line 53: Why was `<Disabled>` removed? it'd also be good to document the reason in the commit message.

ryelle: For the same reason it's been removed in other PRs - it doesn't play well with links. I didn't think to explain it again, but I can add it in the merge commit if you think it's necessary.

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/index.js
ryelle: I haven't officially proposed standardizing this, but if we want to follow gutenberg (and eventually get the benefit of having separate editor & general styles), [we should have max 2 .scss files per folder,](https://developer.wordpress.org/block-editor/contributors/develop/coding-guidelines/#scss-file-naming-conventions-for-blocks) one `editor.scss` and one `style.scss`. This is enforced by GB's build process, we could look into doing the same, or opting for the [slightly more complicated version](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0ac1ba488815acc0abfd25451c91652f9c289fa0/webpack.config.js#L78-L95) I used in the Woo blocks.

melchoyce (approved):
LGTM

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/index.js
coreymckrill: +1 for sticking with `style.scss` for now, and eventually moving to a style/editor split like in Gutenberg. I tried to implement that in our original custom webpack config, but I couldn't figure out how to do it efficiently.

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/block-content.js
coreymckrill: More discussion about the removal here:
https://github.com/WordPress/wordcamp.org/pull/171#issuecomment-516200569

I think it's probably worth adding to the merge commit message.

coreymckrill (approved):

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/index.js
iandunn: Ah, splitting between editor and front end sounds like a compelling reason.

Is there a way to prevent duplicating styles, though? If we'd have to duplicate them then I'd be more hesitant.

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/block-content.js
iandunn: Ah, I'd forgotten about that, thanks for the reminder :)

public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/index.js
ryelle: It's not split between editor & front end, it's shared styles (style.scss) and editor-specific styles (editor.scss). If we need to introduce a third, front-end only file I think there's precedence for calling it `theme.css`, but I'm not sure if that's meant to be theme-specific (can look into it more if we need to).

iandunn: Doh, thanks for pointing that out. This sounds good to me then.

PR #178, opened 2019-07-26
Blocks: Add session details to the dropdown

### LINKED ISSUE #92: Sessions: Session results
- [x] Some session results are centered and some are left-aligned? All should be left-aligned. 

image

- [x] Session results styling should match mockup

image
image

- [x] Individual session results should return both session title and session details (see above)
- [x] Individual session results should not show icons (the icons were only there to help differentiate taxonomies from individual results)

ryelle:
The individual sessions should also show the details (date, time, and track) in the select dropdown. To do this, I've extracted a function to get the session details in a human readable format from a session post object, which is then used when building up the select options if the item is a session. This could also allow us to pass more "details" for other post types, if we decide we need to.

Fixes #92.

Screen Shot 2019-07-26 at 5 58 29 PM

**To test**

1. Add a sessions block.
2. Open the dropdown or start searching to show the list.
3. Individual sessions in the list should have more details now.
4. Session tracks & categories should still show only the term names.
5. Other blocks shouldn't be affected by the change.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/components/item-select/utils.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/components/item-select/utils.js
@@ -13,7 +13,6 @@ import { __ } from '@wordpress/i18n';
  * Internal dependencies
  */
 import { filterEntities }    from '../../data';
-import { getSessionDetails } from '../../blocks/sessions/utils';
 
 const buildOptionGroup = ( entityType, type, label, items ) => {
 	items = items.map( ( item ) => {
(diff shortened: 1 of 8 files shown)


melchoyce (approved):
LGTM

public_html/wp-content/mu-plugins/blocks/source/components/item-select/utils.js
```diff
@@ -28,6 +29,11 @@ const buildOptionGroup = ( entityType, type, label, items ) => {
 
 				parsedItem.avatar = get( item, 'avatar_urls[\'24\']', '' );
 				parsedItem.image  = get( item, '_embedded[\'wp:featuredmedia\'][0].media_details.sizes.thumbnail.source_url', '' );
+
+				if ( 'wcb_session' === item.type ) {
+					parsedItem.details = getSessionDetails( item );
+				}
```
coreymckrill: I'd love it if we could avoid importing functionality and adding conditionals into our reusable components that are specific to one block.

Is there possibly a reasonable way to inject the output of `getSessionDetails` into each of the items in `entities.wcb_session` (which is used as `items` here)? That way this function could just look for `item.details` without having to know about specific post types.

ryelle: Good idea, I've moved the `getSessionDetails` into the initial select call so it's all rendered just the once & stored for all sessions - and now we don't need to call the whole function again when rendering in the content, too.

coreymckrill: 🎉

coreymckrill (approved):

PR #181, opened 2019-07-29
Docker: Allow more file types to be loaded from theme or plugin directories

ryelle:
The nginx config blocks non-safelisted extensions from being found on the server. The safelist is pretty restricted though, and prevents some theme & plugin assets from loading. This was causing problems with Jetpack's icon font (and the ability to load SVGs from a theme).

I'm not sure if this was an oversight or an intentional reflection of the WordCamp infrastructure, but thought I'd PR it.

To test

- Activate Jetpack
- The Jetpack menu item should have an icon

Screen Shot 2019-07-29 at 11 39 26 AM

### DIFF
--- a/.docker/config/nginx.conf
+++ b/.docker/config/nginx.conf
@@ -46,10 +46,10 @@ server {
 		fastcgi_pass php;
 	}
 
-	location ~* /wp-content/.*\.(js|css|png|jpg|jpeg|gif|ico)$ {
-	    root /usr/src/public_html;
-	    expires max;
-	    log_not_found off;
+	location ~* /wp-content/.*\.(js|css|png|jpg|jpeg|gif|ico|svg|woff|ttf)$ {
+		root /usr/src/public_html;
+		expires max;
+		log_not_found off;
 	}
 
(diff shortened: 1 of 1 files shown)


coreymckrill:
> I'm not sure if this was an oversight or an intentional reflection of the WordCamp infrastructure

I'm not sure either. But on production, it looks like the Jetpack icon is an icon font instead of an svg. I wonder why that's different in the Docker container...

ryelle:
Yeah, Jetpack loads a font (woff, then if that fails a ttf), but I actually noticed this while trying to work on the WCUS site, and some of our theme's SVGs wouldn't load. WCUS has a workaround since we can't upload SVGs to wordcamp.org, but other plugins might use SVGs for icons or display, like Jetpack, again– in the Jetpack dashboard, you can see a failed SVG under "Site Stats"

Screen Shot 2019-07-30 at 10 14 40 AM

ryelle:
Since this is working on production, the font + the SVGs (ex: in Jetpack Dashboard -> My Plan), I think this is just that the docker is misconfigured.

coreymckrill (approved):
Sounds good to me.

PR #185, opened 2019-07-31
Tools: Update how linting runs on travis

ryelle:
Add phpcodesniffer & the WP rules to the composer install (installed at `public_html/wp-content/mu-plugins/vendor`), and set up a composer script to run phpcs & phpcbf. We can now use this both locally and in travis to make sure everyone's got phpcs installed and on the correct versions.

Travis has also been updated to use [jobs,](https://docs.travis-ci.com/user/job-lifecycle/) which will let us run multiple checks in parallel. Here we're able to run the JS & SCSS linters on the blocks project while the PHP Code Sniffer checks run, and we could set this up to include more checks in other projects later.

The travis file should be well documented, but let me know if I should add anything else.

**To test**

- Travis does the testing :) Verify that the checks passed.
- Read through the .travis.yml file to make sure my changes make sense.


.nvmrc
```diff
@@ -0,0 +1 @@
+lts/*
```
iandunn, line 1: `MIGHT` - Is it possible to put this in an existing `webpack.js`, or `.config/nvmrc`, etc?

.travis.yml
```diff
@@ -8,36 +8,46 @@ php:
...
-  # -v flag: Print verbose output
-  # -n flag: Do not print warnings
-  # -p flag: Show progress of the run
-  # --extensions=php: Only sniff php files
-  # tmp.php: So that phpcs do not assume that it has to scan all files in case no php files are changed.
-  - wpcs/vendor/bin/phpcs -v -n -p --extensions=php $TRAVIS_BUILD_DIR/tmp.php $(echo $CHANGED_FILES)
+  # For debugging builds.
+  - node --version
+  - npm --version
+  - timedatectl
```
iandunn, line 14: Is this here to check that the commands are installed, and if the exit status is `0` then the Travis will automatically fail the build?

`SHOULD` - If so, I think it'd be helpful to document that. 

`SHOULD` - I also don't understand what "for debugging builds" means, can you clarify that?

.travis.yml
```diff
@@ -8,36 +8,46 @@ php:
...
+
+jobs:
+  include:
+    # To keep things clean, we set up separate jobs for travis per project.
```
iandunn:
```suggestion
    # To keep things clean, we set up separate jobs for Travis per project.
```

iandunn: Kudos for separating these out, that'll be helpful!

.travis.yml
```diff
@@ -8,36 +8,46 @@ php:
...
+    - name: "Blocks"
+      install:
+        - nvm install && nvm use
+        - npm install npm -g
+        # `npm ci` runs a "clean install"
+        - npm --prefix=$TRAVIS_BUILD_DIR/public_html/wp-content/mu-plugins/blocks ci
+      script:
+        - cd public_html/wp-content/mu-plugins/blocks
+        # Run eslint & stylelint over all JS/SCSS files
+        # @todo Find a way to only run on changed files
```
iandunn, line 28: It's fine to leave this as a future iteration, but just as FYI, there may be some helpful things in `bin/.../sync-svn-wth-git`, `.githooks/prepare-commit-msg`, and `.githooks/pre-commit`,

.travis.yml
```diff
@@ -8,36 +8,46 @@ php:
...
+        # Make sure the build completes.
+        - npm run build
+
+    - name: "PHP Code Standards"
+      # Only run if we're on the "pull request build", otherwise CHANGED_FILES is empty.
+      if: branch = production
+      install:
+        - composer install
+      script:
+        # Create an empty php file, which will be passed to the linter along with all the changed files.
```
iandunn:
```suggestion
        # Create an empty PHP file, which will be passed to the linter along with all the changed files.
```

composer.json
```diff
@@ -131,5 +133,13 @@
 		"wpackagist-theme/twentysixteen": "2.0",
 		"wpackagist-theme/twentyseventeen": "2.2",
 		"wpackagist-theme/twentynineteen": "1.4"
+	},
+	"scripts": {
+		"phpcs": [
+			"phpcs --extensions=php -s -p"
```
iandunn, line 139: `MAYBE`: `-a` would be useful when running manually IMO

iandunn (approved):

composer.json
ryelle: What does `-a` do?

.travis.yml
ryelle: No, it's just to make sure the correct versions of node & npm are installed (hence, for debugging -- if something goes wrong later, we can check that the versions are as expected). These are copied over from previous travis configs, I think `timedatectl` comes in handy when running unit tests to make sure the system clock is set as expected.

.nvmrc
ryelle: No, I checked - `nvm` only looks in the one place for it (at the project level).

composer.json
iandunn: It's the interactive mode, so after you fix a bunch you can hit enter and it re-checks everything. If there are more left it the current file then it shows you the updated list, and if there aren't then it moves on the next file.

.travis.yml
iandunn: Ah, ok. Something like "Print debugging information" feels like a clearer comment to me then.

.nvmrc
iandunn: Why is the file needed? Can we pass command line args instead?

_edit:_ I'm sure there's a reason it's needed, I'm just curious what that is.

ryelle: It's used by nvm to indicate what version of node we should be using on a project-- so as a dev, you can jump into a project and just type `nvm use` and you know you're getting the correct node version. We could specify something for travis, but then it looses the developer helpfulness part :)

iandunn: Since we can easily avoid it for Travis, and it often works fine w/ any recent version, and it's not hard to test a few when it doesn't, I'd vote for not cluttering the root w/ another file.

If you and @coreymckrill are in favor, though, I can disagree and commit.

.travis.yml
coreymckrill: +1

.nvmrc
coreymckrill: FWIW, it appears that in the [original issue](https://github.com/nvm-sh/nvm/issues/110) that spawned the .nvmrc file, a lot of people didn't want the extra file either, and would have preferred just using a value in package.json. Obviously that's not what ended up happening, and I couldn't quite figure out how things played out reading through the issue comments.

Given that variations in a developer's node version can change the package-lock file and/or cause some of our package scripts to not run correctly, it seems worthwhile to specify the correct node version for our project, and the .nvmrc file is a common convention that contributors are likely to understand, so I'm in favor.

composer.json
ryelle: Since this is the command that's also used for travis, I don't think we should add it. You can pass extra flags through to phpcs with the `--` separator, though, so this will work for you:

`composer run-script phpcs [path] -- -a`

PR #187, opened 2019-08-02
Blocks: Add a setting for name/title alignment

### LINKED ISSUE #141: Blocks: Allow centering of content in addition to images
Currently, it's possible to center the image for example, in the Speakers block:

Screen Shot 2019-07-09 at 6 04 08 PM

But if I want the speaker name to be centered above that (for example, [the Boston speakers list](https://2019.boston.wordcamp.org/speakers/)), I need to add custom CSS. We could maybe use the alignment in the toolbar to set the block's text alignment? Like paragraph:

Screen Shot 2019-07-09 at 6 04 54 PM

Should this be content-type specific (so title and excerpt are individually aligned), or should it always be the same (if it's centered, both title and content will be centered)? @melchoyce, thoughts?

ryelle:
Adds an alignment setting to each block, which will control the name/title. The setting is added to the Content Settings panel, and lets you left/right/center align the content.

Fixes #141 

**Screenshots**

In the sidebar, this setting shows up first.

Screen Shot 2019-08-02 at 5 39 04 PM

Except in sponsors, because the Name toggle can hide the setting (if name is hidden, there's no point to the alignment setting).

Screen Shot 2019-08-02 at 5 43 22 PM

In the editor preview, this controls the heading:

Screen Shot 2019-08-02 at 5 43 27 PM

**To test**

1. Add any WC block
2. Toggle the different alignments, it should update correctly across all content items.
3. Save the block, and make sure it works correctly on the frontend too.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/components/item/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/components/item/controller.php
@@ -12,10 +12,11 @@ defined( 'WPINC' ) || die();
  * @param string $link
  * @param int    $heading_level
  * @param array  $classes
+ * @param string $align
  *
  * @return false|string
  */
-function render_item_title( $title, $link = '', $heading_level = 3, array $classes = [] ) {
+function render_item_title( $title, $link = '', $heading_level = 3, array $classes = [], $align = 'none' ) {
 	$valid_heading_levels = [ 1, 2, 3, 4, 5, 6 ];
 
 	if ( ! in_array( $heading_level, $valid_heading_levels, true ) ) {
@@ -29,9 +30,15 @@ function render_item_title( $title, $link = '', $heading_level = 3, array $class
 		$classes
 	) );
 
+	$style = '';
+
+	if ( in_array( $align, [ 'left', 'center', 'right' ], true ) ) {
+		$style = "text-align:$align;";
+	}
+
 	ob_start();
 	?>
-	<<?php echo esc_html( $tag ); ?> class="<?php echo esc_attr( $classes ); ?>">
+	<<?php echo esc_html( $tag ); ?> class="<?php echo esc_attr( $classes ); ?>" style="<?php echo esc_attr( $style ); ?>">
 	<?php if ( $link ) : ?>
 		<a href="<?php echo esc_url( $link ); ?>">
 	<?php endif; ?>
--- a/public_html/wp-content/mu-plugins/blocks/includes/definitions.php
+++ b/public_html/wp-content/mu-plugins/blocks/includes/definitions.php
@@ -37,6 +37,11 @@ function get_shared_definitions( $keys, $type ) {
 					'enum'    => wp_list_pluck( get_shared_definition( 'align_block', 'option' ), 'value' ),
 					'default' => '',
 				],
+				'align_content'     => [
+					'type'    => 'string',
+					'enum'    => wp_list_pluck( get_shared_definition( 'align_content', 'option' ), 'value' ),
+					'default' => '',
+				],
 				'align_image'       => [
 					'type'    => 'string',
 					'enum'    => wp_list_pluck( get_shared_definition( 'align_image', 'option' ), 'value' ),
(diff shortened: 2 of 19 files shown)


public_html/wp-content/mu-plugins/blocks/source/components/item/controller.php
```diff
@@ -29,9 +30,15 @@ function render_item_title( $title, $link = '', $heading_level = 3, array $class
...
+	$style = '';
+
+	if ( in_array( $align, [ 'left', 'center', 'right' ], true ) ) {
+		$style = "text-align:$align;";
+	}
+
 	ob_start();
 	?>
-	<<?php echo esc_html( $tag ); ?> class="<?php echo esc_attr( $classes ); ?>">
+	<<?php echo esc_html( $tag ); ?> class="<?php echo esc_attr( $classes ); ?>" style="<?php echo esc_attr( $style ); ?>">
```
coreymckrill, line 41: Could we add a class like `align-center` or `aligncenter` (or something else) instead of adding an inline style? Seems like that's usually considered a best practice and would give more flexibility to organizers who are customizing the CSS...

coreymckrill: (Same question applies in the **item-title.js** file)

coreymckrill (changes requested):
Everything worked as expected when testing on my sandbox. A few questions, plus one inline thing on the changeset.

- [ ] Seems like the alignment button in the toolbar that's associated with the default alignment should already be highlighted.
- [ ] I think it looks a little weird with the border around the alignment toolbar expanding to the entire width of the Inspector side panel, even though there are only three buttons. In our custom Image alignment toolbar the border shrinks to fit. I think we should standardize on one or the other, and my personal preference would be the shrink to fit.
- [ ] Perhaps slightly redundant, but maybe we should create a new shared definition for `align_content` that only has `left`, `center`, and `right`. Using the `align_image` one is semantically misleading and it includes the `none` option, which doesn't apply here.

public_html/wp-content/mu-plugins/blocks/source/components/item/controller.php
ryelle: I'm following how the core heading block alignment works, and that makes sense to me - if you've explicitly aligned the heading, you don't want some theme styles accidentally overriding it.

> would give more flexibility to organizers who are customizing the CSS

I'm not sure what the use case would be for someone setting "align left" and then wanting to change that to right/center by the theme, where they couldn't just click the right alignment button.

In any case, it is possible to override with `!important` in their CSS (not that I'd encourage that, but it is possible)

ryelle:
> Seems like the alignment button in the toolbar that's associated with the default alignment should already be highlighted

There is no default alignment, so nothing is selected.

> I think it looks a little weird with the border around the alignment toolbar expanding to the entire width of the Inspector side panel, even though there are only three buttons. 

That's a core bug, it's the same on the heading block where the same `AlignmentToolbar`  component is used. For the image alignment, there's kind of a hack where the component is wrapped in a `PanelRow` to take advantage of the flex CSS rule there to change how the `Toolbar` is laid out. Personally, I'd rather not have the `PanelRow` hack and instead leave the `Toolbar`s full width.

> Perhaps slightly redundant, but maybe we should create a new shared definition for align_content that only has left, center, and right. Using the align_image one is semantically misleading and it includes the none option, which doesn't apply here.

Sure - I started down that way but it did feel redundant. I'll make a commit adding that back in.

public_html/wp-content/mu-plugins/blocks/source/components/item/controller.php
coreymckrill: I wasn't thinking about flexibility as overriding the style so much as being able to select things easily. E.g. `.wordcamp-block__item-title.align-center { ... }`

But, I don't feel too strongly, and following the same behavior as Core's heading block sounds good to me.

coreymckrill (approved):

PR #193, opened 2019-08-05
Blocks: Set the fallback sort order for sessions

### LINKED ISSUE #192: Sessions Block: Order by date/time uses different order on front & back end
When multiple sessions are at the same time, the sort order falls back to different metrics on the front and back end, which means the preview isn't technically accurate.

For example, all 4 of these sessions are at 10:10am, but "Playing Ball with Plugins" is first in the editor, and 3rd on the frontend.

Screen Shot 2019-08-05 at 4 20 50 PM

Screen Shot 2019-08-05 at 4 20 57 PM

ryelle:
Sessions sorted by date have a fallback sorting of title on the frontend, but until now there wasn't a fallback on the backend leading to potentially different previews. This adds a title comparison to the custom sorting function, used when the date is equal.

Fixes #192

**To test**

- Create a few sessions at the same time as each other, best if they're not created in alphabetical order
- Handpick those sessions into a session block
- Confirm that the Sorting order is "Day and Time"
- Publish the post
- The editor preview and frontend should have the same order of items.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/session-list.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/session-list.js
@@ -150,9 +150,7 @@ class SessionList extends Component {
 			];
 		}
 
-		if ( 'session_time' === sort ) {
-			args.sort = 'title_asc';
-		} else {
+		if ( 'session_time' !== sort ) {
 			args.sort = sort;
 		}
 
@@ -160,6 +158,11 @@ class SessionList extends Component {
 
 		if ( Array.isArray( filtered ) && 'session_time' === sort ) {
 			filtered = filtered.sort( ( a, b ) => {
+				if ( Number( a.meta._wcpt_session_time ) === Number( b.meta._wcpt_session_time ) ) {
+					const title = get( a, 'title.rendered', '' );
+					return title.localeCompare( b.title.rendered );
+				}
+
 				return Number( a.meta._wcpt_session_time ) - Number( b.meta._wcpt_session_time );
 			} );
 		}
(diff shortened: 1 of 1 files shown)


iandunn (approved):

public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/session-list.js
```diff
@@ -150,7 +150,9 @@ class SessionList extends Component {
 			];
 		}
 
-		if ( 'session_time' !== sort ) {
+		if ( 'session_time' === sort ) {
+			args.sort = 'title_asc';
+		} else {
```
coreymckrill: This seems a little bit confusing since there is a separate clause right below this that does a custom sort if `session_time` is the `sort` value. What if instead of essentially sorting twice, we just added more logic to the custom sort function? Something like:

```js
if ( Array.isArray( filtered ) && 'session_time' === sort ) {
	filtered = filtered.sort( ( a, b ) => {
		if ( Number( a.meta._wcpt_session_time ) === Number( b.meta._wcpt_session_time ) ) {
			return a.title.localeCompare(b.title);
		}

		return Number( a.meta._wcpt_session_time ) - Number( b.meta._wcpt_session_time );
	} );
}
```

_(I hadn't heard of `localeCompare` before, but saw it [here](https://stackoverflow.com/a/16481400))_

ryelle: Yeah, this method makes more sense - I've updated the PR to use this in e42055866d54b119843f00bc8677fc8229ef9bb0

coreymckrill (approved):

PR #194, opened 2019-08-06
Add github issue and PR templates

ryelle:
This adds a simple PR template, you can [see it here.](https://raw.githubusercontent.com/WordPress/wordcamp.org/add/github-templates/.github/pull_request_template.md) We can add any extra notes or steps, but this seems like a good start.

There's also a set of issue templates, which will give people the ability to pick which template to use, like [woocommerce](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/new/choose) and [gutenberg](https://github.com/wordpress/gutenberg/issues/new/choose). I've added "Bug Report", "Feature Request" and "Support Question" - with the last really linking people off to other places to get help.

Finally, I updated the main readme to let folks know that we address new issues/PRs once a month, and to add a note about the monthly meta bug scrubs.

Outside of this PR, I've also updated [the descriptions of labels](https://github.com/WordPress/wordcamp.org/labels) with some notes we made in team chats– specifically:

- [Status] Needs Decision: Needs full team input, decision should be reached within a week
- [Status] Blocked: Progress is blocked, and needs team input to move forward
- **NEW** [Status] On Hold: External blockers, or a de-prioritized project
- **NEW** [Status] Ready to Merge _I added this one a while ago, it's helpful for taking approved PRs out of the "Needs Review" state_

The On Hold status is to differentiate "blocked by team input/work" from "blocked by an external issue", and to give PRs that aren't a priority right now a place to live.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -5,9 +5,11 @@ about: Let us know if something isn't working as you expect it to.
 ---
 
 **Describe the bug**
+
 A clear and concise description of what the bug is.
 
 **To reproduce**
+
 Steps to reproduce the behavior:
 1. Go to '...'
 2. Click on '....'
@@ -15,19 +17,24 @@ Steps to reproduce the behavior:
 4. See error
 
 **Expected behavior**
+
 A clear and concise description of what you expected to happen.
 
 **Screenshots**
+
 If applicable, add screenshots to help explain your problem.
 
-**WordCamp:**
-If this is a problem on a specific WordCamp's site, list that here.
+**WordCamp**
+
+If this is a problem on a specific WordCamp's site, list the site or page URL here.
 
 **System (please complete the following information):**
+
  - Device: [e.g. laptop, iPhone6]
  - OS: [e.g. iOS]
  - Browser [e.g. chrome, safari]
  - Version [e.g. 22]
 
-**Additional context**
-- To report a security issue, please visit the WordPress HackerOne program: https://hackerone.com/wordpress.
+**Security Issues**
+
+To report a security issue, please visit the WordPress HackerOne program: https://hackerone.com/wordpress.
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,33 @@
+---
+name: "\U0001F41E Bug report"
+about: Let us know if something isn't working as you expect it to.
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**WordCamp:**
+If this is a problem on a specific WordCamp's site, list that here.
+
+**System (please complete the following information):**
+ - Device: [e.g. laptop, iPhone6]
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Additional context**
+- To report a security issue, please visit the WordPress HackerOne program: https://hackerone.com/wordpress.
--- a/.github/ISSUE_TEMPLATE/feature-request.md
+++ b/.github/ISSUE_TEMPLATE/feature-request.md
@@ -1,0 +1,17 @@
+---
+name: "✨ Feature request"
+about: If you have an idea to improve an existing WordCamp feature, please let us know.
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here.
--- a/.github/ISSUE_TEMPLATE/feature-request.md
+++ b/.github/ISSUE_TEMPLATE/feature-request.md
@@ -4,14 +4,10 @@ about: If you have an idea to improve an existing WordCamp feature, please let u
 
 ---
 
-**Is your feature request related to a problem? Please describe.**
-A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+Feature requests should be made on the Make Community P2, to ensure everyone can contribute to new ideas. Please follow this process for new features: https://make.wordpress.org/community/feature-requests-for-community-sites/
 
-**Describe the solution you'd like**
-A clear and concise description of what you want to happen.
+If you have a problem, go back and select the Bug Report option.
 
-**Describe alternatives you've considered**
-A clear and concise description of any alternative solutions or features you've considered.
+---------------
 
-**Additional context**
-Add any other context or screenshots about the feature request here.
+If you've already made your feature proposal on the p2, you can summarize it here and make sure to include a link.
--- a/.github/ISSUE_TEMPLATE/support-request.md
+++ b/.github/ISSUE_TEMPLATE/support-request.md
@@ -1,0 +1,20 @@
+---
+name: "❓ Support Question"
+about: If you have a question please see our docs or try Slack.
+
+---
+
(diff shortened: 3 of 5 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,33 @@
...
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**WordCamp:**
+If this is a problem on a specific WordCamp's site, list that here.
+
+**System (please complete the following information):**
```
iandunn, line 31: `MAYBE` - I'd remove this section, since it feels like a lot of work to fill it in w/ specific version numbers etc, and the vast majority of issues can be reproduced without it. It feels like it adds friction to the process without much benefit. We can always follow up and ask for this info if we can't reproduce something in the browsers we have installed locally.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,33 @@
...
+ - Device: [e.g. laptop, iPhone6]
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Additional context**
+- To report a security issue, please visit the WordPress HackerOne program: https://hackerone.com/wordpress.
```
iandunn: `MAYBE` I'd move this to the top, and maybe make it an HTML comment so people don't leave it in the report through laziness.

Maybe change the `additional context` heading to `Security Issues` or something, so it's more descriptive.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,33 @@
...
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**WordCamp:**
```
iandunn: This is a really good idea. People often report stuff w/out telling us which camp they're working on. 

`MAYBE` - I wonder if it'd be better to ask for a specific URL instead, since they might not make it clear which specific screen they're referring to.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,33 @@
...
+name: "\U0001F41E Bug report"
+about: Let us know if something isn't working as you expect it to.
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To reproduce**
+Steps to reproduce the behavior:
```
iandunn, line 13: When I [preview](https://github.com/WordPress/wordcamp.org/blob/3e573def4a32f2b974eaf39475eaab6ba0d49e25/.github/ISSUE_TEMPLATE/bug-report.md) all of these section headers are missing the blank line that they have here.

`SHOULD` - It seems like we should use `###` instead of `**` so that they're actual headings.

This applies to the other files as well.

.github/ISSUE_TEMPLATE/bug-report.md
iandunn: This heading has a `:` but the others don't. I think it should be consistent. 

`SHOULD` - I'd prefer to remove the `:` since IIRC most English style guides recommend against punctuation in headings, but don't feel strongly either way.

.github/ISSUE_TEMPLATE/support-request.md
```diff
@@ -0,0 +1,20 @@
+---
+name: "❓ Support Question"
+about: If you have a question please see our docs or try Slack.
+
+---
+
+If you have a question about using your site, or contributing to the platform, please email support@wordcamp.org or join the `#meta-wordcamp` channel in the official WordPress Slack workspace: https://make.wordpress.org/chat/
```
iandunn, line 7:
```suggestion
If you have a question about using your site, or contributing to the platform, please email support@wordcamp.org or join the `#meta-wordcamp` channel in [the official WordPress Slack workspace](https://make.wordpress.org/chat/).
```

Same applies to the URLs below.

.github/pull_request_template.md
```diff
@@ -0,0 +1,18 @@
...
+
+<!-- If your change has a visual component, add a screenshot here. A "before" screenshot would also be helpful. -->
+
+### How to test the changes in this Pull Request:
+
+1.
+2.
+3.
+
+<!-- If you can, add the appropriate [Component] and [Status] labels. -->
```
iandunn, line 18: `MAYBE` - I think committers will already know to do this, and everyone else won't be able to add labels, so it'd probably be best to remove this.

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -0,0 +1,17 @@
+---
+name: "✨ Feature request"
+about: If you have an idea to improve an existing WordCamp feature, please let us know.
+
+---
+
+**Is your feature request related to a problem? Please describe.**
```
iandunn: `MUST` - Feature requests need to be approved by the Community Team before we can accept them, so I think we should reflect that here.

`MAYBE` - We could just remove all of these sections and just ask people to follow this instead:

https://make.wordpress.org/community/feature-requests-for-community-sites/

...or, maybe we should add a warning (similar to security in the bug report template) to follow that process first, and then once it's been approved, then they can open an issue here, and the template should ask them to link to the p2 post where it was approved. 

If we do that, there shouldn't be a need to describe to problem or alternatives, since those should already be worked through in the p2 post. We'd only need a description of the solution that was agreed upon on the p2.

iandunn (dismissed):
Overall it's really good, kudos!

IMO the feature request approval is a blocker, but everything else is subjective.

.github/ISSUE_TEMPLATE/support-request.md
ryelle: Ideally they'd never see the content post-markdown-conversion though, because this shows up in the issue text editor and the goal is to prevent people from submitting these as issues. So adding `[]()` just adds visual noise to the editor.

.github/ISSUE_TEMPLATE/bug-report.md
ryelle: I'd rather leave it- I've had a lot of issues created where I follow up to ask for this info and the person never replies.

browser/OS is also more important in JS issues, so as we're building more JS-focused things, this will be more useful.

.github/pull_request_template.md
ryelle: It's a good reminder, I'd rather leave it.

.github/ISSUE_TEMPLATE/bug-report.md
coreymckrill: I agree, this is awesome, and also a specific URL would be more helpful.

.github/ISSUE_TEMPLATE/feature-request.md
coreymckrill: > the template should ask them to link to the p2 post where it was approved

+1 for this.

.github/pull_request_template.md
coreymckrill: Yeah, I think these templates should be used by committers also, so I agree it's a good reminder.

.github/ISSUE_TEMPLATE/support-request.md
iandunn: Ah, that's a great point 👍

.github/ISSUE_TEMPLATE/bug-report.md
iandunn: In cases when you can't reproduce it locally in the current version of the major browsers, I would assume that the most likely cause is a browser extension. Is that true in your experience, or did most of the issues that came down to differences in supported versions of the browsers?

It seems like if we can't reproduce it, and the reporter doesn't care enough about it to respond, then we could just close it as `maybelater` and let another reporter re-open it if someone ever runs into it again.

ryelle: Having them report the browser short-circuits the need to trial-and-error tests among all browsers (start in Chrome, then try again in FF, maybe now try Safari, then boot the VM to try Windows + Chrome, Firefox, IE…).

Adding a note about trying with browser extensions disabled could be a good idea though.

iandunn: To me the amount of friction for the user and the dev feels about the same in those cases, but we'd be adding that friction for users every time in order to save us from that friction in a small percent of cases.

So, personally, it still feels like it hurts them more than it helps us.

ryelle:
I've (finally) addressed the comments on the templates – @coreymckrill do you want to give this another 👀 before I merge it? We can always tweak the templates once merged if we need to.

coreymckrill (approved):
👍

Issue #204, opened 2019-08-07
PWA Discussion: Caching configuration


ryelle:
from [comments](https://github.com/WordPress/wordcamp.org/blob/b58e7e1847564765f3427cf622f0ea844a5891a3/public_html/wp-content/mu-plugins/service-worker-caching.php#L73-L84)

There are questions about how these caching methods work for site assets, if they're good defaults and how they work with expirations.

- Assets are limited to 60 entries and are set to invalidate after a day. Assets will be pulled from the cache by default, if they exist there is no network request. We might want to bump the number of entries since there can be a lot of CSS/JS & images (esp with sponsors) on a page.
- Custom CSS is also cached with 60 entries up to a day, but since the `query` value is part of the cache route rule, as soon as the query updates, the service worker's caching rules change. This essentially deletes the old cache rule, and adds a new one for the current custom CSS. We could probably tweak the rules here to cache this indefinitely, since it will invalidate itself.
- API requests are also cached, for 15min. Again, the response will be pulled from the cache if it exists, and skip the network request until it's invalid. Why did it seem necessary to cache the API?
- Navigation pages are limited to 50 entries, and don't time out, but will only request from cache if offline. #202 has more about this.
- There are some things we're not caching– Remote CSS #200, uploaded files #203

Cachebuster strings do not invalidate cached entries. If the resource changes, the `ETag` would change, but this doesn't seem to help. In the case of the assets, this doesn't trigger a change since the network is never hit when a cached value is found. (this might need more research)

As for side effects, the comments call out Gutenberg & Tagregator. It shouldn't affect gutenberg since the admin service worker is separate and removes any caching from the API.

Caching the API request does "break" tagregator– it continues to ping the API every 30s, but gets the same cached value back for 15min. IMO, tagregator doesn't really make sense in an offline context, so we should skip caching it - if we're worried about bandwidth, we can bump the interval to something in 1-5min, but otherwise this caching is causing JS overhead (as the request is fetched and processed at the tggr interval) without having real "real-time" updates.

So to summarize, the questions here are about whether these caching settings sound appropriate.

iandunn:
> API requests are also cached, for 15min. Again, the response will be pulled from the cache if it exists, and skip the network request until it's invalid. Why did it seem necessary to cache the API?

Looks like it was added in https://github.com/wceu/wordcamp-pwa-page/commit/b15ad18dd8bc6049f9a01b21b5b16fa3cb59e01d

[...]

Issue #202, opened 2019-08-07
PWA Discussion: Caching strategy for navigation (pages a user navigates too)


ryelle:
From [comment](https://github.com/WordPress/wordcamp.org/blob/b58e7e1847564765f3427cf622f0ea844a5891a3/public_html/wp-content/mu-plugins/service-worker-caching.php#L88)

Currently we cache pages a user navigates to automatically, up to 50 pages. We're using the `NetworkFirst` caching strategy which means it hits the network, and if it can't load from the network it loads the cached page.

An alternate would be `StaleWhileRevalidate`, which would serve up the cached page while running the network request in the background. Faster for initial view, still makes network request. If the page is then loaded via network, the cache is updated. Either way, you make a network request. If the user is offline, they'd only hit the cache.

The only downside I see to `StaleWhileRevalidate` is that it's a little strange for the browser to "reload" when the network request comes back.

westonruter:
> The only downside I see to `StaleWhileRevalidate` is that it's a little strange for the browser to "reload" when the network request comes back.

Something to consider here would be to show a message like “A new version of the page is available” if an update happened in the background. See Workbox docs on that: https://developers.google.com/web/tools/workbox/modules/workbox-broadcast-cache-update

iandunn:
I like the idea of showing a notice instead of trying to refresh automatically. If that's more trouble than it's worth for the beta release, though, then I don't think it's a high priority.

westonruter:
The `NetworkFirst` strategy is the safest option, at the price of some additional latency. But you'll never get a stale page if you are truly connected to the internet. If you aren't connected to the internet, then you'll likely get served the cached page quickly. The biggest painpoint would be users with very-slow connectivity where they are technically connected but timeouts cause more latency than ideal to then fallback to the cached page.

ryelle:
Yeah, the idea is that we're optimizing for folks who are probably on slow conference wifi or maybe poor data connections - which is why I went with `StaleWhileRevalidate` in #246.

ryelle:
Closed with #246.

Issue #128, opened 2019-06-21
Finish offline template


iandunn:
Picking up where https://github.com/wceu/wordcamp-pwa-page/issues/9 left off.

We've already got it created from a plugin, so now we need to iterate on the content per the discussion in that thread, and fix any remaining `@todo` notes.

--- 

- [ ] Use theme classes on elements to keep theme style
- [ ] Add date/time, location, schedule (use blocks?)
- [ ] Look into `{{{error_details_iframe}}}` or `wp_service_worker_error_details_template()` ?
- [ ] Maybe list all of the pages that are available offline, see https://github.com/WordPress/wordcamp.org/issues/201

PR #208, opened 2019-08-08
PWA: Remove comments that have been moved to github

ryelle:
Leaving comments in files is good for future developers, but it's hard to have discussions or take action based on them. I've copied out all the issues described in these comments into issues with the [PWA](https://github.com/WordPress/wordcamp.org/issues?q=is%3Aissue+is%3Aopen+label%3A%22%5BComponent%5D+PWA%22) and [Theme Templates](https://github.com/WordPress/wordcamp.org/issues?q=is%3Aissue+is%3Aopen+label%3A%22%5BComponent%5D+Theme+Templates%22) labels, or in some cases into trello cards.

The issues with "PWA Discussion", #202 & #204, specifically need discussion, while the rest are tasks to complete. 

I also cleaned up some phpcs errors as I went 🧹

**To test**

- There should be no functionality change anywhere

### DIFF
--- a/public_html/wp-content/mu-plugins/service-worker-caching.php
+++ b/public_html/wp-content/mu-plugins/service-worker-caching.php
@@ -4,90 +4,12 @@ namespace WordCamp\PWA\Caching;
 use WP_Service_Worker_Caching_Routes;
 
 add_action( 'wp_front_service_worker', __NAMESPACE__ . '\register_caching_routes' );
-	// todo is this the most appropriate hook to register these?
-		// seemed to have problems when using `default_service_workers`, like infinite loop.
-	// But do caching routes always affect both workers regardless? Or should they?
-		// Even if the caching is restricted to front end worker, that would still affect Tagregator and others,
-		// so still need to worry about side-effects.
 add_action( 'wp_front_service_worker',    __NAMESPACE__ . '\set_navigation_caching_strategy' );
-
-
-// todo prompt to install app to home screen is automatically showing on mobile
-	// not sure we want it to
-	// not really related to this file, but nothing closer at the moment
-	// maybe want it to be eventually, but not until we do more work to really make the site use pwa features well?
-	// otherwise risk giving users bad impression of pwas
-
-// also, says "install wordcamp" instead of "install wordcamp europe" - maybe manifest issue?
-
-// after adding to home screen, there's a delay and a "wordcamp" interstitial screen before the site appears.
-	// why? can remove that? it should load instantly, right? that's the whole idea
-
-// experience on mid-level phone (moto e4) and fast bandwidth is pretty poor
-	// what's the cause? big images? too much markup to parse? server slow ttfb?
-	// maybe just needs to have better perceived speed rather than better actual speed? like transitions between screens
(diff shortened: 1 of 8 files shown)


iandunn (commented):
Were all of the removed comments added to issues, or only some? There are some where I'm not seeing when I search, like:

> this doesn't have headlines for separate days. why not? will probably be fixed by replacing block with shortcode though

> maybe don't need this file, just embed into maincontroller?

> maybe it should be an explicit/intentional action on the organizer's part? but then how to make sure that	organizers know it's available beyond just handbook documentation?

If those are in issues somewhere, then I think it's fine to merge this. If not, though, then I think all of the removed lines should have corresponding issues before this is merged, because it's important to not lose any of the notes.

ryelle:
All comments that had actions or needed multiple-dev discussions were created as issues (or trello cards), unless it seemed to be part of an existing issue.

> this doesn't have headlines for separate days. […]

Adding headlines and/or using the block would be part of the content tweaks, so  I considered this part of https://github.com/WordPress/wordcamp.org/issues/128

> maybe don't need this file, just embed into maincontroller?

I think leaving the file is fine. I prefer having the API functions in their own file anyway - so no action had to come out of this. It didn't seem like the kind of thing we should spend time discussing.

> maybe it should be an explicit/intentional action on the organizer's part? […]

This one is covered by a trello card, "Add "Day of" page to new sites"

iandunn:
> Adding headlines and/or using the block would be part of the content tweaks, so I considered this part of #128

I don't see any mention of headlines in #128, though, so I worry that that would get missed.

> It didn't seem like the kind of thing we should spend time discussing.

I'm open to that, I do tend to be overly detail-oriented at times -- wanting code to be "perfect" rather than just good enough to make the desired impact -- which can slow things down unnecessarily.

Ideally, though, I'd prefer to know _which ones_ are being completely removed; which ones were  moved to an issue verbatim or in spirit; and which ones were assumed to be covered by an issue but without being explicitly stated there. Right now it seems impractical to see that, though, since I'd need to manually go through the diff line-by-line and search all the issues looking for that line (including accounting for it being rephrased). 

Since it's already done, though, it's probably not worth the time to go back and separate those into different commits or something. You're right that some of those were minor nitpicks, and we'll also have the chance to do a final design- and code-review before we open it up to more sites, so we can open GH issues for anything we find there.

iandunn (approved):

ryelle:
Yeah #128 is pretty sketchy, but I consider it part of this checkbox: "Add date/time, location, schedule (use blocks?)" since "schedule" would also cover the day headlines. Is there another reason you called it out specifically in the notes?

> Ideally, though, I'd prefer to know which ones are being completely removed; which ones were moved to an issue verbatim or in spirit; and which ones were assumed to be covered by an issue but without being explicitly stated there. 

Next time, if there's things you don't want lost, maybe you can create issues for them yourself? I did a lot of reading through the notes, parsing out what you meant, testing things, and then using my own judgement to determine what was an issue that needs creation, what's covered by other things, and what needs further discussion. It wasn't a linear process, so removing things doesn't _exactly_ map 1:1 to new issues.

PR #209, opened 2019-08-08
Meetup Client: Refactor to stop using v2 API endpoints

coreymckrill:
Meetup will apparently shut down all of the v2 API endpoints on 2019-08-15. Our client was still using the v2 `events` endpoint, so this refactor changes that to get as close as possible to the same dataset using v3 endpoints. It's not a perfect match, though, and the v3 API is not friendly towards our use case of pulling events from a large number of different groups all at once.

Client changes:
* Refactor the `get_events` method
* Remove unused signing method
* Add a request header for accepting json content type
* Adjust the throttle mechanism to be a bit more conservative
* Sanitize group slugs before using them in API endpoint requests

This also refactors the Meetup Events report in the WordCamp Reports plugin, since the refactored `get_events` method isn't a dropin replacement.

**To test**

Load this changeset into your wporg sandbox and then try running the [Meetup Events report](https://central.wordcamp.org/wp-admin/index.php?page=wordcamp-reports&report=meetup-events). It will take a really long time, but should produce results.

### DIFF
--- a/public_html/wp-content/mu-plugins/utilities/class-meetup-client.php
+++ b/public_html/wp-content/mu-plugins/utilities/class-meetup-client.php
@@ -178,20 +178,6 @@ class Meetup_Client extends API_Client {
 	}
 
 	/**
-	 * Sign a request URL with our API key.
-	 *
-	 * @param string $request_url
-	 *
-	 * @return string
-	 */
-	protected function sign_request_url( $request_url ) {
-		return add_query_arg( array(
-			'sign' => true,
-			'key'  => $this->api_key,
-		), $request_url );
-	}
-
-	/**
 	 * Generate headers to use in a request.
 	 *
 	 * @return array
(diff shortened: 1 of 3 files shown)


iandunn:
If you haven't already, it'd be good to compare the results for the same report queries before and after applying the branch, to ensure that they match exactly. I've caught unexpected errors w/ this type of change in the past that way, and if they're missed they could mess up stats, especially w/ year-to-year comparisons.

coreymckrill:
Good idea.

I used the Meetup Events report to compare total event numbers for 2017 and 2018 between the current, old endpoint and the new one on this branch. The numbers were a bit different, but that turned out to be due to a change in the report logic that is also made in this branch. After adjusting for that, the numbers are nearly identical (2018 is identical, 2017 is off by 8). I don't know how to account for that minor discrepancy, but it could be something within the logic of the v3 endpoint. I'm inclined not to worry about it.

PR #211, opened 2019-08-12
Blocks: Add jest infrastructure for component testing

ryelle:
We need to add a few dev packages: `babel-preset-default` is needed locally for jest to detect it. The `@wordpress/*` packages are needed by jest since we're not loading the whole WP admin context for these tests. As we write more tests, we will probably have to add more packages (but these are not used by webpack, so they do not add to the bundle size).

This PR also adds a trivial test suite for `ItemTitle`, which tests each available prop using snapshot testing. The snapshot file is also version controlled, because it should be the same for everyone (though we probably don't need to commit snapshots to SVN). For more about testing, I've also updated [the readme for this folder.](https://github.com/WordPress/wordcamp.org/tree/try/blocks-add-testing/public_html/wp-content/mu-plugins/blocks/)

**To test**

1. `npm install` to update the dev dependencies
2. `npm test` to run the tests

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/package.json
+++ b/public_html/wp-content/mu-plugins/blocks/package.json
@@ -21,11 +21,10 @@
 		"rememo": "3.0.0"
 	},
 	"devDependencies": {
-		"@wordpress/babel-preset-default": "4.4.0",
 		"@wordpress/components": "8.1.0",
 		"@wordpress/element": "2.6.0",
 		"@wordpress/eslint-plugin": "^2.3.0",
-		"@wordpress/scripts": "^3.3.0",
+		"@wordpress/scripts": "3.4.0",
 		"css-loader": "1.0.1",
 		"mini-css-extract-plugin": "0.4.4",
 		"node-sass": "4.11.0",
--- a/public_html/wp-content/mu-plugins/blocks/package.json
+++ b/public_html/wp-content/mu-plugins/blocks/package.json
@@ -21,6 +21,9 @@
 		"rememo": "3.0.0"
 	},
 	"devDependencies": {
+		"@wordpress/babel-preset-default": "4.4.0",
+		"@wordpress/components": "8.1.0",
+		"@wordpress/element": "2.6.0",
 		"@wordpress/eslint-plugin": "^2.3.0",
 		"@wordpress/scripts": "^3.3.0",
 		"css-loader": "1.0.1",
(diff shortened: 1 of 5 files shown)


iandunn (approved):
Look's great!

public_html/wp-content/mu-plugins/blocks/package.json
```diff
@@ -21,6 +21,9 @@
 		"rememo": "3.0.0"
 	},
 	"devDependencies": {
+		"@wordpress/babel-preset-default": "4.4.0",
```
gziolo: You don't need to import this Babel preset anymore if you use `@wordpress/scripts` v3.4.0. This was a bug which should be resolved now.

I opened https://github.com/WordPress/gutenberg-examples/pull/84 to explore all of this myself and I identified a couple of things which we can simplify on Gutenberg side to make it more seamless. I will work on it in the following weeks.

ryelle: Thanks for pointing that out – I started out by updating the `@wordpress/*` packages but was running into some `eslint` issues. Tried again just now and it's fine - I think part of it might've been that I wanted to add `eslint-plugin-jest` to our config, but then had trouble replicating it to debug.

gziolo: > I think part of it might've been that I wanted to add `eslint-plugin-jest` to our config, but then had trouble replicating it to debug.

I will handle it as part of https://github.com/WordPress/gutenberg/issues/17016. It might take a few weeks until it gets released though. As soon as you want to override config, you have to bring defaults and this might mean more dependencies. That's why it's so important to do the recommended configs correctly. We will refine `@wordpress/scripts` to make it seamless for real :)

gziolo:
I guess this looks great as is. I can't think of anything we can improve without doing direct changes to `@wordpress/scripts`. Thank you for opening an issue in Gutenberg so we could have more use cases to think about improvements :)

public_html/wp-content/mu-plugins/blocks/package.json
ryelle: Yep, that's why I've been working to remove as much of our overrides as we can 😁

ryelle: (but getting the lint errors in the test file was why I started down the eslint-jest path, I ended up just adding a comment to the file in the end, like you did with https://github.com/WordPress/gutenberg-examples/pull/84 )

PR #213, opened 2019-08-14
 Blocks: Load blocks on all sites

ryelle:
Remove the feature gate, to load new blocks on all WordCamp sites. This also adds a check before we run `registerBlockType`, so we can conditionally load individual blocks (this way the Schedule block can be merged in pieces and tested on a few real sites).

We'll feature gate specific blocks by conditionally loading their controller (which populates `WordCampBlocks.[blockName]` with options and the schema). If the block is not set up in `WordCampBlocks`, we can assume it should not be loaded.

**To test**

- Switch your test site to production: `define( 'WORDCAMP_ENVIRONMENT', 'production' );`
- On production, the WC blocks should no longer be loading
- Using this branch, you get all the blocks 🎉 
- Feature gate a specific controller to test conditional block loading, for example we can do the following in `blocks/blocks.php`

```
if (
	( defined( 'WORDCAMP_ENVIRONMENT' ) && 'production' !== WORDCAMP_ENVIRONMENT )
	|| in_array( get_current_blog_id(), [ 928 ], true ) // 2017.testing
) {
	require_once $blocks_dir . 'sessions/controller.php';
}
```

- If you site is still in production mode, you should only see Speakers, Sponsors, and Organizers.
- Turn development mode back on
- You should see all 4 blocks now.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks.js
@@ -10,6 +10,10 @@ import './_z-index.scss'; // Have z-index values, similar to https://github.com/
 import './styles.scss'; // Common styles for WordCamp Blocks.
 import { BLOCKS } from './blocks/'; // Trailing slash required to differentiate the folder from the file.
 
-BLOCKS.forEach( ( { NAME, SETTINGS } ) => {
+const enabledBlocks = BLOCKS.filter( ( b ) =>
+	window.WordCampBlocks.hasOwnProperty( b.NAME.replace( 'wordcamp/', '' ) )
+);
+
+enabledBlocks.forEach( ( { NAME, SETTINGS } ) => {
 	registerBlockType( NAME, SETTINGS );
 } );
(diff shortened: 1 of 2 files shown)


public_html/wp-content/mu-plugins/blocks/source/blocks.js
```diff
@@ -10,6 +10,10 @@ import './_z-index.scss'; // Have z-index values, similar to https://github.com/
 import './styles.scss'; // Common styles for WordCamp Blocks.
 import { BLOCKS } from './blocks/'; // Trailing slash required to differentiate the folder from the file.
 
-BLOCKS.forEach( ( { NAME, SETTINGS } ) => {
+const enabledBlocks = BLOCKS.filter( ( b ) =>
```
iandunn:
```suggestion
const enabledBlocks = BLOCKS.filter( ( block ) =>
```

`SHOULD` - Despite it's proximity to `BLOCKS`, I still don't think `b` is meaningful/descriptive enough.

Same for the line below.

iandunn (approved):
I think the var name should change, but other than that it LGTM

public_html/wp-content/mu-plugins/blocks/source/blocks.js
coreymckrill: +1

coreymckrill (approved):

PR #214, opened 2019-08-14
WC Post Types: "Deprecate" speaker, session, sponsor, organizer shortcodes

ryelle:
As part of launching the new content blocks, we want to "deprecate" the existing shortcodes. This PR moves the shortcode functions to a new "deprecated" file, which is still loaded on all sites, but is a signal that they're going away.

Since we use the shortcodes for the widgets, we can't just turn them off on all new sites. I started a PR to replace the `do_shortcode` call with the blocks' `render` functions, only the blocks don't support the arguments of the widgets (count, random display). I've left these as using the "deprecated" shortcode (with the assumption that we'll deprecate the widgets too, when blocks in sidebars happens).

**To test**

1. Try using the shortcodes on a post or page.
2. There should be no change.


public_html/wp-content/plugins/wc-post-types/inc/deprecated.php
```diff
@@ -0,0 +1,616 @@
+<?php
+/**
+ * Deprecated functionality.
+ *
+ * This is still loaded on all WordCamp sites, but is considered unsupported. Instead of using shortcodes to
+ * display content, the corresponding blocks should be used.
+ */
+
+class WordCamp_Post_Types_Plugin_Deprecated {
```
coreymckrill: This may not currently be anywhere in our dev docs, but we've been trying to move away from old-school namespacing via singleton classes. The [privacy.php](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wc-post-types/inc/privacy.php) and [rest-api.php](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php) files are examples of starting to do that in this plugin.

- [ ] `[SHOULD]` Use something like `WordCamp\Post_Types\Deprecated` as a namespace instead of having everything as methods of a class.

public_html/wp-content/plugins/wc-post-types/inc/deprecated.php
```diff
@@ -0,0 +1,616 @@
...
+	}
+
+	/**
+	 * Convert a string representation of a boolean to an actual boolean
+	 *
+	 * @param string|bool $value
+	 *
+	 * @return bool
+	 */
+	public function str_to_bool( $value ) {
```
coreymckrill: I wonder if this function is really needed anymore. Core has `wp_validate_boolean` now, which is very similar, except that it treats more things as truthy.

- [ ] `[MIGHT]` It may not be worth the time to test for behavior changes, but if it's quick, it would be nice to eliminate this redundant custom function in favor of Core's `wp_validate_boolean`.

coreymckrill (commented):
> Since we use the shortcodes for the widgets, we can't just turn them off on all new sites.

D'oh. I forgot about this. This seems like a good first step, which will make it simpler to feature flag things later once, as you say, we can do blocks in sidebars.

A couple of other comments below inline.

public_html/wp-content/plugins/wc-post-types/inc/deprecated.php
ryelle: Good to know– I saw the pattern but didn't make that connection :) I've updated the PR.

public_html/wp-content/plugins/wc-post-types/inc/deprecated.php
ryelle: `wp_validate_boolean` seems more in line with expected behavior too, so I've switched that over.

coreymckrill (approved):

PR #216, opened 2019-08-14
WC Shortcodes: Update link detection to find block content

ryelle:
(note this PR is based off #214)

Currently, the schedule block links off to sessions and speakers by searching the content to locate the first post/page with `[speakers` or `[sessions` shortcodes. It then uses this as the permalink for the canonical Sessions/Speakers page. When we move the base content to blocks, it will no longer be able to find these pages. 

This PR updates the search for the block comment delimiter, but only if the site doesn't have the skip flag of `content_blocks`. This way, current sites can get the flag set, and they will continue to search for shortcode content. New sites won't have the skip flag set, and will search for block content.

❓  Does this feature flag name/flow make sense? I feel like I have a tenuous idea of how this should work and want to make sure this makes sense to other devs too.

**To test**

- Set up a new site using #215, so it has blocks (or add blocks to an existing site)
- Check that the speaker link goes to the post/page with the speaker block when `[sessions speaker_link="permalink" show_meta="true"]` or with the schedule shortcode
- Switch to another site with shortcode content
- Set the flag: `wp wc-misc skip-feature-flag set content_blocks [siteId]`
- Check that the speaker link goes to the post/page with the speaker shortcode

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -673,12 +673,20 @@ class WordCamp_Post_Types_Plugin {
 
 		switch ( $anchor_target->post_type ) {
 			case 'wcb_speaker':
-				$permalink = has_block( 'wordcamp/speakers', $post->post_content ) ? get_permalink( $post->id ) : $this->get_wcpt_permalink( 'speakers' );
+				if ( ! wcorg_skip_feature( 'content_blocks' ) ) {
+					$permalink = has_block( 'wordcamp/speakers', $post->post_content ) ? get_permalink( $post->id ) : $this->get_wcpt_permalink( 'speakers' );
+				} else {
+					$permalink = has_shortcode( $post->post_content, 'speakers' ) ? get_permalink( $post->id ) : $this->get_wcpt_permalink( 'speakers' );
+				}
 				$anchor_id = $anchor_target->post_name;
 				break;
 
 			case 'wcb_session':
-				$permalink = has_block( 'wordcamp/sessions', $post->post_content ) ? get_permalink( $post->id ) : $this->get_wcpt_permalink( 'sessions' );
+				if ( ! wcorg_skip_feature( 'content_blocks' ) ) {
+					$permalink = has_block( 'wordcamp/sessions', $post->post_content ) ? get_permalink( $post->id ) : $this->get_wcpt_permalink( 'sessions' );
+				} else {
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -673,12 +673,20 @@ public function get_wcpt_anchor_permalink( $target_id ) {
 
 		switch ( $anchor_target->post_type ) {
 			case 'wcb_speaker':
-				$permalink = has_shortcode( $post->post_content, 'speakers' ) ? get_permalink( $post->id ) : $this->get_wcpt_permalink( 'speakers' );
+				if ( ! wcorg_skip_feature( 'content_blocks' ) ) {
```
coreymckrill: - [x] `[MIGHT]` For grokability, it would be nice if the three usages of `wcorg_skip_feature( 'content_blocks' )` in conditional tests were consistent. My personal preference would be to standardize on the "true" value, (so without the `!`)

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -654,9 +654,9 @@ public function get_speaker_wporg_permalink( $speaker_id ) {
 	/**
 	 * Returns an anchor permalink for a Speaker or Session
 	 *
-	 * Any page with the [speakers | sessions] shortcode will contain IDs that can be used as anchors.
+	 * Any page with the Speakers or Sessions block will contain IDs that can be used as anchors.
 	 *
-	 * If the current page contains the corresponding shortcode, we'll assume the user wants to link there.
+	 * If the current page contains the corresponding block, we'll assume the user wants to link there.
 	 * Otherwise, we'll attempt to find another page that contains the shortcode.
```
coreymckrill: - [x] Change `shortcode` to `block` here too
- [x] `[MIGHT]` It might be worth explaining the feature flag in a doc comment here, as a reminder to future devs that these blocks used to be shortcodes.

coreymckrill (approved):
> Does this feature flag name/flow make sense?

I think it does. As mentioned inline below, it will be a bit easier to understand at a glance if all the conditionals are all written the same way.

I'm not sure how this will be implemented on the JS side for the Schedule block content preview, but that's a separate discussion.

ryelle:
I've updated the PR, and in doing so I changed the logic flow a bit so it's more clear how the skip feature flag works. The function docs have also been updated to reflect blocks & add notes about the feature flag.

coreymckrill (approved):

PR #217, opened 2019-08-19
Blocks: Add IDs to items for anchor links

ryelle:
Updates the blocks frontend render code (the `view.php` files) for Speaker & Session to add IDs matching the format of the shortcodes so that the `[schedule speaker_link="anchor" session_link="anchor"]` shortcode output will still work for blocks.

⚠️  This can mean that we have invalid HTML if someone adds more than one block with the same person/session on a page. It seems that we don't have this option at all on the new schedule block, so maybe blocks _shouldn't_ add the IDs to support anchor links?

There's also a note in the shortcode output, which looks like it was added because of [the invalid HTML issue.](https://meta.trac.wordpress.org/changeset/779)

```
<!-- Organizers note: The id attribute is deprecated and only remains for backwards compatibility, please use the corresponding class to target individual speakers -->
```

Right now I think we shouldn't merge this PR, and instead should just silently "deprecate" anchor links since they won't be possible once the schedule block comes out. But I'd already tested this code before coming to that conclusion, so I figured I'd write it up to get other opinions 🙂 

To test:

- Add a Speaker block to a page on your site
- On another page/post, add a schedule shortcode with `speaker_link="anchor"`
- View the schedule, the speaker link should have something like `speakers/#wcorg-speaker-full-name`
- Clicking the link should scroll you down to that specific speaker in the block's list.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/view.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/sessions/view.php
@@ -14,7 +14,7 @@ defined( 'WPINC' ) || die();
 setup_postdata( $session );
 ?>
 
-<div class="wordcamp-session wordcamp-sessions__post slug-<?php echo sanitize_html_class( $session->post_name ); ?>">
+<div id="wcorg-session-<?php echo esc_attr( $session->ID ); ?>" class="wordcamp-session wordcamp-sessions__post slug-<?php echo sanitize_html_class( $session->post_name ); ?>">
 	<?php echo wp_kses_post(
 		render_item_title(
 			get_the_title( $session ),
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/view.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/speakers/view.php
@@ -14,7 +14,7 @@ defined( 'WPINC' ) || die();
 setup_postdata( $speaker ); // This is necessary for generating an excerpt from content if the excerpt field is empty.
 ?>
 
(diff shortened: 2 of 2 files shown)


coreymckrill:
> Right now I think we shouldn't merge this PR, and instead should just silently "deprecate" anchor links since they won't be possible once the schedule block comes out.

I'm in favor of this. If it turns out that linking to a page full of speakers or sessions from the schedule is still needed (as opposed to just linking to the single post), we can figure out a different way to do it as a future iteration of those blocks.

ryelle:
It looks like the `speaker_link` in the schedule shortcode defaults to `anchor`. I could change this default to `permalink`, but if I change it, it will change the behavior for all sites that haven't set this param.

Existing sites are fine (they're using the shortcode), but new sites (until we get the schedule block) will have their default schedules linking speakers to "broken" pages (it'll still link to the speaker page, but won't correctly scroll to that speaker).

coreymckrill:
Could we use a skip feature flag to change the default only for new sites going forward, and maybe even deprecate the `anchor` option altogether?

iandunn:
I don't think the duplicate IDs is a big deal in practice, especially since it's an edge case. The browser would just jump to the first one on the page, which would still meet the user's expectations.

I'm fine w/ deprecating it, though, in the name of limiting support to the most common use cases.

At this point, though, would it be easier to just add it? It sounds like deprecating it may be more work. I'm fine either way, though.

ryelle:
Turns out deprecating it wasn't a big deal: #219  – I only bothered to deprecate it for schedule, since if you're using the other shortcodes, you're using the shortcodes and will have the anchors available 🙂

PR #218, opened 2019-08-20
Blocks: Remove unnecessary API request

ryelle:
The `siteSettings` value isn't used in this block, so we can remove this request to cut down on one API request.

**To test**

- Add the sponsors block
- It should only request the post type & taxonomy endpoints, not the general settings endpoint


coreymckrill (approved):
Good catch. I vaguely remember thinking we didn't need that request anymore, and then I forgot to create an issue for it.

PR #219, opened 2019-08-20
Shortcodes: Deprecate anchor links for the schedule shortcode 

ryelle:
For all new sites that use blocks, we prevent use of the `anchor` setting for session and speaker links. The new default for `speaker_link` is now permalink. Any `schedule` shortcode that tries to use anchor will be silently switched to permalink.

Closes #217.

Note: this still uses the `content_blocks` flag rather than creating a new one, since it relies on whether blocks exist, which that flag will tell us.

**To test**

You can use the following to toggle the skip flag on your site

```
wp wc-misc skip-feature-flag set content_blocks [siteId]
wp wc-misc skip-feature-flag unset content_blocks [siteId]
```

1. Add the `[schedule]` shortcode to your site
2. If you have the flag set, speakers should link to your Speakers page, with anchors pointing to the individual speaker
3. If you don't have the flag set, speakers should link to the individual speaker pages.
4. Try adding `session_link="anchor"`
5. If you have the flag set, each session should be anchor links
6. If you don't have the flag set, sessions will link to the individual session pages (overriding the param set in the shortcode).

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
@@ -164,7 +164,7 @@ function preprocess_schedule_attributes( $attr ) {
 		array(
 			'date'         => null,
 			'tracks'       => 'all',
-			// Default to anchor links only if we're not using blocks on this site.
+			// Sites without the `content_blocks` skip flag use blocks, these do not support anchor links.
 			'speaker_link' => wcorg_skip_feature( 'content_blocks' ) ? 'anchor' : 'permalink',
 			'session_link' => 'permalink',
 		),
@@ -183,7 +183,7 @@ function preprocess_schedule_attributes( $attr ) {
 		$attr['session_link'] = 'permalink';
 	}
 
-	// Sites without the `content_blocks` skip flag use blocks, so we need to prevent these from using `anchor`.
+	// See above re: `content_blocks`.
 	if ( ! wcorg_skip_feature( 'content_blocks' ) ) {
 		$attr['speaker_link'] = 'anchor' !== $attr['speaker_link'] ? $attr['speaker_link'] : 'permalink';
 		$attr['session_link'] = 'anchor' !== $attr['session_link'] ? $attr['session_link'] : 'permalink';
--- a/public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
@@ -164,9 +164,11 @@ function preprocess_schedule_attributes( $attr ) {
 		array(
 			'date'         => null,
 			'tracks'       => 'all',
-			'speaker_link' => 'anchor',    // anchor|wporg|permalink|none
-			'session_link' => 'permalink', // permalink|anchor|none
-		), $attr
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
```diff
@@ -164,9 +164,11 @@ function preprocess_schedule_attributes( $attr ) {
 		array(
 			'date'         => null,
 			'tracks'       => 'all',
-			'speaker_link' => 'anchor',    // anchor|wporg|permalink|none
-			'session_link' => 'permalink', // permalink|anchor|none
-		), $attr
+			// Default to anchor links only if we're not using blocks on this site.
```
iandunn: `SHOULD` - I don't think this comment needs to explain _what_ the statement is doing, because that's fairly easy to see. Instead, I think it should explain _why_ it's doing it.

Although, explaining what the `content_blocks` feature flag does would be good, since that's not obvious from the name.

public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
```diff
@@ -181,6 +183,12 @@ function preprocess_schedule_attributes( $attr ) {
 		$attr['session_link'] = 'permalink';
 	}
 
+	// Sites without the `content_blocks` skip flag use blocks, so we need to prevent these from using `anchor`.
```
iandunn: `SHOULD` - I think it'd be good to explain the reasoning here more. A new developer reading this -- or ourselves in 6 months -- probably wouldn't get enough information from the comment in order to understand the problem that's being solved.

It might be best to add all that explanation in the comment above, though, and make this comment just say something like "see above for details".

public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
```diff
@@ -181,6 +183,12 @@ function preprocess_schedule_attributes( $attr ) {
...
+	if ( ! wcorg_skip_feature( 'content_blocks' ) ) {
+		$attr['speaker_link'] = 'anchor' !== $attr['speaker_link'] ? $attr['speaker_link'] : 'permalink';
+		$attr['session_link'] = 'anchor' !== $attr['session_link'] ? $attr['session_link'] : 'permalink';
```
iandunn, line 189:
```suggestion
		$attr['session_link'] = 'anchor' === $attr['session_link'] ? 'permalink' : $attr['session_link'];
```

`MAYBE` - This feels like a more straight-forward way to write it to me; if there's a negation there then I have to do an extra mental cycle to reverse it before the statement clicks for me. 

Same for the one above it.

iandunn (approved):
Left a few minor comments, but not blockers. Overall it LGTM

public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
ryelle: I disagree- I think about it a different way: we want to fall through to the selected link type, unless it's anchor.

I've left this, but updated the two comments above.

iandunn: WFM :shipit:

PR #220, opened 2019-08-20
Schedule shortcode: Improve accessibility of favorite buttons

ryelle:
This starts to address [ticket 4247,](https://meta.trac.wordpress.org/ticket/4247) which lists a few accessibility issues with favoriting sessions. In particular, it makes the stars focusable by keyboards, and gives them some text for screen reader users. The stars are also slightly darker now, but still don't technically pass a "good" contrast ratio (IMO, 3.0+ would be best).

The favorite button is now moved _after_ the session information, so visual tab order is consistent and screen reader users hear the session title and speakers before being given the option to favorite it.

Lastly I want to address that this isn't an ideal solution– we're overloading this `a` with aria, rather than using a button. Basically it's for backwards compatibility with existing sites.

- Buttons are more likely to have aggressive styling from the theme, and it would take more work to reset this, and each theme does something slightly different
- The current CSS uses the elements in styling, so any organizer looking to override this probably copied the styles here, and if we swap it out for a button, current sites that have styled this will break

On Twenty Thirteen (Zero Hour Networking is selected, Opening Remarks is hovered):

On Twenty Sixteen (Zero Hour Networking is selected, Opening Remarks is hovered):

**To test**

1. View a schedule
2. Use a keyboard to tab through the sessions and try favoriting some
3. Try using a screen reader (Safari + [VoiceOver](https://webaim.org/articles/voiceover/) is a good test), navigate through the content, you should hear the favorite buttons, and it should tell you when one is selected/pressed.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -4,10 +4,10 @@
  * Plugin Description: Sessions, Speakers, Sponsors and much more.
  */
 
-require( 'inc/back-compat.php' );
-require_once( 'inc/favorite-schedule-shortcode.php' );
-require_once( 'inc/privacy.php' );
-require_once( 'inc/deprecated.php' );
+require 'inc/back-compat.php';
+require_once 'inc/favorite-schedule-shortcode.php';
+require_once 'inc/privacy.php';
+require_once 'inc/deprecated.php';
 
 class WordCamp_Post_Types_Plugin {
 	protected $wcpt_permalinks;
@@ -77,9 +77,9 @@ class WordCamp_Post_Types_Plugin {
 				'wcb_sponsor_level_order',
 				array(
 					'sanitize_callback' => array( $this, 'validate_sponsor_options' ),
-					'show_in_rest' => array(
+					'show_in_rest'      => array(
 						'schema' => array(
-							'type' => 'array',
+							'type'  => 'array',
 							'items' => array(
 								'type' => 'integer',
 							),
@@ -101,7 +101,7 @@ class WordCamp_Post_Types_Plugin {
 	 * Runs during init, because rest_api_init is too late.
 	 */
 	public function rest_init() {
-		require_once( 'inc/rest-api.php' );
+		require_once 'inc/rest-api.php';
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -4,10 +4,14 @@
  * Plugin Description: Sessions, Speakers, Sponsors and much more.
  */
 
-require( 'inc/back-compat.php' );
-require_once( 'inc/favorite-schedule-shortcode.php' );
-require_once( 'inc/privacy.php' );
-require_once( 'inc/deprecated.php' );
+// Disable phpcs rules for strict comparisons, since fixing these might break edge cases.
+// phpcs:disable WordPress.PHP.StrictComparisons.LooseComparison
+// phpcs:disable WordPress.PHP.StrictInArray.MissingTrueStrict
```
iandunn: `SHOULD` - I don't think we should ignore CS violations simply because they require work/testing to fix. I'd rather leave them there until we have time to properly fix them, or just fix them with minimal testing, and then come back and fix it if that introduces a bug that's discovered later on.

iandunn (approved):

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
ryelle: I disagree, I think leaving the failing tests in leads to everyone ignoring failing tests. We have PHPCS running on all PRs, and IMO we should not merge anything [with failing checks.](https://github.com/WordPress/wordcamp.org/pull/220/checks)

I'm happy to fix these now, in this PR, but I didn't want to get flagged for unrelated changes.

> I'd rather leave them there until we have time to properly fix them

That's essentially what this does– now we can come back, search for `phpcs:disable`, remove the disable rule and fix the issues.

iandunn: >  I think leaving the failing tests in leads to everyone ignoring failing tests

I agree with that.

> IMO we should not merge anything with failing checks.

If the failure is just because of CS stuff, merging it wouldn't bother me personally.

> I didn't want to get flagged for unrelated changes

I'm fine w/ CS fixes in a PR, as long as they're a separate commit.

> That's essentially what this does– now we can come back, search for phpcs:disable, remove the disable rule and fix the issues.

I don't think that _fixes_ the issue, it just _hides_ it. I think a better solution would be to [make the linters only run on changed lines](https://make.wordpress.org/core/handbook/best-practices/commit-messages/#before-a-commit), or fix the CS issues in a separate commit, or merge the PR despite the CS failures, or make CS failures a warning instead of a failure condition (if that's possible).

ryelle: I've removed the comments– these are just warnings, so we'll let it through.

PR #222, opened 2019-08-23
Jetpack Tweaks: Add supplemental list of safe CSS properties

coreymckrill:
There are many properties that have wide browser support but are not currently included in CSSTidy's validator. This adds a handful of these that are popular and supported so that WordCamp designers can use them while customizing the CSS of their sites.

**To test**

* Open `plugins/wordcamp-remote-css/app/synchronize-remote-css.php`.
* In the `synchronize_remote_css` method, comment out the first line that starts with `$unsafe_css`, and then reassign the `$unsafe_css` variable a static string of CSS that includes some of the newly safelisted properties.
* On your test site, go to the Remote CSS admin screen: `wp-admin/themes.php?page=remote-css`.
* Enter a fake URL that will still validate, `https://github.com/WordPress/style.css`, and click Update.
* In another tab, go to this URI on your test site: `wp-admin/admin-ajax.php?action=wordcamp_remote_css`.
* The safelisted properties should be included in the stylesheet.

### DIFF
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
@@ -1,13 +1,13 @@
 <?php
-
 namespace WordCamp\Jetpack_Tweaks;
+
+use Exception;
 use WordCamp\RemoteCSS;
 use WordCamp\Logger;
-use Exception;
 
-defined( 'WPINC' ) or die();
+defined( 'WPINC' ) || die();
 
-add_filter( 'update_custom_css_data',     __NAMESPACE__ . '\sanitize_custom_css', 15       );	// After Jetpack_Custom_CSS_Enhancements::update_custom_css_data()
+add_filter( 'update_custom_css_data',     __NAMESPACE__ . '\sanitize_custom_css', 15       ); // After Jetpack_Custom_CSS_Enhancements::update_custom_css_data().
 add_action( 'csstidy_optimize_postparse', __NAMESPACE__ . '\sanitize_csstidy_parsed_rules' );
 add_action( 'admin_notices',              __NAMESPACE__ . '\notify_import_rules_stripped'  );
 add_action( 'csstidy_optimize_subvalue',  __NAMESPACE__ . '\sanitize_csstidy_subvalues'    );
@@ -33,7 +33,7 @@ function sanitize_custom_css( $post ) {
 		 * There's no way to gracefully abort the process and show an error message, so just die.
 		 */
 		Logger\log( 'custom_css_sanitization_failed', compact( 'post', 'exception' ) );
-		wp_die( $exception->getMessage() );
+		wp_die( wp_kses_data( $exception->getMessage() ) );
 	}
 
 	return $post;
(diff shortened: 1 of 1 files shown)


public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
```diff
@@ -154,3 +174,46 @@ function sanitize_urls_in_css_properties( $url, $property ) {
...
+ * Add the custom safelisted properties to CSSTidy's config.
+ *
+ * Hooked to `safecss_parse_pre` to try and ensure that the `csstidy` global exists before trying to modify it.
+ *
+ * @return void
+ */
+function update_csstidy_safelist() {
+	$properties = get_custom_css_properties_safelist();
+
+	$properties_for_csstidy = array_fill_keys( $properties, 'CSS3.0' );
```
ryelle, line 218: I feel like this needs a comment explaining how/why CSSTidy uses this value, but it might not be explainable in a short comment.

ryelle (approved):
Looks good – tested with a few properties & tried data-uris with the new properties, all working as expected 👍

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
```diff
@@ -132,9 +133,22 @@ function sanitize_csstidy_subvalues( $safecss ) {
  * @return string
  */
 function sanitize_urls_in_css_properties( $url, $property ) {
-	$allowed_properties = array( 'background', 'background-image', 'border-image', 'border-image-source', 'content', 'cursor', 'list-style', 'list-style-image' );
+	$allowed_properties = array(
```
iandunn, line 142: `MAYBE` - 🤔 I wonder if we should have unit tests for this, and `get_custom_css_properties_safelist()`? 

https://github.com/WordPress/wordcamp.org/tree/a4d3dc043b34e4149f9367e26023619a93f68b42/public_html/wp-content/plugins/wordcamp-remote-css/tests

The existing tests show examples of trying to get JS to pass though. We could test a few of these new properties as well. 

It may not really be much different, though, since at this point it's just treating them as strings, before the browser has a chance to parse them. So, the results may not be any different for one property than for another. What do y'all think?

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
iandunn: What were your thoughts on the security aspects?

-----

_EDIT_: Er, I meant this to be a reply to:

> Looks good – tested with a few properties & tried data-uris with the new properties, all working as expected 👍

...not the part about the CSSTidy comment.

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
iandunn: ( this was initially a pending comment that i was in the middle of editing when I left [this other comment](https://github.com/WordPress/wordcamp.org/pull/222#discussion_r317706955), which for some reason published the pending review. The substance of this comment is different than when it was initially published, based on further reflection )

iandunn: This might be useful in seeing what payloads would look like, and how they should be sanitized:

https://github.com/cure53/DOMPurify/blob/master/test/fixtures/expect.js

DOMPurify is by far the most widely respected, well-tested, and properly-audited sanitizer out there. It's JS, so that it can use the browser itself to reliably parse payloads, which is impossible to do without an actual JS engine. Unfortunately, that means we can't use it server-side (without Node yada yada).

coreymckrill: > I wonder if we should have unit tests for this

This doesn't add any additional sanitizing functions, just new properties. I don't think testing more properties in the same ways that existing tests work would add any beneficial coverage.

iandunn (approved):
I can't think of anything left to try re:security, so this LGTM.

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
coreymckrill: Comment added in 51cee830

Issue #205, opened 2019-08-07
Offline Template: Test with all themes


ryelle:
Our supported themes are:

- [x] CampSite 2017
- [x] Twenty Thirteen
- [x] Twenty Sixteen
- [x] Twenty Seventeen
- [x] Twenty Nineteen
- [x] WordCamp Base

We should also test with custom CSS, and with remote CSS.

There's a note about testing with child themes, but WCs can't have child themes, so do we need to do this?

---

The rest we don't test with, but will fix issues on

- Twenty Ten
- Twenty Eleven
- Twenty Twelve
- Twenty Fourteen
- Twenty Fifteen
- WordCamp Base Redux

westonruter:
See also https://github.com/xwp/pwa-wp/issues/199 where the PWA plugin could bundle offline templates for the core themes.

ryelle:
We would still want to inject our own, so we can show the schedule and other info related to the WordCamp.

westonruter:
Good point. At least we could share the same base, or the PWA templates could trigger an action for the WordCamp sites to inject the content. Nevertheless, the content could also be injected via script.

westonruter:
The need to inject content into the offline template is also something that PWA plugin isn't properly facilitating. Right now there is a [`wp_service_worker_error_message_placeholder()`](https://github.com/xwp/pwa-wp/blob/092902c4a2c335053b076956493d987fdc5b5f1b/wp-includes/template.php#L163-L168) function that the `offline.php` template is expected to include. Perhaps this function should do an action as well? 

Perhaps `wp.hooks` could also be incorporated into the service worker to allow filtering the offline template response prior to sending to the browser, which would better enable dynamically-populating the template with content (e.g. list of pages available offline https://github.com/WordPress/wordcamp.org/issues/201#issuecomment-519274408): https://github.com/xwp/pwa-wp/issues/142 

See also https://github.com/xwp/pwa-wp/issues/211

ryelle:
I think the specifics would have to depend on how the the template itself works, we might still want to override the whole thing. For more "average user" cases, an action before & after `wp_service_worker_error_message_placeholder` sounds good (as does the wp.hooks suggestion, though I personally haven't used it as much).

PR #223, opened 2019-08-27
PWA: Update Offline template

### LINKED ISSUE #128: Finish offline template

ryelle:
An alternate approach to the offline template– This PR adds a new Offline page to the default site content, that the organizer can then update with relevant info. We can set up an outline of content (using headings, shortcode/block placeholders, etc) and let organizers decide what's relevant.

This also updates the offline template itself with theme classes & markup (mirroring CampSite, but works with other themes - post merge of this PR I'll check against the other themes listed in #205). 

Fixes #128 

The default page on a newly-created site:

offline-page

If there's no page, we show… basically nothing, since ideally everyone will have this page. The exact content comes directly from the PWA 

offline-no-page

**To test**

- Create a new site to populate this default content
- Or create a new page on an existing site with the stub content, and add `wc_page_offline` = `yes` as a custom field.
- View the offline template by going to `site.com/?wp_error_template=offline`
- Cache the page by viewing the frontend while online (I _think_ the above will do that)
- Turn off your internet (Chrome: in dev tools, the Application tab, Service Workers, you can click the "Offline" checkbox)
- View any page, you should get the offline page.

- Test also on a site with no Offline page set up.

❓  Some other thoughts…

- Assuming we activate the PWA plugin only on new sites, they should always have the new offline content, so we can leave the fallback content to be very simple, only used if an org trashes the post. In that case, I think the very simple content is fine - but open to counterpoints.
- What do we think about the stub content as it is? Should/can I pull content out of the WordCamp post type (date, location, etc) to pre-populate the location section, or does this not usually exist yet?
- Should there be more content in the offline stub page?

### DIFF
--- a/public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
+++ b/public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
@@ -201,7 +201,7 @@ function add_offline_template_cachebuster( $entry ) {
 	$page = get_offline_content();
 
 	if ( $entry && isset( $entry['revision'] ) ) {
-		$entry['revision'] .= ';' . md5( $page['content'] );
+		$entry['revision'] .= ';' . filemtime( __DIR__ . '/templates/offline.php' ) . md5( $page['content'] );
 	}
 
 	return $entry;
--- a/public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
+++ b/public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
@@ -147,7 +147,6 @@ function get_offline_page() {
 		'order'          => 'asc',
 		'meta_key'       => 'wc_page_offline',
 		'meta_value'     => 'yes',
-		'hierarchical'   => 0,
 	) );
 	if ( count( $found_pages ) ) {
 		return $found_pages[0];
@@ -170,7 +169,7 @@ function get_offline_content() {
 	$page = get_offline_page();
 	if ( $page ) {
 		return array(
-			'title'   => $page->post_title,
+			'title'   => apply_filters( 'the_title', $page->post_title, $page->ID ),
 			'content' => BlockUtilities\get_all_the_content( $page ),
 		);
 	}
--- a/public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
+++ b/public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
@@ -6,6 +6,7 @@
 namespace WordCamp\Theme_Templates;
 use WP_Post;
 use WP_Service_Worker_Scripts;
+use WordCamp\Blocks\Utilities as BlockUtilities;
 
 defined( 'WPINC' ) || die();
 
@@ -132,7 +133,60 @@ function inject_offline_template( $template_path ) {
 }
 
 /**
- * Add a cache-buster to the offline template's pre-cache entry.
+ * Get the offline page.
+ *
+ * This is a page created when the WordCamp site is created, so it should exist with this post meta.
+ *
+ * @return WP_Post|false
+ */
+function get_offline_page() {
+	$found_pages = get_posts( array(
+		'posts_per_page' => 1,
+		'post_type'      => 'page',
+		'orderby'        => 'date',
+		'order'          => 'asc',
+		'meta_key'       => 'wc_page_offline',
+		'meta_value'     => 'yes',
+		'hierarchical'   => 0,
+	) );
+	if ( count( $found_pages ) ) {
+		return $found_pages[0];
+	}
+
+	return false;
+}
+
+/**
+ * Get the offline page content, if the page exists. Otherwise, use simple defaults.
+ *
+ * @return array {
+ *     Content for the offline template.
+ *
+ *     @type string $title   Title of the page, or a default title.
+ *     @type string $content Content of the page, or the PWA error callback.
+ * }
+ */
+function get_offline_content() {
+	$page = get_offline_page();
+	if ( $page ) {
+		return array(
+			'title'   => $page->post_title,
+			'content' => BlockUtilities\get_all_the_content( $page ),
+		);
+	}
+
+	ob_start();
+	wp_service_worker_error_message_placeholder();
+	$offline_content = ob_get_clean();
+
+	return array(
+		'title'   => esc_html__( 'Offline', 'wordcamporg' ),
+		'content' => $offline_content,
+	);
+}
+
+/**
+ * Add the Offline page content as a revision parameter, to bust the cache when the page is changed.
  *
  * @param array|false $entry {
  *     Offline error precache entry.
@@ -144,8 +198,10 @@ function inject_offline_template( $template_path ) {
  * @return array|false
  */
 function add_offline_template_cachebuster( $entry ) {
+	$page = get_offline_content();
+
 	if ( $entry && isset( $entry['revision'] ) ) {
-		$entry['revision'] .= ';' . filemtime( __DIR__ . '/templates/offline.php' );
+		$entry['revision'] .= ';' . md5( $page['content'] );
 	}
 
 	return $entry;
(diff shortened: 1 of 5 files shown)


public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
```diff
@@ -144,8 +198,10 @@ function inject_offline_template( $template_path ) {
  * @return array|false
  */
 function add_offline_template_cachebuster( $entry ) {
+	$page = get_offline_content();
+
 	if ( $entry && isset( $entry['revision'] ) ) {
-		$entry['revision'] .= ';' . filemtime( __DIR__ . '/templates/offline.php' );
+		$entry['revision'] .= ';' . md5( $page['content'] );
```
westonruter: Perhaps include both?

```suggestion
		$entry['revision'] .= ';' . filemtime( __DIR__ . '/templates/offline.php' ) . md5( $page['content'] );
```

This would ensure the revision is bumped whenever the content changes _or_ the template is updated.

iandunn:
I like the idea of letting organizers customize it. I'm leery of letting them delete it entirely, though, or remove essential information from it.

What do you think about using a block template to ensure that the Schedule, Location, and Date info always remains, but can be rearranged? I don't see any use case for removing those. New blocks could be added of course.

> should/can I pull content out of the WordCamp post type (date, location, etc) 

Definitely. That'll require a small change in `get_required_fields()` to make sure they always exist when the site is created, but that's easy.

ryelle:
> What do you think about using a block template to ensure that the Schedule, Location, and Date info always remains, but can be rearranged? I don't see any use case for removing those. New blocks could be added of course.

Unfortunately, block templates don't work like that– you can lock a template down entirely so blocks can't be added, removed, or moved; or lock insertion, so blocks can't be added or removed, but can be rearranged. There's no way to lock just removal.

I think it's okay to leave this control in the hands of the organizers - it's not like they can't re-enter the content, or revert to an earlier revision. And maybe the format we've chosen doesn't work for the style of every camp, so locking the template down would just get frustrating.

ryelle:
> That'll require a small change in get_required_fields() to make sure they always exist when the site is created, but that's easy.

I'm going to add that in another PR, since that starts touching other areas– so this is ready for a proper review now.

public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
```diff
@@ -132,7 +133,60 @@ function inject_offline_template( $template_path ) {
...
+ */
+function get_offline_page() {
+	$found_pages = get_posts( array(
+		'posts_per_page' => 1,
+		'post_type'      => 'page',
+		'orderby'        => 'date',
+		'order'          => 'asc',
+		'meta_key'       => 'wc_page_offline',
+		'meta_value'     => 'yes',
+		'hierarchical'   => 0,
```
iandunn: I don't see that in `WP_Query::parse_query()`, is that an undocumented param?

public_html/wp-content/plugins/wcpt/stubs/page/offline.php
```diff
@@ -0,0 +1,23 @@
...
+<!-- wp:heading -->
+<h2><?php esc_html_e( 'Location', 'wordcamporg' ); ?></h2>
+<!-- /wp:heading -->
+
+<!-- wp:heading -->
+<h2><?php esc_html_e( 'Schedule', 'wordcamporg' ); ?></h2>
+<!-- /wp:heading -->
+
+<!-- wp:shortcode -->
+[schedule date="YYYY-MM-DD" tracks="example-track,another-example-track,yet-another-example-track"]
```
iandunn:
```suggestion
[schedule]
```

`MAYBE` - Since the organizers aren't likely to revisit this page when they add/remove/edit tracks, it might be safest to leave off all the options and just display everything by default.

public_html/wp-content/plugins/wcpt/stubs/page/offline.php
```diff
@@ -0,0 +1,23 @@
...
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p><?php esc_html_e( "This page couldn't be loaded because you appear to be offline. Please try again once you have a network connection.", 'wordcamporg' ); ?></p>
+<!-- /wp:paragraph -->
+
+<!-- wp:paragraph -->
+<p><?php esc_html_e( 'In the mean time, hopefully this information is useful:', 'wordcamporg' ); ?></p>
+<!-- /wp:paragraph -->
+
```
iandunn: `MAYBE` - I think it'd be helpful to include the dates as well. We can probably pre-populate them from the `wordcamp` post type.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
```diff
@@ -1,11 +1,14 @@
 <?php
 
+// Nonces are verified in Event_Admin::metabox_save (if applicable), we don't need to re-verify.
+// phpcs:disable WordPress.Security.NonceVerification.Missing
+
```
iandunn, line 4: `SHOULD` - Disabling this for the entire file makes me nervous, because we could add a new function in the future which should have a specific check, but we wouldn't be warned about it. What do you think about disabling it for specific functions instead?

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
```diff
@@ -176,14 +187,19 @@ public function maybe_create_new_site( $wordcamp_id ) {
 		) );
 
 		if ( is_int( $this->new_site_id ) ) {
-			update_post_meta( $wordcamp_id, '_site_id', $this->new_site_id );    // this is used in other plugins to map the `wordcamp` post to it's corresponding site
+			// This is used in other plugins to map the `wordcamp` post to it's corresponding site.
```
iandunn:
```suggestion
			// `_site_id` is used in other plugins to map the `wordcamp` post to it's corresponding site.
```

`SHOULD` - Since the comment is no longer tied to the line specifically, it seems helpful to make "this" explicit.

public_html/wp-content/mu-plugins/theme-templates/templates/offline.php
```diff
@@ -16,31 +16,22 @@
...
-		</h3>
+$offline_page = get_offline_content();
+?>
 
-		<?php echo do_shortcode( '[schedule]' ); ?>
+	<main id="main" class="site-main">
+		<section class="error-offline">
+			<header class="page-header">
+				<h1 class="page-title">
+					<?php echo esc_html( $offline_page['title'] ); ?>
```
iandunn:
```suggestion
					<?php echo wp_kses_post( $offline_page['title'] ); ?>
```

`SHOULD` - Sometimes people use HTML in post titles.

iandunn (approved):
I still think it'd be better ensure that essential info is always included, but can disagree and commit.

Left some suggestions, but no blockers.

public_html/wp-content/mu-plugins/theme-templates/bootstrap.php
ryelle: No, not sure where that came from… removed.

public_html/wp-content/plugins/wcpt/stubs/page/offline.php
ryelle: I've added `get_wordcamp_location` & `get_wordcamp_date_range`, and am using them here now.

PR #224, opened 2019-08-28
Use CLDR for country code and name data

coreymckrill:
This adds the [WP CLDR plugin](https://wordpress.org/plugins/wp-cldr/) to the WordCamp codebase for use in places where we need a list of country codes and/or names, replacing the hard-coded ISO-3166 list we had previously.

See https://meta.trac.wordpress.org/ticket/4638 for details and reasoning behind this change.

Note that we won't actually activate the WP CLDR plugin network-wide or on individual sites, because we don't need or want the [WP Admin UI](https://ps.w.org/wp-cldr/assets/screenshot-1.png?rev=1375816) that ships with the plugin. Instead, the `wcorg_get_countries` function loads the necessary class file directly.

**To test**

* This PR doesn't actually add the wp-cldr plugin directly. You can either run `composer update` to install it, or just do it manually from your testing environment of choice. Don't activate the plugin after installing it, though.
* Visit a page that outputs a `<select>` element containing all the countries. `https://central.wordcamp.org/wordcamp-organizer-application/` is a good choice if you're using your sandbox.
* The `<select>` should contain all the countries, with some naming differences compared to what is currently on Production (see the Meta ticket for specific examples).

**Pre-deploy tasks**

* [x] Add the wp-cldr plugin to our list of plugin externals
* [x] Add the `cldr-countries` feature flag
* [x] Make sure WC Central doesn't haven't the flag

**Post-deploy tasks**

* [x] Update the country field for all Multi Event Sponsors' contact info


coreymckrill:
@iandunn @ryelle I need help thinking through these changes, especially b2399db0a1974fefe12b3bc1c890b602c542ee60. It seems like the skip feature flag will  sufficiently handle the meta value change from country name to country alpha2 code, but there could be some implications I'm not thinking of.

composer.json
```diff
@@ -121,6 +121,7 @@
 		"wpackagist-plugin/supportflow": "dev-trunk",
 		"wpackagist-plugin/tagregator": "dev-trunk",
 		"wpackagist-plugin/wordpress-importer": "0.6.4",
+		"wpackagist-plugin/wp-cldr": "1.0",
```
iandunn: I thought that, post-#225, these would go in `wp-content/composer.json` and use `*` as the version, but I'm probably missing something?

public_html/wp-content/mu-plugins/3-helpers-misc.php
```diff
@@ -55,17 +55,28 @@ function wcorg_get_user_by_canonical_names( $name ) {
 }
 
 /**
- * Get ISO-3166 country names and codes
+ * Get CLDR country names and codes
  *
  * @todo move the real functionality from get_valid_countries_iso3166() to here, then have the Budgets plugin,
  * QBO, etc call this.
```
iandunn: It looks like this comment can be removed, since it was done in 6e725da83b84c83eb8adeb35d1b9034c8f6b0d44.

public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
```diff
@@ -2306,7 +2307,7 @@ public static function validate_save_payment_method_fields( $post_id, $meta_key_
 				case 'interm_bank_country_iso3166':
 				case 'beneficiary_country_iso3166':
 				case 'check_country':
-					if ( array_key_exists( $unsafe_value, self::get_valid_countries_iso3166() ) ) {
```
iandunn, line 2309: At a quick glance it doesn't look like there are any calls left to `get_valid_countries_iso3166()`; can it be removed?

public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
```diff
@@ -2199,19 +2199,20 @@ public static function get_valid_countries_iso3166() {
...
-	 * Get a country name from an alpha-2 or alpha-3 code
+	 * Get a country name from the alpha2 or alpha3 code.
 	 *
 	 * @param string $country_code
 	 *
 	 * @return string
 	 */
 	public static function get_country_name( $country_code ) {
-		$countries = self::get_valid_countries_iso3166();
+		$countries = wcorg_get_countries();
```
iandunn: I think this needs `include_alpha3 => true`, right?

public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
```diff
@@ -2199,19 +2199,20 @@ public static function get_valid_countries_iso3166() {
...
 		$name      = '';
 
 		foreach ( $countries as $country ) {
 			if ( $country_code === $country['alpha2'] || $country_code === $country['alpha3'] ) {
 				$name = $country['name'];
+				break;
```
iandunn: Nice catch!

public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
```diff
@@ -1130,7 +1130,7 @@ public static function _generate_payment_report_jpm_checks( $args ) {
 
 			$payable_to = WCP_Encryption::maybe_decrypt( get_post_meta( $post->ID, '_camppayments_payable_to', true ) );
 			$payable_to = html_entity_decode( $payable_to ); // J&amp;J to J&J
-			$countries = WordCamp_Budgets::get_valid_countries_iso3166();
+			$countries = wcorg_get_countries( array( 'include_alpha3' => true ) );
 			$vendor_country_code = get_post_meta( $post->ID, '_camppayments_vendor_country_iso3166', true );
```
iandunn, line 1134: There are a bunch of these meta field names that reference 3166. I guess they're _technically_ still accurate, since CLDR uses 3166 for the alpha-2/3 code; is that right?

It might still be good to have a comment somewhere explaining that they're just a legacy name, though, and that CLDR should be considered authoritative throughout the codebase now?

iandunn (approved):
Left a few minor comments, but overall LGTM

It'd be good to test a QBO sponsor invoice if you haven't already, or at least glance to see if it shouldn't be affected.

The only other thing that might be worth looking into is stats, in case any of our reports would be thrown off by the differences in sites with `cldr-countries` applied vs not.

composer.json
ryelle: It should, looks like this PR was branched before #225 merged. After this is rebased that will need to be updated, and the composer.lock file should also get added.

ryelle (approved):
> It seems like the skip feature flag will sufficiently handle the meta value change from country name to country alpha2 code, but there could be some implications I'm not thinking of.

I think this makes sense since flags are a per-site setting, you shouldn't have a case where a site uses both formats… central would be the only exception, but it looks like only the Multi-Event Sponsors use a separate country field and you've already addressed that in the post-deploy.

This needs a rebase, but I tested it with & without the feature flags and it's working as expected 👍

ryelle:
I forgot to ask - I noticed this doesn't change the camptix field, which is what the original issue is asking about (since that's user-facing), will there be a follow-up PR that swaps out the CampTix list too?

public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
coreymckrill: Yep, now fixed in b5854aa5

public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
coreymckrill: Good idea. Addressed in d933d594

public_html/wp-content/mu-plugins/3-helpers-misc.php
coreymckrill: Addressed in d933d594

composer.json
coreymckrill: Now rebased. This is fixed.

coreymckrill:
Changing the CampTix field requires first incorporating that plugin into our repo, which doesn't yet have a PR. I'm on the fence about delaying this PR until that other change is complete, or just shipping this as-is and opening another PR specifically for the country field in CampTix.

ryelle:
> I'm on the fence about delaying this PR until that other change is complete, or just shipping this as-is and opening another PR specifically for the country field in CampTix.

I would ship this and open another PR once CampTix is here, since this part doesn't rely on CampTix. I just wanted to make sure that fix was still in the works.

coreymckrill:
> It'd be good to test a QBO sponsor invoice if you haven't already, or at least glance to see if it shouldn't be affected.

Good idea. Made a change in 121908f7 to accommodate this.

PR #225, opened 2019-08-29
Tooling: Changes to composer and docker

### LINKED ISSUE #159: Versioned Composer dependencies add unnecessary maintenance burden
Continuing discussion from https://github.com/WordPress/wordcamp.org/pull/152#discussion_r303564592 so that it doesn't get lost in a closed PR.

We need to version things that run on production, like `league/oauth2-client`, but the plugins/themes are just dev dependencies, and production uses `svn:externals`.

If we version the dev dependencies, then we have to update the version number in 4 places instead of just 1. Since non-committers don't need it to match exactly, and it's running the latest 99% of the time, can we just setup Composer dev dependencies to pull `latest.zip` or `trunk` instead of a specific SVN tag?

coreymckrill:
I was frustrated by having the 3rd-party plugins and themes included in our composer.json file, even though they're only necessary for the Docker environment. It felt cumbersome to work around them, since as a committer, I have the svn repo, and don't need them to be installed by composer anyway. So I set out to split the plugins and themes into a separate composer.json file that only runs within the Docker container, as discussed in #159. Along the way, I had to significantly refactor the provisioning scripts used by Docker. I also discovered some unrelated problems with some of our apt-get packages in the php-fpm container, so I had to fix those to get the environment to function. So here's a list of all the changes:

* Remove 3rd-party plugins and themes from main composer.json file.
* Add a new composer.json file with only plugins and themes, that is only used during Docker container provisioning.
* Update docker configs to install from the new composer.json file during `docker-compose up`, but only if wp-content does not contain an .svn directory.
* Replace apt-get packages installed during docker provisioning that are no longer available (mysql-client and ssmtp)
* Add setup and config for the alternate smtp package (msmtp).
* Update the readme for Docker.

Fixes #159 

**To test with the svn repo**

* Go through the setup steps in .docker/readme.md. You might already have it all set up, but you'll probably need to re-run `composer update` from the project root.
* Run `docker-compose up --build` (without the `-d`). This will show you the output coming from within the containers after they have finished building. You should see a line that says "Starting up..." followed by one that says "The wp-content directory appears to be a Subversion repository. Skipping additional setup...". This means it didn't try to install the 3rd-party plugins and themes via composer.

**To test without the svn repo**

* Clone a fresh copy of this repo somewhere.
* Follow the instructions in .docker/readme.md.
* Run `docker-compose up --build` (without the `-d`). This will show you the output coming from within the containers after they have finished building. You should see a line that says "Starting up..." followed by several lines from composer as it installs 3rd-party plugins and themes.

**To test MailCatcher with the new msmtp package**

* Using either setup from above, with the containers up and running, open the MailCatcher dashboard in a new tab: `http://localhost:1080/`
* Get WP to send an email. One way to do this is via `https://central.wordcamp.test/wp-admin/post.php?post=2935&action=edit&classic-editor`. Check the box to "Manually send this email" and pick one of the test WordCamps from the dropdown. Then click Update.
* Back in the MailCatcher dashboard, the email should appear.


.docker/readme.md
```diff
@@ -8,8 +8,6 @@ Follow these steps to setup a local WordCamp.org environment using [Docker](http
     ```bash
     git clone git@github.com:WordPress/wordcamp.org.git wordcamp.test
     cd wordcamp.test
-	rm -rf .git/hooks
-	ln -s .githooks .git/hooks
     ```
```
iandunn, line 11: Why is this part removed?

composer.json
```diff
@@ -12,127 +12,12 @@
...
-		"wpackagist-theme/twentyten": "2.9",
-		"wpackagist-theme/twentyeleven": "3.3",
-		"wpackagist-theme/twentytwelve": "3.0",
-		"wpackagist-theme/twentythirteen": "2.9",
-		"wpackagist-theme/twentyfourteen": "2.7",
-		"wpackagist-theme/twentyfifteen": "2.5",
-		"wpackagist-theme/twentysixteen": "2.0",
-		"wpackagist-theme/twentyseventeen": "2.2",
-		"wpackagist-theme/twentynineteen": "1.4"
+		"wp-coding-standards/wpcs": "2.*"
```
iandunn, line 20: Good idea!

iandunn (approved):

.docker/readme.md
coreymckrill: Sorry, I should have made this change as a separate commit and explained it.

While I think it's a good idea to use hooks to ensure better code quality in PRs we get from contributors, the hooks aren't functional out-of-the-box. The scripts make assumptions that tools like phpcs are already installed, and about where they are installed, which may not work for everyone (they didn't work for me). So I think the scripts need some updates (using `$(which phpcs)` for example), and we should probably add the installation and configuration of the necessary tool dependencies as a separate step in the Initial Setup process.

In the mean time, I removed them because in the process of doing this refactor, I was setting up the Docker environment from scratch and going through these steps, and i realized that those lines might be a stumbling point for other folks who are trying to get the environment up and running.

coreymckrill: > we should probably add the installation and configuration of the necessary tool dependencies as a separate step in the Initial Setup process.

Or, this could potentially be automated in composer and/or the Docker script.

iandunn: 🤔 those are good considerations. I'm a bit leery of removing them entirely, though. 

I wonder if there's a middle ground, like marking them as optional steps, and mentioning that  the documentation is still a work in progress, so that people feel free to skip them, or give up if they can't get work them working quickly?

.docker/readme.md
```diff
@@ -33,21 +31,28 @@ Follow these steps to setup a local WordCamp.org environment using [Docker](http
     git checkout 5.2
     ```
 
-1. Install the 3rd-party plugins, themes, and additional packages used on WordCamp.org. For this, you must have [Composer](https://getcomposer.org/doc/00-intro.md) installed. Once it is, change to the root directory of the project where the **composer.json** file is located.
+1. Install 3rd-party PHP packages used on WordCamp.org, including for development and testing. For this, you must have [Composer](https://getcomposer.org/doc/00-intro.md) installed. Once it is, change to the root directory of the project where the **composer.json** file is located.
```
ryelle: This threw me off, I wasn't sure if it meant the wp-content composer or root composer – we can assume the person going through this is in `mu` still, so why not say `cd ../..` in the bash instructions?

Also, the "including for development and testing" seems unnecessary, I think it could just be removed.

ryelle (approved):
Looks good & the updated documentation is great. Tested with both updating my local install and creating a new, non-SVN'd install; the SVN install correctly did not run composer over my checked out plugins, and the brand new install correctly _did_ 🙂

.docker/readme.md
ryelle: Could we use something like [husky](https://www.npmjs.com/package/husky) for this? I know we don't currently use `npm` in the root of the project, but this might be a good reason?

Also, since phpcs has been in composer since #185, we can say `composer run phpcs …` to run the vendor script, instead of relying on a global install.

iandunn: > husky

I looked at the docs, but they don't really explain what it does. What about it do you find useful?

> composer run phpcs

Great point :)

ryelle: > I looked at the docs, but they don't really explain what it does.

It just handles setting up git hooks automatically so we don't need to tell people to sync githooks like this. We can then use custom scripts or npm commands, the hooks work just like the `scripts` part of package.json.

.docker/readme.md
coreymckrill: Updated in e516dba7

.docker/readme.md
coreymckrill: Added the git hook commands back in as an optional step in f343eb1e

I think it's worth looking into automating the git hook stuff, either with husky or something else, but I'd rather open an issue to discuss it along with other improvements to the precommit hook scripts, and not deal with it in this PR.

coreymckrill: Issue: https://github.com/WordPress/wordcamp.org/issues/228

PR #229, opened 2019-09-04
Fonts: Support Font Awesome kits

ryelle:
As of Font Awesome 5, the CDN uses a kit system like typekit - a javascript file that in turn loads the necessary CSS. This adds a new field to the fonts section for this kit ID.

Screen Shot 2019-09-03 at 7 42 28 PM

**To Test**

- Sign up for a kit at [fontawesome.com](https://fontawesome.com/)
- You'll get a kit script like `<script src="https://kit.fontawesome.com/[kit id].js"></script>`
- Use that ID for the input,
- Save, it should stay in the input field
- View the frontend of your site, you should see the script tags in the source, and see the JS and CSS files loading.

### DIFF
--- a/public_html/wp-content/plugins/wc-fonts/wc-fonts.php
+++ b/public_html/wp-content/plugins/wc-fonts/wc-fonts.php
@@ -57,11 +57,14 @@ class WordCamp_Fonts_Plugin {
 	 * Provides the <head> output for Font Awesome
 	 */
 	public function wp_head_font_awesome() {
-		if ( empty( $this->options['font-awesome-url'] ) ) {
-			return;
+		if ( $this->options['font-awesome-url'] ) {
+			printf( "<style>@import url( '%s' );</style>", esc_url( $this->options['font-awesome-url'] ) );
 		}
 
-		printf( "<style>@import url( '%s' );</style>", esc_url( $this->options['font-awesome-url'] ) );
+		if ( $this->options['font-awesome-kit'] ) {
+			// phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+			printf( '<script src="https://kit.fontawesome.com/%s.js"></script>', esc_attr( $this->options['font-awesome-kit'] ) );
+		}
 	}
 
 	/**
@@ -100,6 +103,14 @@ class WordCamp_Fonts_Plugin {
 			'font-awesome-url',
 			__( 'Font Awesome', 'wordcamporg' ),
 			array( $this, 'field_font_awesome_url' ),
+			'wc-fonts-options',
+			'general'
+		);
+
+		add_settings_field(
+			'font-awesome-kit',
+			__( 'Font Awesome Kit', 'wordcamporg' ),
+			array( $this, 'field_font_awesome_kit' ),
 			'wc-fonts-options',
 			'general'
 		);
@@ -200,6 +211,21 @@ class WordCamp_Fonts_Plugin {
 	}
 
 	/**
+	 * Settings API field for the Font Awesome kit ID
+	 */
+	public function field_font_awesome_kit() {
+		$value = isset( $this->options['font-awesome-kit'] ) ? $this->options['font-awesome-kit'] : '';
+		?>
+
+		<input type="text" name="wc-fonts-options[font-awesome-kit]" value="<?php echo esc_attr( $value ); ?>" class="regular-text code" />
+		<p class="description">
+			<?php esc_html_e( 'Enter your Font Awesome Kit ID only. Do not add any URLs or JavaScript.', 'wordcamporg' ); ?>
+		</p>
+
+		<?php
+	}
+
+	/**
 	 * Settings API field for the Dashicons checkbox
 	 */
 	public function field_dashicons() {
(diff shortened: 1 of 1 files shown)


ryelle:
I also thought about allowing for unpkg links, since you can get to fontawesome's CSS via the npm package, but then we'd want to safelist just this package, otherwise folks can link to _any_ css in npm, and that would require more than just checking the host. Seems like the right way to use Font Awesome is by using the kit script anyway.

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
```diff
@@ -243,7 +291,7 @@ public function validate_options( $input ) {
 			$url = wp_parse_url( $input['font-awesome-url'] );
 
 			if ( isset( $url['host'] ) && isset( $url['path'] ) ) {
-				$valid_hostname  = 'maxcdn.bootstrapcdn.com' === $url['host'];
+				$valid_hostname  = in_array( $url['host'], [ 'maxcdn.bootstrapcdn.com', 'stackpath.bootstrapcdn.com' ] );
```
ryelle, line 294: Looks like Bootstrap CDN uses `stackpath.bootstrapcdn.com` by default now, though maxcdn will still work. If someone wants to use 4.x of Font Awesome, they'll likely get the stackpath link which would be stripped on save currently.

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
```diff
@@ -48,18 +49,22 @@ public function wp_head_google_web_fonts() {
...
 		printf( '<style>%s</style>', $this->options['google-web-fonts'] );
 	}
 
 	/**
 	 * Provides the <head> output for Font Awesome
 	 */
 	public function wp_head_font_awesome() {
-		if ( empty( $this->options['font-awesome-url'] ) ) {
-			return;
+		if ( $this->options['font-awesome-url'] ) {
```
coreymckrill: I'm getting a `Notice: Undefined index: font-awesome-url` (and same for `font-awesome-kit`). This should probably either be wrapped in a `! empty()` call, or prefill all the correct keys up in the `init()` method.

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
```diff
@@ -48,18 +49,22 @@ public function wp_head_google_web_fonts() {
...
+			printf( "<style>@import url( '%s' );</style>", esc_url( $this->options['font-awesome-url'] ) );
 		}
 
-		printf( "<style>@import url( '%s' );</style>", esc_url( $this->options['font-awesome-url'] ) );
+		if ( $this->options['font-awesome-kit'] ) {
```
coreymckrill: Same here as on line 60.

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
```diff
@@ -177,6 +210,21 @@ public function field_font_awesome_url() {
...
+	/**
+	 * Settings API field for the Font Awesome kit ID
+	 */
+	public function field_font_awesome_kit() {
+		$value = isset( $this->options['font-awesome-kit'] ) ? $this->options['font-awesome-kit'] : '';
+		?>
+
+		<input type="text" name="wc-fonts-options[font-awesome-kit]" value="<?php echo esc_attr( $value ); ?>" class="regular-text code" />
+		<p class="description">
+			<?php esc_html_e( 'Enter your Font Awesome Kit ID only. Do not add any URLs or JavaScript.', 'wordcamporg' ); ?>
```
coreymckrill: Might be worth adding a line here, or on the other Font Awesome field (or both) explaining that you only need to use one field or the other.

coreymckrill: And maybe that the kit is only relevant for v5 and above.

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
```diff
@@ -36,7 +36,8 @@ public function wp_head_typekit() {
 			return;
 		}
 
-		printf( '<script type="text/javascript" src="https://use.typekit.com/%s.js"></script>' . "\n", $this->options['typekit-id'] );
+		// phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+		printf( '<script type="text/javascript" src="https://use.typekit.com/%s.js"></script>' . "\n", esc_attr( $this->options['typekit-id'] ) );
```
coreymckrill: `sanitize_key` might be a better function here than `esc_attr`, although I don't know if phpcs would recognize it as an escaping function. Same for the Font Awesome kit output on line 66.

coreymckrill (changes requested):
One thing about a PHP notice, everything else is just minor suggestions.

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
ryelle: Yep, I forgot `empty` has that protection, thanks for the catch

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
ryelle: Technically you _can_ use both, v5 has some kind of version conflict magic ✨  (i mean, you still shouldn't.)

You're just suggesting adding some text, not doing any detection on whether one field is filled out or not, right?

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
ryelle: ah, yeah - I need a cheatsheet of all the sanitization functions (oh hey, [there is one](https://developer.wordpress.org/themes/theme-security/data-sanitization-escaping/))

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
coreymckrill: Yeah, just some text

coreymckrill (approved):

PR #230, opened 2019-09-04
Payments: Convert amount into zero-decimal format for Stripe

ryelle:
Stripe expects payment amounts in the smallest amount in a given currency – cents in USD, yen in JPY, etc. This function formats a given amount into the "zero decimal" amount. So, 10.00 USD would become 1000, while 10 JPY would stay 10. Using this function for Sponsor Payments fixes issue with JPY payments being inflated by 100.

**To test**

- Go to https://central.wordcamp.test/sponsorship-payment/
- Try submitting a payment for an invoice using Japanese Yen
- Click through to the stripe "Make a Payment" button, check that the correct value is listed in the modal.
- Try another payment for an invoice using USD
- Again, the value in the stripe modal should be correct.

### DIFF
--- a/public_html/wp-content/mu-plugins/utilities/class-stripe-client.php
+++ b/public_html/wp-content/mu-plugins/utilities/class-stripe-client.php
@@ -155,4 +155,56 @@ class Stripe_Client {
 
 		return $match;
 	}
+
+	/**
+	 * Convert an amount in the currency's base unit to its equivalent fractional unit.
+	 *
+	 * Stripe wants amounts in the fractional unit (e.g., pennies), not the base unit (e.g., dollars).  The data
+	 * here comes from https://stripe.com/docs/currencies.
+	 *
+	 * Note: This uses different data than `get_fractional_unit_multiplier` above, these have different data
+	 * sources. These should be reconciled in the future.
+	 *
+	 * @param string $order_currency
+	 * @param int    $base_unit_amount
+	 *
+	 * @return int
+	 * @throws Exception
+	 */
+	public static function get_fractional_unit_amount( $order_currency, $base_unit_amount ) {
+		$fractional_amount = null;
+
+		$currency_multipliers = array(
+			// Zero-decimal currencies.
+			1   => array(
+				'BIF', 'CLP', 'DJF', 'GNF', 'JPY', 'KMF', 'KRW', 'MGA', 'PYG', 'RWF', 'UGX', 'VND', 'VUV', 'XAF',
+				'XOF', 'XPF',
+			),
+			100 => array(
+				'AED', 'AFN', 'ALL', 'AMD', 'ANG', 'AOA', 'ARS', 'AUD', 'AWG', 'AZN', 'BAM', 'BBD', 'BDT', 'BGN',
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-payment-stripe.php
```diff
@@ -199,15 +199,23 @@ function _handle_post_data( &$data ) {
...
-				'description'  => $description,
-				'currency'     => $currency,
-				'amount'       => $amount,
+				'payment_type'   => $payment_type,
+				'wordcamp_id'    => $wordcamp_id,
+				'invoice_id'     => $invoice_id,
+				'description'    => $description,
+				'currency'       => $currency,
+				'amount'         => $amount,
+				'decimal_amount' => $decimal_amount,
```
coreymckrill, line 218: Why add the converted amount in a separate field instead of the normal `amount` field? Even if the Stripe API handles this correctly, I'm a bit leery of including both the converted and unconverted amounts in the payload.

public_html/wp-content/mu-plugins/utilities/class-stripe-client.php
```diff
@@ -155,4 +155,56 @@ public static function get_fractional_unit_multiplier( $order_currency ) {
...
 		return $match;
 	}
+
+	/**
+	 * Convert an amount in the currency's base unit to its equivalent fractional unit.
+	 *
+	 * Stripe wants amounts in the fractional unit (e.g., pennies), not the base unit (e.g., dollars).  The data
+	 * here comes from https://stripe.com/docs/currencies.
+	 *
+	 * Note: This uses different data than `get_fractional_unit_multiplier` above, these have different data
```
coreymckrill: For consistency with our other code comments that expect a followup, I'd change `Note` to `@todo`

ryelle: Feel free to use the suggestion functionality for that, then I can just commit it.

```suggestion
	 * @todo This uses different data than `get_fractional_unit_multiplier` above, these have different data
```

public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-payment-stripe.php
ryelle: This isn't the payload sent to Stripe, this is just the data sent to the template – `amount` is still used in the display of the table:

Screen Shot 2019-09-04 at 4 15 15 PM

coreymckrill: Oh, I see. Makes sense, carry on.

coreymckrill (approved):

shusei-t:
I think there is still some error.
I tried paying but I noticed things shown below:
- tried to pay using the invoice number sent by the WordCamp Tokyo team
- the currency and amount shown on Stripe is correct
- but the amount claimed was x100
- got our company credit card stopped
- the day we tried was September 16

Could you guys check please?

ryelle:
Thanks for letting us know @shusei-t, and I'm sorry you got overcharged. There was still a bug, but it's fixed now (in 8fd9e801cef2baf66367621fecd0afafe250f705). I've tested it with both USD and JPY, going through the whole payment process, and it will charge the correct amount now.

PR #236, opened 2019-09-15
Blocks: Allow `noscript` tags in block output

ryelle:
When processing each image, Jetpack adds a `noscript` wrapper around the original image tag, so that users without JS will still get the original image - in addition to injecting the lazy-loaded image. When we run `wp_kses_post`, this strips out that `noscript`, so both the lazy image and original image are both included. This change permits noscript, so the original image is only shown is JS is disabled.

This is a draft PR: We use `wp_kses_post` all over the place, so the noscript is being stripped over and over again 🙃 

To test

- Turn on lazy images: Jetpack > Settings > Performance > "Enable Lazy Loading for images"
- View a block with avatars, ex the Speakers block
- There should be no duplicate avatars
- View the Sponsors block
- There should be no duplicate sponsor logos
- View a block with images in the item content
- There should be no duplicate images

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/blocks.php
+++ b/public_html/wp-content/mu-plugins/blocks/blocks.php
@@ -100,25 +100,3 @@ function register_assets() {
 }
 
 add_action( 'init', __NAMESPACE__ . '\register_assets', 9 );
-
-/**
- * Update kses filter.
- *
- * Allow `noscript`: this is used by Jetpack's lazy-loading before we output the content, and by default is
- * stripped by the `wp_kses_post` function, causing duplicate images.
- *
- * @param array $tags
- * @return array
- */
-function allow_noscript_blocks( $tags, $context ) {
-	global $post;
-
-	// Only allow noscript through if we're showing a post with blocks.
-	if ( 'post' === $context && isset( $post, $post->post_content ) && has_blocks( $post->post_content ) ) {
-		$tags['noscript'] = array();
-	}
-
-	return $tags;
-}
-
-add_action( 'wp_kses_allowed_html', __NAMESPACE__ . '\allow_noscript_blocks', 10, 2 );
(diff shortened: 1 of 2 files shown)


coreymckrill:
🤔 Instead of updating all the places that we use `wp_kses_post`, could we just add `noscript` to the allowed tags via the `wp_kses_allowed_html` filter?

ryelle:
@coreymckrill Good point– I've updated the PR to do that, and it should only safelist `noscript` when we're working in posts with blocks, but more 👀 s would be good to check that this doesn't open up any kind of security issues.

coreymckrill (approved):
I have only found [one security issue](https://securethoughts.com/hacking-for-xss-inside-noscript-html-tags/) related to the `<noscript>` tag, and it wouldn't be relevant to us, because it requires also being able to use the `<script>` tag, which we are not adding an exception for.

* [ ] `MAYBE` We could make the `has_blocks` part more explicit and specify only our own blocks, instead of any block. This would reduce the number of cases where the `<noscript>` tag would be allowed, but I'm not sure that would add any practical benefit.
* [ ] `MAYBE` It might make sense for this code to live in **mu-plugins/jetpack-tweaks**, since it's directly related to Jetpack. But I can see the case for keeping it in **blocks** as well. I don't have a strong preference.

ryelle:
> We could make the has_blocks part more explicit and specify only our own blocks, instead of any block. 

I think that would overcomplicate this, since `has_blocks` doesn't accept a block type parameter, we'd need to manually check it.

> It might make sense for this code to live in mu-plugins/jetpack-tweaks, since it's directly related to Jetpack

I like that, I'll move it over tomorrow– will you want to re-review, or can I consider it "pre-approved" to merge once I move it?

coreymckrill:
Pre-approved! ™️

PR #241, opened 2019-09-25
New Sites: Add a "Day of Event" page to new sites

ryelle:
This creates a new page that will use the Day of Event template on new sites. It includes the Live Schedule block from #243, and the live Latest Posts block from #250. Now that all this content is blocks, organizers can fully control the information & style of this page.

**To test**

- Create a new WordCamp site
- The new site should pre-populate with a page called "Day of Event", with useful information for during the camp (date, location, live schedule, latest posts).

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
@@ -615,6 +615,16 @@ class WordCamp_New_Site {
 					'wc_page_offline' => 'yes',
 				),
 			),
+
+			array(
+				'title'   => __( 'Day Of Event', 'wordcamporg' ),
+				'content' => $this->get_stub_content( 'page', 'day-of-event', $wordcamp ),
+				'status'  => 'publish',
+				'type'    => 'page',
+				'meta'    => array(
+					'_wp_page_template' => 'day-of-event', // Set the page template.
+				),
+			),
 		);
 
 		return $pages;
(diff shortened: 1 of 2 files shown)


ryelle:
This will wait on #243 & #250 – once those are merged, we can use the new blocks to replicate the existing Day of Event template, and remove the duplicate JS code.

ryelle:
Live "Latest Posts" block requires the Gutenberg plugin to be active.

ryelle:
Now that these blocks are enabled for everyone, this is _actually_ ready for review :)

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
```diff
@@ -615,6 +615,16 @@ protected function get_stub_pages( $wordcamp, $meta ) {
 					'wc_page_offline' => 'yes',
 				),
 			),
+
+			array(
+				'title'   => __( 'Day Of Event', 'wordcamporg' ),
+				'content' => $this->get_stub_content( 'page', 'day-of-event', $wordcamp ),
+				'status'  => 'publish',
```
coreymckrill: Should this be `draft` instead of `publish`, since the page is only intended to be used near the time of the actual event?

coreymckrill (approved):
👍 Tested locally

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
ryelle: oh, that's a good idea 👍

Issue #130, opened 2019-06-21
Finish Day of Event template


iandunn:
- [x] Add dates & location
- [x] The URL to the Schedule page should be determined dynamically
- [x] Add some fallback content in case JS doesn't load
- [x] Remove `stripTagsAndEncodeText` and support diacritics. See https://github.com/wceu/wordcamp-pwa-page/issues/6#issuecomment-499634671
- [x] CSS is too opinionated, remove some styles and let that be handled by remote or custom CSS

_(rewritten by @ryelle to focus just on the content in the Day of Event template, other checklist items have been moved to new issues or trello cards)_

Issue #207, opened 2019-08-08
PWA: Investigate API requests & caching


ryelle:
> We're currently pulling ~615k of data every 60 seconds, which will use a ton of cellular data. That should be paired down to only the fields that actually get used.

This depends on the API response, but we are attempting to call 3 API endpoints every 60s. Right now this is mitigated by the service worker, so it's only passing through to the server every 15min.

> Probably better to have day-of-event.php register a caching route specifically for this endpoint, so that the dependencies are clear and explicit. that may make some of service-worker-caching.php redundant for this request, but that's ok b/c better to make the code self-documenting. add an inline comment w/ that route registration though, noting that it complements things done in service-worker-caching.php

[source](https://github.com/WordPress/wordcamp.org/blob/b58e7e1847564765f3427cf622f0ea844a5891a3/public_html/wp-content/mu-plugins/theme-templates/templates/day-of-event/src/components/main-controller.js#L102-L106)

API requests themselves can be slow, too:

> this is REALLY slow, takes almost 10 seconds on reasonably fast connection need to pair down to only fields being used, maybe other optimizations

[source](https://github.com/WordPress/wordcamp.org/blob/b58e7e1847564765f3427cf622f0ea844a5891a3/public_html/wp-content/mu-plugins/theme-templates/templates/day-of-event/src/api.js#L8-L9)

westonruter:
The service worker gating requests to 15 minute intervals can be eliminated in favor of implementing conditional requests. If the REST API response includes a `Last-Modified` header (or `ETag`) then requests from the page can include a `If-Modified-Since` (or `If-None-Match`)  request header, allowing the REST API to respond with a `304 Not Modified` response, which would have no body, thus saving bandwidth. It would also allow content updates to appear faster for the user.

Refs:
* https://github.com/wceu/wordcamp-pwa-page/issues/6#issuecomment-499366120:
* https://github.com/WordPress/wordcamp.org/issues/129#issuecomment-507057618
* https://core.trac.wordpress.org/ticket/47676

ryelle:
I like the idea of `If-Modified-Since`, but if I'm reading correctly it would still need to hit the server to get the modified time, so we'll probably want to have some extra server caching in place too.

> A good `ETag` could be the max `post_modified` date of whatever is being queried.

@westonruter is there a difference between using `If-Modified-Since` vs an `ETag`? I see you've recommended both/either, but TBH I'm not sure if there's a reason to choose one over the other.

[...]

PR #243, opened 2019-09-27
Blocks: Create a Live Schedule block

### LINKED ISSUE #127: Day of Event Template: Decide on strings for "on now" & "up next" headings
We should use the texts 'On now' and 'Up next' in the day-of-event template here:

https://github.com/WordPress/wordcamp.org/blob/42a361e9f20e642f8d4d6eb037049265e3e60143/public_html/wp-content/mu-plugins/theme-templates/templates/day-of-event/src/components/schedule.js#L115

and here:

https://github.com/WordPress/wordcamp.org/blob/42a361e9f20e642f8d4d6eb037049265e3e60143/public_html/wp-content/mu-plugins/theme-templates/templates/day-of-event/src/components/schedule.js#L122

ryelle:
Add a new block called "Live Schedule", which can be used on the Day of Event template (or any other page). Combined with #241, this will let organizers have total control over the content of the Day of Event template.

This also removes the need for a separate build script/process in `mu-plugins/theme-templates`, that will be removed in a following PR. It does remove the self-updating "Latest Posts" section, but organizers can use the Latest Posts block if they want, and this reduces the repeat API calls. If more organizers want this functionality we can see about hooking into the Latest Posts block.

Fixes #127 – You can customize the heading text and level in the block UI.
See #130 – Addresses a few issues from this issue

**Screenshots**

In wp-admin, it only displays a placeholder (for now). The headings are customized in the sidebar. 

Screen Shot 2019-09-27 at 6 46 16 PM

No sessions upcoming:

Screen Shot 2019-09-27 at 6 51 33 PM

List of sessions (faked "now" to be 10am):

Screen Shot 2019-09-27 at 6 56 07 PM

-----

Still to do (in following PRs)

- Fix issues with time
  - If we're after the start of the last event for a day, it says nothing's happening (we could check if there were other sessions on this day, and put up a "done for the day" message?)
  - If we're before the first event on the second day, it considers the last session of the prev day "on now".
- Remove `stripTagsAndEncodeText` and support diacritics.
- Audit the imports from the frontend script so we're not loading too much to the page
- Add a better, interactive preview in the editor
- (as mentioned) Remove the JS code from day of theme-templates.

------

**To test**

- Create a few sessions that are happening now (or add a property to `$data['live-schedule']` called `nowOverride` which should be a fake "now" in miliseconds).
- Add the Live Schedule block to a page.
- View it on the frontend, it should show you the current and upcoming sessions.


public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/block.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+class Block extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = {
+			sessions: [],
+			isFetching: true,
+		};
+		this.update = this.update.bind( this );
+		this.interval = setInterval( this.update, 60 * 1000 ); // 60 seconds in ms.
```
ryelle: We could tweak this a little, have it only fetch the data from the API every 10min or so but update the schedule display every minute. The schedule probably isn't changing much during the event, a 5-10min wait for fresh data would probably be fine - but we would still want the display to update quickly so when a session ends, the next is immediately visible. Thoughts?

melchoyce:
An idea for a placeholder state, using obviously fake session data:

image

ryelle:
A future improvement could be to add a setting to restrict to a single track, so this block could be used on a livestream page.

public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/block.js
iandunn: If the API endpoint is cached on at the server/network layer -- see https://github.com/WordPress/wordcamp.org/issues/204#issuecomment-535231721 -- then I think leaving it this way is fine, and simplest from an application perspective.

If not, though, then I agree that it'd be ideal to fetch new data every 5-10 min, but have the client update the interface every minute, regardless of whether new data has come in.

`MUST` - If the endpoint isn't cached at the network layer, then changing this to only fetch every 5+ min should probably be a blocker, to avoid a situation similar to the Tagregator bug at WCUS several years ago.

ryelle: > If the API endpoint is cached on at the server/network layer -- see #204 (comment) -- then I think leaving it this way is fine, and simplest from an application perspective.

I think it's a good idea to cache this at the server side, and eventually see if we can get the `If-Modified-Since` headers working (since this would still require a server hit, it won't protect against the tagregator issue alone). Let's move any more API cache discussion over to #207.

In any case, I've updated this PR to fetch the API every 5 minutes, while still re-rendering every minute.

ryelle:
Rebased to remove the eslint `require-jsdoc` exceptions , and added the editor preview – so now it looks like this in the editor:

Screen Shot 2019-10-11 at 3 42 34 PM

The data we use to populate that is in [`data/mock.js`.](https://github.com/WordPress/wordcamp.org/blob/update/blocks-live-schedule/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/data/mock.js) It doesn't pull any API data, since the site probably doesn't have any active sessions when someone is editing the page.

PR #244, opened 2019-09-30
PWA: Add caching support for uploaded files and Jetpack files.

### LINKED ISSUE #203: PWA: Add caching for uploaded files
File uploads are loaded from the `/files/` directory, while the service working is only caching `wp-content` or `wp-includes` assets.

Example URLs:

https://2019.europe.wordcamp.org/files/2019/06/blog-2-1024x512.jpg
https://2019.annarbor.wordcamp.org/files/2018/08/woocommerce_logo.png

Sites can also opt into Jetpack's CDN, so these might not be cached - I think it depends on CORS + cache settings? Needs more investigation.

ryelle:
Uploaded files on a multisite install use `files/…` as the path, and are not caught by the `wp-content` rule. Files cached by Jetpack (either core Jetpack assets or Photon-enabled images) use the wp.com domain, so these must also be added to the cache rules.

Fixes #203 

**To test**

- Try this on a site with uploaded files by navigating to a page with uploads
- Turn on offline mode/turn off your wifi
- Reload the page
- The page should still load, images and all
- You can also look for the images in the wp-assets cache (for uploaded filed) or `wc-jetpack` (for jetpack assets).

### DIFF
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
@@ -1,6 +1,8 @@
 <?php
 
 namespace WordCamp\Jetpack_Tweaks;
+use WP_Service_Worker_Caching_Routes, WP_Service_Worker_Scripts;
+
 defined( 'WPINC' ) || die();
 
 // Allow Photon to fetch images that are served via HTTPS.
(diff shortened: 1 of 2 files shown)


ryelle:
Forgot to mention two things:

1. it doesn't really play well with Jetpack's Lazy Images (by design?). We only cache things when we make the network request, but lazy images aren't requested until they're in view. Ex: you can open a page, and if the sponsor logos are at the footer but you never scroll there, they're not cached. so now you go offline, reload the page and scroll – the images are not there. I think this is actually "expected" behavior, but we might want to think about notifying folks who have Lazy Images + PWA active.
2. Something wonky is happening with the EU Cookie banner CSS & JS, but maybe that's just my local environment? Or might be related to how we call the enqueue function. I can look into this more once the PR is merged.

PR #250, opened 2019-10-07
Blocks: add "Live update" option to Latest Posts block

ryelle:
_Note: this is built off the branch in #243_ 

This will enable realtime updates of news posts on the frontend of the site, without refreshing the page. It can be used on the Day of Event template (or anywhere on the site) to always show the latest posts.

This hooks into the Latest Posts block to add a new "Live Update" toggle. We also need to hook into the `register_block_type` code to unregister the core block, and re-register it with this new attribute and to add our front-end script.

enabled-off

enabled-on

If the order is not "newest to oldest", we don't add the live-update script, so we should give the post author a warning about that:

disabled-off

On the front-end, we filter the block content to inject the data-attribute and a special class, if live update is enabled and supported. This is used to trigger the JS rendering.

In JS, we take advantage of the ServerSideRender component & API to render the block using the same `render_callback` that's used to render the block in the editor. The API endpoint needs to be safelisted for anonymous visitor access, the alternative is building our own SSR component and endpoint.

loading

list

cc @melchoyce for a quick design/copy review.

**To test**

- Add the Latest Posts block to your post/page
- Change any settings
- Turn on "Live update"
- Publish, and view on the frontend
- As the content loads, it shows the loading view
- Once loaded, it should look the same as the static Latest Posts block
- But after a minute, the block will reload, showing any posts that were published in the last minute.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/blocks.php
+++ b/public_html/wp-content/mu-plugins/blocks/blocks.php
@@ -15,6 +15,7 @@ function load_includes() {
 	$includes_dir   = PLUGIN_DIR . 'includes/';
 	$blocks_dir     = PLUGIN_DIR . 'source/blocks/';
 	$components_dir = PLUGIN_DIR . 'source/components/';
+	$hooks_dir      = PLUGIN_DIR . 'source/hooks/';
 
 	require_once $includes_dir . 'definitions.php';
 
@@ -37,6 +38,9 @@ function load_includes() {
 		|| in_array( get_current_blog_id(), [ 928 ], true ) // 2017.testing
 	) {
 		require_once $blocks_dir . 'live-schedule/controller.php';
+
+		// Hooks.
+		require_once $hooks_dir . 'latest-posts/controller.php';
 	}
 
 }
--- a/public_html/wp-content/mu-plugins/blocks/package.json
+++ b/public_html/wp-content/mu-plugins/blocks/package.json
@@ -39,8 +39,8 @@
 		"*.min.js"
 	],
 	"scripts": {
-		"start": "wp-scripts start blocks=./source/blocks.js live-schedule=./source/blocks/live-schedule/front-end.js",
-		"build": "wp-scripts build blocks=./source/blocks.js live-schedule=./source/blocks/live-schedule/front-end.js",
+		"start": "wp-scripts start blocks=./source/blocks.js live-schedule=./source/blocks/live-schedule/front-end.js live-posts=./source/hooks/latest-posts/front-end.js",
+		"build": "wp-scripts build blocks=./source/blocks.js live-schedule=./source/blocks/live-schedule/front-end.js live-posts=./source/hooks/latest-posts/front-end.js",
 		"lint:js": "wp-scripts lint-js",
 		"lint:css": "wp-scripts lint-style '**/*.scss'",
 		"lint:pkg-json": "wp-scripts lint-pkg-json",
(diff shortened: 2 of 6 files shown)


iandunn (approved):

melchoyce:
I think I want to rewrite the copy.

What's one reason someone would or would not want to use this feature?

shaunandrews:
The accordion label could be "Live Updates" with the label on the switch being more descriptive, like "Add new posts as they're published".

I find it strange that the label and description are dynamic based on the state of the switch — I don't think thats a consistent pattern.

ryelle:
> I find it strange that the label and description are dynamic based on the state of the switch — I don't think thats a consistent pattern.

Hm, I thought it was… maybe it changed? The "Drop Cap" control changes the help text to describe the current state… but nothing else I can find right now does that. I personally like when the help text describes the current state, but if we have better labels/copy maybe it's not necessary.

> "Add new posts as they're published"

This might make it sound like it's totally static when off, like the posts you publish the page with are the posts there forever– not that it'll still pull in the new posts once you refresh.

> What's one reason someone would or would not want to use this feature?

If your posts are very important/timely, you might want this so people are always aware of the latest news. The use case specifically is on the day of a WordCamp, you want to make sure someone looking at the site will always see the latest posts, without the extra step to refresh the page.

melchoyce:
Okay, so maybe...

-----

**Live Updates**

This feature helps your attendees keep up-to-date with your WordCamp's latest news. When active, new posts will be loaded as they're published without your attendees needing to refresh the page.

(o  ) Live update posts

-----

And then we could put the panel underneath Sorting & Filtering — or even in Sorting & Filtering, and the option only shows when posts are set to News to Oldest?

enriquesanchez:
>you want to make sure someone looking at the site will always see the latest posts, without the extra step to refresh the page.

@ryelle Is there a situation where you wouldn't want this? I'm wondering if Live Updates should be ON by default or if this shouldn't be a setting at all?

>I find it strange that the label and description are dynamic based on the state of the switch

Would a dynamic label affect accessibility?

ryelle:
> Is there a situation where you wouldn't want this? I'm wondering if Live Updates should be ON by default or if this shouldn't be a setting at all?

I don't think we want to ping the API every minute for all instances of this block. I'd also say for the vast majority of sites, you don't need to "push" notifications out like this, it's only really relevant on the day of the event.

> Would a dynamic label affect accessibility?

I don't think it does, IIRC, but I haven't tested it in a while. I'm happy to un-dynamic-ify it though – I like @melchoyce's copy in https://github.com/WordPress/wordcamp.org/pull/250#issuecomment-540167912

enriquesanchez:
>I don't think we want to ping the API every minute for all instances of this block

Oh, I had not thought of that! Thanks for the clarification. 

I wonder if this important detail is something that should be somehow communicated. If someone's not aware of the technical implications of this setting (like me), they could probably end up keeping this setting on all the time.

>I like @melchoyce's copy in #250 (comment)

Agreed, I like that option too.

ryelle:
Update: I haven't changed any behavior, just the copy - it's no longer dynamic, the only change is when you switch away from "Newest to Oldest" it swaps the toggle for the unsupported message.

I can't put it into the Sorting & Filtering panel because there are no hooks there, so instead it's right below.

Screen Shot 2019-10-11 at 2 43 19 PM Screen Shot 2019-10-11 at 2 49 11 PM

PR #253, opened 2019-10-15
Templates: Clean up Day of Event template

ryelle:
Update the template structure to better match core themes, and use post content instead of custom partials for displaying data. Since there are no currently-running camps using the day of pages, we can safely remove the custom code from the `day-of-event` folder, and simplify the template to just show the page content.

**To test**

- Create a page with the "Day of Event" page template, view it
- It should not have any PHP errors
- It should not have `<main id="day-of-event">Loading…</main>`
- It should display the page title
- It should display only the content you've added to the page


coreymckrill (approved):
Everything checks out in testing 👍

PR #256, opened 2019-10-30
Blocks: Live Schedule: Decode html entities before displaying on schedule

ryelle:
If a session name has an HTML entity, like an em-dash, it is now correctly decoded before being displayed to the site visitor.

This replaces `stripTagsAndEncodeText` with `decodeEntities( stripTags( … ) )`. I think `stripTags` alone does enough sanitization, and we want the entities decoded, not encoded, but I'm asking for a review just to confirm that. I'd like to merge this in the next 24hrs so we can deploy it before WCUS, since we have some sessions it affects.

To test: 

- Add a Live Schedule block to a page
- Add or update some sessions to be happening "now", add an emdash or some other fancy html entity
- Check that it displays correctly on the frontend of your site


iandunn (approved):
LGTM, couldn't find a way to bypass it.

PR #258, opened 2019-10-31
Tomjn/vvv provisioner

tomjn:
Holy moly was this a pain in the arse to get started, but I've gotten most of the way. It still needs some adjustments to get the WP install right, and it needs to install the database

A word of note: `/usr/src` is not a great place to put a WWW folder, for example linux headers get put in there amongst other things packages do during installation. Otherwise my job would have been much easier with a nice symlink

Once this is merged I'd like to eliminate the WordCamp sites from the WP Meta environment. It's clear they aren't maintained whereas this repo is, so it's better to consolidate

**Note:** The other option is to fix the meta environment so that it works on contributor days. Otherwise we'd have to bundle a tonne of various docker images, and write a tonne of docs on Hyper-V, command line, etc

### DIFF
--- a/.docker/bin/extra-setup.sh
+++ b/.docker/bin/extra-setup.sh
@@ -21,7 +21,9 @@ function install_composer() {
 }
 
 function do_extra_setup() {
-  PKG_PATH="/usr/src"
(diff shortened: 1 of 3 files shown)


pkevan:

PR #269, opened 2019-11-03
Enhancement: Textfield resize for WordCamp and Meetup post types

terriann:
Allows all textareas in meta boxes for both the WordCamp and Meetup custom post types to be resized both vertically and horizontally

**To test**

* Edit or Add a new WordCamp or Meetup through the admin
* The textareas in all the meta boxes should be able to resize vertically and horizontally

### DIFF
--- a/public_html/wp-content/plugins/wcpt/css/applications/admin.css
+++ b/public_html/wp-content/plugins/wcpt/css/applications/admin.css
@@ -12,6 +12,11 @@ span.swag-needed-icon {
     width: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


terriann:
I noticed while testing locally I had to hard refresh to see my CSS changes - I wasn't sure the process of updating the plugin version number and the associated PHP constant but I'm happy to update that if needed!

https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wcpt/wcpt-loader.php#L17

PR #274, opened 2019-11-04
Packages: Set up yarn workspaces for sub-projects

ryelle:
_Note: This is just a proposal for discussion._

This PR sets up [yarn workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) for our projects. The benefit here is mostly for us as devs– as we're creating more packages that use React, build processes, wordpress components, etc, we need to keep re-installing these packages in each subfolder. Using `yarn install` in the root of the project now will hoist all the sub-projects' packages into `/node_modules/`, so we only have one copy of `@wordpress/scripts` , for example.

Currently there is only the one project (I've left `public_html/wp-content/plugins/wcpt` as a stand-alone since it has so many unique dependencies that won't be shared), but as we add more we can share the dependencies across them assuming they're all the same version (so it might be good to bring up the dependabot or renovate discussion again).

Another possible benefit that's pretty far off - workspaces can depend on one another too, so if we created any shared components or other shared code, it will be easier to link those together than using `npm link`.

More info:
- [Introducing workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/)
- [`yarn workspace` cli docs](https://yarnpkg.com/lang/en/docs/cli/workspace/)


iandunn:
Sounds good to me 👍

coreymckrill:
Still thinking through this, but at first glance this seems like a great idea. I actually ran across Yarn a while back when I was working on some enhancements to the Docker environment. Yarn [scripts](https://yarnpkg.com/en/docs/package-json#scripts-) could be really useful for fully setting up Docker with one command, because it could execute several things both inside and outside the containers.

ryelle:
Are yarn scripts different than [npm scripts?](https://docs.npmjs.com/misc/scripts) Not to say we shouldn't use either, but i don't think that's a specifically yarn thing.

coreymckrill:
It's not just a yarn thing, but I think it's more useful for Docker setup, etc., when the package is in the root of the repo, which so far hasn't been the case for our npm stuff.

ryelle:
Tests were not working due to modules moving around, I think— jest needs to be told where the `rootDir` of the whole project is, and it also needs a defined `babel` property otherwise it thinks we don't use babel & `SyntaxError`s at the first sign of JSX. Those things have been updated, and I think this PR is ready for merge.

Any objections to merge, @coreymckrill @iandunn ?

.docker/readme.md
```diff
@@ -36,6 +36,13 @@ Follow these steps to setup a local WordCamp.org environment using [Docker](http
 	composer install
 	```
 
+1. Install 3rd-party JS packages and build the CSS & JS needed for some projects. You'll need [node](https://nodejs.org/) & [yarn](https://yarnpkg.com/). Optionally you can use [nvm](https://github.com/nvm-sh/nvm) to keep your node version up to date. Run the following to install and build each project (omit `nvm` command if you're not using it).
```
coreymckrill:
```suggestion
1. Install 3rd-party JS packages and build the CSS & JS needed for some projects. You'll need [node](https://nodejs.org/) & [yarn](https://yarnpkg.com/). Optionally you can use [nvm](https://github.com/nvm-sh/nvm) to keep your node version up to date. Running the following will install and build all of the projects in one step (omit `nvm` command if you're not using it).
```

When I first read this I thought it meant I would need to run these commands for each project. So this is just a minor wording change that I think makes it clearer, but you can take it or leave it.

coreymckrill (approved):

.docker/readme.md
ryelle: Hm, I see what you mean— the yarn workspaces run build command does actually run through each project (workspace) itself, which is why I said it like that, but this is more clear to folks who don't know that 👍🏻

PR #281, opened 2019-11-13
CampTix: Notify: Add Attended segment

### LINKED ISSUE #280: CampTix: Notify: Can't send notification message to just people that attended WordCamp
_Copied from Meta Trac [#4641](https://meta.trac.wordpress.org/ticket/4641)._

Under CampTix Tools, the notify feature is very helpful. However, it would be especially helpful if we could filter by people that were marked as attended. It's annoying (and annoying to them) to send follow-up surveys to everyone that bought a ticket, instead of just those that attended.

CdrMarks:
Can now segment attendees by if they attended the event.
Fixes #280.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -3313,6 +3313,25 @@ class CampTix_Plugin {
 				?>
 			}));
 
+			camptix.collections.segmentFields.add( new camptix.models.SegmentField({
+				caption: 'Attended the event',
+				option_value: 'attended',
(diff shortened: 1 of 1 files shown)


ryelle:
@CdrMarks Thanks for working on this. It looks like it only works for the "has attended" segment though, because `tix_attended` only exists when an attendee is marked as attended. Maybe you can use `EXISTS`/`NOT EXISTS` instead?

CdrMarks:
@ryelle, thank you for the feedback.  I had trouble getting my existing PR updated, so I closed this one and opened #288.

PR #285, opened 2019-11-14
QuickBooks: OAuth2 authentication flow

### LINKED ISSUE #283: QBO: Create OAuth2 authentication flow in network admin
Similar to the current OAuth v1 flow, there is a required step where a user must be redirected to the QBO website to login and consent, and then get redirected back to our app. However, for v2 the process is a bit different behind the scenes, so we'll build it out separately, rather than refactoring the v1 code. In our v1 app the redirection flow happens on the WordCamp Central site, but since the authentication applies to the whole network, it makes more sense to have this live in the network admin.

Since we're using the QBO v3 PHP SDK, but not distributing it, we also need to make sure this fails gracefully (but still with a notification and/or error log) if the SDK is not available.

coreymckrill:
Our current QuickBooks authentication uses OAuth 1.0, however Intuit is deprecating that method on December 17th in favor of OAuth 2.0. This adds tools to authenticate with OAuth 2.0, using Intuit's QuickBooks V3 PHP SDK. This will also set us up for using more of the SDK's functionality in the future for improvements to our payments and reporting.

Fixes #283 

**To test**

* Testing will probably only work on the sandbox since the Redirect URI is hard coded in the app created on the Intuit site.
* Make sure your sandbox is up-to-date so that it has the SDK installed.
* Navigate to Network Admin->Settings->QuickBooks
* It may already be connected. Try both connecting and disconnecting.
* When connecting, use our WordCamp Developer credentials to log into the Intuit site.


composer.json
```diff
@@ -12,9 +12,6 @@
 		},
 		"vendor-dir": "public_html/wp-content/mu-plugins/vendor"
 	},
-	"require": {
-		"league/oauth2-client": "2.4.1"
```
ryelle, line 16: This is removed because you ended up not needing it for Meetup oauth either, right?

public_html/wp-content/mu-plugins/quickbooks/includes/client.php
```diff
@@ -0,0 +1,382 @@
...
+		} catch ( SdkException $exception ) {
+			$this->add_error_from_exception( $exception );
+		}
+
+		if ( isset( $config['refreshTokenKey'] ) ) {
+			try {
+				/**
+				 * The OAuth access token expires after an hour, so whenever we instantiate with saved token data, we
+				 * need to refresh it with the refresh token to get a current access token. According to the SDK docs,
+				 * the refresh token may change when this happens, so it needs to be re-saved during this process.
```
ryelle, line 75: Interesting. So, to make sure I understand this– this code will fetch a new access token whenever the client is initialized, which should ensure that any later call has a valid token. Is that right?

public_html/wp-content/mu-plugins/quickbooks/includes/client.php
```diff
@@ -0,0 +1,382 @@
...
+		 *
+		 * @param array $config {
+		 *     See https://intuit.github.io/QuickBooks-V3-PHP-SDK/configuration.html
+		 *
+		 *     @type string $auth_mode
+		 *     @type string RedirectURI
+		 *     @type string $scope
+		 *     @type string $ClientID
+		 *     @type string $ClientSecret
+		 *     @type string $baseUrl
```
ryelle: It might be worth adding some short docs for each param here, just in case the above link changes or goes away. Especially for `scope` & `auth_mode`, the others are more self-explanatory.

.docker/wp-config.php
```diff
@@ -108,6 +108,11 @@
 define( 'WORDCAMP_QBO_CONSUMER_SECRET', '' );
```
ryelle, line 108: The edits on this file are making it fail the phpcs check – maybe add `// phpcs:ignoreFile` to the start of the file?

ryelle (approved):
Looks good! I tested the disconnect/connect flow and it works as expected. There is no code for making an oauth'd request yet, right? Just want to make sure I'm not missing it.

Locally, I ended up with this error displayed twice when on the QBO connection page - looks like it's from the `maybe_show_disconnection_warning` hook, so you could either disable that on this page, or leave it; TBH it's not a big issue.

Screen Shot 2019-11-19 at 1 50 44 PM

.docker/wp-config.php
coreymckrill: Hrm, none of the errors are related to lines that I actually changed. Is there a way to limit the phpcs check to only changed lines? `// phpcs:ignoreFile` seems drastic.

public_html/wp-content/mu-plugins/quickbooks/includes/client.php
coreymckrill: Good idea. Added: 3b8ab198

public_html/wp-content/mu-plugins/quickbooks/includes/client.php
coreymckrill: Yes, it will ensure that all calls on the current session have a valid access token (unless the session somehow lasts more than an hour).

This might be slightly overkill, but it's simpler than storing the access token expiration timestamp and then writing logic to decide whether there's enough time left on the token or not before refreshing it.

composer.json
coreymckrill: Correct. It wasn't sufficient for Meetup, and the QBO SDK was a lot closer to turnkey.

.docker/wp-config.php
ryelle: yeah, that was a todo item we decided to come back to– I've mostly been fixing issues as they come up so that the whole file passes, as a way to clean up as we go, but I figured wp-config would be a special case that could be ignored.

coreymckrill:
> There is no code for making an oauth'd request yet, right?

Correct. This just establishes and maintains the connection that makes requests possible.

> Locally, I ended up with this error displayed twice when on the QBO connection page

Oh, good catch. That admin notice should only display the one particular error when QBO is disconnected, not all the other errors that might be causing the disconnection. Fixed in a48fcd50

.docker/wp-config.php
coreymckrill: There were only a few errors, so I just fixed the file: c1cbb35c

CdrMarks:
Hi @coreymckrill,  I was going to work on #281 today, but found an error after pulling down the recent commits from the the `production` branch into my local environment.

> Parse error: syntax error, unexpected '|', expecting variable (T_VARIABLE) in /srv/www/wordpress-meta-environment/github_wordcamp_repo/public_html/wp-content/mu-plugins/quickbooks/includes/client.php on line 81

coreymckrill:
@CdrMarks what version of PHP are you running? The single pipe is a valid separator between multiple exception types in a catch statement [as of 7.1](https://www.php.net/manual/en/language.exceptions.php).

PR #288, opened 2019-11-23
Camptix: Notify: Add attended segment

### LINKED ISSUE #280: CampTix: Notify: Can't send notification message to just people that attended WordCamp

CdrMarks:
Can now segment attendees by if they attended the event.
Fixes #280.

Reattempt after feedback on #281.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -3314,6 +3314,13 @@ class CampTix_Plugin {
 			}));
 
 
+			camptix.collections.segmentFields.add( new camptix.models.SegmentField({
+				caption: 'Attended the event',
+				option_value: 'attended',
+				type: null,
+				ops: [ 'Yes', 'No' ]
+			}));
+
 			// Add POST'ed conditions.
 			<?php if ( ! empty( $conditions ) ) : ?>
 				<?php foreach ( $conditions as $condition ) : ?>
@@ -3390,7 +3397,10 @@ class CampTix_Plugin {
 		);
 
 		foreach ( $conditions as $condition ) {
-			if ( empty( $condition['field'] ) || empty( $condition['op'] ) || ! isset( $condition['value'] ) )
+			if ( empty( $condition['field'] ) || empty( $condition['op'] ) )
+				continue;
+
+			if ( "attended" != $condition['field'] && ! isset( $condition['value'] ) )
 				continue;
 
 			// Purchased ticket.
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/camptix/camptix.php
```diff
@@ -3460,6 +3470,26 @@ public function get_segment( $relation, $conditions ) {
...
 				continue;
 			}
+
+			// Attended the event
+			if ( 'attended' == $condition['field'] ) {
+				$meta_query = array(
+					'key' => 'tix_attended',
+					'value' => 1,
+				);
+				switch ( $condition['op'] ) {
```
ryelle, line 3480: This block might be more readable as an `if`/`else` statement - what do you think?

public_html/wp-content/plugins/camptix/camptix.php
```diff
@@ -3390,7 +3397,10 @@ public function get_segment( $relation, $conditions ) {
 		);
 
 		foreach ( $conditions as $condition ) {
-			if ( empty( $condition['field'] ) || empty( $condition['op'] ) || ! isset( $condition['value'] ) )
+			if ( empty( $condition['field'] ) || empty( $condition['op'] ) )
+				continue;
+
+			if ( "attended" != $condition['field'] && ! isset( $condition['value'] ) )
```
ryelle, line 3403: Some PHP code style nitpicks: this should use single quotes around `attended`, and use strict comparisons: 
```suggestion
			if ( 'attended' !== $condition['field'] && ! isset( $condition['value'] ) )
```

public_html/wp-content/plugins/camptix/camptix.php
```diff
@@ -3460,6 +3470,26 @@ public function get_segment( $relation, $conditions ) {
 				$post_query_conditions[] = $condition;
 				continue;
 			}
+
+			// Attended the event
+			if ( 'attended' == $condition['field'] ) {
```
ryelle, line 3475: Again, this should use a strict comparison:
```suggestion
			if ( 'attended' === $condition['field'] ) {
```

ryelle (commented):
I left a few code style changes (you should be able to merge those from the github UI, if you want), and one comment about the `switch` statement, that one's up to you :)

public_html/wp-content/plugins/camptix/camptix.php
CdrMarks: I agree an `if`/`else` statement would be more readable.  I was using the same pattern I had seen for the other segments.
Here are links to those: 
* https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/camptix.php#L3403
* https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/camptix.php#L3421
* https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/camptix.php#L3441

Would you suggest I change the others as well?

ryelle: Yeah, I saw the others– I also think those would be more readable as if/else, but no, I wouldn't touch those in this PR. It's okay if this one looks different, the idea is to iteratively make this code better, not refactor everything all at once.

ryelle:
Hey @CdrMarks, are you still working on this issue? Let me know if you'd like to me to take it over and merge it 🙂

PR #291, opened 2019-12-04
Payments: Add "uncollectible" status for sponsor invoices

### LINKED ISSUE #282: Payments: Add status "Uncollectable" for Sponsor Invoices
For accounting and Quickbooks-connection reasons, we need a status for Sponsor Invoices that indicates that we're writing off the invoice as "uncollectable."

ryelle:
We need an "uncollectible" status for sponsor invoices that are not going to be paid. This adds the new post status, and adds an action in the Sponsor Invoices list table to mark invoices uncollectible. It also adds an Uncollectible tab to the network Sponsor Invoices dashboard.

Fixes #282 

### Screenshots

The Sponsor Invoices list on a WordCamp. Uncollectible invoices are flagged as such, and other invoices are able to be marked as uncollectible.

sponsor-list

The network dashboard tab.

network-dashboard

1. Add some test invoices to your dev environment
2. Mark them uncollectible
3. The invoice should be correctly updated, and show as uncollectible both in the list table and in the edit screen.
4. The network dashboard should show this invoice in the uncollectible tab.

There should be no interaction with QBO by using this status.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments-network/includes/sponsor-invoices-dashboard.php
+++ b/public_html/wp-content/plugins/wordcamp-payments-network/includes/sponsor-invoices-dashboard.php
@@ -109,26 +109,26 @@ function get_submenu_page_sections() {
 	$sections        = array();
 	$current_section = get_current_section();
 
-	foreach ( $statuses as $status_slug => $status_name ) {
-		$status_slug = str_replace( 'wcbsi_', '', $status_slug );    // make the URL easier to read
+	foreach ( $statuses as $slug => $status ) {
+		$slug = str_replace( 'wcbsi_', '', $slug );    // make the URL easier to read.
 
 		$classes = 'nav-tab';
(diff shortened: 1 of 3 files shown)


coreymckrill (approved):
👍 Everything worked in testing.

I tried it using Editor and Author roles too. Editors can mark things uncollectible, but Authors can't. But I wonder if we should limit changing to that status to only admins, or even only network admins, since the determination that an invoice is indeed uncollectible would only come from someone at WPCS.

ryelle:
>  I wonder if we should limit changing to that status to only admins, or even only network admins, since the determination that an invoice is indeed uncollectible would only come from someone at WPCS.

Yeah, and since it's not easily undo-able, I think it makes sense to restrict it to just network admins— made that change in https://github.com/WordPress/wordcamp.org/pull/291/commits/bc4def476f19762daad059beb416f551083dcdd7

I'll merge this tomorrow morning unless I hear otherwise 🙂

PR #292, opened 2019-12-04
QBO: Refactor to use OAuth2 token for REST request authentication

### LINKED ISSUE #284: QBO: Use OAuth2 access token in API requests
This should refactor all of our requests to the QBO API to use the OAuth2 access token. It also needs to take into account that access tokens expire after an hour, and thus need to be refreshed without interrupting API operations.

Depends on the resolution of #283

coreymckrill:
With #283 merged, we have a new client class that provides a functional OAuth2 connection flow and token that can be used to authenticate requests to the QBO API. The client also has the full QBO PHP SDK, which has helper methods for making requests to the API. However, in this PR, I'm not attempting to switch everything over to use the SDK (except in WordCamp Reports, where it was a fairly simple change). Rather, I'm just modifying our existing methods that send requests so that they use our new OAuth2 token, and the correct realm ID. This also removes the markup and methods that provided the flow for making an OAuth v1 connection, since that will no longer be possible.

In a future issue/PR we can explore using more of the SDK's capabilities.

Fixes #284


ryelle (approved):
I've read through the code changes, and it all makes sense. Not testing based on the slack convo.

It looks like you've got a bunch of phpcs issues, which is causing [the CI failure.](https://travis-ci.com/WordPress/wordcamp.org/jobs/263585278) I think some of that is on the new code, so it'd be nice to see that fixed, but don't consider it a blocker to merge.

coreymckrill:
Some phpcs cleanup here: db0c8a33a52c97b13d137682e721c691f8dfeb88

The **wordcamp-qbo.php** is a big mess, and I don't want to worry about introducing regressions by fixing all of its errors, so I'm going to skip that file in this PR.

ryelle:
@coreymckrill looks good, thanks for fixing those up!

PR #294, opened 2019-12-07
Add id to organizer block (Trac #4893)

vdwijngaert:
See https://meta.trac.wordpress.org/ticket/4893


ryelle:
Closing this based on the comment in https://meta.trac.wordpress.org/ticket/4893#comment:2 — but thanks for jumping in with a PR!

PR #298, opened 2019-12-13
Payments: Send a receipt to the sponsor after paying an invoice

### LINKED ISSUE #297: Payments: Send a receipt email after sponsor payment
We currently don't send any email receipt after sponsors pay the invoice via stripe on the [Sponsorship Payment](https://central.wordcamp.org/sponsorship-payment/) page. Stripe will send a receipt if we pass [a `receipt_email` to the charge API call.](https://stripe.com/docs/receipts#receipts-charges)

We need to get the email (maybe via the stripe popup), and pass that through to the charge call.

ryelle:
Pass the email back to stripe when sending the charge request. Stripe expects a specific parameter in the API request in order to send a receipt.

Fixes #297 

You technically can't, since the sandbox uses the stripe test environment, and the test env doesn't trigger the automatic receipt. You can see in the API log for the request that the `receipt_email` is passed in. I'm basically trusting that the API is correct :|

If you could test, you would…

1. Submit a payment on https://central.wordcamp.org/sponsorship-payment/
2. You should get a receipt


iandunn (approved):
LGTM 👍

PR #299, opened 2019-12-16
Add end times for Session post type

iandunn:
Fixes https://meta.trac.wordpress.org/ticket/3623

Will be needed for the (full) Schedule block (see https://meta.trac.wordpress.org/ticket/3117), and the Live Schedule block.

I've tested it with the full Schedule block and it works well. @ryelle , can you test it w/ the Live Schedule block, and make sure it meets your needs for that? It adds new fields to the `Session Info` metabox.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -26,8 +26,11 @@ function expose_public_post_meta() {
 		'single'       => true,
 	];
 
+	$default_integer = wp_parse_args( array( 'type' => 'integer' ), $meta_defaults );
+
 	// Session.
-	register_post_meta( 'wcb_session', '_wcpt_session_time', wp_parse_args( [ 'type' => 'integer' ], $meta_defaults ) );
+	register_post_meta( 'wcb_session', '_wcpt_session_time',     $default_integer );
+	register_post_meta( 'wcb_session', '_wcpt_session_duration', $default_integer );
 	register_post_meta( 'wcb_session', '_wcpt_session_type', $meta_defaults );
 	register_post_meta( 'wcb_session', '_wcpt_session_slides', $meta_defaults );
 	register_post_meta( 'wcb_session', '_wcpt_session_video', $meta_defaults );
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -12,6 +12,8 @@ require_once 'inc/deprecated.php';
 class WordCamp_Post_Types_Plugin {
 	protected $wcpt_permalinks;
 
+	const SESSION_DEFAULT_DURATION = 50 * MINUTE_IN_SECONDS;
+
 	/**
 	 * Fired when plugin file is loaded.
 	 */
@@ -307,8 +309,9 @@ class WordCamp_Post_Types_Plugin {
 			case 'edit-wcb_sponsor':
 			case 'edit-wcb_session':
 			case 'wcb_sponsor':
+			case 'wcb_session':
 			case 'dashboard':
-				wp_enqueue_style( 'wcpt-admin', plugins_url( '/css/admin.css', __FILE__ ), array(), 1 );
+				wp_enqueue_style( 'wcpt-admin', plugins_url( '/css/admin.css', __FILE__ ), array(), 2 );
 				break;
 			default:
 		}
@@ -1271,6 +1274,10 @@ class WordCamp_Post_Types_Plugin {
 		$session_minutes  = ( $session_time ) ? date( 'i', $session_time )     : '00';
 		$session_meridiem = ( $session_time ) ? date( 'a', $session_time )     : 'am';
 
+		$session_duration         = $post->_wcpt_session_duration ?? self::SESSION_DEFAULT_DURATION;
+		$session_duration_hours   = floor( $session_duration / HOUR_IN_SECONDS );
+		$session_duration_minutes = floor( ( $session_duration / MINUTE_IN_SECONDS ) % MINUTE_IN_SECONDS );
+
 		$session_type     = get_post_meta( $post->ID, '_wcpt_session_type',   true );
 		$session_slides   = get_post_meta( $post->ID, '_wcpt_session_slides', true );
 		$session_video    = get_post_meta( $post->ID, '_wcpt_session_video',  true );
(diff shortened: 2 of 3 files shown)


public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -26,8 +26,12 @@ function expose_public_post_meta() {
 		'single'       => true,
 	];
 
+	$default_integer = wp_parse_args( array( 'type' => 'integer' ), $meta_defaults );
+
 	// Session.
-	register_post_meta( 'wcb_session', '_wcpt_session_time', wp_parse_args( [ 'type' => 'integer' ], $meta_defaults ) );
+	register_post_meta( 'wcb_session', '_wcpt_session_time',           $default_integer );
+	register_post_meta( 'wcb_session', '_wcpt_session_length_hours',   $default_integer );
+	register_post_meta( 'wcb_session', '_wcpt_session_length_minutes', $default_integer );
```
ryelle: Could this be saved (or just exposed) as one value, `_wcpt_session_length`? This way we can avoid repeated calculations, since we'll almost always want to know the full duration, not just hours/minutes.

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -1300,6 +1310,36 @@ public function metabox_session_info() {
...
+			</label>
+
+			<select id="wcpt-session-length-hours" name="wcpt-session-length-hours" aria-label="<?php esc_html_e( 'Session Length: Hours', 'wordcamporg' ); ?>">
+				<?php for ( $i = 0; $i <= 23; $i++ ) : ?>
+					<option value="<?php echo esc_attr( $i ); ?>" <?php selected( $i, $session_length_hours ); ?>>
+						<?php echo esc_html( $i ); ?>
+					</option>
+				<?php endfor; ?>
+			</select>
+			<label for="wcpt-session-length-hours">
```
ryelle: It looks like while two labels for the same input is technically valid html, [screen readers might not read this correctly.](https://stackoverflow.com/questions/2829936/can-an-input-field-have-two-labels/3992160#3992160) A more standard approach would be a `fieldset` to wrap both inputs, and `legend` for the first label; and then you don't need the `aria-label`s either (the legend will set the context for both selects, then read their labels).

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -1495,13 +1535,16 @@ public function save_post_session( $post_id, $post ) {
 				sanitize_text_field( $_POST['wcpt-session-date'] ),
 				absint( $_POST['wcpt-session-hour'] ),
 				absint( $_POST['wcpt-session-minutes'] ),
-				'am' == $_POST['wcpt-session-meridiem'] ? 'am' : 'pm'
+				'am' === $_POST['wcpt-session-meridiem'] ? 'am' : 'pm'
 			) );
 			update_post_meta( $post_id, '_wcpt_session_time', $session_time );
 
+			update_post_meta( $post_id, '_wcpt_session_length_hours',   absint( $_POST['wcpt-session-length-hours']   ) );
+			update_post_meta( $post_id, '_wcpt_session_length_minutes', absint( $_POST['wcpt-session-length-minutes'] ) );
```
ryelle: Same comment as on the API side - I think it would make more sense to calculate the duration (in seconds, probably) and save that in one single field.

ryelle (commented):
I left feedback on this as-is, but the new UI control could use some design feedback too.

@melchoyce we're talking about just the "Length" fields in the screenshots here:

Artboard

melchoyce:
Is this in the sidebar? Hard to tell but looks like maybe no?

Would be great to simplify and streamline these fields — maybe something like:

image

Instead of dropdowns, it seems like it would be easier for folks to enter numbers directly into a number input field. Then they could also use arrow keys to adjust increments.

Ideally, we'd also use the calendar/datepicker component in addition to the field, like the publish field:

image

One of the things I struggle with when entering sessions is, well, dates — numbers tend to get jumbled in my head. Having a visual calendar for scheduling vastly improves the overall experience for me.

ryelle:
> Ideally, we'd also use the calendar/datepicker component in addition to the field, like the publish field

yeah, I'd love to get this using gutenberg stuff instead of a plain old metabox, but i think that's an issue for another time 🙂

melchoyce:
Ah, it's a metabox. Fair enough.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
iandunn: 🤔 IIRC there was a reason I kept them separate, but maybe that was just dictated by the UI being two separate fields. I'll see if I can remember.

iandunn:
Added the `fieldset` and `number` fields in d9bdd2acd053c97a58c4de834ef29ad117254f4f, but I haven't tested how well that works w/ the full Schedule block yet.

I'll wait to see what happens with #303 before doing anything else, to avoid spending time on things that would need to be refactored.

ryelle:
@iandunn What is this blocked by?

iandunn:
Did you notice the comment above? Let me know if there's anything missing that you need.

ryelle:
Yes, but it's not clear from the comment what's blocking. #303 is just an experiment, no one's even replied to it yet— when this merges I'll just update it.

iandunn:
If #303 can be set to `blocked` until this is merged, then I'm happy to pick this up again. 

Otherwise I'd prefer to wait until #303 is merged before restarting this, to avoid the potential for having to redo the UI.

ryelle:
#303 does not block this though— there's no timeline on merging it, and this PR is more important. The UI in this PR is fine, the only change that still needs to happen is merging the fields into one "length" field.

iandunn:
I think we just have different opinions, but it seems like you're fine with waiting to merge #303 until this is done, so I'll go ahead and remove the `blocked` label here and add it to #303. Feel free to reverse that if I'm misunderstand your perspective.

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
iandunn: That's a great catch, thanks! Fixed in d9bdd2acd053c97a58c4de834ef29ad117254f4f

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
iandunn: I think it was just a result of the UI. I combined them in 4aa7acb286d75469cc05253ee976fb5f38ed05f4, and that seems to work fine for the purposes of the Schedule block.

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
iandunn: Done in 4aa7acb286d75469cc05253ee976fb5f38ed05f4

iandunn:
I think this is ready to merge, but could you test it w/ the changes you want to make to the Live Schedule block, to make sure it works for that as well?

iandunn:
[ The merge conflict is just b/c of cherry picking the coding standards commit to production, I'll rebase ]

ryelle (approved):
Looks good! I haven't built anything for the Live Schedule yet, but this matches what I was thinking.

I see there are still some failing phpcs checks, were you going to fix those too, or leave as-is? (not a blocker, just unclear from your comment above)

PR #300, opened 2019-12-18
Update text domain in CampTix to wordcamporg

coreymckrill:
Now that CampTix is part of the WordCamp codebase, it should share the same text domain as the rest of the code.

This makes two changes:

* Changes the text domain in every usage of a translation function from `camptix` to `wordcamporg`
* Removes the method that loads a text domain from the CampTix plugin, since that is taken care of in mu-plugins

**To test**

It would be difficult and very time consuming to try and make sure every translatable string is correctly getting translated with its new text domain. I don't think we should attempt that here. However, there were some false positives in the code when doing a search and replace for the text domain. We should make sure none of those false positives got replaced. So:

* The handle for the main JS and CSS files for the plugin is the same as the old text domain. So make sure scripts and styles are loading and working correctly. A camp's Attendees page is a good place to do this, because it loads both **camptix.js** and **camptix.css**.
* There is a global instance of the `CampTix_Plugin` class that is assigned via `$GLOBALS['camptix']`, and the variable is also occasionally referenced this way. So click around the CampTix UI and make sure there aren't any fatal errors from missing variables.


public_html/wp-content/plugins/camptix/camptix.php
```diff
@@ -1123,35 +1113,35 @@ function register_post_types() {
 
 	function register_post_statuses() {
 		register_post_status( 'cancel', array(
-			'label'                     => _x( 'Cancelled', 'post', 'camptix' ),
-			'label_count'               => _nx_noop( 'Cancelled <span class="count">(%s)</span>', 'Cancelled <span class="count">(%s)</span>', 'camptix' ),
+			'label'                     => _x( 'Cancelled', 'post', 'wordcamporg' ),
+			'label_count'               => _nx_noop( 'Cancelled <span class="count">(%s)</span>', 'Cancelled <span class="count">(%s)</span>', 'wordcamporg' ),
```
ryelle: This function call isn't right (and wasn't in the old code either) – [`_nx_noop` needs a context,](https://developer.wordpress.org/reference/functions/_nx_noop/) or we should use `_n_noop` instead. IMO, it should be [`_n_noop`.](https://developer.wordpress.org/reference/functions/_n_noop/)

(the same is true for the next few changes too)

iandunn:
🤔 

Are you seeing strings like `Click here to enter a coupon code` being localized into `fr_FR`? [It's been translated](https://translate.wordpress.org/projects/meta/wordcamp/fr/default/?filters%5Bterm%5D=click+here&filters%5Buser_login%5D=&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filter=Filter&sort%5Bby%5D=priority&sort%5Bhow%5D=desc), but it's not localizing for me. 

Most of `wp-admin > CampTix > Tools` isn't either, but some of it is (e.g., `Interface de participation`).

My `wp-content/languages/wordcamporg` folder is up to date. I'm getting an error running `lang-update.sh`, though, so the repo version may be outdated.

iandunn:
From the commit history, it looks like `languages` has been getting updated automatically, and you manually updated it last week, but `languages/wordcamporg` hasn't been updated in 9 months (manually or automatically).

ryelle (commented):
If you add the following to the end of the phpcs.xml.dist, you can use phpcs to find any textdomain issues– running that, I don't see any misses, so good job catching all of them :)

```xml
	<rule ref="WordPress.WP.I18n">
		<properties>
			<property name="text_domain" type="array" value="wordcamporg" />
		</properties>
	</rule>
```

You can run just the i18n sniffs using this: 

```bash
composer run phpcs public_html/wp-content/plugins/camptix -- --sniffs=WordPress.WP.I18n
```

It found a few places where things aren't using any text domain (including the `_nx_noop` issue), and a few that could be improved with ordered placeholders (`%1$s` vs just `%s`). It would be nice to fix those issues too.

---

Trying this locally, I'm seeing the same thing as @iandunn (though I tested with `es_ES`).

coreymckrill:
> From the commit history, it looks like languages has been getting updated automatically, and you manually updated it last week, but languages/wordcamporg hasn't been updated in 9 months (manually or automatically).

Well I'll be ding dang'd. The **lang-update.sh** script was broken, but it [should be fixed now](https://wordcamp.trac.wordpress.org/changeset/3239/bin).

`languages/wordcamporg` should be updated now, but that hasn't fixed the issue. Still investigating...

public_html/wp-content/plugins/camptix/camptix.php
coreymckrill: Fixed in 5e2032f5711fd71d22b0791dc9e97df9381fb15a

coreymckrill:
@ryelle Good suggestions for the phpcs fixes. Done in 97b214856ef17ac7b393ece8a9d966baab89668f

coreymckrill:
@iandunn @ryelle After fixing **lang-update.sh**, I am now seeing translations correctly in `wp-admin > CampTix > Tools` and elsewhere. Can you double check on your end? (Make sure you pull the latest in `wp-content/languages` first)

iandunn (approved):
WFM now :)

ryelle (approved):
Also works for me 👍

PR #106, opened 2019-05-01
Organizer Reminders: Split manual send onto a separate screen for UX.

iandunn:
This doesn't yet reflect all of the design changes, but the direction that it should go in is to make the new screen a Gutenberg block.

Supercedes #59

PR #303, opened 2019-12-19
Sessions: Gutenberg-ify Session Info & Speakers metaboxes

ryelle:
This is an attempt at switching over to using a gutenberg pattern for the session meta info (date, time, etc). It sets up a gutenberg plugin which slots into the Document Settings sidebar, and creates panels for Session Info & Speakers.

The idea is that we could phase out the metabox & meta saving functionality, and instead expose the meta via API + `register_post_meta` and use the UI patterns of Gutenberg.

### Screenshots

Screen Shot 2020-01-14 at 4 56 17 PM

1. Edit a session (either new, or existing)
2. If existing, the session info should correctly populate the new panel
3. If new, it should be empty
4. Edit the session info, save
5. It should save correctly

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
 	"workspaces": [
 		"public_html/wp-content/mu-plugins/blocks",
 		"public_html/wp-content/mu-plugins/virtual-embeds",
+		"public_html/wp-content/plugins/wc-post-types",
 		"public_html/wp-content/plugins/wordcamp-forms-to-drafts",
 		"public_html/wp-content/plugins/wordcamp-speaker-feedback"
 	],
--- a/public_html/wp-content/mu-plugins/gutenberg-tweaks.php
+++ b/public_html/wp-content/mu-plugins/gutenberg-tweaks.php
@@ -31,7 +31,7 @@ function disable_editors_by_post_type( $editors, $post_type ) {
 	 * @todo Uncomment these as the metaboxes are converted into gutenberg-native panels.
 	 */
 	$gutenberg_only = array(
-		// 'wcb_session',
+		'wcb_session',
 		// 'wcb_speaker',
 		// 'wcb_sponsor',
 		// 'wcb_organizer',
--- a/public_html/wp-content/plugins/wc-post-types/.prettierrc.js
+++ b/public_html/wp-content/plugins/wc-post-types/.prettierrc.js
@@ -1,0 +1,7 @@
+// Import the default config for core compatibility, but enable us to add some overrides as needed.
+const defaultConfig = require( '@wordpress/scripts/config/.prettierrc.js' );
+
+module.exports = {
+	...defaultConfig,
+	printWidth: 115,
+};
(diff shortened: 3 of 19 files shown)


public_html/wp-content/plugins/wc-post-types/.eslintrc.js
```diff
@@ -0,0 +1,139 @@
+/*
```
ryelle: If we use core's eslint rules instead of our own customizations, I think we can avoid adding these .eslintrc files to every sub-project… or maybe we can add it to the top-level of the project? I can look into this more.

iandunn:
Blocked per #299, which should be merged first to avoid redoing work.

iandunn:
Removed `blocked` because #299 is merged now.

coreymckrill (commented):
👍 Worked in local testing

The Session Length number fields I'm seeing aren't sized the same way as in the screenshot; they're way too big and each take up their own line in the panel. Maybe there's a build command somewhere I'm missing? (see screenshot, using Firefox)

session-length

The big conundrum I'm facing is if we could merge this given that WordCamp sites can still choose to default to the Classic Editor. Since this PR removes a lot of code necessary for the old metaboxes to work, sessions won't be editable in the Classic view anymore.

Maybe we could add one of those tutorial tool tips pointing at the Session Info panel the first time someone loads an Edit Session screen, so they know where it moved to?

iandunn:
> if we could merge this given that WordCamp sites can still choose to default to the Classic Editor.

I think it's fine for us to gradually move support for our _custom_ work to Gutenberg-only. I don't think it's worth it to support both editors for anything custom.

We've already got [code that disables G for some CPTs](https://github.com/WordPress/wordcamp.org/blob/7644a12/public_html/wp-content/mu-plugins/gutenberg-tweaks.php#L20-L37), and [an unfinished PR that expands that in the other direction](https://github.com/WordPress/wordcamp.org/pull/106/commits/622d23a6b868f9dbe8f830db5b6969df8dae9573#diff-addb2c8c6a22887efcd71302eeaaeb94R38-R42).

ryelle:
Oh, I forgot to add the build step to my testing instructions 🤦‍♀  This introduces a new project, so you need to run `yarn` & `yarn workspace wc-post-types build`. I'm not sure how you're even seeing any session fields, without the build step.

ryelle:
> The big conundrum I'm facing is if we could merge this given that WordCamp sites can still choose to default to the Classic Editor. 

I've got a branch locally to disable the classic editor on select post types, I'll push that up so we can merge it before this.

> Maybe we could add one of those tutorial tool tips pointing at the Session Info panel the first time someone loads an Edit Session screen, so they know where it moved to?

I'll look into this.

ryelle:
Looks like the tooltips were deprecated in favor of this style of guide-modal, but I don't think it'll work for us… what do you think?

Screen Shot 2020-04-07 at 12 41 48 PM

coreymckrill:
Hmm, yeah that seems like a bit much. Unless it can be a lot smaller without an image?

coreymckrill (approved):
👍 Looks good in Firefox now.

ryelle:
Rebased, moved some folders, and fixed an issue with sessions not being correctly tagged as "regular sessions". I don't think this needs a re-review, so will merge when the tests pass.

PR #304, opened 2019-12-20
CampTix Tweaks: Fix keyboard accessibility of payment options

ryelle:
This PR changes the tab markup to use buttons, which natively catch keyboard actions. The "preferred payment option" label can be a button, since the input itself is hidden, and a toggle button matches the expected behavior better. All payment options are now output in the "Payment Options" list. If there is only one payment option, it is now just output as text, not a clickable button.

This also improves the labelling of the other payment options by wrapping them in a fieldset.

Fixes https://meta.trac.wordpress.org/ticket/4421

### Screenshots

Screen Shot 2019-12-20 at 3 51 57 PM
Screen Shot 2019-12-20 at 2 18 51 PM

1. Turn on both Stripe & PayPal in the payment settings
2. Go try to buy a ticket
3. Using just a keyboard, tab through the page to the end
4. Hit space/enter on "Other payment methods", you can now select either stripe or paypal
5. You can also shift-tab back up to the "Pay with Stripe" button, and hit space/enter to flip back to that.

Basically, everything you can click with a mouse should now be activated by the keyboard too.

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
@@ -5,7 +5,7 @@
 
 namespace WordCamp\CampTix_Tweaks;
 
-defined( 'WPINC' ) or die();
+defined( 'WPINC' ) || die();
 
 use CampTix_Addon;
 
@@ -21,15 +21,15 @@ class Payment_Options extends CampTix_Addon {
 	/**
 	 * Initialize Payment_Options class
 	 */
-	function camptix_init() {
-		add_filter( 'tix_render_payment_options', array( $this, 'generate_payment_options'), 15, 4 );
+	public function camptix_init() {
+		add_filter( 'tix_render_payment_options', array( $this, 'generate_payment_options' ), 15, 4 );
 		$this->enqueue_scripts_and_styles();
 	}
 
 	/**
 	 * Enqueue styles and scripts needed for the addon to work
 	 */
-	function enqueue_scripts_and_styles() {
+	public function enqueue_scripts_and_styles() {
 		wp_register_script(
 			'payment_options',
 			plugins_url( 'js/payment-options.js', __FILE__ ),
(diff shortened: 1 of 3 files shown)


public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
```diff
@@ -201,41 +224,37 @@ private function only_one_payment_method( $payment_methods ) {
...
-		// Pre-select first payment method if stripe is not available
+		// Pre-select first payment method if stripe is not available.
 		if ( ! $this->is_stripe_available( $payment_methods ) && ! isset( $selected_payment_method ) ) {
 			$selected_payment_method = array_keys( $payment_methods )[0];
 		}
 
-		foreach ( $payment_methods as $payment_method_key => $payment_method ) {
-			if ( 'stripe' === $payment_method_key ) {
-				continue;
-			}
```
coreymckrill, line 220: Why also include Stripe in the alternatives list when it already has its own button?

coreymckrill (commented):
> If there is only one payment option, it is now just output as text, not a clickable button.

I noticed that if there are multiple payment options, but none of them are Stripe, the "Payment methods" heading is still a button, even though there is not a second one to toggle to.

public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
ryelle: Well, it's a list of all payment options, and it might not be immediately clear that you can click the stripe button again.

There's more potential for confusion for a screen reader user — reading through the list of "Payment Methods", it's more straightforward to select stripe from this list of options, vs knowing that they need to leave the list, and reverse-navigate back to the first toggle button.

ryelle:
> I noticed that if there are multiple payment options, but none of them are Stripe, the "Payment methods" heading is still a button, even though there is not a second one to toggle to.

Good catch, fixed that in  https://github.com/WordPress/wordcamp.org/pull/304/commits/3b042a1cd3ece748f356bed48370c998f792ff98

public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
coreymckrill: That makes sense. What if we changed the label to "All payment methods" or something? It just seems confusing to me to see "other" and then have "Credit Card" be listed in there again.

ryelle: "All payment methods" sounds good to me 👍

ryelle:
Updated the label for "other payment methods" to "All", so it now looks like…

Screen Shot 2019-12-20 at 5 33 01 PM
Screen Shot 2019-12-20 at 5 33 11 PM

coreymckrill (approved):

PR #305, opened 2020-01-06
Enable Live Schedule and Live Post blocks

ryelle:
This removes the site filter on Live Schedule & the live Latest Posts block, so these features are available to all sites.

This is part of opening the PWA features to all sites, and is needed before merging #241. 

1. Set `WORDCAMP_ENVIRONMENT` to `"production"`
2. Check that the Live Schedule block is still available
3. Check that Latest Posts has a "live updates" toggle setting


coreymckrill (approved):
👍 Tested

Issue #279, opened 2019-11-11
Live Schedule block: Add option to display specific track(s)


ryelle:
When embedding a Live Schedule block, I might want to only show a single track – for example, on a livestream page.

Could allow for picking multiple tracks, too?

We might want to change how we mock display the "live preview" to better reflect the chosen content.

zex2911:
This is also needed for "multilingual" WordCamps, where we have different track names for different languages.

Issue #261, opened 2019-11-01
Live Schedule block: Collapse sessions across multiple tracks into one row


ryelle:
When a sessions spans multiple tracks, like lunch, breaks, or the afterparty, could we collapse this down into one row?

Currently it repeats the session for every track:

<img width="669" alt="Screen Shot 2019-11-01 at 4 19 46 PM" src="https://user-images.githubusercontent.com/541093/68057394-bc54b100-fcc3-11e9-833a-93d405a4cac0.png">

Issue #260, opened 2019-10-31
Live Schedule block ignores sessions that aren't assigned to tracks


iandunn:
Single-track camps usually don't bother creating a formal track, and it's implied that all of the published sessions are in the "main" (only) track.

For those camps, the block will incorrectly say that there aren't any sessions happening today.

Issue #226, opened 2019-08-29
Sessions: Start times are not saved with the correct timezone


iandunn:
Currently we jump through some hoops to get the correct timezone:

https://github.com/WordPress/wordcamp.org/blob/cb99e19/public_html/wp-content/mu-plugins/theme-templates/templates/day-of-event/src/components/schedule.js#L42-L62

That's overly complicated, and I don't think it always works correctly, depending on which time zone is set.

We can simplify it to just using `gmdate()`, like the Schedule block does:

https://github.com/WordPress/wordcamp.org/blob/add/schedule-block/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js#L371

Because the value is already in the local timezone, there's no need to convert it. `gmdate()` will leave it the way it should be.

coreymckrill:
Not sure if this is relevant, since you're talking about JS, but there is a PHP timezone function coming to WP in 5.3:

https://core.trac.wordpress.org/changeset/45853

ryelle:
Well, this was   ⤵️🕳 

We currently save `_wcpt_session_time` using `strtotime` patching together the separate fields for date & time, which seems to technically save the value as that datetime UTC, regardless of the site's timezone. So if you have a site in EST, and you save a session for March 1st at 4pm, it will _actually_  save it as "March 1st 2020 16:00 **UTC**" (but as a unix timestamp), which is not actually when that talk is. In most places we don't care, because php's `date` also assumed UTC, but once we want to start working with this in JS, we need to care about timezones again.

_you can also see the issue if you switch `date` to the timezone-aware `wp_date` [on this line.](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wc-post-types/wc-post-types.php#L2047)_

This is especially a problem for the live schedule, because we need to test against `Date.now()` (or `new Date()`), which use the real current timestamp. So a talk that's on at 4pm EST would be "on now" at 4pm UTC. And all that is why we need to "account" for the site timezone here:

https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/data/index.js#L52

Because that corrects the UTC-saved date into the expected, that time in the site's timezone (so a site set to EST is corrected).

So… we should probably think about updating how we save session dates to use the correct timestamp, that correctly reflects the site's timezone, but I have no idea how much else that would impact, or how to handle back-compat…?

[...]

PR #313, opened 2020-01-08
Live Schedule: Add tests for "current session" functionality

ryelle:
This will test the `getCurrentSessions` function to make sure it returns the expected sessions at given times. This will be helpful once we start changing the Live Schedule functionality (adding session durations, tweaking the date conversion #226, other fixes like #260, #261, #279).

1. Update packages by running `yarn` (new local packages are needed to run the tests)
2. Run the tests with `yarn workspace wordcamp-blocks run test`


public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/data/tests/__mock__/session_track.json
```diff
@@ -1 +1,38 @@
...
\ No newline at end of file
+[
+  {
```
iandunn, line 2: Great idea to make this more readable 👍

iandunn (approved):
Code LGTM, and the tests pass locally 👍 

---

_Non-blocking aside:_ Should `yarn test` also work? 

Thats [in our docs](https://github.com/WordPress/wordcamp.org/blob/e6f354e/public_html/wp-content/mu-plugins/blocks/readme.md), but doesn't WFM:

```sh
> yarn test
yarn run v1.21.1
error Command "test" not found.
```

ryelle:
> Should yarn test also work?

@iandunn What directory are you in when you run that? If you're in `mu-plugins/blocks` that should work. If you're not, you need to tell yarn which workspace to run in, because we don't have a global `test` command.

iandunn:
Ah, that was probably it. Thanks!

PR #314, opened 2020-01-10
WordCamp Central: Add global sponsors "block" to homepage

### LINKED ISSUE #306: WordCamp Central: Add global sponsors widget to homepage
Replace the Latest Tweets widget on the WordCamp Central homepage with a Global Sponsors widget that shows Gold global sponsor logos. Two different ideas for presentation:

* An auto-advancing slider (every 3-5 seconds) 
* Either a random single sponsor logo or a randomized list of three every time the page is reloaded.

More info:
https://make.wordpress.org/community/2020-wordpress-global-community-sponsorship-program/

Screenshot of location for the widget:
central_hp

coreymckrill:
Swaps out the Latest Tweets column on the WordCamp Central [homepage](https://central.wordcamp.org) for a randomized list of three Gold-level sponsors. The seed for the randomize function changes every hour.

The code that outputs the sponsor list is structured similarly to WordCamp's Sponsor block, so that it would be easier to convert this into an actual block in the future.

Fixes #306

**Screenshot (Updated):**

widget-logos

~Right now we are lucky that all of the Gold-level sponsors have logos with similar aspect ratios. However, I think further design thinking is needed here about how best to accommodate one or more logos that aren't necessarily all in landscape orientation. @karmatosed~

~Also, does the heading make sense, given the link to the rest of the sponsors at the bottom? Should that link be there? @andreamiddleton~

**To test (Updated):**

* Load the WordCamp Central homepage in your sandbox.
* The widgets should already be there, I think, because I added them via my sandbox.
* Ensure that the third column below the hero contains Gold-level sponsors' logos, plus a link to the Global Sponsorship page.
* Try reloading the page a couple of times and make sure the order of the logos changes.

**Pre-deploy**

* [ ] Receive and confirm square logo assets from each of the Gold-level global sponsors. Update the widgets accordingly.


karmatosed:
Just one note, would it be possible to have 'needs design feedback' as a label if this just needs feedback? That's the flow we use on trac and other GitHub repos and might enable easier feedback.

Assuming this just needs design feedback, for me this works but as you note for now the logos fit. I don't think 'some of our sponsors' is the best copy but I will leave that to others to refine.

ryelle:
@karmatosed  Done— we now have [Needs Design Feedback](https://github.com/WordPress/wordcamp.org/labels/%5BStatus%5D%20Needs%20Design%20Feedback) & [Needs Design](https://github.com/WordPress/wordcamp.org/labels/%5BStatus%5D%20Needs%20Design) 👍

public_html/wp-content/plugins/multi-event-sponsors/views/block-sponsor.php
```diff
@@ -0,0 +1,26 @@
...
+
+use WP_Post;
+use function WordCamp\Blocks\Components\{ render_featured_image, render_item_title, render_item_content, render_item_permalink };
+
+defined( 'WPINC' ) || die();
+
+/** @var array   $attributes */
+/** @var WP_Post $sponsor */
+
+setup_postdata( $sponsor ); // This is necessary for generating an excerpt from content if the excerpt field is empty.
```
ryelle: Is this used in this block?

public_html/wp-content/plugins/multi-event-sponsors/views/block-sponsor.php
```diff
@@ -0,0 +1,26 @@
+<?php
+namespace WordCamp\Multi_Event_Sponsors\Block;
+
+use WP_Post;
+use function WordCamp\Blocks\Components\{ render_featured_image, render_item_title, render_item_content, render_item_permalink };
```
ryelle: Are all of these functions needed? I think you can just include `render_featured_image`.

public_html/wp-content/themes/wordcamp-central-2012/js/central.js
```diff
@@ -54,7 +54,7 @@ var WordCampCentral = ( function( $ ) {
 
 		try {
 			toggleNavigation();
-			populateLatestTweets();
+			//populateLatestTweets();
```
ryelle: Why not remove the whole function?

public_html/wp-content/themes/wordcamp-central-2012/template-home.php
```diff
@@ -105,7 +107,23 @@
...
+		<div class="wc-global-sponsors last">
+			<h3><strong>Some Of Our Sponsors</strong></h3>
+
+			<?php
+			echo wp_kses_post( MES_Block\render( array(
+				'number'      => 3,
+				'region_id'   => 'all',
+				'level_id'    => 3040794,
+				'image_align' => 'center',
+				'image_width' => 300,
```
ryelle: This should probably be `600` so that it displays correctly on retina screens, the logos look fuzzy on my laptop. If you do, you'll want to add `height: auto` to the CSS for the `img` itself, otherwise it skews.

public_html/wp-content/themes/wordcamp-central-2012/template-home.php
```diff
@@ -105,7 +107,23 @@
...
+			) ) );
+			?>
+
+			<a href="<?php echo esc_attr( get_permalink( get_page_by_path( 'global-community-sponsors' ) ) ); ?>" class="more">More sponsors &rarr;</a>
+		</div>  <!-- .wc-global-sponsors -->
+
+		<!--<div class="wc-tweets last">
```
ryelle: Can we delete this code, rather than commenting it out?

public_html/wp-content/themes/wordcamp-central-2012/style.css
```diff
@@ -1203,6 +1204,17 @@ li:after { content: "."; display: block; height: 0; clear: both; visibility: hid
 	padding-bottom: 20px;
 }
 
+/* !Featured Sponsor */
+.wordcamp-mes-sponsors > li {
+	width: 100%;
+	/*min-height: 300px;
```
ryelle: And here too— if we need to bring back the code, we can grab it from revision history.

public_html/wp-content/themes/wordcamp-central-2012/style.css
```diff
@@ -1203,6 +1204,17 @@ li:after { content: "."; display: block; height: 0; clear: both; visibility: hid
...
+	display: flex;
+	flex-direction: column;
+	justify-content: center;*/
+}
+.wordcamp-mes-sponsors .wordcamp-mes-sponsor__logo.align-center {
```
ryelle: I think this is redundant, since the blocks CSS is output on all sites. I see this style already applied with `.wordcamp-image__featured-image-container.align-center`.

ryelle (changes requested):
Good idea about making this the start of a block 👍 I have a few suggestions, nothing major.

I did notice that none of the logos I was looking at had alt text. I think that's something we can fix in the media library, so while these are being updated, we should make sure to do that (can probably just say the company name).

public_html/wp-content/plugins/multi-event-sponsors/views/block-sponsor.php
coreymckrill: Per [this comment](https://github.com/WordPress/wordcamp.org/issues/306#issuecomment-573198287), we're going to try using some other elements besides just the logo, but I'll be sure to remove the unused ones when the display is finalized.

public_html/wp-content/themes/wordcamp-central-2012/style.css
coreymckrill: Fair point, I can remove it all.

coreymckrill: 91c2d44b

public_html/wp-content/themes/wordcamp-central-2012/template-home.php
coreymckrill: 91c2d44b

public_html/wp-content/themes/wordcamp-central-2012/js/central.js
coreymckrill: 91c2d44b

public_html/wp-content/themes/wordcamp-central-2012/style.css
coreymckrill: Good catch. be693910

public_html/wp-content/themes/wordcamp-central-2012/template-home.php
coreymckrill: 🤔 I assumed that the output of `render_featured_image` would include a `srcset` attribute, since it's using `get_the_post_thumbnail` under the hood. But you're right that it's not doing it in this case. Maybe my assumption is wrong? Looking into it.

ryelle: yeah, that's interesting— i wonder if it's a theme-support thing, since this is based on 2010?

karmatosed:
@ryelle thanks 🙌

public_html/wp-content/themes/wordcamp-central-2012/template-home.php
coreymckrill: After some digging, I suspect it's because things go through multiple layers of `wp_kses_post`. Astonishingly, `srcset` and `sizes` are not considered valid attributes on the `img` tag with that set of allowed tags...

coreymckrill: Looks related: https://core.trac.wordpress.org/ticket/29807

coreymckrill: I guess for now we can just double the set width 😕

coreymckrill: 8f1e1c7e

andreamiddleton:
I'd like to change the title of the widget to "Gold Global Sponsors," please, and... is there a technical or design reason that it's preferable to show multiple sponsors in the widget, rather than a rotating display of one sponsor at a time? That's how we had been imagining it, but if there's a strong reason against doing it that way, I'm open to hearing about it! :)

ryelle (commented):
Code looks good now, though it sounds like the design/content might change, so I can give this another look over once it's ready to merge 👍

coreymckrill:
@andreamiddleton 

> is there a technical or design reason that it's preferable to show multiple sponsors in the widget, rather than a rotating display of one sponsor at a time? That's how we had been imagining it, but if there's a strong reason against doing it that way, I'm open to hearing about it! :)

There's some discussion about a rotating display ("slider") here: #306 

Basically, sliders/carousels are considered kind of an anti-pattern. That being said, if we're going to [try switching to just one logo at a time instead of a list of three](https://github.com/WordPress/wordcamp.org/issues/306#issuecomment-573198287) anyway, the slider functionality does have the added benefit of eventually displaying all of the gold sponsor logos on a single page load. If that's an important factor, we can explore adding the slider part. It shouldn't be hard to implement since we already have one further up on the page.

coreymckrill:
Here's what a cycling logo would look like. Not technically "sliding" since the transition is a fade, but that's the only transition we can do while using the slideshow script that's already loaded on the page for the WordCamp slideshow at the top.

logo-cycle

andreamiddleton:
That is what I asked for, and I see some problems with it, as you probably also do. :)

Is it possible to use specific files for this, rather than the logo that's a featured image on the global sponsor's listnig? We've had at least one global sponsors offer to create a special logo for this placement, with a concise call to action below the logo, and would like to offer that benefit to all Global Golds. 

Presumably if we're specifying a logo size, that will help resolve the "swimming in white space" challenge of fade/rotating logos idea?

coreymckrill:
A logo with a different aspect ratio (probably square) would definitely look better in that space.

Adding a separate input/file upload for this is not trivial, but doable. Is there anywhere else we would potentially use this alternate logo, or would it be only on the homepage? I worry that this would be adding a lot of complexity for something that was intended as just a small change.

Another alternative might be to put the Gold Sponsor logos in a row _beneath_ the three columns, that spans the entire content width. Then the landscape-oriented logos would have more space, could all be shown at once, and the height of the logo area would only need to accommodate one or two rows of logos.

coreymckrill:
Here's what it might look like with the logos beneath the columns:

global-sponsors-footer

iandunn:
Adding another uploader for this seems like overkill, personally. 

@andreamiddleton, what are you concerned about with the proposed approach?

andreamiddleton:
Two concerns are motivating me here:

The general vision had been for one sponsor to be displayed at a time, even if it's just a short amount of time, which is preferable to the sponsoring companies of course. :) 

And then, in discussion of this benefit with one global sponsor, they asked whether they could add a 3-5 word tagline/call to action under their logo for this particular case, which the global sponsorship stewards (me, Courtney, and Kevin) approved. I don't know whether they were hoping to use that logo treatment on all WordCamp sites -- I suspect not -- so we'd need a unique logo for this particular widget. 

I did not convey this information very clearly, or possibly at all, and for that I'm really sorry. :(

iandunn:
🤔 

What if we output a hidden `div` with each sponsor's existing image/text when PHP renders the page, and then use JS to toggle through them every `n` seconds?

That way only 1 would be visible at a time, but we wouldn't have to deal w/ the slideshow script. IIRC jQuery is already on the page, and could maybe provide a transition, even it's it's just a quick `hide()` and then `show()`.

( we can't randomize in PHP b/c of static page caching, not sure if that's been mentioned anywhere explicitly yet )

andreamiddleton:
That could work, I suppose, though I suspect that a sponsor would prefer to have the tagline/call to action in a font that works best with their logo, so it could get complicated to serve up that text for them. I guess I didn't realize that it would be tricky to pull the images from this widget from the media library rather than the global sponsor listing? Or maybe it wouldn't have been tricky if I had mentioned the idea beforehand. 😒

iandunn:
>  I guess I didn't realize that it would be tricky to pull the images from this widget from the media library rather than the global sponsor listing? 

Er, yeah, that's a great clarification. I think we maybe got off track thinking that this needed to be 100% automated. 

We could just upload those images manually, and manually create an HTML Widget (or an Image widget in a custom sidebar, or whatever). It's only 4 companies, and they only update once a year.

Then, we can use the JS bit to show/hide them?

andreamiddleton:
> It's only 4 companies, and they only update once a year.

Yes. 😄 It's ok if we have to do a little manual labor to set the right logos, here I think. The pool of images will naturally be small, and change very infrequently.

iandunn:
Might even be able to use this to play around w/ FSE if the Gutenberg plugin supports it.

coreymckrill:
Here's what it would look like with the column as a widget area, and each sponsor as an image widget, with a (hopefully square) logo image and a caption underneath as the short call to action.

widget-logos

iandunn:
👍 

> caption underneath as the short call to action

It sounds like [they wanted the blurb to be a part of the image](https://github.com/WordPress/wordcamp.org/pull/314#issuecomment-574881599), so that they can use specific fonts, etc.

That shouldn't change anything AFAICT, just FYI.

I guess the caption actually is nice, though, b/c it gives people the option of putting it in the image, or in text.

coreymckrill:
Marking this as "On Hold" until we receive all the correct logo assets. In the mean time, it can still be reviewed.

public_html/wp-content/themes/wordcamp-central-2012/style.css
```diff
@@ -1100,6 +1100,7 @@ li:after { content: "."; display: block; height: 0; clear: both; visibility: hid
 	margin-bottom: 9px;
 }
 #wc-content-blocks .wc-news h3 { background-position: 0 -71px }
+#wc-content-blocks .wc-global-sponsors h3 { background-position: 0 -147px; }
 #wc-content-blocks .wc-tweets h3 {
```
ryelle: There are still a few styles related to twitter (`wc-tweet`) in here, those could also be removed

ryelle:
Looks good on desktop, but smaller screens are a little janky.

When it switches to a single column (660px), the images overlap the footer:

Screen Shot 2020-01-17 at 9 36 05 AM

And when the columns are less than 300px (screen width about 960px), the logos get cut off on the right:

Screen Shot 2020-01-17 at 9 47 04 AM

public_html/wp-content/themes/wordcamp-central-2012/style.css
coreymckrill: Good catch. Fixed with 09cade7f

coreymckrill:
Ok, the small screen issues should be fixed with 15e5700a

ryelle (approved):

Issue #129, opened 2019-06-21
Finish service worker caching


iandunn:
See the `@todo` notes in `mu/-plugins/service-worker-caching.php` and this:

* -[ ] Rather than pulling down fresh data every time, even when it hasn't changed, can we detect when it's changed and only pull new data then? IIRC service workers provide a way to do that. See https://github.com/wceu/wordcamp-pwa-page/issues/6#issuecomment-499366120

This can be used as a tracking issue, but we'll probably want lots of small PRs, and maybe a few smaller issues as well. One giant PR would be difficult to review.

vdwijngaert:
This should be pretty straight forward using the stale while revalidate strategy. If the right http headers are used (and exposes for cors use), this should reduce server load and client side traffic quite a bit!

westonruter:
Yes, see also https://github.com/wceu/wordcamp-pwa-page/issues/6#issuecomment-499366120:

> The way to do this would be to send an `ETag` in the REST API response, and then when making  the request include the value in an `If-None-Match` request header. The REST API controller should then compare the value in the `If-None-Match` request header with the the `ETag` that would be sent, and if they are the same, send a `304 Not Modified` response. This would prevent wasting client data pulling down the same data that has not changed. 
>
> A good `ETag` could be the max `post_modified` date of whatever is being queried.

ryelle:
Responding to some of the comments from that file so I can get a better handle on the tasks:

> prompt to install app to home screen is automatically showing on mobile

According to [the criteria for showing the install badge,](https://developers.google.com/web/fundamentals/app-install-banners/#criteria) this is because we've added a manifest. I think it's fine– it's just like a "bookmark" to the site while you're online, and of course if you're offline we want the UX to be better but we're improving that now.

> also, says "install wordcamp" instead of "install wordcamp europe" - maybe manifest issue?

Created an issue for this: #196

> after adding to home screen, there's a delay and a "wordcamp" interstitial screen before the site appears.

That's [the splash screen](https://developers.google.com/web/fundamentals/web-app-manifest/#splash-screen) that loads while the browser is booting. We can customize the icon, name, and background color. The background color comes from theme settings, a setting called `theme-color` (no theme supports) and then the custom-background theme setting (which only CampSite & Twenty Ten support) - so we might want to add in a customizer setting on Site Identity to control that. Created an issue for this: #197

> experience on mid-level phone (moto e4) and fast bandwidth is pretty poor

Created issue: #198

[...]

PR #315, opened 2020-01-10
PWA: Only allow the install bar when the WordCamp is approaching/running

ryelle:
This prevents Chrome from displaying the mini-infobar ("Add [WordCamp] to your home screen") when the event is more than 2 weeks away. This is only added to sites that have the PWA plugin active.

See #129 for discussion around removing vs limiting the banner.

As far as I can tell, the banner only shows up on Android + Chrome, so to functionally test, you need an Android & to proxy your sandbox so you can access it from mobile. I used Charles for this, but you could also just verify that the code to prevent the banner appears when expected (2 weeks before until 1 week after the start date of the camp, for any site that has the PWA plugin active).


coreymckrill (approved):
Verified locally that the code appears when camp start date is outside of the window 👍

PR #316, opened 2020-01-10
Theme Templates: Fix display issues of Offline & Day of Event templates across themes

### LINKED ISSUE #205: Offline Template: Test with all themes

### LINKED ISSUE #206: Day of Event Template: Test with all themes
Our supported themes are:

- [x] CampSite 2017
- [x] Twenty Thirteen
- [x] Twenty Sixteen
- [x] Twenty Seventeen
- [x] Twenty Nineteen
- [x] Twenty Twenty
- [x] WordCamp Base

---

The rest we don't test with, but will fix issues on

- Twenty Ten
- Twenty Eleven
- Twenty Twelve
- Twenty Fourteen
- Twenty Fifteen
- WordCamp Base Redux

ryelle:
The templates on some default themes use different markup, which causes display issues when trying to use the same templates across themes. To fix this, we should stop trying to fully emulate all themes. This PR removes the global Day of Event template, since the Day of Event stub is created and can use a regular page template. The offline page is called differently, but instead of trying to match the theme directly, we can create a very simple page, which inherits the basic style of the active theme.

Fixes #205, fixes #206 

Twenty Thirteen
twentythirteen

Twenty Fifteen
twentyfifteen

Twenty Seventeen
twentyseventeen

Twenty Twenty
twentytwenty

Day of Event:

1. Before switching to this branch, set a page to have the Day of Event template
2. Switch to the branch
3. Check that the page from step 1 still displays correctly

Offline:

1. Make sure the PWA plugin is active
2. View the offline page: `yoursite.wordcamp.test/?wp_error_template=offline`
3. Try with multiple themes, it should not look broken (just simple, see screenshots for example)


coreymckrill (dismissed):
Tested locally:

* Twenty Nineteen's layout looks broken, probably what you're describing as "unexpectedly full width"
* Twenty Fourteen doesn't have the full width problem per se, but it has some margin issues:

twentyfourteen

ryelle:
Okay, I've removed the Day of Event template, and tweaked the offline template so that it's very, very simple. I'll update the description to reflect the actual PR now.

coreymckrill (commented):
Day of Event: 👍 while testing locally

Offline: It looks good while testing locally on everything except Twenty Twelve (which, maybe we don't care about so much at this point?)

twenty-twelve

ryelle:
> It looks good while testing locally on everything except Twenty Twelve 

I think that's fine, only 3 sites in my "2019+" dataset use Twenty Twelve… and it's not totally unreadable, just not great.

PR #319, opened 2020-01-15
Live Schedule: Update API endpoint and call to reduce bandwidth and TTFB

### LINKED ISSUE #207: PWA: Investigate API requests & caching

ryelle:
Currently, the `sessions` endpoint used to fetch sessions for the Live Schedule needs to request the `_embed` version of the endpoint, which causes a much slower response (around 10s on 2019.us.wordcamp.org). We needed to do this to get the speaker names and category labels, but the rest of the data it outputs is unnecessary.

This PR pre-computes speakers and category names, and adds them to the API, so the request can be made for the default endpoint instead. It also restricts the data requested for the live schedule to only the fields needed for display, which drops the size of content downloaded significantly.

Fixes #207 by improving API size/response time, API responses are still currently being cached by the SW in the `rest-api` bucket.

**Metrics:**

I measured the TTFB and download size for 3 sites (2019.us.wc.org, 2017.testing.wc.org, and a local site on the docker setup). First is the current request, which is both large (size) and slow (time), second is just the pre-computed content, which cuts all the requests to just 2 seconds. The third data point is the field subset request, which reduces the filesize down to less than 10% of the current response.

Screen Shot 2020-01-15 at 2 05 46 PM

1. Query the sessions endpoint, and make sure `session_speakers` shows all selected (published) speakers.
2. Query the sessions endpoint, and make sure `session_cats_rendered` shows all selected categories 
3. Load up a site with the Live Schedule block, it should display as expected (no JS errors)

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/components/session.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/components/session.js
@@ -16,13 +16,13 @@ export default function( { headingLevel = 3, session, track } ) {
 	const Heading = `h${ headingLevel }`;
 
 	// Pull details out of the session object.
-	const { link, terms = {} } = session;
 	const title = get( session, 'title.rendered', '' );
+	const link = get( session, 'link', '' );
+	const categories = get( session, 'session_cats_rendered', '' );
 	const type = get( session, 'meta._wcpt_session_type', '' );
-	const categories = session.session_category || [];
 	const time = get( session, 'session_date_time.time', '' );
 
-	const speakers = get( session, '_embedded.speakers', [] );
+	const speakers = get( session, 'session_speakers', [] );
 	const validSpeakers = speakers.filter( ( speaker ) => !! speaker.id );
 
 	const cleanTitle = decodeEntities( stripTags( title ) );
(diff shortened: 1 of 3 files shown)


coreymckrill (approved):
👍 Tested locally

PR #323, opened 2020-01-20
Application tracker: improve log section notes

### LINKED ISSUE #322: Application tracker: improve log section notes
[Initially reported on Trac](https://meta.trac.wordpress.org/ticket/3839). Even tho Trac ticket is marked as closed, there are some things that are not working as expected.

- `wpautop` does not add paragraphs because it's output is runned thru too strict `wp_kses_post`
- Hugh aksed if we could automatically add a-tag for links in the note message

timiwahalahti:
Changeset [8080](https://meta.trac.wordpress.org/changeset/8080) made improvements to log section view but did not fully complete the changes requested.

In the changeset `wp_kses_post` uses ruleset `data` which is too strict and does not allow `p` or `br` tags so in the end result of using `wpautop` reverts. Let's allow those tags to allowed html and really use `wpautop`. This make reading of long vetting notes much nicer.

Also linkify links in notes for slightly better user experience. Add `target="blank"` to ensure not losing application view and `rel=nofollow` so search engines don't use link for ranking, crawling or indexing.

Fixes #322 

1. Write a note to application with one of two paragraphs and links
2. Note should be outputted nicely instead of one long chunck of text

### DIFF
--- a/public_html/wp-content/plugins/wcpt/views/common/metabox-log.php
+++ b/public_html/wp-content/plugins/wcpt/views/common/metabox-log.php
@@ -1,4 +1,29 @@
-<?php defined( 'WPINC' ) || die(); ?>
+<?php defined( 'WPINC' ) || die();
+
+/**
+ * Slightly modify allowed html tags in entry message because default
+ * "data" set is too restrictive and "post" set too broad.
+ */
+$allowed_html = wp_kses_allowed_html( 'data' );
+$allowed_html['p'] = array(); // allow paragraph for easier reading
+$allowed_html['br'] = array(); // allow line changes for easier reading
+$allowed_html['a']['target'] = true; // target to ensure not losing application view
+$allowed_html['a']['rel'] = true; // we want to set nofollow so search engines don't use link for ranking, crawling or indexing
+
+/**
+ * Loop all entries and search for notes. In notes, turn plaintext links
+ * to linkified html links for easier usage. Add always target="_blank" and
+ * rel="nofollow".
+ */
+if ( $entries ) {
+	foreach ( $entries as $entry_key => $entry ) {
+		if ( 'note' !== $entry['type'] ) {
+			continue;
+		}
+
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wcpt/views/common/metabox-log.php
```diff
@@ -15,10 +40,10 @@
 
 			<?php foreach ( $entries as $entry ) : ?>
 				<tr class="<?php echo esc_attr( str_replace( '_', '-', $entry['type'] ) ); ?>">
-					<th><p><?php echo esc_html( date( 'Y-m-d h:ia', $entry['timestamp'] ) ); ?></p></th>
+					<th><p><?php echo esc_html( gmdate( 'Y-m-d h:ia', $entry['timestamp'] ) ); ?></p></th>
 					<th><p><?php echo esc_html( ucwords( str_replace( '_', ' ', $entry['type'] ) ) ); ?></p></th>
 					<th><p><?php echo esc_html( $entry['user_display_name'] ); ?></p></th>
-					<th><?php echo wp_kses( wpautop( $entry['message'] ), wp_kses_allowed_html( 'data' ) ); ?></th>
+					<th><?php echo wp_kses( wpautop( $entry['message'] ), $allowed_html ); ?></th>
```
ryelle: Could you use [`make_clickable`](https://developer.wordpress.org/reference/functions/make_clickable/) here, so that you don't need to create a new regex above? It also handles more edge-cases (for example, I added punctuation to my link when testing, and it considered that part of the link, ex `https://wordpress.org.`)

ryelle (changes requested):
Thanks for jumping in and fixing this! It's working as expected, though I have one suggestion about the link code.

public_html/wp-content/plugins/wcpt/views/common/metabox-log.php
timiwahalahti: Wow, wasn't aware of that kind of function! Will definitely to change that to core way.

ryelle (approved):
Sorry I missed your update here. This looks good to merge now 👍

PR #324, opened 2020-01-21
Block Live Schedule: Use duration to find currently running sessions

ryelle:
In #299, a new field was added for session duration. We can use this to accurately get the running sessions. If a session does not have a duration, we default to `WordCamp_Post_Types_Plugin::SESSION_DEFAULT_DURATION`, 50 minutes.

Currently, the Live Schedule block does not correctly display the last session of the day, because it uses the following session to gauge what is "now." Since we have duration now, the logic for sorting and selecting "current sessions" has been updated to fetch the session that starts before now, and ends after now.

The tests have also been updated to be a bit more accurate, and the test mock data has been trimmed down to one day of sessions.

### Screenshots

In this screenshot, the "afterparty" session is happening "now", with nothing after, so we hide the "Up Next" header.

Screen Shot 2020-01-21 at 11 11 44 AM

1. Check that the tests pass
2. Set up some sessions to be running "now", or add `'nowOverride' => strtotime( 'your time' ) * 1000,` to the live schedule controller (in `add_script_data`).
3. View the live schedule and check that it is showing the right sessions

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/components/schedule.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/components/schedule.js
@@ -24,9 +24,14 @@ export default function( { attributes, isFetching, sessions } ) {
 	const { level = 2 } = attributes;
 	const Heading = `h${ level }`;
 
+	const hasNow = !! sessions.filter( ( { now } ) => !! now ).length;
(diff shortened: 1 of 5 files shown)


coreymckrill (approved):
👍 Tests pass
👍 Shows correct sessions with several different values in `nowOverride`

PR #325, opened 2020-01-21
Block Live Schedule: Display sessions with no tracks

### LINKED ISSUE #260: Live Schedule block ignores sessions that aren't assigned to tracks

ryelle:
The Live Schedule block uses tracks to display the currently running sessions, which does not work if a WordCamp has no tracks. In this case, we can assume it's a one track camp and show all sessions in one default track. 

Fixes #260.

### Screenshots

Since there is no track, we also hide the "track" cell in the display

Screen Shot 2020-01-21 at 10 44 58 AM

1. Check that the tests pass
2. On a site with no tracks, add some sessions
3. Optionally use `'nowOverride' => strtotime( 'your time' ) * 1000`, in the live schedule controller (in add_script_data), to set a "now"
4. View the live schedule and check that it is showing the right sessions

⚠️  _Note_ that this won't work on a site with tracks, even if you don't assign them to sessions… that could be a future fix if it turns out to be a problem.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/components/session.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/components/session.js
@@ -29,7 +29,9 @@ export default function( { headingLevel = 3, session, track } ) {
 
 	return (
 		<div className={ `wordcamp-live-schedule__session type-${ type }` }>
-			<span className={ `wordcamp-live-schedule__session-track track-${ track.slug }` }>{ track.name }</span>
+			{ !! track.slug && (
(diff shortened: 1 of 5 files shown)


coreymckrill (approved):
👍 Tests pass locally
👍 Live schedule block is showing the right sessions when I have one current and one next

Probably not a big deal, since this is an edge case and not intended use, but I noticed that if you have two sessions at the same time (and no tracks), the block only displays one of them.

Issue #321, opened 2020-01-20
Changes to the Meetup organizer application form 


timiwahalahti:
_[Moved from Trac](https://meta.trac.wordpress.org/ticket/4177)._

 Improving the Meetup organizer application form was proposed and discussed in the community P2. As a result of the discussion, the following changes to the form are agreed on and needs to be done at some point. I'm opening this trac ticket, because I guess the new meetup status tracker should be taken into account when making changes to the form.

Questions to add to the form:
- Could you introduce yourself and tell us more about your connection with WordPress?
- Would you mind sharing your blog with us, and links to your Twitter, Facebook and Instagram accounts?
- Why do you want to start a WordPress Meetup in your city? And what are your immediate plans?
- Have you talked with people in your community to check the level of interest?
- Anything else you want us to know while we're looking over your application? 

Fields to remove from the form:
- How would you describe yourself? (connection with WP responds to this)
- Anything you’d like to tell us about yourself, or what you hope to do with a meetup group?

For social media accounts, we have one textarea-field instead of locked fields for each platform. Label for the field contains information about what links we would like to see and it is easily changed, unlike fields which need coding the logic every time platforms change.

timiwahalahti:
_Moved from Trac._

At the same time while doing these changes to the form, would it be possible to also require wordpress.org login or do some kind of check that entered username is correct? That would also help much while vetting applications.

timiwahalahti:
_Moved from Trac._

Ian: consensus from latest Slack discussion:
- require login
- show the form, but disable it, and show a log-in modal on top of it
- reuse the speaker application code for the above, after refactoring it to be generic

timiwahalahti:
FYI: I started on working with this.

PR #333, opened 2020-01-26
Changes to the Meetup organizer application form and generic WordPress.org login required functionality

timiwahalahti:
This PR contains Meetup organizer application changes as per #321.

It also contains refactored WordPress.org login prompt. Code for that is based on functionality that was previously in `wordcamp-forms-to-drafts` plugin and bound only to the speaker submission form. PR introduces new plugin `wordcamp-forms-worg-login-requirement` that drives the same thing but all content behind a login is configured via filter. The plugin should be network activated in order to work.

Currently, this PR enables filters to speaker submission and meetup organizer application forms. For logged-in users, the prompt is never shown.

The only downside with the way how this plugin works is if `the_content` has multiple forms. Then things can break. But we can't detect reliably the forms otherwise than using `<form` tag if we don't make this with a javascript that has its own downsides.

Also slight style change on the prompt for better readability.

### Screenshots

**Old prompt**

**New prompt**

1. Network activate the `WordCamp Forms WordPress.org login requirement` plugin
2. Go to https://central.wordcamp.test/meetup-organizer-application in private browsing and you should see a login prompt
3. Log-in to wordcamp.test
4. Now visit the meetup organizer application page again and form should be visible with few fields prefilled

To test the speaker submission form:

2. Create a new testing WordCamp site
3. Publish the call for speakers post
4. Visit the call for speakers post in private browsing and you should see a login prompt
5. Log-in to wordcamp.test
5. Now you should see the form

### DIFF
--- a/public_html/wp-content/plugins/wcpt/views/applications/meetup/shortcode-application.php
+++ b/public_html/wp-content/plugins/wcpt/views/applications/meetup/shortcode-application.php
@@ -122,6 +122,14 @@ function render_meetup_application_form( $countries ) {
 			<div class="PDF_questionDivide"></div>
 			<div class="PDF_question">
 				<label>
+					Why do you want to start a WordPress Meetup in your city? And what are your immediate plans?
+					<span class="required-indicator">(required)</span>
+					<textarea name="q_reasons_plans" rows="5" required ></textarea>
+				</label>
+			</div>
+			<div class="PDF_questionDivide"></div>
+			<div class="PDF_question">
+				<label>
 					Your <a href="https://wordpress.org" target="_blank">WordPress.org</a> username
 					<span class="required-indicator">(required)</span>
 					<input type="text" name="q_wporg_username" required/>
--- a/public_html/wp-content/plugins/wcpt/views/applications/meetup/shortcode-application.php
+++ b/public_html/wp-content/plugins/wcpt/views/applications/meetup/shortcode-application.php
@@ -105,21 +105,6 @@ function render_meetup_application_form( $countries ) {
 			<div class="PDF_questionDivide"></div>
 			<div class="PDF_question">
 				<label>
-					How would you describe yourself?
-					<span class="required-indicator">(required)</span>
-
-					<select name="q_describe_yourself" required>
-						<option value=""></option>
-						<option>WordPress professional</option>
-						<option>Current WordPress user or developer</option>
-						<option>New to WordPress</option>
-						<option>I don't use WordPress</option>
-					</select>
-				</label>
-			</div>
-			<div class="PDF_questionDivide"></div>
-			<div class="PDF_question">
-				<label>
 					Your <a href="https://wordpress.org" target="_blank">WordPress.org</a> username
 					<span class="required-indicator">(required)</span>
 					<input type="text" name="q_wporg_username" required/>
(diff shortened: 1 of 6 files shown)


timiwahalahti:
😅 Will fix all the coding standards fails if this PR and the approach I've taken look good.

ryelle:
I've taken a quick look here, and I have a few bits of feedback. 

If we're going to require wp.org login on more forms, then pulling that functionality out as a new plugin makes sense; but I think this should be added as an `mu-plugin`, so that it can't be accidentally deactivated.

Overall this approach to the issue seems pretty fragile, but I'd need to investigate further to see if there's an alternative.

In the future, these two features would be easier to review as separate PRs — one for the form login fix, and one for the meetup form changes. In fact, if you want the meetup form questions to be updated faster, you could pull those changes out into a new PR, while we figure out the best approach here.

timiwahalahti:
> If we're going to require wp.org login on more forms, then pulling that functionality out as a new plugin makes sense; but I think this should be added as an mu-plugin, so that it can't be accidentally deactivated.

👍 This was the consensus in one Slack chat long while ago (see [this comment](https://github.com/WordPress/wordcamp.org/issues/321#issuecomment-576403030) and [Slack](https://wordpress.slack.com/archives/meta-wordcamp/p1558026060222100)).

> Overall this approach to the issue seems pretty fragile, but I'd need to investigate further to see if there's an alternative.

Totally agree. I did try to find the best solution here for making the plugin easy to use in the future for other forms (WordCamp organizer application for example) but at the same time as specific as possible. The issue here is that forms are rendered in to to the `the_content` instead of them having A) their own shortcode B) totally isolated template files.

If we'd have each form as own template file, that would solve the fragility of the current approach. In that scenario, the plugin would just have one or two functions to call in the template to achieve the login prompt. For example `prefix_needs_worg_login` and `prefix_get_wcorg_login_prompt( $custom_message )`.

Moving the forms to own files on central.wordcamp.org isn't a problem, but moving speaker application form is because those are handled very differently with Jetpack.

> In fact, if you want the meetup form questions to be updated faster, you could pull those changes out into a new PR, while we figure out the best approach here.

Good point - will do!

coreymckrill:
@timiwahalahti just checking on this. What is needed to move this forward? It seems like the required login part of this would be really useful for mitigating application spam.

timiwahalahti:
@coreymckrill If we speak about the WordPress.org login requirement for WordCamp and Meetup application forms, it's stuck to finding if there is a less fragile way of adding the login prompt. Now it's injected in quite an ugly way (see https://github.com/WordPress/wordcamp.org/pull/333/commits/a2903efc239dd4f91d1e036740d40499be096b66).

If the application login prompt does not need to take the speaker application form into consideration, which is IIRC block nowadays and the issue is fixed there, I'd probably rather do custom WordCamp.org page template which does the same in a prettier way. If agreed, I can start new PR for that.

coreymckrill:
@timiwahalahti I personally think the best approach here would be to start some new PRs that separate out the different changes attempted here:

* One for the wording changes in the Meetup application
* One to add a wp.org login requirement for WordCamp and Meetup application forms
* One to tweak the wp.org login overlay for the forms-to-drafts plugin (but honestly this is a lower priority)

timiwahalahti:
> One for the wording changes in the Meetup application

This has been actually done already.

> 

> - One to add a wp.org login requirement for WordCamp and Meetup application forms
> - One to tweak the wp.org login overlay for the forms-to-drafts plugin (but honestly this is a lower priority)

Will do! Maybe not this week, but after WCEU asap. If WCEU CD goes well, I might have some time during the day to move this forward.

timiwahalahti:
@coreymckrill See #523.

Closing this PR as most of the things are already covered in other PR's. Let check the login prompt styiling another time in another PR.

PR #348, opened 2020-01-29
Sessions: Save & display date using the correct timezone

### LINKED ISSUE #226: Sessions: Start times are not saved with the correct timezone

ryelle:
This PR updates session date (`_wcpt_session_time`) displays to output a timezone-aware date. It also replaces `strtotime` with `date_create` when saving the timestamp, to generate the correct timestamp value. Lastly, it removes the (now unnecessary) timezone computation from the Live Schedule block.

Fixes #226 

This will need to be merged right before running the session time fixer script (see https://github.com/WordPress/wordcamp.org/issues/226#issuecomment-577916924), but this means sites will display the wrong time until all are fixed. On the other hand, waiting until after will have the sites that get fixed displaying incorrectly as the script runs.  Thoughts❓

Edit: I think the script will take about 10-15min to run, so it's not a large window either way.

1. Set a timezone on your site, if not done already
2. Add/edit some sessions with times
3. Check that the right value is saved:
    - `wp post meta get <id> _wcpt_session_time`
    - Go to `https://www.unixtimeconverter.io/[value]`
    - It should show you the timestamp value in UTC & your timezone, unless you set the site timezone to UTC+0, the UTC version _should not_ be the session time
4. Check that the date displays correctly…
    - in the datepicker for the session
    - in the session list table
    - in the schedule shortcode
    - anywhere else a human-readable date is output
5. Check that the live schedule is calculating correctly (see #324 for LS test instructions)

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/data/index.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/data/index.js
@@ -6,7 +6,6 @@ import { findLastIndex, reverse, sortBy } from 'lodash';
 /**
  * WordPress dependencies
  */
-import { __experimentalGetSettings } from '@wordpress/date';
 import { addQueryArgs } from '@wordpress/url';
 import apiFetch from '@wordpress/api-fetch';
 
@@ -52,7 +51,6 @@ function fetchFromAPI() {
  * @return {Array} A list of objects, `{track, now, next}`.
  */
 export function getCurrentSessions( { sessions, tracks } ) {
-	const tzOffset = __experimentalGetSettings().timezone.offset * ( 60 * 60 * 1000 );
 	const nowTimestamp = window.WordCampBlocks[ 'live-schedule' ].nowOverride || Date.now();
 
 	const trackListWithSessions = tracks.length ?
@@ -82,7 +80,7 @@ export function getCurrentSessions( { sessions, tracks } ) {
 		// Check if we're more than a day out from the earliest session (so that we don't show "up next" before
 		// the WordCamp starts).
 		const firstSession = sessionsInTrack[ sessionsInTrack.length - 1 ];
-		const firstSessionTimestamp = ( firstSession.meta._wcpt_session_time * 1000 ) - tzOffset;
+		const firstSessionTimestamp = firstSession.meta._wcpt_session_time * 1000;
 		const dayInMiliseconds = 24 * 60 * 60 * 1000;
 		if ( nowTimestamp < ( firstSessionTimestamp - dayInMiliseconds ) ) {
 			return {};
@@ -90,7 +88,7 @@ export function getCurrentSessions( { sessions, tracks } ) {
 
 		const index = sessionsInTrack.findIndex( ( { meta } ) => {
 			const duration = ( meta._wcpt_session_duration || window.WordCampBlocks[ 'live-schedule' ].fallbackDuration ) * 1000;
-			const startTimestamp = ( meta._wcpt_session_time * 1000 ) - tzOffset;
+			const startTimestamp = meta._wcpt_session_time * 1000;
 			const endTimestamp = startTimestamp + duration;
 
 			// Start time before now, end time after now.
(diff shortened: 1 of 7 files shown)


coreymckrill (approved):
👍 All the tests worked on my local environment... after I remembered to rebuild the `wordcamp-blocks` scripts 🤦‍♂ 

> I think the script will take about 10-15min to run, so it's not a large window either way.

Yeah, as long as we do it on a day when there aren't any WordCamps happening, it shouldn't be a big deal.

iandunn:
Great effort here so far, kudos 👍 

How will these external clients be affected?

* the Events Widget
* make.w.org/community/events (probably the same as ^, but worth checking)
* https://github.com/wordpress-mobile/WordCamp-Android
* https://github.com/ctarda/WordCamp (probably the same as ^, but worth checking)

At first glance it seems like this would break back-compat with all those API clients, is that right? If so, a safer approach might be to:

* keep the existing code and values, but treat `_wcpt_session_time` as deprecated
* save the new values as something like `_wcpt_session_time_with_tz`
* update all wordcamp.org code to use the new field
* update api.w.org/events to also serve `_wcpt_session_time_with_tz`, but continue serving `_wcpt_session_time`
* if there are any Events Widget bugs related to TZ, then update it to use `_wcpt_session_time_with_tz`. older versions would still always need `_wcpt_session_time` to remain in the API though.
* add lots of documentation in code, api responses, written docs, etc that `_wcpt_session_time` is deprecated and buggy

ryelle:
> How will these external clients be affected?

Most of those clients don't use **session times.** The events widget & make.w.org page only seem to be showing times for meetup.com events, not individual WordCamp sessions. The iOS WordCamp app doesn't show session info. It _would_ affect the android app— but it seems [they're still on the v1 endpoints,](https://github.com/wordpress-mobile/WordCamp-Android/blob/9a7f60b96017114a02d6dcca264855513713229e/app/src/main/java/org/wordcamp/android/networking/WPAPIClient.java#L16-L24) so I don't know if we need to guarantee compatibility. On the other hand, I'm fine with dropping in a filter so that v1 requests for `_wcpt_session_time` are actually served `_wcpt_session_time_utc`.

iandunn:
You're right, those were bad examples. It seems like back-compat is still an issue, though?


_( Note for our future selves: If we do [make similar changes for the WCPT start date/time](https://github.com/WordPress/wordcamp.org/issues/226#issuecomment-580368588) then it seems like that would affect the Events Widget, etc, since in some cases the date would be shifted by a day. )_



> _wcpt_session_time_utc

That's a much better name 👍

ryelle:
I dug a little more and was unable to find an active project that uses the sessions API (using github's global code search). I'll acknowledge that there might still be projects that use this value, but I think the added technical debt of defaulting to the incorrect value outweighs any potential impact of changing this. So, here's what I'm planning on doing:

- Update the v1 endpoint to continue using the incorrect value, calculated from the real value. This way, the WordCamp Android app will continue to work.
- Add a parameter to the v2 endpoint to switch into “UTC mode”, which will return the incorrect value, again calculated from the real value — so anyone using the v2 endpoint can toggle this switch on while they fix their code.
- Draft something summarizing all this to post on make/meta
- Post that, then wait a few days for anyone to raise issues (can bring it up in the bug scrub on Thurs, too)
- Unless someone brings up a new issue, merge this PR & run the script (probably not until the 10th)

coreymckrill:
This sounds like a reasonable approach to me. This way we're not "silently" changing the API and possibly breaking backcompat for someone.

iandunn:
I definitely agree that adding more technical debt is undesirable. We have a lot of it already, and avoiding/minimizing it is important.

I'm still really concerned that this would break things that others have built, though. I'm not confident that a GitHub search would find enough projects.

Do you think there's a way that plan can be tweaked to avoid breaking back-compat, while still minimizing the technical debt?

Issue #340, opened 2020-01-27
Speaker Feedback: Add permissions-checks to feedback view


ryelle:
Speakers should only be able to see their own feedback, so the comments will need to be filtered such that anyone not an admin on the site can only view feedbacks on sessions where they're selected as the speaker.

Once this is done, organizers will only have to send out a single URL to speakers, since the main comments view should only show their personal feedback.

coreymckrill:
This also needs to include ensuring that feedback comments are not included in responses to `GET` requests to the standard comments REST endpoint.

PR #350, opened 2020-01-30
Speaker Feedback: Add base plugin file and comment type functions

### LINKED ISSUE #338: Speaker Feedback: Create custom comment type
Create the custom comment type for saving feedback data. This will be attached to the session post type.

coreymckrill:
Gotta start somewhere.

This scaffolds the main plugin file and creates CRUD functions for managing comments of our custom type.

There isn't a function for the "D" in "CRUD" because `wp_delete_comment` is sufficient, and needs no wrapper method.

Fixes #338 

**To test**

1. Activate the plugin on a site in your local dev environment
1. Use `wp shell` to try running the CRUD functions:
  * `\WordCamp\SpeakerFeedback\Comment\get_feedback()` (do this one first to see that there aren't any submissions yet, then again after adding a submission)
  * `\WordCamp\SpeakerFeedback\Comment\add_feedback( {post ID}, array( 'name' => 'Foo', 'email' => 'foo@example.org' ), array( 'bar' => 'baz' ) )`
  * `\WordCamp\SpeakerFeedback\Comment\update_feedback( {comment ID}, array( 'bar' => 'bop', 'fizz' => 'bang' ) )`
  * `\WordCamp\SpeakerFeedback\Comment\delete_feedback( {comment ID} )`
1. To check that comment meta is getting added/updated, try accessing them as properties on the returned comment objects:
  * `$test = \WordCamp\SpeakerFeedback\Comment\get_feedback()`
  * `$test[0]->bar`

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
@@ -91,3 +91,14 @@ function get_feedback( array $status = array( 'hold', 'approve' ), array $post__
 
 	return $feedback;
 }
+
+/**
+ * Trash a feedback submission.
+ *
+ * @param int $comment_id The ID of the comment to delete.
+ *
+ * @return bool
+ */
+function delete_feedback( $comment_id ) {
+	return wp_delete_comment( $comment_id );
+}
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
@@ -7,7 +7,7 @@ use WordCamp\SpeakerFeedback\Feedback;
 
 defined( 'WPINC' ) || die();
 
-const COMMENT_TYPE = 'speaker-feedback';
+const COMMENT_TYPE = 'wordcamp-speaker-feedback';
 
 /**
  * Add a new feedback submission.
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
@@ -11,9 +11,9 @@ const COMMENT_TYPE = 'speaker-feedback';
 /**
  * Add a new feedback submission.
  *
- * @param int           $post_id
- * @param array|WP_User $feedback_author
- * @param array         $feedback_meta
+ * @param int       $post_id         The ID of the post to attach the feedback to.
+ * @param array|int $feedback_author Either an array containing 'name' and 'email' values, or a user ID.
+ * @param array     $feedback_meta   An associative array of key => value pairs.
  *
  * @return false|int
  */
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
```diff
@@ -0,0 +1,93 @@
...
+	if ( ! empty( $meta_query ) ) {
+		$args['meta_query'] = $meta_query;
+	}
+
+	$feedback = get_comments( $args );
+
+	// This makes loading meta values for comments much faster.
+	wp_queue_comments_for_comment_meta_lazyload( $feedback );
+
+	return $feedback;
```
ryelle: Can this also return the comment meta? That'll prevent an extra step for us later.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
```diff
@@ -0,0 +1,93 @@
...
+ * @param array     $feedback_meta   An associative array of key => value pairs.
+ *
+ * @return false|int
+ */
+function add_feedback( $post_id, $feedback_author, array $feedback_meta ) {
+	$args = array(
+		'comment_approved' => 0, // "hold".
+		'comment_post_ID'  => $post_id,
+		'comment_type'     => COMMENT_TYPE,
+		'comment_meta'     => $feedback_meta,
```
ryelle, line 26: Maybe when we figure out what the meta we need is, we can add an abstraction here to sanitize/format & prefix the meta keys & values… (nothing to do now, just thinking out loud)

ryelle (commented):
Works as expected, and these "comments" already show up in wp-admin (we'll probably want to fix that for regular comment screens, maybe as part of #340 )

> There isn't a function for the "D" in "CRUD" because wp_delete_comment is sufficient, and needs no wrapper method.

For consistency I would add it in anyway, but I don't feel too strongly about that.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
coreymckrill: I thought about that, but it seemed really inefficient to have to call `get_comment_meta` on every result in the comment array. And then I wasn't sure if I could attach the meta directly to the results, since the `WP_Comment` class is `final`. But I didn't actually test it... I'll do some more research on this.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
coreymckrill: Yeah, I was thinking we'd do a separate file with all the meta-related functionality, but it might make sense here to run the data through a normalizing/sanitizing function once we have that available.

coreymckrill:
> For consistency I would add [the D] in anyway

Fair enough: 9274c7d6

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
coreymckrill: Ok, taking a slightly different approach in 711e063d

I've updated the testing instructions to match.

ryelle (approved):
Looks good, all methods work as expected 👍

PR #105, opened 2019-05-01
Travis: Run unit test suite

iandunn:
I started working on this, but didn't get very far. @vedanshujain, how did you test the original config before committing it? I couldn't find any good info on that.

Also, do you have any feedback on this?

In addition to running the tests, I pulled in some stuff from Core's config.

Supercedes #29

PR #352, opened 2020-01-31
Add infrastructure to run PHPUnit 

ryelle:
This adds a job to run PHPUnit on travis, and adds an optional docker file to set up tests & PHPUnit locally. Previously it was not clear how to run the few tests we have, so this also updates [the documentation.](https://github.com/WordPress/wordcamp.org/tree/add/travis-phpunit/.docker#testing-with-phpunit) Finally, the Remote CSS test was failing after the change in #222, so the test has been fixed.

Having this testing infrastructure will make it easier to write & run tests in the future. New tests will need to be added to the `phpunit.xml.dist` file, following the other `<testsuite>` definitions.

Some past effort here: #105

1. Verify that the tests ran & passed on travis by looking at the Checks tab ^
2. Try installing this locally by following [the instructions](https://github.com/WordPress/wordcamp.org/tree/add/travis-phpunit/.docker#testing-with-phpunit)
3. If you already had `phpunit` running some other way, make sure it still works

### DIFF
--- a/.docker/Dockerfile.phpunit
+++ b/.docker/Dockerfile.phpunit
@@ -1,0 +1,26 @@
+FROM php:7.2-fpm
+
+# Run some Debian packages installation.
+ENV PACKAGES="curl git mariadb-client subversion wget zlib1g-dev"
+RUN apt-get update && \
+    apt-get install -yq --no-install-recommends $PACKAGES && \
+    apt-get clean && \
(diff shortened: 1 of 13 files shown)


iandunn (approved):
I didn't test the Docker/Travis bits, but the code itself LGTM 👍

PR #353, opened 2020-02-04
Speaker Feedback: Register API endpoints for feedback comments

### LINKED ISSUE #337: Speaker Feedback: Create an API endpoint for feedback-comments
We can't use the core `/comments` endpoint, because it doesn't support custom comment types. We'll need to create our own - though we can use comments `WP_REST_Comments_Controller` as a base.

This endpoint only needs to create comments, since we don't have a view for editing or deleting existing feedback (from the frontend).

Comments will have meta, this should also be registered with this endpoint:

- The star rating, integer 1-5
- The question answers, an array of objects `[ { question, answer }, … ]` where `question` & `answer` are strings.
- Anything else TBD?

Make sure it still uses whatever hooks are necessary for Akismet to check the content (and meta).

coreymckrill:
Extends the `WP_REST_Comments_Controller` class to create a new custom endpoint for creating feedback comments.

We can't use the standard Comments endpoint, because it won't allow creation of comments with a custom type or without anything in the content field.

This also introduces schema and functions for the comment meta values that we'll be using to store feedback content, as well as sets up this plugin for phpunit testing.

Fixes #337, Refs #340 

**To test:**

1. Run all the phpunit tests and make sure all are passing.
1. It's easiest to use a tool like [Insomnia](https://insomnia.rest/) to craft payloads to send to the endpoint. The endpoint URL will be `[your-test-site].wordcamp.org/wp-json/wordcamp-speaker-feedback/v1/feedback`
1. The parameters required for a successful payload are:
    * `post` (the ID of the post to attach the feedback comment to)
    * `author` (the numeric ID of a user)
    * `meta[rating]` (an integer representation of 1 to 5 stars)
    * `meta[q1]` (a string less than 5000 characters)
1. You can use `author_name` and `author_email` (string values) in lieu of `author`.
1. You can also include `meta[q2]` and `meta[q3]` parameters, but these are not currently required fields.
1. The response to a successful request to the endpoint should be a `201` status. For an unsuccessful request, it'll be a JSON object with an error code and message. Try to get some of each! Try multiple invalid fields in one request!
1. It would probably also be good to check that the comments and their meta are actually being created in the database upon successful request.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace WordCamp\SpeakerFeedback;
+
+use WP_Error, WP_REST_Request, WP_REST_Response;
+use WP_REST_Comments_Controller;
+use const WordCamp\SpeakerFeedback\Comment\COMMENT_TYPE;
+
+defined( 'WPINC' ) || die();
+
+
+class REST_Feedback_Controller extends WP_REST_Comments_Controller {
+	/**
+	 * REST_Feedback_Controller constructor.
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->namespace = 'wordcamp-speaker-feedback/v1';
+		$this->rest_base = 'feedback';
+	}
+
+	/**
+	 * Ensure a request has the correct parameters set for our feedback comment type.
+	 *
+	 * @param WP_REST_Request $request
+	 *
+	 * @return WP_REST_Request
+	 */
+	protected function sanitize_request( $request ) {
+		$request['type'] = COMMENT_TYPE;
+
+		return $request;
+	}
+
+	/**
+	 * Retrieves a list of feedback items.
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 *
+	 * @return WP_Error|WP_REST_Response Response object on success, or error object on failure.
+	 */
+	public function get_items( $request ) {
+		$request = $this->sanitize_request( $request );
+
+		return parent::get_items( $request );
+	}
+
+	/**
+	 * Checks if a given request has access to read comments.
+	 *
+	 * @since 4.7.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 *
+	 * @return WP_Error|bool True if the request has read access, error object otherwise.
+	 */
+	public function get_items_permissions_check( $request ) {
+		$request = $this->sanitize_request( $request );
+
+		$default_comment_check = parent::get_item_permissions_check( $request );
+
+		if ( is_wp_error( $default_comment_check ) ) {
+			return $default_comment_check;
+		}
+
+		// TODO See #340.
+
+		return true;
+	}
+}
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
@@ -208,6 +208,16 @@ class REST_Feedback_Controller extends WP_REST_Comments_Controller {
 			);
 		}
 
+		if ( ! post_type_supports( get_post_type( $post ), 'wordcamp-speaker-feedback' ) ) {
+			return new WP_Error(
+				'rest_feedback_post_not_supported',
+				__( 'Sorry, this post type does not support feedback.', 'wordcamporg' ),
+				array(
+					'status' => 403,
+				)
+			);
+		}
+
 		if ( 'publish' !== $post->post_status ) {
 			return new WP_Error(
 				'rest_feedback_post_unavailable',
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
@@ -4,12 +4,16 @@ namespace WordCamp\SpeakerFeedback;
 
 use WP_Error, WP_REST_Request, WP_REST_Response, WP_REST_Server;
 use WP_REST_Comments_Controller;
-use function WordCamp\SpeakerFeedback\Comment\{ add_feedback };
+use function WordCamp\SpeakerFeedback\Comment\add_feedback;
 use function WordCamp\SpeakerFeedback\CommentMeta\validate_feedback_meta;
 
 defined( 'WPINC' ) || die();
 
-
+/**
+ * Class REST_Feedback_Controller
+ *
+ * @package WordCamp\SpeakerFeedback
+ */
 class REST_Feedback_Controller extends WP_REST_Comments_Controller {
 	/**
 	 * REST_Feedback_Controller constructor.
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
@@ -129,7 +129,7 @@ class REST_Feedback_Controller extends WP_REST_Comments_Controller {
 			$error_code = $check_comment_lengths->get_error_code();
 			return new WP_Error(
 				$error_code,
-				__( 'Comment field exceeds maximum length allowed.' ),
+				__( 'Comment field exceeds maximum length allowed.', 'wordcamporg' ),
 				array(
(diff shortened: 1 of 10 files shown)


coreymckrill:
Bah, the tests are passing locally...

ryelle:
> Bah, the tests are passing locally...

for what it's worth, the test failed for me the same was as travis the first time I ran them, but have worked each time since.

ryelle:
I'm having trouble testing this with Insomnia, it keeps giving me this response even though I know the ID I've given it is a real session.

```json
{
  "code": "rest_feedback_invalid_post_id",
  "message": "Sorry, you are not allowed to create this comment without a post.",
  "data": {
    "status": 403
  }
}
```

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
```diff
@@ -0,0 +1,225 @@
...
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 *
+	 * @return WP_Error|bool True if the request has access to create items, error object otherwise.
+	 */
+	public function create_item_permissions_check( $request ) {
+		if ( empty( $request['post'] ) ) {
+			return new WP_Error(
+				'rest_feedback_invalid_post_id',
+				__( 'Sorry, you are not allowed to create this comment without a post.' ),
```
ryelle: This is missing a textdomain

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
```diff
@@ -0,0 +1,225 @@
...
+					'status' => 403,
+				)
+			);
+		}
+
+		$post = get_post( (int) $request['post'] );
+		if ( ! $post ) {
+			return new WP_Error(
+				'rest_feedback_invalid_post_id',
+				__( 'Sorry, you are not allowed to create this comment without a post.' ),
```
ryelle: This message confused me at first since I did give it a post — maybe something about needing a valid post. Should it also check for Session post type? (also, missing textdomain here too)

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
```diff
@@ -0,0 +1,207 @@
...
+/**
+ * Define the properties of the feedback meta fields.
+ *
+ * @param string $key Optional. A specific key to get the schema for.
+ *
+ * @return array
+ */
+function get_feedback_meta_field_schema( $key = '' ) {
+	$schema = array(
+		'version' => array(
```
ryelle, line 64: Good idea with versioning the questions! 💯

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
```diff
@@ -0,0 +1,207 @@
...
+			array(
+				'missing_fields' => array_keys( $missing_fields ),
+			)
+		);
+	}
+
+	$integer_fields = array_keys( wp_list_pluck( $fields, 'type' ), 'integer', true );
+
+	foreach ( $integer_fields as $key ) {
+		if ( isset( $meta[ $key ] ) && ! is_numeric( $meta[ $key ] ) ) {
```
ryelle: Can we set some custom values in the schema? Thinking it might be good to check against a "max" so that no one tries to enter in a rating of 100 or something (in a future PR is fine)

ryelle (commented):
some notes from looking through the code — overall, great improvements here; dunno why it's not working for me. It seems like the endpoint `$request` is not actually receiving the parameters. I'll try testing again when I'm back on Friday.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
coreymckrill: Oops, good catch. Fixed here: 025f85f8

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
coreymckrill: I'm not sure why it's not recognizing your post ID. I just tried it again with Insomnia and it's still working for me...

> Should it also check for Session post type?

Yeah, I was thinking of including that as part of another validation function that would go where the `TODO` note currently is. I've been thinking I would add that as part of a separate PR for #340, but I guess that one is specifically about permissions for viewing/getting, so perhaps it would be better to do it here. Stay tuned.

coreymckrill: Added post type support and a check for it here: 2c9036af

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
coreymckrill: See what you think of this: 42180384

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
```diff
@@ -50,6 +60,20 @@ function activate() {
 	flush_rewrite_rules();
 }
 
+/**
+ * Add post type support for supported post types. Only for the types that are supported though.
+ *
+ * This makes it easy to check if a particular post type can have feedback comments using Core functionality, rather
+ * than having to import a namespaced constant or function.
```
ryelle, line 67: Also a great idea 😁

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
```diff
@@ -0,0 +1,295 @@
...
+			'show_in_rest'      => false,
+			'default'           => 0,
+			'required'          => true,
+			'attributes'        => array(
+				'min'  => 1,
+				'max'  => 5,
+				'step' => 1,
+			),
+		),
+		'q1'      => array(
```
ryelle, line 87: The first question is also required:

> The talk selection, star rating and first text area will be required fields with the rest being optional. [(source)](https://make.wordpress.org/community/2020/01/27/speaker-feedback-tool-final-outline/)

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
```diff
@@ -0,0 +1,235 @@
...
+			return new WP_Error(
+				'rest_feedback_creation_failed',
+				__( 'Feedback submission failed.', 'wordcamporg' ),
+				array(
+					'status' => 400,
+				)
+			);
+		}
+
+		$response = __( 'Success!', 'wordcamporg' );
```
ryelle: Will a user ever see this? I'd expect this endpoint to only be called by javascript, so we'd want a stable return value to check against — a string that might be in a different language could make that tricky.

ryelle (approved):
Turns out my API issues were user-error 🤦‍♀  All working as expected now. I really like the validation approach 💯 

Two minor comments, then this is ready to merge 👍

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
coreymckrill: Fixed: f92715ae

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
coreymckrill: Good point. It does also send a `201` response code, which should be stable. But `Success!` probably isn't necessary.

The default comment endpoint returns data about the comment on successful creation. I wasn't sure we would want to do that in this case since feedback comments are not public. What do you think would be best?

ryelle: yeah, I'm agreed on not returning comment data here, I don't think that's necessary. I didn't think about the response code, so that should be enough. Does the response need content? we might be fine with an empty response (unless that does unexpected things).

coreymckrill: I'll test and find out...

coreymckrill: Seems to work fine. I updated the response here: 3415a674

PR #355, opened 2020-02-04
Subroles: Omit user caps that have been added via usermeta values

coreymckrill:
On WordCamp.org we grant additional capabilities to users via the `map_meta_cap` filter and a hard-coded list of users/caps. This way the granting/revoking of capabilities has an audit log in the form of revision history.

When a `WP_User` object is instantiated, it populates a `caps` property from a user meta value that is blog-specific. Normally this value only contains the user's role, but it can also store additional capabilities if they are added via `add_cap`. These additional capabilities are then merged with additions from `map_meta_cap` when evaluating something like `current_user_can`.

Possibly related: https://core.trac.wordpress.org/ticket/10201

This adds a filter that omits capabilities that have been incorporated from user meta when evaluating `has_cap`, so that only the `map_meta_cap` filter can add caps.

### DIFF
--- a/public_html/wp-content/mu-plugins/wcorg-subroles.php
+++ b/public_html/wp-content/mu-plugins/wcorg-subroles.php
@@ -6,6 +6,9 @@
  */
 
 namespace WordCamp\SubRoles;
+
+use WP_User;
+
 defined( 'WPINC' ) || die();
 
 /**
@@ -53,10 +56,10 @@ function is_proxied() {
 /**
  * Add capabilities to a user depending on their subroles.
  *
- * @param array    $allcaps The original list of caps for the given user.
- * @param array    $caps    Unused.
- * @param array    $args    Unused.
- * @param \WP_User $user    The user object.
+ * @param array   $allcaps The original list of caps for the given user.
+ * @param array   $caps    Unused.
+ * @param array   $args    Unused.
+ * @param WP_User $user    The user object.
  *
  * @return array The modified list of caps for the given user.
  */
@@ -178,3 +181,31 @@ function map_subrole_caps( $primitive_caps, $meta_cap, $user_id, $args ) {
 }
 
 add_filter( 'map_meta_cap', __NAMESPACE__ . '\map_subrole_caps', 10, 4 );
+
+/**
+ * Remove capabilities that are "additional" i.e. stored in user meta.
+ *
+ * Additional capabilities should only be granted via `map_meta_cap`, not via values stored in the user meta table.
+ *
+ * See `additional_capabilities_display` filter.
+ *
+ * @param bool[]   $allcaps
+ * @param string[] $caps
+ * @param array    $args
+ * @param WP_User  $user
+ *
+ * @return bool[]
+ */
+function omit_usermeta_caps( $allcaps, $caps, $args, $user ) {
+	if ( $user instanceof WP_User && count( $user->caps ) > count( $user->roles ) ) {
+		$extraneous_caps = array_diff_key( array_keys( $user->caps ), $user->roles );
+
+		foreach ( $extraneous_caps as $cap ) {
+			unset( $allcaps[ $cap ] );
+		}
+	}
+
+	return $allcaps;
+}
+
+add_filter( 'user_has_cap', __NAMESPACE__ . '\omit_usermeta_caps', 10, 4 );
--- a/public_html/wp-content/mu-plugins/wcorg-subroles.php
+++ b/public_html/wp-content/mu-plugins/wcorg-subroles.php
@@ -189,10 +189,11 @@ add_filter( 'map_meta_cap', __NAMESPACE__ . '\map_subrole_caps', 10, 4 );
  *
  * See `additional_capabilities_display` filter.
  *
- * @param bool[]   $allcaps
- * @param string[] $caps
- * @param array    $args
- * @param WP_User  $user
+ * @param bool[]   $allcaps Array of key/value pairs where keys represent a capability name and boolean values
+ *                          represent whether the user has that capability.
+ * @param string[] $caps    Unused. Required primitive capabilities for the requested capability.
+ * @param array    $args    Unused. Arguments that accompany the requested capability check.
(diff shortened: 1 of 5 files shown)


ryelle (approved):
Works as described. It seems strange that it discards a dynamic cap if the same one exists in the database, but i _think_ that's enough of an edge case to not be a problem.

public_html/wp-content/mu-plugins/wcorg-subroles.php
```diff
@@ -178,3 +181,31 @@ function map_subrole_caps( $primitive_caps, $meta_cap, $user_id, $args ) {
 }
 
 add_filter( 'map_meta_cap', __NAMESPACE__ . '\map_subrole_caps', 10, 4 );
+
+/**
+ * Remove capabilities that are "additional" i.e. stored in user meta.
```
iandunn:
```suggestion
* Ignore capabilities that are "additional" i.e. stored in user meta.
```

`MIGHT` - "Remove" could be confused for deleting from the database.

public_html/wp-content/mu-plugins/wcorg-subroles.php
```diff
@@ -178,3 +181,31 @@ function map_subrole_caps( $primitive_caps, $meta_cap, $user_id, $args ) {
...
+ * See `additional_capabilities_display` filter.
+ *
+ * @param bool[]   $allcaps
+ * @param string[] $caps
+ * @param array    $args
+ * @param WP_User  $user
+ *
+ * @return bool[]
+ */
+function omit_usermeta_caps( $allcaps, $caps, $args, $user ) {
```
iandunn, line 200: `MIGHT` - I think something like `$capability_grant_map` and `$required_primitives` would be more descriptive.

I know the others are coming from Core, but I often have to jump to the definition to make sure I'm remembering what they mean correctly.

Another approach would be copy/pasting the filter's docblock descriptions, which may we helpful either way.

public_html/wp-content/mu-plugins/wcorg-subroles.php
```diff
@@ -178,3 +181,31 @@ function map_subrole_caps( $primitive_caps, $meta_cap, $user_id, $args ) {
...
+	if ( $user instanceof WP_User && count( $user->caps ) > count( $user->roles ) ) {
+		$extraneous_caps = array_diff_key( array_keys( $user->caps ), $user->roles );
+
+		foreach ( $extraneous_caps as $cap ) {
+			unset( $allcaps[ $cap ] );
```
iandunn, line 205:
```suggestion
			$allcaps[ $cap ] = false;
```

`SHOULD` - I haven't tested, but I think this is technically the correct way to do it. 

If I understand it right, `$allcaps` isn't a _list_ of all the capabilities that have been granted to the user; it's a _map_ of all the capabilities that exist, and whether or not they've been granted to the user. This is a good example of why Core's variable aren't descriptive enough.

I'm guessing that unsetting will work the same in most cases because of how `empty()` works, but might introduce some subtle and hard-to-diagnose edge-case bugs.

iandunn (approved):
Left some suggestions, but no blockers 👍

iandunn:
> discards a dynamic cap if the same one exists in the database

Would this solve that?

```diff
- add_filter( 'user_has_cap', __NAMESPACE__ . '\add_subrole_caps', 10, 4 );
+ add_filter( 'user_has_cap', __NAMESPACE__ . '\add_subrole_caps', 12, 4 );

- add_filter( 'user_has_cap', __NAMESPACE__ . '\omit_usermeta_caps', 10, 4 );
+ add_filter( 'user_has_cap', __NAMESPACE__ . '\omit_usermeta_caps', 11, 4 );
```

We'd probably want `omit_usermeta_caps` to run after most plugins anyway.

iandunn:
Er, actually, maybe we'd want those callbacks to run at priority `1` and `2`? That'd make sure the correct caps get assigned _before_ plugins have a chance to grant access to something the user shouldn't have access to.

public_html/wp-content/mu-plugins/wcorg-subroles.php
coreymckrill: Good call. 23c31bd6

coreymckrill: Oops, should have just used GitHub's UI to commit your suggestion :/

public_html/wp-content/mu-plugins/wcorg-subroles.php
coreymckrill: 👍 dbc547bf

public_html/wp-content/mu-plugins/wcorg-subroles.php
coreymckrill: I would actually expect that setting to `false` would potentially introduce edge case bugs. When using `map_meta_cap`, the extra capabilities are not present in the `$allcaps` array at all (at least, not when this filter is applied). So using `unset` ensures that the array is the same way it would be if there weren't extraneous caps.

coreymckrill:
> maybe we'd want those callbacks to run at priority 1 and 2? That'd make sure the correct caps get assigned before plugins have a chance to grant access to something the user shouldn't have access to.

As long as plugins are adding caps via the `map_meta_cap` filter, it shouldn't matter when it runs. This won't affect those additions, because it's only running when the User object `caps` property doesn't match the `roles` property. The contents of those properties comes only from usermeta, and are not affected by the `map_meta_cap` filter.

public_html/wp-content/mu-plugins/wcorg-subroles.php
iandunn: Huh, TIL. That WFM then 👍

coreymckrill: Er, actually I misspoke. Caps added via `map_meta_caps` _are_ in the `$allcaps` array, but they aren't in the `caps` or `roles` properties in the `WP_User` object. Since this filter removes things based on what's in those properties, things added via `map_meta_cap` won't be affected.

PR #356, opened 2020-02-06
Add/schedule block scaffolding

iandunn:
This is branched off of #104, to make reviewing easier. This one is just the scaffolding for the block, while 104 contains all the logic for the grid, etc.

With this one you should be able to add the block to a page, and see the date and "todo" in the place of the grid.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/.eslintrc.js
+++ b/public_html/wp-content/mu-plugins/blocks/.eslintrc.js
@@ -1,10 +1,17 @@
 /*
- * @todo
+ * @todo discuss these potential changes
  *
- * should use hasOwnProperty or Object.getOwnPropertyDescriptors(). the latter usually makes code more readable, but sometimes the former first better
+ * `arrow-parens` - skipping them when only have 1 arg remove unnecessary clutter, which helps readability
+ * `no-multiple-empty-lines` - somestimes 2 lines useful to separate sections of code, like `import`s at top of file from `export` of components
+ * `require-jsdoc` - not for `render` and lifecycle methods b/c it's obvious what they are
+ *
  * assignment and control structures/returns/etc should be separate by a blank line for readability.
  *      same for div and other block-level html elements
+ *
+ * should use hasOwnProperty or Object.getOwnPropertyDescriptors(). the latter usually makes code more readable, but sometimes the former first better
+ *
  * disable `no-console` b/c valid use case. if can make exception for `log` function without disabling, then do that. don't want console used for temporary debugging, but there are valid cases where you want to provide the user some insight into what went wrong
+ *
  */
 
 module.exports = {
--- a/public_html/wp-content/mu-plugins/blocks/package.json
+++ b/public_html/wp-content/mu-plugins/blocks/package.json
@@ -52,6 +52,7 @@
 	"scripts": {
 		"start": "wp-scripts start blocks=./source/blocks.js live-schedule=./source/blocks/live-schedule/front-end.js live-posts=./source/hooks/latest-posts/front-end.js",
 		"build": "wp-scripts build blocks=./source/blocks.js live-schedule=./source/blocks/live-schedule/front-end.js live-posts=./source/hooks/latest-posts/front-end.js",
+		"lint:js:comment": "echo 'Usage: `npm run lint:js source/blocks/schedule -- --fix`. The path has to be from the root `mu-plugins/blocks` folder.' ",
 		"lint:js": "wp-scripts lint-js",
 		"lint:css": "wp-scripts lint-style '**/*.scss'",
 		"lint:pkg-json": "wp-scripts lint-pkg-json",
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/controller.php
@@ -1,0 +1,118 @@
+<?php
+
+namespace WordCamp\Blocks\Schedule;
+
+defined( 'WPINC' ) || die();
+
+
+/**
+ * Register block types and enqueue scripts.
+ *
+ * @return void
+ */
+function init() {
+	register_block_type(
+		'wordcamp/schedule',
+		array(
+			'attributes'      => get_attributes_schema(),
+			'render_callback' => __NAMESPACE__ . '\render',
+			'editor_script'   => 'wordcamp-blocks',
+			'editor_style'    => 'wordcamp-blocks',
+			'style'           => 'wordcamp-blocks',
+		)
+	);
+}
+add_action( 'init', __NAMESPACE__ . '\init' );
+
+/**
+ * Render the block on the front end.
+ *
+ * @param array $attributes Block attributes.
+ *
+ * @return string
+ */
+function render( $attributes ) {
+	// @todo Render via JS instead?
+	return '';
+}
+
+/**
+ * Add data to be used by the JS scripts in the block editor.
+ *
+ * @param array $data
+ *
+ * @return array
+ */
+function add_script_data( array $data ) {
+	$data['schedule'] = array(
+		'schema'  => get_attributes_schema(),
+		'options' => get_options(),
+	);
+
+	return $data;
+}
+add_filter( 'wordcamp_blocks_script_data', __NAMESPACE__ . '\add_script_data' );
+
+/**
+ * Get the schema for the block's attributes.
+ *
+ * @return array
+ */
+function get_attributes_schema() {
+	// todo need to update this when building inspector controls, to make them DRY
+	// this one will be different than others in that it wont have a 'mode', so need to make that not a shared attribute?
+		// or maybe leave it shared since most blocks use it, but explicitly remove it here?
+
+	return array(
+		'item_ids' => array(
+			'type'    => 'array',
+			'default' => array(),
+			'items'   => array(
+				'type' => 'integer',
+			),
+		),
+
+		'className' => array(
+			'type'    => 'string',
+			'default' => '',
+		),
+
+		'show_categories' => array(
+			'type'    => 'bool',
+			'default' => false,
+		),
+
+		'chosen_days' => array(
+			'type'    => 'array',
+			'default' => array(),
+			'items'   => array(
+				'type' => 'string',
+			),
+		),
+
+		'chosen_tracks' => array(
+			'type'    => 'array',
+			'default' => array(),
+			'items'   => array(
+				'type' => 'integer',
+			),
+		),
+	);
+}
+
+/**
+ * Get the label/value pairs for all options or a specific type.
+ *
+ * @param string $type
+ *
+ * @return array
+ */
+function get_options( $type = '' ) {
+	$options = array();
+
+	if ( $type ) {
+		return empty( $options[ $type ] ) ? array() : $options[ $type ];
+	}
+
+	return $options;
+}
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
@@ -1,0 +1,55 @@
+/**
+ * WordPress dependencies
+ */
+import { withSelect } from '@wordpress/data';
+/* eslint-disable-next-line */ /* todo tmp */
+import { Component, Fragment } from '@wordpress/element';
+
+/**
+ * Internal dependencies
+ */
+import { WC_BLOCKS_STORE } from '../../data';
+import { ScheduleGrid } from './schedule-grid';
+import InspectorControls from './inspector-controls';
+import { ICON } from './index';
+
+const blockData = window.WordCampBlocks.sessions || {};
+
+/**
+ * Top-level component for the editing UI for the block.
+ *
+ * @param {Object} props
+ * @param {Object} props.attributes
+ * @param {Object} props.entities
+ *
+ * @return {Element}
+ */
+function ScheduleEdit( { attributes, entities } ) {
+	return (
+		<Fragment>
+			<ScheduleGrid
+				icon={ ICON }
+				attributes={ attributes }
+				entities={ entities }
+			/>
+
+			<InspectorControls />
+		</Fragment>
+	);
+}
+
+const scheduleSelect = ( select ) => {
+	const { getEntities, getSiteSettings } = select( WC_BLOCKS_STORE );
+
+	const sessionArgs = {
+	};
+
+	const entities = {
+		sessions: getEntities( 'postType', 'wcb_session', sessionArgs ),
+		settings: getSiteSettings(),
+	};
+
+	return { blockData, entities };
+};
+
+export const Edit = withSelect( scheduleSelect )( ScheduleEdit );
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
@@ -1,0 +1,116 @@
+/**
+ * WordPress dependencies
+ */
+import { InspectorControls } from '@wordpress/block-editor';
+import { CheckboxControl, PanelBody, ToggleControl } from '@wordpress/components';
+import { Component, Fragment } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+
+/*
+ * @todo
+ * make this dynamic
+ * add default schema and options, like speakers?
+ *
+ * commit w/ props mark, mel, corey. look for any others as well
+ */
+
+/**
+ * Render the inspector Controls for the Schedule block.
+ */
+export default class extends Component {
+	/**
+	 * Render the component.
+	 *
+	 * @return {Element}
+	 */
+	render() {
+		const { setAttributes } = this.props;
+		const show_categories = false;
+		const choose_days = true;
+		const choose_tracks = true;
+
+		return (
+			<InspectorControls>
+				<PanelBody title={ __( 'Display Settings', 'wordcamporg' ) } initialOpen={ true }>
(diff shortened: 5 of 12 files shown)


public_html/wp-content/mu-plugins/blocks/.eslintrc.js
```diff
@@ -1,10 +1,17 @@
 /*
- * @todo
+ * @todo discuss these potential changes
```
ryelle, line 2: Can you make a GH issue for this instead?

public_html/wp-content/mu-plugins/blocks/package.json
```diff
@@ -52,6 +52,7 @@
 	"scripts": {
 		"start": "wp-scripts start blocks=./source/blocks.js live-schedule=./source/blocks/live-schedule/front-end.js live-posts=./source/hooks/latest-posts/front-end.js",
 		"build": "wp-scripts build blocks=./source/blocks.js live-schedule=./source/blocks/live-schedule/front-end.js live-posts=./source/hooks/latest-posts/front-end.js",
+		"lint:js:comment": "Usage: `npm run lint:js source/blocks/schedule -- --fix`. The path has to be from the root `mu-plugins/blocks` folder.",
```
ryelle: Is this meant to just be a code comment? I ran it, and it actually ran `npm run lint:js source/blocks/schedule -- --fix`, which was not expected. (it then attempted to run `mu-plugins/blocks` but that's not a script).

IMO this kind of documentation should be in the readme, but it would just be a copy of [the wp-scripts readme…](https://github.com/WordPress/gutenberg/tree/master/packages/scripts/) we could link off to that instead?

Also an aside, since we're using yarn workspaces it'd be better to run commands with `yarn workspace wordcamp-blocks run [script]`, then you can run from any place in the project (it still expects paths to be relative to the blocks dir).

public_html/wp-content/mu-plugins/blocks/.eslintrc.js
```diff
@@ -132,7 +139,7 @@ module.exports = {
 		 */
 		'valid-jsdoc' : [ 'error', {
 			'requireParamDescription'  : false,
-			'requireReturnDescription' : false,
+			'requireReturnDescription' : false, // todo this isn't working
```
ryelle: Found the issue & fixed it in 6044c68d2d9f766f30e2891689ac1cd6cce4f456 👍

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
```diff
@@ -0,0 +1,58 @@
...
+import { ScheduleGrid } from './schedule-grid';
+import InspectorControls from './inspector-controls';
+import { ICON } from './index';
+
+const blockData = window.WordCampBlocks.sessions || {};
+
+/**
+ * Top-level component for the editing UI for the block.
+ */
+class ScheduleEdit extends Component {
```
ryelle: This could be simplified to just a functional component ([like this,](https://github.com/WordPress/wordcamp.org/blob/6044c68d2d9f766f30e2891689ac1cd6cce4f456/public_html/wp-content/mu-plugins/blocks/source/blocks/live-schedule/components/session.js#L12) the args for the function are the props), which would reduce the redundant comments too.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
```diff
@@ -0,0 +1,58 @@
...
+			</Fragment>
+		);
+	}
+}
+
+const scheduleSelect = ( select ) => {
+	const { getEntities, getSiteSettings } = select( WC_BLOCKS_STORE );
+
+	const sessionArgs = {
+		_embed: true,
```
ryelle: Does this need the `_embed`? It really inflates the response time of the endpoint, and we've [already added extra info](https://github.com/WordPress/wordcamp.org/commit/8c56b352123118f62bc8f163922cf6dc7149df43) to the sessions endpoint to avoid the full embed call.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @param {string} date In a format acceptable by `wp.date.gmdate()`.
+ * @param {Array} sessions
+ * @param {Array} tracks
+ * @param {string} dateFormat
+ * @param {string} timeFormat
+ *
+ * @return {Element}
+ */
+function ScheduleDay( { date, sessions, tracks, dateFormat, timeFormat } ) {
```
ryelle, line 29: Can you move this to its own file? We've been trying to stick to one-component-per-file.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/controller.php
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Render the block on the front end.
+ *
+ * @param array $attributes Block attributes.
+ *
+ * @return string
+ */
+function render( $attributes ) {
+	// @todo Render via JS instead?
```
ryelle, line 35: 🤔  Some combination of a static (saved HTML) block grid & dynamic php-rendered child blocks for session-cells could be interesting… like the relationship between [core/social-links](https://github.com/WordPress/gutenberg/tree/master/packages/block-library/src/social-links) and [core/social-link.](https://github.com/WordPress/gutenberg/tree/master/packages/block-library/src/social-link)

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/controller.php
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Get the schema for the block's attributes.
+ *
+ * @return array
+ */
+function get_attributes_schema() {
+	// todo need to update this when building inspector controls, to make them DRY
+	// this one will be different than others in that it wont have a 'mode', so need to make that not a shared attribute?
+		// or maybe leave it shared since most blocks use it, but explicitly remove it here?
```
ryelle, line 64: I think it's fine if this block doesn't use the shared attributes— live-schedule doesn't, and like that one, this is totally different "view" for data.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
```diff
@@ -0,0 +1,103 @@
...
+
+					<div className="wordcamp-schedule__control-container">
+						<ToggleControl
+							label={ __( 'Choose specific days', 'wordcamporg' ) }
+							checked={ choose_days }
+							onChange={ ( value ) => setAttributes( { choose_days: value } ) }
+						/>
+
+						{ choose_days &&
+							<Fragment>
```
ryelle: This should probably be a fieldset+legend, so it's clear what these checkboxes do (pick those days).

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
```diff
@@ -0,0 +1,103 @@
...
+
+					<div className="wordcamp-schedule__control-container">
+						<ToggleControl
+							label={ __( 'Choose specific tracks', 'wordcamporg' ) }
+							checked={ choose_tracks }
+							onChange={ ( value ) => setAttributes( { show_tracks: value } ) }
+						/>
+
+						{ choose_tracks &&
+							<Fragment>
```
ryelle: Same fieldset comment here.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
```diff
@@ -0,0 +1,58 @@
...
+
+		return (
+			<Fragment>
+				<ScheduleGrid
+					icon={ ICON }
+					attributes={ attributes }
+					entities={ entities }
+				/>
+
+				<InspectorControls />
```
ryelle: You need to pass `attributes` & `setAttributes` down here, it's causing an error in the InspectorControls now when I try to click any of the toggles/checkboxes.
```suggestion
				<InspectorControls
					attributes={ attributes }
					setAttributes={ setAttributes }
				/>
```

ryelle (changes requested):
It does create a block, but since it's not really functional there isn't much to test.

ryelle:
Screenshot of the block:

Screen Shot 2020-02-06 at 11 40 12 AM

cc'ing @melchoyce  & adding the needs design review label for some UI feedback.

iandunn:
> adding the needs design review label for some UI feedback.

I'd personally wait until future PRs for that, since this isn't intended to be a perfect/final design. I just wanted to get the scaffolding merged in order to make this and #104 easier to review.

The spacing in the inspector controls was [a requested change](https://make.wordpress.org/community/2018/10/26/wordcamp-block-schedule/#comment-27080).

public_html/wp-content/mu-plugins/blocks/package.json
iandunn: It's intended to be a comment, but `npm` will strip actual comments out. There are [a lot of workarounds for that](https://stackoverflow.com/questions/14221579/how-do-i-add-comments-to-package-json-for-npm-install), and this one seemed to have the best tradeoffs in this context.

I didn't intend people to run it, but I'll add an `echo` before it for anyone who does.

iandunn: I think it's necessary. I added it because I tried a lot of variations and couldn't get any to work. After digging through a lot of stuff I realized the problem was that the path had to be relative to the root folder, which didn't feel intuitive to me. I don't think the external documentation is close enough or detailed enough to save people from running into the same problems that I did.

iandunn: I personally prefer `npm run {command}`, because it's shorter and more memorable, and I rarely run a `blocks/` command from outside the `blocks/` folder.

If you both prefer the `yarn` variation, though, I'm happy to use that for the comment.

public_html/wp-content/mu-plugins/blocks/.eslintrc.js
iandunn: Thanks!

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
iandunn: Good catch, I think I just did it that way b/c the others are, but we should probably change most of them too?

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
iandunn: I didn't think we were doing that for functional components, it seems like that'd add a ton of clutter.

https://github.com/WordPress/wordcamp.org/pull/167#discussion_r306520040

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
iandunn: 👍

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
iandunn: It's only meant to be a static scaffolding at this point, but I'm happy to add that.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
ryelle: I wouldn't bother changing the other blocks if that's what you mean, but for new components you're adding, sure.

public_html/wp-content/mu-plugins/blocks/package.json
ryelle: > It's intended to be a comment, but npm will strip actual comments out.

Right, that's why this sort of documentation usually goes in a readme. I think this clutters the list of scripts, and is confusing since it's not a real command. I'm ➕  for adding documentation, but in the readme.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
iandunn: The full branch is currently using it, but during the review of that PR we could look into ways it to optimize it.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
iandunn: b30cca71ff04d2a3d5fa52e4e0649b5d0d4d9bb0

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
iandunn: Done for this block in b30cca71ff04d2a3d5fa52e4e0649b5d0d4d9bb0

I wasn't planning to do the other blocks in this PR, but we've had a lot of conversations about consistency between all the blocks, so it seems odd to only change this in one of them.

cc @coreymckrill in case you have any thoughts.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/controller.php
iandunn: That'll be done in a future PR, so I think we should discuss it then.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/controller.php
iandunn: That'll be done in a future PR, so I think we should discuss it then.

iandunn:
b30cca7 makes all the changes that I think we'll find consensus on. Does anyone see any remaining merge blockers?

ryelle:
> Does anyone see any remaining merge blockers?

I'm checking out your changes now, but I see the tests & lint-checks are still failing on travis — Can you fix those before merging?

iandunn:
Which tests do you see failing? I only see linting stuff. I ran through those several times yesterday and they're mostly temporary side-effects of 80cba3a. They'll be "fixed" in #104, and changing them here would make rebasing more of a headache.

ryelle:
I scanned [the checks page](https://github.com/WordPress/wordcamp.org/pull/356/checks?check_run_id=430767380) quickly & saw the ❌s, but you're right it's both JS & PHP linters failing, not the tests. I do consider this a blocker though. Rebasing is difficult but not impossible, and will keep these code standards changes out of that PR.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
```diff
@@ -0,0 +1,111 @@
...
+					<ToggleControl
+						label={ __( 'Show categories', 'wordcamporg' ) }
+						checked={ show_categories }
+						onChange={ ( value ) => setAttributes( { show_category: value } ) }
+					/>
+
+					<div className="wordcamp-schedule__control-container">
+						<fieldset>
+							<legend>
+								<ToggleControl
```
ryelle, line 44: Technically this is invalid HTML, a legend should only contain "phrasing content", but it _does_ seem to be working. I'll test this more when you've got it actually functional in your next PR.

ryelle:
> The spacing in the inspector controls was a requested change.

I don't follow the linked conversation since i think it's based on older mockups/patterns(?), but you are planning on fixing the janky spacing here, right?

Screen Shot 2020-02-07 at 1 29 50 PM

Edit: to summarize my feedback, there are still some unresolved comments inline, there's the failing lint issues, and this visual issue that I consider varying levels of blockers.

iandunn:
I'm happy to revert that UI change in a future PR, but if I understand it correctly, the current UI is what Mel wanted.

It's not based on older mockups, but maybe we misunderstood each other, and you both thought that the screenshot was how Gutenberg would space it by default? The screenshot was what it looked like _before_ the switch to Gutenberg's defaults.

melchoyce:
If that screenshot isn't current, can you take an up-to-date screenshot I can preview?

iandunn:
Kelly's screenshot above shows the way that it currently looks (with Gutenberg's defaults), and the screenshot in the p2 post shows the way it looked before.

melchoyce:
Okay, looks like I totally misunderstood your original message in that p2 post. The spacing _should_ look like this:

image

Only thing I might change is adding an additional 4px of spacing between the toggle and its checkboxes:

image

(I think I also assumed Gutenberg had a toggle —> checkbox/radio pattern, which was incorrect, because this latest spacing is bananas)

iandunn:
👍 happy to do that in the future PR for the inspector controls.

melchoyce:
Rad, thanks 👍

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
coreymckrill: If I recall, we already converted a bunch of class components to functional in the other blocks when we were iterating on them. But if there's more to do, it's probably worth opening an issue for it. I definitely don't think it should be done as part of the Schedule block.

public_html/wp-content/mu-plugins/blocks/package.json
coreymckrill: I agree that it's super confusing to have a comment here. I'd prefer to have it in the readme.

public_html/wp-content/mu-plugins/blocks/.eslintrc.js
coreymckrill: +1 for making a GitHub issue, and then we can resolve this one.

iandunn:
Closing in favor of #104

PR #357, opened 2020-02-07
Speaker Feedback: Add main feedback page

### LINKED ISSUE #336: Speaker Feedback: Create `/feedback` page on WordCamp sites
This should be a (dynamically created?) page on every WordCamp site, `site.wordcamp.org/feedback`, which will direct the visitor to pick a session to give feedback on. It should list every session with speaker name(s), in a dropdown (possibly autocomplete, if an acceptable accessible solution exists).

Once "Give Feedback" is clicked, this will either navigate to the session page's form, or load the session info & form onto this page. For accessibility reasons, navigation should _not_ happen on dropdown change.

ryelle:
Fixes #336 — Add a page called "Leave Feedback" with a list of sessions. This will navigate to the session feedback form (TBD). The page is created on activation, and removed on deactivation, and it's otherwise not possible for most users to delete the page (it can be done through wp-cli).

### Screenshots

Screen Shot 2020-02-06 at 6 41 01 PM

The page has a special label in the list table:

Screen Shot 2020-02-06 at 7 21 17 PM

1. The page is created on activation, so deactivate/reactivate the plugin
2. You should see a page called Leave Feedback, at site.wordcamp.org/feedback
3. This will list all sessions, and when "give feedback" is activated, navigate to that session page.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/page.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/page.php
@@ -4,6 +4,7 @@ namespace WordCamp\SpeakerFeedback\Page;
 
 add_filter( 'pre_trash_post', __NAMESPACE__ . '\prevent_deletion', 10, 2 );
 add_filter( 'pre_delete_post', __NAMESPACE__ . '\prevent_deletion', 10, 3 );
+add_filter( 'display_post_states', __NAMESPACE__ . '\add_label_to_page', 10, 2 );
 
 /**
  * Prevent deletion of the Feedback page.
@@ -25,3 +26,20 @@ function prevent_deletion( $check, $post, $force_delete = false ) {
 	return $check;
 }
 
+/**
+ * Add label to the Feedback page, to indicate it's a "special" page.
+ *
+ * @param string[] $post_states An array of post display states.
+ * @param WP_Post  $post        The current post object.
+ *
+ * @return array The filtered list of post display states.
+ */
+function add_label_to_page( $post_states, $post ) {
+	$feedback_page = (int) get_option( 'feedback_page' );
+
+	if ( $feedback_page === $post->ID ) {
+		$post_states['sft-page'] = _x( 'Speaker Feedback', 'wordcamporg' );
+	}
+
+	return $post_states;
+}
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/page.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/page.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace WordCamp\SpeakerFeedback\Page;
+
+add_filter( 'pre_trash_post', __NAMESPACE__ . '\prevent_deletion', 10, 2 );
+add_filter( 'pre_delete_post', __NAMESPACE__ . '\prevent_deletion', 10, 3 );
+
+/**
+ * Prevent deletion of the Feedback page.
+ *
+ * @param bool|null $check Whether to go forward with trashing/deletion.
+ * @param WP_Post   $post  Post object.
+ * @param bool      $force_delete Whether to bypass the trash, set when deactivating the plugin to clean up.
+ */
+function prevent_deletion( $check, $post, $force_delete = false ) {
+	if ( $force_delete ) {
+		return $check;
+	}
+
+	$feedback_page = (int) get_option( 'feedback_page' );
+	if ( $feedback_page === $post->ID ) {
+		return false;
+	}
+
+	return $check;
+}
+
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/page.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/page.php
@@ -1,6 +1,7 @@
 <?php
 
 namespace WordCamp\SpeakerFeedback\Page;
+use const WordCamp\SpeakerFeedback\OPTION_KEY;
 
 add_filter( 'pre_trash_post', __NAMESPACE__ . '\prevent_deletion', 10, 2 );
 add_filter( 'pre_delete_post', __NAMESPACE__ . '\prevent_deletion', 10, 3 );
@@ -18,7 +19,7 @@ function prevent_deletion( $check, $post, $force_delete = false ) {
 		return $check;
 	}
 
-	$feedback_page = (int) get_option( 'feedback_page' );
+	$feedback_page = (int) get_option( OPTION_KEY );
 	if ( $feedback_page === $post->ID ) {
 		return false;
 	}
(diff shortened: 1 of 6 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
```diff
@@ -25,4 +27,64 @@
...
+function add_feedback_page() {
+	$page_id = wp_insert_post( array(
+		'post_title'  => __( 'Leave Feedback', 'wordcamporg' ),
+		/* translators: Page slug for the feedback page. */
+		'post_name'   => __( 'feedback', 'wordcamporg' ),
+		'post_status' => 'publish',
+		'post_type'   => 'page',
+	) );
+	if ( $page_id > 0 ) {
+		update_option( 'feedback_page', $page_id );
```
coreymckrill: The option key is hard-coded in a bunch of places. Might be better to define a constant for it.

Also, I was thinking that `feedback` is generic enough that we might run into some namespace issues with other plugins (e.g. Jetpack) if we're not careful. In some other places I see you're using `sft_` as a prefix. Could be worth doing for the option key too.

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
```diff
@@ -25,4 +27,64 @@
...
+	require_once PLUGIN_DIR . 'includes/form.php';
+	require_once PLUGIN_DIR . 'includes/page.php';
+}
+
+/**
+ * Create main Feedback page.
+ */
+function activate() {
+	add_feedback_page();
+}
```
coreymckrill, line 43: I'm wondering if this will work if this plugin ends up being network activated. I.e., when a new site is created, and this is already network activated, will it still run this and create the page on the new site?

Unfortunately, when I tried to test this on my Docker, I kept running into memory limits and max execution times, which is probably (hopefully) an unrelated problem, so I don't know the answer yet. I did notice that when I network activated the plugin, existing sites in the network did not have a feedback page added.

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
```diff
@@ -25,4 +27,64 @@
...
+
+/**
+ * Create the Feedback page, save ID into an option.
+ */
+function add_feedback_page() {
+	$page_id = wp_insert_post( array(
```
coreymckrill, line 60: Since the content is getting overwritten by a filter, I wonder if it would be helpful to insert an admin note into the post content here explaining how this page works.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/page.php
```diff
@@ -0,0 +1,45 @@
...
+ * @param string[] $post_states An array of post display states.
+ * @param WP_Post  $post        The current post object.
+ *
+ * @return array The filtered list of post display states.
+ */
+function add_label_to_page( $post_states, $post ) {
+	$feedback_page = (int) get_option( 'feedback_page' );
+
+	if ( $feedback_page === $post->ID ) {
+		$post_states['sft-page'] = _x( 'Speaker Feedback', 'wordcamporg' );
```
coreymckrill: Missing the context parameter in the `_x()` here.

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-select-sessions.php
```diff
@@ -0,0 +1,36 @@
...
+		<div class="speaker-feedback__field">
+			<select name="sft_session" id="sft-session">
+				<?php while ( $sessions->have_posts() ) {
+					$sessions->the_post();
+					printf(
+						'<option value="%s">%s</option>',
+						esc_attr( get_the_ID() ),
+						wp_kses_post( get_the_title() )
+					);
+				} ?>
```
coreymckrill: Looks like it's not actually required in the WP [coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/#brace-style), but I generally think it's a good idea to use the alternative control structure syntax when mixing HTML and PHP.

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
ryelle: Good point, I didn't test this network-wide.

ryelle: I've fixed it so that it will also check for the page on `init` and add it if necessary… do you think I should also update the deactivation cleanup to account for network deactivations? seems like that would probably time out on wc.org, looping through every site just to delete a page.

coreymckrill: > do you think I should also update the deactivation cleanup to account for network deactivations? seems like that would probably time out on wc.org, looping through every site just to delete a page.

Yeah, probably not worth it to worry about deactivation. This did make me think that we should probably add a skip feature flag here though, and only add the page on sites past a certain ID.

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
```diff
@@ -16,13 +16,92 @@
...
+	}
+}
+
+/**
+ * Remove the feedback page.
+ */
+function deactivate() {
+	$page_id = get_option( OPTION_KEY );
+	wp_delete_post( $page_id, true );
+}
```
coreymckrill, line 80: Need to delete the option here as well, since the page is being permanently deleted.

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
```diff
@@ -16,13 +16,92 @@
...
+
+/**
+ * Create the Feedback page, save ID into an option.
+ */
+function add_feedback_page() {
+	$page_id = get_option( OPTION_KEY );
+	if ( $page_id ) {
+		return;
+	}
```
coreymckrill, line 52: It might be a good idea check and make sure an actual page is attached to this ID. Although, if the option gets deleted when the plugin is deactivated (per my other suggestion), that might eliminate any edge cases where the option still exists but the page doesn't...

I guess if this is running on every page load on `init` having the extra query for a specific page isn't great...

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
ryelle: > skip feature flag

should the entire plugin functionality be on a skip feature flag? or would we still want the `/session/blah/feedback` form to work on every site?

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
ryelle: I've updated it so that when the page is deleted, the option is deleted too — so we _shouldn't_ run into a case where the option still exists for a missing page.

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
coreymckrill: Good point, it should probably just be the entire plugin.

ryelle: We can do that in a separate PR, then 👍

ryelle:
@coreymckrill I think I've addressed all your feedback here ~is this ready to merge now?~ Edit: going to merge this since you're AFK, but can address any issues in following PRs 👍

PR #359, opened 2020-02-07
Add feedback form pages to each session

### LINKED ISSUE #334: Speaker Feedback: Create session feedback form
This is the form for sending in session feedback. It will be shown on the session page, below the session content. It should submit the answers to an API endpoint (#337), and show a success/failure message (can be mocked until the endpoint exists).

The star rating & first question are required, the rest are optional. If the user is not logged in, there should also be name & email fields, which are required.

Session Form

ryelle:
See #334 — This PR adds a new endpoint for `[session]/feedback`, which will have the feedback form on it. This branches off #357, and updates the navigation action to redirect to these new pages.

I haven't styled anything yet, or done any submission work, but am looking for feedback on this approach & flow.

### Screenshots

Screen Shot 2020-02-07 at 3 46 27 PM

1. Deactivate/Activate the plugin, or flush permalinks
2. Visit the `site.wordcamp.org/feedback` page from #357 
3. Pick a session, submit
4. It should redirect you to a session page with a form on it (URL is wonky right now, but the page loads correctly)

Alternately, visit the session feedback directly, by adding `feedback` to a session URL, ex:

```
https://site.wordcamp.test/session/session-name/feedback
```

You should see the feedback form.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -9,7 +9,10 @@ require_once 'inc/favorite-schedule-shortcode.php';
 require_once 'inc/privacy.php';
 require_once 'inc/deprecated.php';
 
-define( 'EP_SESSIONS', 8192 );
+// Bitwise mask for the sessions CPT, to add endpoints to the session pages. This should be a unique power of 2
+// greater than the core-defined ep_masks, but could potentially conflict with another plugin.
+// See https://developer.wordpress.org/reference/functions/add_rewrite_endpoint/.
+define( 'EP_SESSIONS', 16384 );
 
 class WordCamp_Post_Types_Plugin {
 	protected $wcpt_permalinks;
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -9,6 +9,8 @@ require_once 'inc/favorite-schedule-shortcode.php';
 require_once 'inc/privacy.php';
 require_once 'inc/deprecated.php';
 
+define( 'EP_SESSIONS', 8192 );
+
 class WordCamp_Post_Types_Plugin {
 	protected $wcpt_permalinks;
 
@@ -1732,6 +1734,7 @@ class WordCamp_Post_Types_Plugin {
 				'rewrite'         => array(
 					'slug'       => 'session',
 					'with_front' => false,
+					'ep_mask'    => EP_SESSIONS,
 				),
 				'supports'        => array( 'title', 'editor', 'excerpt', 'author', 'revisions', 'thumbnail', 'custom-fields' ),
 				'menu_position'   => 21,
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
@@ -8,9 +8,9 @@ add_filter( 'the_content', __NAMESPACE__ . '\render' );
 add_filter( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 
 /**
- * Check if the current page is the feedback form.
+ * Check if the current page should include the feedback form.
  */
-function is_feedback_page() {
+function has_feedback_form() {
 	global $wp_query;
 	return false !== $wp_query->get( QUERY_VAR, false );
 }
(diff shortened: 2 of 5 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
```diff
@@ -1,16 +1,29 @@
...
+use const WordCamp\SpeakerFeedback\QUERY_VAR;
 use function WordCamp\SpeakerFeedback\{ get_views_path, get_assets_url };
 
 add_filter( 'the_content', __NAMESPACE__ . '\render' );
 add_filter( 'wp_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 
+/**
+ * Check if the current page is the feedback form.
+ */
+function is_feedback_page() {
```
coreymckrill: The name of this function is confusing since it actually is for a different purpose than `is_page( get_option( 'feedback_page' ) )`. Maybe something like `wants_feedback_form()`?

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -9,6 +9,8 @@
 require_once 'inc/privacy.php';
 require_once 'inc/deprecated.php';
 
+define( 'EP_SESSIONS', 8192 );
```
coreymckrill: My understanding of rewrite endpoints and EP masks is quite limited, but since this is the next power of 2 after the last mask defined in Core, how likely are we to have a mask collision here with another plugin?

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
ryelle: Yeah, I went back and forth a bit on it. Maybe `is_session_feedback_page()`? I try to stick to `is_`/`has_` names when a function returns a boolean.

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
ryelle: I can bump it up, but right now nothing else uses a custom ep_mask in our codebase or installed plugins (and there's no guarantee the other plugin wouldn't do the same).

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
coreymckrill: I think `is_session_feedback_page()` has the same problem, that it sounds like it is referring to the actual page, rather than a single session that includes the feedback form.

`has_feedback_form()`?

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
coreymckrill: This just seems fragile. I guess it's a limitation in Core though, that there isn't a good way to check for the existence of a custom EP mask integer before registering a new one. And/or dynamically assign the integers for the masks at runtime.

Maybe just add a code comment here explaining what this number means, and linking to EP mask documentation, so it doesn't just look like a magic arbitrary number.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
ryelle: > that it sounds like it is referring to the actual page

It is referring to the feedback form page though, with this approach there is no "main" feedback form, just the navigation dropdown from #357  & the feedback forms on the session pages.

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
ryelle: I agree, I think it is a core limitation. I'm open to a different approach if you think it's too fragile, otherwise I'll add the comment in.

coreymckrill: I don't have any other ideas at the moment, so comment it is!

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
coreymckrill: By "actual page" I mean `site.worcamp.org/feedback/` where the dropdown is.

I think I'm getting tripped up by having "page" be in the name here, since that means something very specific in the WP context.

All the function is actually doing is checking for the existence of a particular query parameter, right? And the intention the query parameter is signaling is that we want to append a feedback form to the current view. In the future, that might even be for other post types besides sessions (e.g. `workshop` or something). That's why I think it would be best to use something like `has_feedback_form` or `wants_feedback_form`.

ryelle: Good point about other session types. I've updated it to `has_feedback_form`.

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
ryelle: Added the comment (and bumped the EP one more power-of-two just in case).

coreymckrill (approved):

PR #360, opened 2020-02-08
Refactor sponsor invoice update/notify system to avoid issues caused by `switch_to_blog`

### LINKED ISSUE #358: Sponsor invoice reminder emails sent from wrong site
Every site on the network has a cron job set that queries the sponsor invoices index for invoices that have had the "pending" status for a certain amount of time, and sends out email reminders to the camp organizers asking them to follow up with the sponsors.

The problem is, these emails are sent from whichever site happened to trigger the cron job, using the locale of that site. This sometimes causes an issue, such as when an email about an invoice for WordCamp Chicago is sent to the WC Chicago lead organizer, and appears to come from WordCamp Wuerzberg, and the content is in German 😬 

```
From: wordpress@2020.wuerzburg.wordcamp.org <wordpress@2020.wuerzburg.wordcamp.org> On Behalf Of WordCamp Central
Sent: Sunday, February 2, 2020 5:25 AM
To: {organizer email address}
Subject: Ausstehende Rechnung: [390] CWM Sponsor Invoice
```

coreymckrill:
Removes `switch_to_blog` wherever possible in the ecosystem of functions that processes and notifies about changes to sponsor invoices.

**To do:**

* [ ] ~Maybe make email message strings translatable that currently aren't~ This PR is complicated enough, we can do this later.
* [x] Add more temporary logging or email to devs for tracking/troubleshooting

Fixes #358 

**To test:**

1. Follow all the testing steps in #371 to create two or three invoices and check that they have all been sent over to the QBO sandbox. When you click the "Approve" button on an invoice in the network Invoices dashboard, you should be able to refresh the screen and see that it has moved from the "Submitted" tab to the "Sent" tab. Then also in the Sponsor Invoices list table over on the test site, you should see that the status of the invoice changes as well. (Success here indicates that both the post status and the entry in the index table are being updated correctly)
1. In the QBO sandbox company dashboard, on the Invoices screen, choose one of the invoices that you just created and click the button to "Receive payment"
1. Open up MailCatcher so you can see email notifications being sent from the test environment.
1. In order to test the cron job that checks for paid invoices, you can use wp-cli to immediately trigger a scheduled event. So first, test that running the cron job on a different site from the one where the invoices are does _not_ mark anything as paid:
    `wp --allow-root --url=https://wrongsite.wordcamp.test cron event run wcbdsi_check_for_paid_invoices`
1. Then go to the site that does have the sponsor invoices:
    `wp --allow-root --url=https://rightsite.wordcamp.test cron event run wcbdsi_check_for_paid_invoices`
1. The invoice that received payment should move from "Sent" to "Paid" in the network Invoices dashboard, and the status of the invoice post in the test site should also update to Paid. MailCatcher should have an email about the invoice being paid. Note that this email notification is not translatable, so it will be in `en_US`.
1. In order to test the pending invoice reminder email cron job, you'll need to add a temporary hack to the code. In the `send_invoice_pending_reminder()` function, comment out lines 551 through 561, so it will send for all pending invoices, even though you just created them today, and also not mark them as defaulted in the case that your test WordCamp start date is already in the past.
1. Change the locale to something other than `en_US` (to test whether this email notification, which is translatable, is sent using the correct locale).
1. As with the above tests for finding paid invoices, try first on the wrong site to make sure no reminder emails are sent, and then from the correct site to make sure the emails _are_ sent, and in the correct locale:
    `wp --allow-root --url=https://wrongsite.wordcamp.test cron event run send_invoice_pending_reminder`
    `wp --allow-root --url=https://rightsite.wordcamp.test cron event run send_invoice_pending_reminder`
1. When finished, don't forget to remove the temporary hack 😄

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments-network/includes/sponsor-invoices-dashboard.php
+++ b/public_html/wp-content/plugins/wordcamp-payments-network/includes/sponsor-invoices-dashboard.php
@@ -496,7 +496,7 @@ function delete_index_row( $invoice_id ) {
  *
  * This runs as a cron job on every site, so it only needs to look for invoices that are from the current site.
  * (Previously it pulled all pending invoices from the index table, which caused weird issues like the email about
- * an invoice in Chicago coming from a WordCamp site in Germany.)
+ * an invoice in Chicago coming from a WordCamp site in Germany, translated in `de_DE`.)
  *
  * @return void
  */
--- a/public_html/wp-content/plugins/wordcamp-payments-network/includes/sponsor-invoices-dashboard.php
+++ b/public_html/wp-content/plugins/wordcamp-payments-network/includes/sponsor-invoices-dashboard.php
@@ -1,7 +1,9 @@
 <?php
 
 namespace WordCamp\Budgets_Dashboard\Sponsor_Invoices;
+
 use \WordCamp\Logger;
+use const WordCamp\Budgets\Sponsor_Invoices\POST_TYPE;
 
 defined( 'WPINC' ) or die();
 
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wordcamp-payments-network/includes/sponsor-invoices-dashboard.php
```diff
@@ -481,32 +483,29 @@ function schedule_sent_invoice_reminder() {
 
 /**
  * Send reminder to organizer about the unpaid invoice.
+ *
+ * This runs as a cron job on every site, so it only needs to look for invoices that are from the current site.
+ * (Previously it pulled all pending invoices from the index table, which caused weird issues like the email about
+ * an invoice in Chicago coming from a WordCamp site in Germany.)
```
iandunn:
```suggestion
 * an invoice in Chicago coming from a WordCamp site in Germany, translated in `de_DE`.)
```

iandunn (commented):
The code itself LGTM 👍 

You mentioned some testing problems on Slack, how did you end up testing it?

Beyond verifying that the emails are sent w/ the expected "From" and locale, one side-effect I'd check would be to make sure the index table is kept in sync w/ the site's `wp_N_posts` table.

coreymckrill:
This now has testing instructions since #371 has been merged.

> Beyond verifying that the emails are sent w/ the expected "From" and locale, one side-effect I'd check would be to make sure the index table is kept in sync w/ the site's wp_N_posts table.

This is a good point. The instructions include verification of this.

PR #361, opened 2020-02-10
CampTix: Improve ticket form accessibility

ryelle:
Fixes multiple instances of unlabelled inputs in the registration flow for screen reader users. Add headings & table structure hints to the available ticket table, and label for the quantity dropdown. Add & attach labels to the registration form inputs, including addons & custom fields. Wrap the checkboxes & radios into `fieldset`s, using [hidden labels](https://codepen.io/stevef/post/fielsdet-and-legend-semantics) to attach the question text.

Fixes https://meta.trac.wordpress.org/ticket/1591

### Screenshots

There should be no visual changes.

1. Go through the ticket buying process using a screen reader
2. Every input you hit should have an audible label

(alternately you can use dev tools to check the accessible name of each input)

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/addons/privacy.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/addons/privacy.php
@@ -58,6 +58,51 @@ class Privacy_Field extends CampTix_Addon {
 	}
 
 	/**
+	 * Render the field, used on both Registration and when editing an existing ticket.
+	 *
+	 * @param string $name
+	 * @param string $value
+	 */
+	public function render_field( $name, $value ) {
+		?>
+
+		<tr class="tix-row-<?php echo esc_attr( self::SLUG ); ?>">
+			<td class="tix-left">
+				<?php echo wp_kses_post( $this->question ); ?>
+				<span aria-hidden="true" class="tix-required-star">*</span>
+			</td>
+
+			<td class="tix-right">
+				<fieldset class="tix-screen-reader-fieldset" aria-label="<?php echo esc_attr( $this->a11y_label ); ?>">
+					<label>
+						<input
+							name="<?php echo esc_attr( $name ); ?>"
+							type="radio"
+							value="yes"
+							<?php checked( 'yes', $value ); ?>
+							required
+						/>
+						<?php echo esc_html( $this->options['yes'] ); ?>
+					</label>
+					<br />
+					<label>
+						<input
+							name="<?php echo esc_attr( $name ); ?>"
+							type="radio"
+							value="no"
+							<?php checked( 'no', $value ); ?>
+							required
+						/>
+						<?php echo esc_html( $this->options['no'] ); ?>
+					</label>
+				</fieldset>
+			</td>
+		</tr>
+
+		<?php
+	}
+
+	/**
 	 * Render the new field for the registration form during checkout.
 	 *
 	 * @param array $form_data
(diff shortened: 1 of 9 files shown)


public_html/wp-content/mu-plugins/camptix-tweaks/addons/privacy.php
```diff
@@ -78,15 +79,17 @@ public function render_registration_field( $form_data, $i ) {
...
-				<label>
-					<input name="tix_attendee_info[<?php echo esc_attr( $i ); ?>][<?php echo esc_attr( self::SLUG ); ?>]" type="radio" value="yes" <?php checked( 'yes', $current_data[ self::SLUG ] ); ?> required />
-					<?php echo esc_html( $this->options['yes'] ); ?>
-				</label>
-				<br />
-				<label>
-					<input name="tix_attendee_info[<?php echo esc_attr( $i ); ?>][<?php echo esc_attr( self::SLUG ); ?>]" type="radio" value="no" <?php checked( 'no', $current_data[ self::SLUG ] ); ?> required />
-					<?php echo esc_html( $this->options['no'] ); ?>
-				</label>
+				<fieldset class="tix-screen-reader-fieldset" aria-label="<?php echo esc_attr( $this->a11y_label ); ?>">
```
coreymckrill, line 76: Firefox's a11y tool is [telling me](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names) that the label for groups like this should be in a `<legend>` element instead of an `aria-label`.

But it does pick up the correct name of the group. I'm not suggesting this has to be updated, just commenting in case you think this is an optimization that would be worth making.

public_html/wp-content/mu-plugins/camptix-tweaks/addons/privacy.php
```diff
@@ -78,15 +79,17 @@ public function render_registration_field( $form_data, $i ) {
 			</td>
```
coreymckrill, line 73: Seems like the `tix-required-star` in the line above should also have `aria-hidden`.

coreymckrill (commented):
👍 LGTM

Given that there are some minor DOM hierarchy changes (e.g. wrapping some stuff in labels and fieldsets), do you think there's a possibility that this could break the CSS styles for any current WordCamp sites? If so, we should probably post about the changes in Make/Community and maybe proactively test current sites and offer CSS fixes for anything that breaks because of this.

public_html/wp-content/mu-plugins/camptix-tweaks/addons/privacy.php
ryelle: Yeah, a `legend` is the proper HTML approach, but that's a visible element, so I would need to intentionally hide it with `screen-reader-text`. `<fieldset>` is "just" a special group with the legend automatically labelling it. According to [this post,](https://codepen.io/stevef/post/fielsdet-and-legend-semantics) the same result can be achieved with an `aria-label` when you don't want a visible label.

public_html/wp-content/mu-plugins/camptix-tweaks/addons/privacy.php
ryelle: Thanks for pointing that out — I also noticed that the edit form used a different render function, so I've also gone through and updated the "Edit Ticket" fields too.

coreymckrill (approved):

ryelle:
> do you think there's a possibility that this could break the CSS styles for any current WordCamp sites?

I thought this was pretty remote, since most camps don't do anything fancy on these pages, but I checked WCEU (since they are highly customized), and yep— it's janky with this PR thanks to direct-children selectors. I'll write up a note for the community P2 before I merge this.

PR #364, opened 2020-02-13
Blocks: Update packages, fix deprecations & new configurations

ryelle:
Update the `@wordpress/*` packages, which includes some configuration changes and component prop deprecations. Remove deprecated `isDefault` prop from `<Button />`. Update eslint config to use `@wordpress/eslint-plugin/recommended-with-formatting`, which uses eslint for formatting rules instead of prettier. This prevents the prettier ruleset from running before we're ready to use it. Lastly, this adds the prettier config & format command, so we can test out whether we want to adopt the auto-formatting features.

1. Run `yarn` to update local packages
2. Building, testing, etc should all work
3. Linters should pass
4. Try running `yarn workspace wordcamp-blocks run format:js` to test out the prettier commands

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/.eslintrc.js
+++ b/public_html/wp-content/mu-plugins/blocks/.eslintrc.js
@@ -8,7 +8,7 @@
  */
 
 module.exports = {
-	extends : 'plugin:@wordpress/eslint-plugin/recommended',
+	extends : 'plugin:@wordpress/eslint-plugin/recommended-with-formatting',
 
 	globals : {
 		wp : true,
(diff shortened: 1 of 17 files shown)


coreymckrill:
Haven't fully tested this yet, but just wanted to note that the Button component's `isLarge` is being deprecated in WP 5.4:

https://make.wordpress.org/core/2020/02/13/an-updated-button-component-in-wordpress-5-4/

So may be worth taking that into account for this PR, since we're making other button changes here as well.

ryelle:
Fixed the merge conflict, and removed `isLarge`. It's not officially deprecated, which is why I missed it — but it doesn't do anything anymore, so it should be removed.

coreymckrill (approved):

PR #365, opened 2020-02-14
Changes to the Meetup organizer application form

timiwahalahti:
This PR contains Meetup organizer application changes as per #321 and discussed to separate into own PR in #333.

## How to test the changes in this Pull Request:

* Go to https://central.wordcamp.test/meetup-organizer-application
* You should see new fields
* Submit an application
* Data inputted to new fields is visible on "Original application" metabox

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
@@ -181,7 +181,7 @@ class Meetup_Application extends Event_Application {
 			'q_mtp_loc'             => '',
 			'q_already_a_meetup'    => '',
 			'q_existing_meetup_url' => '',
-			'q_introduction',				=> '',
+			'q_introduction'				=> '',
 			'q_socialmedia'					=> '',
 			'q_reasons_plans'				=> '',
 			'q_community_interest'	=> '',
--- a/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
@@ -141,6 +141,7 @@ class Meetup_Application extends Event_Application {
 			'q_introduction',
 			'q_socialmedia',
 			'q_reasons_plans',
+			'q_community_interest',
 			'q_wporg_username',
 		);
 
@@ -183,6 +184,7 @@ class Meetup_Application extends Event_Application {
 			'q_introduction',				=> '',
 			'q_socialmedia'					=> '',
 			'q_reasons_plans'				=> '',
+			'q_community_interest'	=> '',
 			'q_wporg_username'      => '',
 			'q_wp_slack_username'   => '',
 		);
--- a/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
@@ -138,6 +138,7 @@ class Meetup_Application extends Event_Application {
 			'q_country',
 			'q_mtp_loc',
 			'q_already_a_meetup',
+			'q_introduction',
 			'q_wporg_username',
 		);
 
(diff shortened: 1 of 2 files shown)


public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
```diff
@@ -228,11 +234,9 @@ public function create_post( $data ) {
 		add_post_meta( $post_id, 'Address', $organizer_address );
 		add_post_meta( $post_id, 'Already a meetup', $data['q_already_a_meetup'] );
 		add_post_meta( $post_id, 'Meetup URL', $data['q_existing_meetup_url'] );
-		add_post_meta( $post_id, 'Organizer description', $data['q_describe_yourself'] );
```
ryelle, line 231: This & the "Extra Comments" meta are removed here, but still show up in the application in wp-admin. I'm not familiar with the application flow, but maybe this should be filled in with the content from `q_introduction`?

public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
```diff
@@ -228,11 +234,9 @@ public function create_post( $data ) {
...
 		add_post_meta( $post_id, 'Primary organizer WordPress.org username', $data['q_wporg_username'] );
 		add_post_meta( $post_id, 'Slack', $data['q_wp_slack_username'] );
 		add_post_meta( $post_id, 'Date Applied', time() );
-		add_post_meta( $post_id, 'Extra Comments', $data['q_additional_comments'] );
```
ryelle, line 235: Same question - should this save the content from `q_anything_else`?

ryelle (commented):
This is working as you described 👍 

I have two questions below about saving the answers to post meta so they can be displayed in the "Organizer Information" metabox in wp-admin, but other than that this is basically ready to merge 🙂

public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
timiwahalahti: IMO those fields are a bit unnecessary because you need to scroll down to the original application nevertheless for seeing all the information in the application. Removing this field makes the flow more similar to WordCamp applications.

public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
timiwahalahti: Same answer than above :)

public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-application.php
ryelle: Like I said, I don't know this flow, but my intuition says this should go one of two ways:

- If these fields are not used for anything, they should be removed from the UI entirely
- If these fields are used, for example if Application Reviewers edit things here, we should still pre-fill them, even if you use the raw data to review.

What do you think?

(We can also open a new issue to better format these answers so that you don't need to look at a raw array dump 🙂 )

timiwahalahti: > If these fields are not used for anything, they should be removed from the UI entirely

I'm not 100% sure if someone uses these fields for something. Usually, these fields shouldn't be edited. I'd agree on removing these fields from UI after some period after the changes to the application form.

> We can also open a new issue to better format these answers so that you don't need to look at a raw array dump

I also thought this and making a PR at some point. For now, we can work with raw array dump like with WordCamp applications - changes to Meetup application form are more urgent as we need to ask more information almost on every application sent.

ryelle: Started another issue to track this idea of improving those metaboxes: https://github.com/WordPress/wordcamp.org/issues/372

ryelle:
Well, I merged this outside of github, and it didn't attach correctly to close this PR, but it has been 
merged (https://github.com/WordPress/wordcamp.org/commit/4a4ad193509e14f1cbc9205619b69930808e42a8) & deployed.

Thanks for making the PR!  🎉

PR #367, opened 2020-02-14
Speaker Feedback: Create a yarn workspace, use wp-scripts & node-sass

ryelle:
This adds a new workspace to the project, `wordcamp-speaker-feedback`. Currently it only builds the sass -> css file, and runs linters.  Eventually we may use wp-scripts to build the javascript file. This sets the groundwork for that.

### How to test the changes in this Pull Request:

1. Run yarn
2. Try running the new commands in this workspace, ex: 
    - `yarn workspace wordcamp-speaker-feedback run lint:css`
    - `yarn workspace wordcamp-speaker-feedback run lint:js`
    - `yarn workspace wordcamp-speaker-feedback run build`
    - `yarn workspace wordcamp-speaker-feedback run start`

PR #368, opened 2020-02-14
Speaker Feedback: Add the star rating UI

ryelle:
This adds a real star rating input to the form.

See #334

Note, this is built on #367 so that I can use sass.

### Screenshots

Screen Shot 2020-02-14 at 3 25 35 PM

1. View a session feedback form
2. Use the star rating
3. It should work with mouse and keyboard controls

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
@@ -1,6 +1,7 @@
 <?php
 
 namespace WordCamp\SpeakerFeedback\View;
+use function WordCamp\SpeakerFeedback\get_assets_path;
 
 ?>
 <hr />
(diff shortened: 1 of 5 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
```diff
@@ -7,8 +7,51 @@
...
+				value="0"
+				checked
+				aria-label="<?php esc_attr_e( 'No stars', 'wordcamporg' ); ?>"
+			/>
+			<input type="radio" id="sft-rating-val-1" name="feedback-rating" value="1" />
+			<label for="sft-rating-val-1">
+				<span class="screen-reader-text"><?php esc_html_e( '1 star', 'wordcamporg' ); ?></span>
+				<svg xmlns="http://www.w3.org/2000/svg" viewbox="0 0 20 20" height="20" width="20" role="img" aria-hidden="true">
+					<path d="M10 1l3 6 6 .75-4.12 4.62L16 19l-6-3-6 3 1.13-6.63L1 7.75 7 7z"/>
+				</svg>
```
coreymckrill: Would it be worth putting this in an include file or variable, since it's the same thing five times? I could see us reusing it elsewhere to display an average rating or something.

ryelle: I've updated this to `require get_assets_path() . 'svg/star.svg';` , which should work fine as long as we [don't add in the XML header tag.](https://sheelahb.com/blog/how-to-get-php-to-play-nicely-with-svg-files/)

coreymckrill: Could it just be in an image tag instead?

ryelle: No, because it uses CSS to change the fill color, and we can't do that unless the SVG is inlined

coreymckrill: Ah, ok. 👍

coreymckrill (approved):

PR #369, opened 2020-02-14
Speaker Feedback: Submit feedback form to API

### LINKED ISSUE #334: Speaker Feedback: Create session feedback form

ryelle:
Submit the feedback form to the API endpoint. Passes in author data (user ID if logged in, name/email if not), rating & questions as meta. Shows a success or failure notice after submission.

Fixes #334 

### Screenshots

Logged in:
Screen Shot 2020-03-18 at 2 28 46 PM

Logged out:
Screen Shot 2020-03-18 at 2 24 03 PM

Success, replaces the form with message:
Screen Shot 2020-03-16 at 5 54 43 PM

Error, message comes from API response:
Screen Shot 2020-04-02 at 6 39 47 PM

1. Go to a feedback form, ex: `/session/[talk]/feedback`
2. Fill out the feedback form, submit it
3. The page should not redirect
4. Open network inspector, You should see a POST to `/wordcamp-speaker-feedback/v1/feedback`
5. A message should appear, success or failure
6. If success, you should also see a comment record in the Session Feedback in wp-admin (unapproved)
7. If failure, it should show you the error message, and let you re-try submission

You can trigger a failure message by not selecting a rating - the other inputs have browser validation 🙂

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -14,7 +14,6 @@
 		var rawData = new FormData( event.target );
 		var data = {
 			post: rawData.get( 'sft-post' ),
-			author: rawData.get( 'sft-author' ),
 			meta: {
 				rating: rawData.get( 'sft-rating' ),
 				q1: rawData.get( 'sft-question-1' ),
@@ -22,6 +21,14 @@
 				q3: rawData.get( 'sft-question-3' ),
 			},
 		};
+
+		var author = rawData.get( 'sft-author' );
+		if ( '0' !== author ) {
+			data.author = author;
+		} else {
+			data.author_name = rawData.get( 'sft-author-name' );
+			data.author_email = rawData.get( 'sft-author-email' );
+		}
 
 		var messageContainer = document.getElementById( 'speaker-feedback-notice' );
 		// Reset the notice before submission.
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -23,11 +23,25 @@
 			},
 		};
 
+		var messageContainer = document.getElementById( 'speaker-feedback-notice' );
+		// Reset the notice before submission.
+		messageContainer.setAttribute( 'class', '' );
+		messageContainer.innerText = '';
+
 		wp.apiFetch( {
 			path: '/wordcamp-speaker-feedback/v1/feedback',
 			method: 'POST',
 			data: data,
-		} );
+		} )
+			.then( function() {
+				messageContainer.setAttribute( 'class', 'speaker-feedback__notice is-success' );
+				messageContainer.innerText = 'Feedback submitted.';
+				event.target.replaceWith( messageContainer );
+			} )
+			.catch( function( error ) {
+				messageContainer.setAttribute( 'class', 'speaker-feedback__notice is-error' );
+				messageContainer.innerText = error.message;
+			} );
 	}
 
 	var navForm = document.getElementById( 'sft-navigation' );
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -13,10 +13,14 @@
 		event.preventDefault();
 		var rawData = new FormData( event.target );
 		var data = {
-			rating: rawData.get( 'sft-rating' ),
-			q1: rawData.get( 'sft-question-1' ),
-			q2: rawData.get( 'sft-question-2' ),
-			q3: rawData.get( 'sft-question-3' ),
+			post: rawData.get( 'sft-post' ),
+			author: rawData.get( 'sft-author' ),
+			meta: {
+				rating: rawData.get( 'sft-rating' ),
+				q1: rawData.get( 'sft-question-1' ),
+				q2: rawData.get( 'sft-question-2' ),
+				q3: rawData.get( 'sft-question-3' ),
+			},
 		};
 
 		wp.apiFetch( {
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -4,13 +4,35 @@
 ( function() {
 	function onFormNavigate( event ) {
 		event.preventDefault();
-		const value = event.target[ 0 ].value;
+		var value = event.target[ 0 ].value;
 		// Use the fact that post IDs will redirect to the right page.
 		window.location = '/?p=' + value + '&sft_feedback=1';
 	}
 
-	const form = document.getElementById( 'sft-navigation' );
-	if ( form ) {
-		form.addEventListener( 'submit', onFormNavigate, true );
+	function onFormSubmit( event ) {
+		event.preventDefault();
+		var rawData = new FormData( event.target );
+		var data = {
+			rating: rawData.get( 'sft-rating' ),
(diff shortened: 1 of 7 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
```diff
@@ -4,13 +4,60 @@
...
+		event.preventDefault();
+		var rawData = new FormData( event.target );
+		var data = {
+			post: rawData.get( 'sft-post' ),
+			meta: {
+				rating: rawData.get( 'sft-rating' ),
+				q1: rawData.get( 'sft-question-1' ),
+				q2: rawData.get( 'sft-question-2' ),
+				q3: rawData.get( 'sft-question-3' ),
+			},
```
coreymckrill: Could this be abstracted so that we don't have to add each meta key explicitly? That would allow us to avoid having to update this script if the question configuration changes in the future.

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
```diff
@@ -6,65 +6,105 @@
...
+			</label>
+			<input type="email" id="sft-author-email" name="sft-author-email" required />
+		</div>
+	</div>
+	<?php else : ?>
+		<div class="speaker-feedback__notice">
+			<p><?php
+			$user = wp_get_current_user();
+			echo wp_kses_post( sprintf(
+				__( 'You are leaving feedback as %s.', 'wordcamporg' ),
```
coreymckrill, line 43: Maybe more of a question for a designer, but I'm wondering if it would make sense here to emphasize the username with `<strong>` or a link to their wporg profile. Right now it blends in a bit too well with the rest of the notification message and my eye skips right over it.

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
```diff
@@ -6,65 +6,105 @@
...
 				<span class="screen-reader-text"><?php esc_html_e( '5 stars', 'wordcamporg' ); ?></span>
 				<?php require get_assets_path() . 'svg/star.svg'; ?>
 			</label>
 		</fieldset>
 	</div>
-	
+
 	<div class="speaker-feedback__field">
 		<label for="sft-question-1">
 			<?php esc_html_e( 'What’s one good thing you’d keep in this presentation?', 'wordcamporg' ); ?>
```
coreymckrill: The question text should probably get populated via `CommentMeta\get_feedback_questions()` now that that is available. Similar to my other comment about abstracting the post data, maybe it would even make sense here to use a `foreach` loop to output the question fields so that if the question configuration changes in the future, we only have to change it in the `CommentMeta\get_feedback_questions()` function (and the schema function) but not several other places as well.

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
```diff
@@ -6,65 +6,105 @@
...
+				$user->display_name
+			) ); ?></p>
+		</div>
+	<?php endif; ?>
+
+	<div id="speaker-feedback-notice" aria-live="polite" aria-relevant="additions text" aria-atomic="true"></div>
+
+	<div class="speaker-feedback__field">
+		<fieldset class="speaker-feedback__field-rating">
+			<legend><?php esc_html_e( 'Rate this talk', 'wordcamporg' ); ?></legend>
```
coreymckrill: Can we populate this label from `CommentMeta\get_feedback_questions()`? And also conditionally include the `required` attribute on the inputs via `CommentMeta\get_feedback_meta_field_schema()`?

public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
```diff
@@ -4,13 +4,60 @@
...
+		} )
+			.then( function() {
+				messageContainer.setAttribute( 'class', 'speaker-feedback__notice is-success' );
+				messageContainer.innerText = 'Feedback submitted.';
+				event.target.replaceWith( messageContainer );
+			} )
+			.catch( function( error ) {
+				messageContainer.setAttribute( 'class', 'speaker-feedback__notice is-error' );
+				messageContainer.innerText = error.message;
+			} );
```
coreymckrill: The API response when there's a meta validation error includes a `data` attribute that contains the key of each field that was invalid and an explanation. Would it make sense to output those as well?

Example, when you skip the rating:

Instead of just "One or more fields has invalid data." it could be

```
One or more fields has invalid data.
* rating: This value is too low
```

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
ryelle: sure (to this & below) — I didn't want to change the questions around here too much, and I started this PR before that function existed. Is changing that a blocker to merge this PR?

coreymckrill: No, fine to do this in a separate PR

public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
ryelle: Not easily, but the API also has these keys hardcoded so I don't think it's worth changing (if we added a `q4`, the API would need to change too).

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
ryelle: Done in 94eb772b99d315dd992dc3bf7aa744580a33e5d6

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
ryelle: Ended up doing this here after all— 69f3a5cad87a6dfe9df2098a1e6c89250ea6d5e2

public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
ryelle: Added these error messages to their relevant inputs in a1b0be0c54ae2ab57bf493bbd2b31a22d77a7532 & updated the screenshot in the description.

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
ryelle: Oh, since we don't identify field type in `get_feedback_questions`, I can't do a straight loop of these questions — I can refactor that (& the JS) in another PR if you think we need that (I don't think we do, but that's a very weakly-held opinion).

public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
coreymckrill: Fair enough. I think it might be worth adding a reminder comment somewhere in `get_feedback_questions()` or `get_feedback_meta_field_schema()` noting that if more fields are added or changed, the JS will have to change as well.

coreymckrill (approved):
👍 Looking good on local dev.

One enhancement for a future PR that I think would be nice would be to add focus or scrolling to the success/error message after the form is submitted. Right now when I successfully submit the form, everything collapses upward and I can't see the green success message unless I scroll back up manually.

ryelle:
> add focus or scrolling to the success/error message after the form is submitted

oh, good idea— the test site i've been using has almost no footer content so that hasn't been a problem

PR #371, opened 2020-02-18
QBO: Make it ~possible~ easier to test in a local dev environment

coreymckrill:
This adds a `local` value for the `WORDCAMP_ENVIRONMENT` constant, and makes various places in the code behave differently when this value is set. This makes it much easier to test interactions with the QBO API and between our own QBO client/server setup.

**To test**

This is specifically designed to work in the Docker environment, but you can probably make it work in other local environments if you change your `WORDCAMP_ENVIRONMENT` constant value to `local` (and maybe update other values in your **wp-config.php** file, see the changeset)

1. After you check out this branch, you may need to rebuild your Docker container to make sure it has the right values in **wp-config.php**.
1. Make sure you have a **sandbox-functionality.php** file in the **mu-plugins** directory.
1. Log in to [Intuit's developer site](https://developer.intuit.com) and go to the My Apps tab. "WordCamp Quickbooks" is the app we want. Go to the Keys & OAuth tab under Development.
1. Add this snippet to your **sandbox-functionality.php** file, using the Client ID and Client Secret from the Intuit app:
    ```php
	// QBO
	add_filter( 'wordcamp_qbo_client_config', function( $config = array() ) {
		$config = array_merge(
			$config,
			array(
				'ClientID'     => 'paste value here',
				'ClientSecret' => 'paste value here',
				'baseUrl'      => 'Development',
			)
		);
	
		return $config;
	}, 99 );
    ```
1. In another browser tab, go to the [Quickbooks settings page](https://wordcamp.test/wp-admin/network/settings.php?page=quickbooks) in the test WordCamp network admin. Click _Connect_ and go through the authentication flow. Choose "Sandbox WPCS" for the company. At the end, you should end up back on the Quickbooks settings page, and it should say "Connected to Sandbox WPCS. Expires in 3 months."
1. Now open a third browser tab and go to one of the other test WordCamp sites. Create a new sponsor invoice, and submit it.
1. Back on the browser tab with the network admin, go to the [Sponsor Invoices Dashboard](https://wordcamp.test/wp-admin/network/admin.php?page=sponsor-invoices-dashboard). You should see the invoice you just created.
1. Click the _Approve_ button on the invoice. This sends the invoice to QBO. You should get a green success notice.
1. Back on the Intuit developer browser tab, under the API Docs & Tools tab, click Sandbox. Then choose the Sandbox WPCS company. This will take you to a QBO dashboard.
1. Click the Invoices item under the Sales tab. You should see the invoice you just approved.

### DIFF
--- a/.docker/wp-config.php
+++ b/.docker/wp-config.php
@@ -103,6 +103,7 @@ define( 'TWITTER_CONSUMER_KEY_WORDCAMP_CENTRAL',    '' );
 define( 'TWITTER_CONSUMER_SECRET_WORDCAMP_CENTRAL', '' );
 define( 'TWITTER_BEARER_TOKEN_WORDCAMP_CENTRAL',    '' );
 
+// These can't be empty, but no real values are needed here for local development.
 define( 'WORDCAMP_QBO_APP_TOKEN',       'localtoken' );
 define( 'WORDCAMP_QBO_CONSUMER_KEY',    'localkey' );
 define( 'WORDCAMP_QBO_CONSUMER_SECRET', 'localsecret' );
--- a/.docker/wp-config.php
+++ b/.docker/wp-config.php
@@ -1,6 +1,6 @@
 <?php
 
-define( 'WORDCAMP_ENVIRONMENT', 'development' );
+define( 'WORDCAMP_ENVIRONMENT', 'local' );
 
 /*
  * Database
--- a/.docker/wp-config.php
+++ b/.docker/wp-config.php
@@ -1,5 +1,18 @@
 <?php
+/**
+ * WordCamp.org configuration file.
+ */
 
+/**
+ * The environment in which the WordCamp codebase is currently running.
+ *
+ * There are a few different values that this might contain that have implications in the code:
+ *
+ * - `production`:  Used on the production server. Should not be used anywhere else.
+ * - `development`: The catchall value for non-production environments. Currently used on wporg sandboxes.
+ * - `local`:       The value used for local development environments, where the domain is wordcamp.test and some
+ *                  functionality that relies on remote connections may not be available.
+ */
 define( 'WORDCAMP_ENVIRONMENT', 'local' );
 
 /*
--- a/.docker/wp-config.php
+++ b/.docker/wp-config.php
@@ -103,13 +103,13 @@ define( 'TWITTER_CONSUMER_KEY_WORDCAMP_CENTRAL',    '' );
 define( 'TWITTER_CONSUMER_SECRET_WORDCAMP_CENTRAL', '' );
 define( 'TWITTER_BEARER_TOKEN_WORDCAMP_CENTRAL',    '' );
 
-define( 'WORDCAMP_QBO_APP_TOKEN',       '' );
-define( 'WORDCAMP_QBO_CONSUMER_KEY',    '' );
-define( 'WORDCAMP_QBO_CONSUMER_SECRET', '' );
-define( 'WORDCAMP_QBO_HMAC_KEY',        '' );
+define( 'WORDCAMP_QBO_APP_TOKEN',       'localtoken' );
+define( 'WORDCAMP_QBO_CONSUMER_KEY',    'localkey' );
+define( 'WORDCAMP_QBO_CONSUMER_SECRET', 'localsecret' );
+define( 'WORDCAMP_QBO_HMAC_KEY',        'localhmac' );
 
-define( 'WORDCAMP_SANDBOX_QBO_CLIENT_ID',        '' );
-define( 'WORDCAMP_SANDBOX_QBO_CLIENT_SECRET',    '' );
+define( 'WORDCAMP_SANDBOX_QBO_CLIENT_ID',        "(Optional) your app's development client ID key goes here" );
+define( 'WORDCAMP_SANDBOX_QBO_CLIENT_SECRET',    "(Optional) your app's development client secret key goes here" );
 define( 'WORDCAMP_PRODUCTION_QBO_CLIENT_ID',     '' );
 define( 'WORDCAMP_PRODUCTION_QBO_CLIENT_SECRET', '' );
 
--- a/public_html/wp-content/mu-plugins/wcorg-misc.php
+++ b/public_html/wp-content/mu-plugins/wcorg-misc.php
@@ -4,23 +4,6 @@
  * Miscellaneous snippets that don't warrant their own file
  */
 
-/**
- * Get the current environment.
- *
- * Defaults to 'development' if the `WORDCAMP_ENVIRONMENT` constant isn't set or is empty.
- *
- * @return string
- */
-function get_wordcamp_environment() {
-	$environment = 'development';
-
-	if ( defined( 'WORDCAMP_ENVIRONMENT' ) && WORDCAMP_ENVIRONMENT ) {
-		$environment = WORDCAMP_ENVIRONMENT;
-	}
-
-	return $environment;
-}
-
 /*
  * Prevents 'index.php' from being prepended to permalink options.
  *
--- a/public_html/wp-content/mu-plugins/wcorg-misc.php
+++ b/public_html/wp-content/mu-plugins/wcorg-misc.php
@@ -412,7 +412,10 @@ add_filter( 'wordcamp_qbo_client_options', function( $options ) {
 	}
 
 	$options['hmac_key'] = WORDCAMP_QBO_HMAC_KEY;
-	$options['api_base'] = 'https://central.wordcamp.org/wp-json/wordcamp-qbo/v1';
+	$options['api_base'] = sprintf(
+		'https://central.wordcamp.%s/wp-json/wordcamp-qbo/v1',
+		( 'local' === get_wordcamp_environment() ) ? 'test' : 'org'
+	);
 
 	return $options;
 });
--- a/public_html/wp-content/mu-plugins/wcorg-misc.php
+++ b/public_html/wp-content/mu-plugins/wcorg-misc.php
@@ -4,6 +4,22 @@
  * Miscellaneous snippets that don't warrant their own file
  */
 
+/**
+ * Get the current environment.
+ *
+ * Defaults to 'development' if the `WORDCAMP_ENVIRONMENT` constant isn't set or is empty.
+ *
+ * @return string
+ */
+function get_wordcamp_environment() {
+	$environment = 'development';
+
+	if ( defined( 'WORDCAMP_ENVIRONMENT' ) && WORDCAMP_ENVIRONMENT ) {
+		$environment = WORDCAMP_ENVIRONMENT;
+	}
+
+	return $environment;
+}
 
 /*
  * Prevents 'index.php' from being prepended to permalink options.
--- a/public_html/wp-content/plugins/wordcamp-qbo-client/wordcamp-qbo-client.php
+++ b/public_html/wp-content/plugins/wordcamp-qbo-client/wordcamp-qbo-client.php
@@ -307,7 +307,7 @@ class WordCamp_QBO_Client {
 			} elseif ( isset( $body->message ) ) {
 				$sent = $body->message;
 			} else {
-				$sent = 'Unknown error.';
+				$sent = print_r( array( $body ), true );
 			}
 		}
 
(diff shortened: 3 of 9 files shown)


ryelle:
About to start testing this, but it looks like `sandbox-functionality.php` is not ignored in git.

coreymckrill:
Oh weird, I didn't even think of that. It looks like I added it to **.git/info/exclude** at some point.

I'll add it to the real .gitignore file.

.docker/wp-config.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-define( 'WORDCAMP_ENVIRONMENT', 'development' );
+define( 'WORDCAMP_ENVIRONMENT', 'local' );
```
ryelle, line 16: Will this affect other things using `WORDCAMP_ENVIRONMENT`? A quick look seems to show that most conditionals are checking for `production` and fall back to dev settings in all other cases, so it might be fine.

ryelle (approved):
After some setup woes, it worked as expected and I was even able to debug my own issues 🙂 

I would feel better if @iandunn goes through it too, since he's got a better grasp on all the edge cases.

.docker/wp-config.php
coreymckrill: Yeah, as far as I can tell there isn't any code that is specifically looking for `development` in that constant.

.docker/wp-config.php
```diff
@@ -103,13 +103,14 @@
...
+// These can't be empty, but no real values are needed here for local development.
+define( 'WORDCAMP_QBO_APP_TOKEN',       'localtoken' );
+define( 'WORDCAMP_QBO_CONSUMER_KEY',    'localkey' );
+define( 'WORDCAMP_QBO_CONSUMER_SECRET', 'localsecret' );
+define( 'WORDCAMP_QBO_HMAC_KEY',        'localhmac' );
 
-define( 'WORDCAMP_SANDBOX_QBO_CLIENT_ID',        '' );
-define( 'WORDCAMP_SANDBOX_QBO_CLIENT_SECRET',    '' );
+define( 'WORDCAMP_SANDBOX_QBO_CLIENT_ID',        "(Optional) your app's development client ID key goes here" );
+define( 'WORDCAMP_SANDBOX_QBO_CLIENT_SECRET',    "(Optional) your app's development client secret key goes here" );
```
iandunn, line 126: Do the apps have separate dev/prod secrets? I only see the production values on the QBO site.

iandunn, line 126: > Add this snippet to your sandbox-functionality.php file, using the Client ID and Client Secret from the [production] Intuit app:

What's the reason to switch to using the production app in local environments, instead of the sandbox app (even if  sandbox companies are used in both cases) ?

How are you weighing the security implications of having production keys in local environments?

.docker/wp-config.php
iandunn: After digging through the codebase, it's starting to feel like there could be a lot of side-effects and added complexity around this. Or it's also possible that we're making different assumptions, and some more explicit code might help clarify that.

What problems did you run into while testing locally that prompted this? I don't remember ever having any, so maybe we've been doing it different ways?

What's the reason for introducing a distinction between local and remote sandboxes?

public_html/wp-content/mu-plugins/wcorg-misc.php
```diff
@@ -4,6 +4,22 @@
...
+ */
+function get_wordcamp_environment() {
+	$environment = 'development';
+
+	if ( defined( 'WORDCAMP_ENVIRONMENT' ) && WORDCAMP_ENVIRONMENT ) {
+		$environment = WORDCAMP_ENVIRONMENT;
+	}
+
+	return $environment;
+}
```
iandunn: What's the reason for falling back to `dev` here instead of `local` ? Same for `set_client_config()`.

`MIGHT` - `3-helpers-misc.php` seems like a more consistent location for this type of thing.

public_html/wp-content/plugins/wordcamp-qbo-client/wordcamp-qbo-client.php
```diff
@@ -302,7 +307,7 @@ public static function send_invoice_to_quickbooks( $invoice_id ) {
 			} elseif ( isset( $body->message ) ) {
 				$sent = $body->message;
 			} else {
-				$sent = 'Unknown error.';
+				$sent = print_r( array( $body ), true );
```
iandunn: Would `$body` ever contain any sensitive values?

.docker/wp-config.php
iandunn: Here are the specific things I think might be good to change, if my assumptions are correct. Or at the very least maybe this will help reveal where we might have different assumptions.

`SHOULD` - If we add `local`, I think we should also change the remote sandbox value to be `staging` or `remote`. `development` seems like it'd be easily confused in that context, since conventionally that means a local environment in my experience.

- [ ] `MUST`: It seems like the `wcorg_sponsor_payment_stripe` and `wordcamp_google_maps_api_key` closures in `wcorg-misc.php` would need to be updated, otherwise the keys won't be populated because there's no `local` case in the `switch`.
- [ ] `MUST`: `wcpt-event/notification.php` should probably return early on local sandboxes instead of remote.

`MIGHT` - Would it be simpler to add `case 'local'` to `set_client_config()` in `quickbooks.php`, instead of devs needing to manually add a new `wordcamp_qbo_client_config` callback in `sandbox-functionality.php`?

`MIGHT` - Right now `set_client_config()` is using the sandbox keys in local and remote sandboxes, but it might make more sense to use the production keys in `staging`/`remote` and the sandbox keys in `local`.

Related in Trello: `/c/rErUVQ6P` (first 2 uncompleted checklist items) and `/c/5exJNacO`

coreymckrill: > What problems did you run into while testing locally that prompted this?

The problem was that there were several places in the code that had to be temporarily modified in order to be able to test workflows involving QBO (hardcoded API URLs, the SSL verification setting, etc). This meant finding and changing things in several different files, leaving the working directory dirty, and introducing the risk of accidentally committing these temporary changes.

> What's the reason for introducing a distinction between local and remote sandboxes?

In a local dev environment, the domain is **wordcamp.test**. In both production and on sandboxes, the domain is **wordcamp.org**. There are other differences as well, such as that sandboxes can have two-way communication with other remote servers (e.g. Jetpack connection stuff), while local environments can't (unless you use ngrok or something)

coreymckrill: > If we add local, I think we should also change the remote sandbox value to be staging or remote.

Ok, although I don't think that needs to be part of this PR, since this is specifically about making it easier to test QBO locally.

> It seems like the wcorg_sponsor_payment_stripe and wordcamp_google_maps_api_key closures in wcorg-misc.php would need to be updated, otherwise the keys won't be populated because there's no local case in the switch.

In both of those closures, the `default` switch case catches "local", so it gets the same treatment as if it was "development".

> wcpt-event/notification.php should probably return early on local sandboxes instead of remote.

Good call. I guess we could just change the line at the top to

```php
if ( 'local' === WORDCAMP_ENVIRONMENT ) {
	return;
}
```

> Would it be simpler to add case 'local' to set_client_config() in quickbooks.php, instead of devs needing to manually add a new wordcamp_qbo_client_config callback in sandbox-functionality.php?

We don't want to publish our QBO sandbox keys in the repo, so the values would need to go in **sandbox-functionality.php** either way. Otherwise you end up putting those values in a file under version control and risk accidentally committing them.

> Right now set_client_config() is using the sandbox keys in local and remote sandboxes, but it might make more sense to use the production keys in staging/remote and the sandbox keys in local.

I think remote sandboxes should still use sandbox keys. If we're testing interactions with the QBO API on a sandbox, we wouldn't want test data to get mixed in with actual transactions.

coreymckrill: Re: this comment in Trello:

> The various usages of `WORDCAMP_ENVIRONMENT` throughout the codebase are starting to feel kind of `WET`, because they're not assuming that it exists. It does (and should/must) exist in all of our environments, though, so that feels like it's adding a lot of needless complexity. What about just removing all `defined()` calls and fallbacks etc, and just referencing the value directly everywhere?

The `get_wordcamp_environment()` function introduced in this PR is meant to address this. I didn't go through and replace all the references to the constant in the whole codebase, because that seemed out of scope, but we could easily do that in a separate PR once this is merged.

.docker/wp-config.php
coreymckrill: They do. Here's where to find them:

qbo-keys

coreymckrill: I guess the testing instructions are a little convoluted since it's spread between steps 3 and 4. But in Step 3 I'm saying to go to the _Development_ keys tab, not _Production_.

So I'm not suggesting that we use production keys in a local environment.

public_html/wp-content/mu-plugins/wcorg-misc.php
coreymckrill: > What's the reason for falling back to dev here instead of local ? Same for set_client_config().

I was thinking that `development` is the catchall non-production value, whereas `local` has specific implications since the domain is different, no 2-way communication, etc.

> 3-helpers-misc.php seems like a more consistent location for this type of thing.

Oh, good call. It's kind of confusing that we have **3-helpers-misc.php** _and_ **wcorg-misc.php**. But that's a conversation for another day 😄

.docker/wp-config.php
coreymckrill: > wcpt-event/notification.php should probably return early on local sandboxes instead of remote.

1bb2e2eca0ba693df20964dd844375e344652835

public_html/wp-content/mu-plugins/wcorg-misc.php
coreymckrill: 1228241190643ce36608c8fee54750db47b68cca

public_html/wp-content/plugins/wordcamp-qbo-client/wordcamp-qbo-client.php
coreymckrill: Not in my testing, but I suppose it's possible it could contain a key or something. I'll rethink the approach here.

public_html/wp-content/mu-plugins/wcorg-misc.php
iandunn: Yeah, I can see that. I don't think the two names were intended to imply a relationship between the two files, just to distinguish miscellaneous helpers from `wcpt`-specific helpers.

Feel free to rename it to whatever makes more sense to you (now or in the future).

iandunn: > I was thinking that `development` is the catchall non-production value, whereas `local` has specific implications since the domain is different, no 2-way communication, etc.

That feels counter-intuitive and unconventional to me. If you'd prefer that, I think all the possible values of `WORDCAMP_ENVIRONMENT` should be documented alongside the `wp-config` definition.

.docker/wp-config.php
iandunn: Doh, you're right, I totally missed that.

iandunn: 👍

public_html/wp-content/plugins/wordcamp-qbo-client/wordcamp-qbo-client.php
coreymckrill: 2111916a

.docker/wp-config.php
iandunn: > I don't think [changing `development` to `staging`] needs to be part of this PR

I didn't realize you were seeing environments being [_both_ `local` and `development` simultaneously](
https://github.com/WordPress/wordcamp.org/pull/371#discussion_r381501047), rather than changing `development` to _only_ mean remote sandboxes. So I guess in that case it shouldn't change 👍 



> In both of those closures, the default switch case catches "local", so it gets the same treatment as if it was "development".

Ah, you're right, I missed that. 👍 


> just change the line at the top to...

👍 


> the values would need to go in sandbox-functionality.php either way. Otherwise [...] version control

Ah, I didn't realize that your `public_html/wp-config.php` gets overwritten when you re-provision. 👍 


> I think remote sandboxes should still use sandbox keys. If we're testing interactions with the QBO API on a sandbox, we wouldn't want test data to get mixed in with actual transactions.

I guess we just use remote sandboxes differently. I only use mine as a staging environment, like trying to reproduce a bug someone reported w/ production data. Either way works 👍

iandunn (approved):

public_html/wp-content/mu-plugins/wcorg-misc.php
coreymckrill: Documented: 6f7dccda

PR #379, opened 2020-03-02
Post Types: Show drafted speakers attached to sessions

ryelle:
Drafted speakers don't currently show up in the Sessions list table (speakers column), only published speakers. This can be confusing while you're setting up a site/schedule, since it seems like the speaker is not correctly attached to the session. This PR updates the query to show "any" speakers, with a "draft" label for unpublished speakers.

### Screenshots

Screen Shot 2020-03-02 at 10 29 28 AM

1. Set up some sessions & speakers, some should be drafts (or private, trash, etc)
2. View the sessions list table, `/wp-admin/edit.php?post_type=wcb_session`
3. Drafted speakers should appear in the Speakers column with ` (draft)` after their name
4. Privately published speakers will also show up here, but trashed should not.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -2069,6 +2069,7 @@ class WordCamp_Post_Types_Plugin {
 				if ( ! empty( $speakers_ids ) ) {
 					$speakers = get_posts( array(
 						'post_type'      => 'wcb_speaker',
+						'post_status'    => 'any',
 						'posts_per_page' => -1,
 						'post__in'       => $speakers_ids,
 					) );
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -2069,6 +2069,7 @@ public function manage_post_types_columns_output( $column, $post_id ) {
 				if ( ! empty( $speakers_ids ) ) {
 					$speakers = get_posts( array(
 						'post_type'      => 'wcb_speaker',
+						'post_status'    => 'any',
```
coreymckrill:
```suggestion
						'post_status'    => array( 'publish', 'draft' ),
```

For clarity I think it would be good to be explicit about which statuses we expect to be included...

coreymckrill (approved):

PR #382, opened 2020-03-05
Add capability checks and admin UI for viewing/moderating feedback

### LINKED ISSUE #339: Speaker Feedback: Display feedback in wp-admin, using comments screen
WordCamp organizers should be able to view feedback in wp-admin, using the comments screen. The meta should be combined on the fly to render a human-friendly view of feedback:

Speaker Feedback

Organizers can spam feedback from this view.

### LINKED ISSUE #340: Speaker Feedback: Add permissions-checks to feedback view

coreymckrill:
* Adds custom meta capabilities for viewing and moderating feedback, and maps them.
* Adds an admin page with a list table for reviewing feedback comments that operates similarly to Core's **edit-comments.php**.
* Adds a column on the Sessions list table to show feedback counts.
* Adds tests.

This is kind of a big PR. I could try and split it up, but I'm not sure how best to do it so the pieces make sense on their own.

Fixes #339 Fixes #340 

**Screenshots**

feedback_screen

sessions_screen

**To test:**

1. Rebuild the assets to get the right styles.
1. Add some normal comments to a blog post or two.
1. Create some session posts and add some feedback comments to one or more of them. The testing instructions on #353 have more details of a way to do that.
1. View the Comments screen from the Admin side menu. No feedback comments should appear, or be included in the counts.
1. Click the Feedback submenu item under Sessions. This should show a list table of all of the feedback you just created, but no non-feedback comments.
1. Try clicking the linked column headings to change the sorting of the feedback. Particularly the Rating column.
1. Try using the row actions and bulk edit to change the status of some feedback comments. Check that the counts in the view links at the top update correctly.
1. Try using the search box to find comments with only a particular word or phrase in the feedback answers.
1. Go to the Sessions list table. Sessions that have feedback comments should show the counts for them in a new column. Clicking on one of the feedback bubbles should go to the Feedback list table, but only show it for the one particular session.
1. Log into the Admin with a user that has Author role or below. The Feedback list table screen should not be accessible.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-feedback-list-table.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-feedback-list-table.php
@@ -19,6 +19,19 @@ defined( 'WPINC' ) || die();
  */
 class Feedback_List_Table extends WP_Comments_List_Table {
 	/**
+	 * Feedback_List_Table constructor.
+	 *
+	 * @param array $args
+	 */
+	public function __construct( $args = array() ) {
+		parent::__construct( $args );
+
+		// This is an ugly hack to prevent the avatar from rendering twice, from both the parent and child class.
+		// See \WordCamp\SpeakerFeedback\Admin\filter_list_table_add_comment_avatar().
+		remove_filter( 'comment_author', array( $this, 'floated_admin_avatar' ), 10 );
+	}
+
+	/**
 	 * Get the screen option for items per page.
 	 *
 	 * @param string $comment_status Unused.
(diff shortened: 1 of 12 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-feedback-list-table.php
```diff
@@ -0,0 +1,259 @@
...
+	public function column_name( $comment ) {
+		$feedback = get_feedback_comment( $comment );
+
+		echo '<strong>';
+		comment_author( $comment );
+		echo '</strong>';
+
+		$user = ( $feedback->user_id ) ? new WP_User( $feedback->user_id ) : null;
+
+		if ( $user ) {
```
ryelle, line 147: If the user isn't a wp.org user, should we output the email here?

ryelle (approved):
Works great! 💯 

Looks likes the front end needs some filtering too, the meta & comment heading is still outputting the replies as if they're "real" — but this isn't a blocker for this PR.

Screen Shot 2020-03-19 at 3 39 35 PM

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-feedback-list-table.php
coreymckrill: Ah good call. 3c2671ad

PR #383, opened 2020-03-05
CampTix: Prevent calling CampTixStripe.form_handler multiple times

ryelle:
There have been reports of attendees who see a failed ticket purchase, but are later charged. These attendees usually have multiple tickets in wp-admin, the first successful, but later tickets failed with the error `token_already_used`. I believe this is due to the registration form submitting multiple times.

When checking out, we intercept the form submission to pop up the stripe modal & take card payment. Once the modal is submitted, the token is added to the form and the form is really-submitted. The intercept code still runs on this second submit. In most cases, the page should navigate off to checkout, but I think `form.submit()` is called again, attempting to charge again with the same token. Both requests to `tix_action=checkout` fly, the first ticket is charged and attendee created, then the second ticket attempts to charge but fails (due to single-use tokens), and the second ticket attendee is created but marked failed. It's this later request that the user is redirected to, making it seem like the entire purchase failed.

So the fix is to make sure we only call the `submit` callback when we intend to. This removes the `form_handler` callback before we programmatically submit the form. It also adds some checks to make sure we're only running when we have the right form, and when we have a stripe object.

Make sure checkout still works 👍  

I haven't been able to reproduce the actual issue locally, so I'm not sure how to test that this fixes it outside of "logically it should".

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.js
+++ b/public_html/wp-content/plugins/camptix/camptix.js
@@ -268,7 +268,11 @@ var CampTixStripe = new function() {
 	self.form = false;
 
 	self.init = function() {
-		self.form = jQuery( '#tix form' );
+		self.form = jQuery( '#tix_checkout_form' );
+		if ( ! self.form.length ) {
+			return;
+		}
+
 		self.form.on( 'submit', CampTixStripe.form_handler );
 
 		// On a failed attendee data request, we'll have the previous stripe token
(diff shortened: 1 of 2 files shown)


coreymckrill (approved):
👍 Tested on sandbox, still works!

I wasn't able to reproduce the issue either. I say go for it with this and we can just monitor to see if the problem continues to crop up or not.

PR #386, opened 2020-03-11
CampTix: Add "Remote" ticket type

### LINKED ISSUE #385: CampTix: Add "remote" ticket type, update required question
> Currently, all WordCamp tickets sold through Camptix ask attendees two questions:
>
> 1. Whether they have any life-threatening allergies
> 2. If they have any special accommodation requests
>
> These questions are very important for attendees who attend the WordCamp in person, but they are unnecessary for people who purchase livestream tickets.

From https://meta.trac.wordpress.org/ticket/4344

The allergy question is unnecessary, the accommodations one might still be, but the examples in the text of the question don't apply. So we could also update the text for this question to reflect remote/livestream needs.

See https://wordpress.slack.com/archives/C08M59V3P/p1583522483036700

To address this, we'll create ticket types, starting with a remote type, and conditionally switch out these required questions.

ryelle:
Set up a "Ticket Type", which pulls from a global list of types (set up in code) and can conditionally alter the registration process. Types are set up as a safelist of slug+name, and the type is saved in ticket post meta. This also updates the allergy & accommodations questions, by adding filters to skip rendering the question and/or reword the question text. Other ticket types are possible in the future, by filtering `camptix_ticket_types` & similarly working with CampTix filters.

Fixes #385 

👩🏼‍💻 _Dev notes…_

I went through 2 architecture iterations before landing here— intentionally not using the `CampTix_Addon` class & instead sticking to plain namespacing so that the ticket-type files can easily access the `get_type*` functions. This way, we can add another file to `ticket-types`, ex `workshop.php`, and just like `remote.php` set it up to filter `camptix_ticket_types`. I tried to make minimal changes to CampTix itself & the hardcoded questions. Overall this works pretty well but I'd still be interested in any feedback.

📝 **I could use copy advice!**

You can see these strings "in context" in the screenshots below, but I'd appreciate better phrasing or even re-worked ideas.

- **Ticket type name:** I went with "Default" & "Remote/Livestream"
- **Describing what the ticket types are:** 
    ```
    The type of ticket will determine how we set up the required questions. This
    cannot be changed once someone buys a ticket.

    Default: Default, in-person attendee.
    Remote/Livestream: Remote attendee, ex: watching on livestream, from home.
    ```
- **Reworded a11y question:** "Do you have any accessibility needs, such as a sign language interpreter, to participate in WordCamp?" — this is still hard-coded, but there was a request that it be re-worded to be more remote-friendly, so I removed the "wheelchair access" bit.

### Screenshots

Creating a new ticket type allows for picking a type
c-default-dd

Once a ticket is purchased, this is disabled
Screen Shot 2020-03-11 at 4 35 21 PM

The ticket type is added to the ticket list table
c-ticket-list

In the ticket form, we don't ask the allergy question, and the accommodations question is updated.
c-remote-ticket-form

1. Create a ticket, pick "Remote" for the type
2. It should work, and be labelled "remote" in the list table in wp-admin
3. Go through buying a remote ticket
4. The allergy question should not be in the form
5. The a11y question should not say anything about physical access
6. Selecting yes for the a11y question should still trigger the email
7. Buy a non-remote ticket
8. Allergy & a11y questions are both present
9. Selecting yes for the allergy question should still trigger the email

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/remote.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/remote.php
@@ -9,6 +9,8 @@ const SLUG = 'remote';
 add_filter( 'camptix_ticket_types', __NAMESPACE__ . '\add_type' );
 add_filter( 'camptix_metabox_questions_default_fields_list', __NAMESPACE__ . '\modify_default_fields_list' );
 add_filter( 'camptix_accommodations_question_text', __NAMESPACE__ . '\accommodations_question_text', 10, 2 );
+add_filter( 'camptix_attendee_form_after_questions', __NAMESPACE__ . '\set_allergy_filter', 10, 2 );
+add_filter( 'camptix_form_edit_attendee_after_questions', __NAMESPACE__ . '\set_allergy_filter', 10 );
 
 /**
  * Add this type to the available ticket types.
@@ -46,3 +48,28 @@ function accommodations_question_text( $question, $ticket_id ) {
 	}
 	return $question;
 }
+
+/**
+ * Set a boolean filter on the "Allergy" question, to skip the question if it's remote.
+ *
+ * @param array    $form_data
+ * @param int|null $i
+ */
+function set_allergy_filter( $form_data, $i = null ) {
+	if ( ! is_null( $i ) ) {
+		$form_data = ( isset( $form_data['tix_attendee_info'][ $i ] ) ) ? $form_data['tix_attendee_info'][ $i ] : array();
+	}
+
+	if ( ! isset( $form_data['ticket_id'] ) ) {
+		return;
+	}
+
+	if ( SLUG === get_type_slug( $form_data['ticket_id'] ) ) {
+		remove_filter( 'camptix_allergy_should_skip', '__return_false' );
+		add_filter( 'camptix_allergy_should_skip', '__return_true' );
+	} else {
+		remove_filter( 'camptix_allergy_should_skip', '__return_true' );
+		add_filter( 'camptix_allergy_should_skip', '__return_false' );
+	}
+}
+
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/remote.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/remote.php
@@ -8,6 +8,7 @@ const SLUG = 'remote';
 
 add_filter( 'camptix_ticket_types', __NAMESPACE__ . '\add_type' );
 add_filter( 'camptix_metabox_questions_default_fields_list', __NAMESPACE__ . '\modify_default_fields_list' );
+add_filter( 'camptix_accommodations_question_text', __NAMESPACE__ . '\accommodations_question_text', 10, 2 );
 
 /**
  * Add this type to the available ticket types.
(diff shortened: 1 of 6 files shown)


iandunn (approved):
LGTM

public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/ticket-types.php
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * Match the given value to an existing type slug, return the slug, or "default" if not a match.
+ */
+function get_slug_from_value( $value ) {
+	$match = wp_list_pluck( wp_list_filter( get_types(), array( 'slug' => $value ) ), 'slug' );
+	if ( count( $match ) > 0 ) {
+		return current( $match );
+	}
+	return 'default';
+}
```
coreymckrill: What if the types were registered as an associative array, with the slug as the key? Then this function might be simpler or even unnecessary.

public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/ticket-types.php
```diff
@@ -0,0 +1,188 @@
...
+		return;
+	}
+
+	// Security check.
+	$nonce_action = 'update-post_' . $post_id;
+	check_admin_referer( $nonce_action );
+
+	if ( isset( $_POST['tix_type'] ) ) {
+		update_post_meta( $post_id, META_KEY, $_POST['tix_type'] );
+	}
```
coreymckrill: I would recommend using `filter_input()` here (everywhere) to retrieve the `$_POST` values, instead of referencing them directly. It gives you some built-in validation and sanitizing.

public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/remote.php
```diff
@@ -0,0 +1,83 @@
...
+	}
+	return $question;
+}
+
+/**
+ * Set a boolean filter on the "Allergy" question, to skip the question if it's remote. This will also skip the
+ * validation step when creating an attendee record.
+ *
+ * @param array    $form_data
+ * @param int|null $i
```
coreymckrill: I'm sure this var is documented or makes sense in context in **camptix.php**, but I think it would be helpful to add a description of it here so we don't have to go hunting in that giant file to figure out what this is.

coreymckrill (approved):
👍 Tested locally.

Some comments, but no blockers.

public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/ticket-types.php
ryelle: The sanitization callback in `register_post_meta` takes care of this, but just to be extra-sure I added it.

public_html/wp-content/mu-plugins/camptix-tweaks/addons/ticket-types/ticket-types.php
ryelle: Good idea — less magic with `wp_list_*` 🙂

timiwahalahti:
For me, copy looks good.

melchoyce:
> Ticket type name: I went with "Default" & "Remote/Livestream"

I'd probably go with either one, or the other. I think "Livestream" is likely more common here, so I'd recommend just using that work as the ticket label.

CdrMarks:
@ryelle I saw you [said](https://wordpress.slack.com/archives/C02RQC7RN/p1584398319379100) in #community-events on Slack the Livestream ticket type was available.  I tested on my local environment and see the new feature related to [ccf415b](https://github.com/WordPress/wordcamp.org/commit/ccf415b9ea5a317f0c2e62af9f381dc90f31066c).  I checked the 2020 Birmingham site where I'm an organizer and I [do not see](https://2020.birmingham.wordcamp.org/wp-admin/post.php?post=1048&action=edit) this feature available.

ryelle:
@CdrMarks It was my mistake— I left the "don't enable this on production" flag in. It's now available for all sites.

PR #390, opened 2020-03-13
CampTix: Add UI for changing attendee status

ryelle:
Adds a status dropdown to the edit attendee screen, mirroring how core's post status works. This adds "publish", "draft", "refund", "cancel" and "failed" as possible statuses. It also updates bulk and quick edit on the list table to allow marking attendees cancelled or refunded in batches.

❓I'm not sure what the status list we want is, but I thought it might be useful to add all (except timeout) just in case we need to manually edit some attendees for … some reason? I can remove these if they feel too much like clutter.

### Screenshots

Screen Shot 2020-03-13 at 6 09 40 PM

Screen Shot 2020-03-13 at 6 10 19 PM

1. Edit an attendee
2. You should be able to change the status
3. Save, it should correctly update
4. Back in the list table, quick edit and/or bulk edit some attendees
5. Refunded and cancelled should be options now
6. Saving correctly updates the attendees

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
@@ -25,7 +25,7 @@ add_filter( 'camptix_shortcode_contents',                    __NAMESPACE__ . '\m
 // Attendees
 add_filter( 'camptix_name_order',                            __NAMESPACE__ . '\set_name_order'                      );
 add_action( 'camptix_form_edit_attendee_custom_error_flags', __NAMESPACE__ . '\disable_attendee_edits'              );
-add_action( 'transition_post_status',                        __NAMESPACE__ . '\log_publish_to_cancel',        10, 3 );
+add_action( 'publish_to_cancel',                             __NAMESPACE__ . '\log_publish_to_cancel',        10, 3 );
 add_filter( 'camptix_privacy_erase_attendee',                __NAMESPACE__ . '\retain_attendee_data',         10, 2 );
 add_action( 'admin_notices',                                 __NAMESPACE__ . '\admin_notice_attendee_privacy'       );
 add_filter( 'wp_privacy_personal_data_erasers',              __NAMESPACE__ . '\modify_erasers',                  99 );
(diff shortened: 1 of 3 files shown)


coreymckrill (commented):
Working for me in local testing.

Two questions/comments:

1. Do we want site admins (ie WordCamp organizers) to be able to change the ticket statuses manually as well, or should it only be super admins? If the intention of this enhancement is to make it easier for community team people to handle edge cases, I would probably lean towards the latter.
1. Looking at the CampTix DB Log metabox after changing the status of an attendee a few times, I noticed that the log doesn't reflect all the changes accurately, or sometimes at all. When I changed from Published to Cancelled via the Edit Attendee screen, the log entry said `Publish to cancel transition, possible bug.` which I suspect is an artifact of the intention that ticket statuses shouldn't be directly editable. When I used Quick Edit to change from Cancelled to Refunded, there was no log entry. And when I switched from Refunded back to Published (also via Quick Edit), the log entry implied the ticket was newly created, with a lot of accompanying data. Ideally, I think, these manual status changes would have their own log entry template, that would be something like `[user] manually changed the status from [old status] to [new status]`. Thoughts?

ryelle:
>  If the intention of this enhancement is to make it easier for community team people to handle edge cases, I would probably lean towards the latter.

I agree, I'll update this.

> CampTix DB Log metabox

I didn't think to check this - all good points.

What do you think about allowing the other statuses (cancelled, failed) - do you think that's worthwhile? Especially with the log issues you noted, I wonder if it might lead to more strange edge cases, vs being useful for admins.

coreymckrill:
> What do you think about allowing the other statuses (cancelled, failed) - do you think that's worthwhile?

I think there's a use case for `cancelled`, if I understand its intention correctly. Say there's a ticket that someone got for free with a coupon. But then the event is cancelled. The ticket can't be set to "refunded" because there was no transaction, but it could be set to `cancelled`.

I can't think of a scenario where it would make sense to manually set to `failed`.

ryelle:
Updated it to only allow changing status to published, refunded, or cancelled. Only trusted deputies & super admins can edit statuses. 

I updated the logging to add an entry when tix_attendees are manually changed, it should now say "Attendee manually changed from [old] to [new] by [user]." I also changed the "possible bug" log entry to only be added if no user is attached (which does mean it would flag for CLI changes, but I think that's fine? I could remove it entirely, if it's not serving a purpose).

I left the "New ticket created" message because i think that's used when tickets are officially purchased, and while confusing I'm not sure how we could easily check "has this ever been published in the past". If it ends up being confusing in practice, we can add some extra logic later.

public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
```diff
@@ -497,19 +497,18 @@ function disable_attendee_edits( $attendee ) {
...
 	global $camptix;
 
-	if ( 'tix_attendee' !== $post->post_type || $to === $from ) {
+	if ( 'tix_attendee' !== $post->post_type ) {
 		return;
 	}
 
-	if ( 'publish' === $from && 'cancel' === $to ) {
+	// Not a manual change, should be logged.
+	if ( 0 === wp_get_current_user_id() ) {
```
coreymckrill:
```suggestion
	if ( 0 === get_current_user_id() ) {
```

Currently throwing a fatal error here.

coreymckrill (approved):
One thing to fix, and then I think it's good to go 👍

PR #393, opened 2020-03-17
Payments: Add Refunded status for Sponsor Invoices

### LINKED ISSUE #392: Payments: Add "Refunded" status
Like #282/#291, we need a status to track refunded invoices.

- Add "Refunded" as action link to invoices
- Mark refunded invoices in the list table
- Add "Refunded" tab to network dashboard (maybe)

This is just for internal tracking, no need to interact with QBO/other 3rd party.

ryelle:
Add a new status for Sponsor Invoices to track which invoices have been refunded. This can be applied to an invoice in the list table, by clicking the new action "Refunded", which is only available on paid invoices. It also adds a Refunded tab to the network Sponsor Invoices dashboard to show all refunded invoices.

Fixes #392 

### Screenshots

Actions in list table, Refunded is only available on paid invoices, Uncollectible is available on everything but already marked "uncollectible": (all actions shown for example)
Screen Shot 2020-03-17 at 5 33 26 PM

Network dashboard tab:
Screen Shot 2020-03-17 at 5 33 58 PM

1. Add some test invoices to your dev environment
2. Mark them paid (I used CLI for this to skip submission)
3. Mark them refunded using the UI
4. The invoice should be correctly updated, and show as refunded both in the list table and in the edit screen.
5. The network dashboard should show this invoice in the refunded tab.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-invoice.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-invoice.php
@@ -109,6 +109,14 @@ function get_custom_statuses() {
 				'wordcamporg'
 			),
 		),
+		'wcbsi_refunded'      => array(
+			'label'       => esc_html__( 'Refunded', 'wordcamporg' ),
+			'label_count' => _nx_noop(
+				'Refunded <span class="count">(%s)</span>',
+				'Refunded <span class="count">(%s)</span>',
+				'wordcamporg'
+			),
+		),
 	);
 }
 
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-invoice.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-invoice.php
@@ -522,27 +522,43 @@ function add_row_action( $actions, $post ) {
 		return $actions;
 	}
 
-	if ( 'wcbsi_uncollectible' === $post->post_status ) {
-		return $actions;
-	}
-
 	$post_type_object = get_post_type_object( $post->post_type );
 
-	$url = add_query_arg(
-		array(
-			'action' => 'wcbsi_update',
-			'status' => 'uncollectible',
-		),
-		admin_url( sprintf( $post_type_object->_edit_link, $post->ID ) )
-	);
+	if ( 'wcbsi_uncollectible' !== $post->post_status ) {
+		$url = add_query_arg(
+			array(
+				'action' => 'wcbsi_update',
+				'status' => 'uncollectible',
+			),
+			admin_url( sprintf( $post_type_object->_edit_link, $post->ID ) )
+		);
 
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-invoice.php
```diff
@@ -514,27 +522,43 @@ function add_row_action( $actions, $post ) {
...
 	$post_type_object = get_post_type_object( $post->post_type );
 
-	$url = add_query_arg(
-		array(
-			'action' => 'wcbsi_update',
-			'status' => 'uncollectible',
-		),
-		admin_url( sprintf( $post_type_object->_edit_link, $post->ID ) )
-	);
+	if ( 'wcbsi_uncollectible' !== $post->post_status ) {
```
coreymckrill, line 527: Seems like `paid` and `refunded` invoices should also not have the option to mark as `uncollectible`...

coreymckrill (commented):
Worked in local testing 👍 

I'm wondering if it would be worth including a way to "go back" from marking something as `refunded` or `uncollectible`.

* For `uncollectible`, I could see a scenario where a sponsor suddenly pays an invoice at a much later date when it has already been written off. But maybe we wouldn't update the original invoice in that case anyway?
* For both `uncollectible` and `refunded` I could see a scenario where someone accidentally clicks the row action and then there's no way to undo it, other than having a dev change the status via CLI.

Could be a lot of extra complexity for a few edge cases, though. Thoughts?

public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-invoice.php
ryelle: I didn't want to make changes to the behavior of uncollectible, and this is what it does currently. I can iterate on this in another PR, maybe?

ryelle:
@coreymckrill  I can add the ability to "go back" from refunded to paid; but uncollectible currently is harder, since it doesn't have a single original status… but if we change that, undo should be easier. Would that be… `wcbsi_approved` (sent)?

public_html/wp-content/plugins/wordcamp-payments/includes/sponsor-invoice.php
coreymckrill: 👍

coreymckrill: Update: sounds like `wcbsi_approved` is the only status when Uncollectible should be an option.

coreymckrill:
> Would that be… wcbsi_approved (sent)?

Sounds like that is the case.

ryelle:
@coreymckrill Are you OK with me merging this PR as-is, and making those improvements in another PR?

coreymckrill (approved):
Yep, sounds good.

PR #399, opened 2020-03-23
WCPT: Add virtual event information to WordCamps

### LINKED ISSUE #398: WordCamp Post Type: Add virtual event fields to WordCamps
We want to be able to manage & track virtual events. We'll need the following added to the WordCamp "Venue Information" metabox:

- A checkbox to indicate a virtual event: "Virtual event only"
- A dropdown to indicate which Crowdcast account:
  - Crowdcast WordCamp 1
  - Crowdcast WordCamp 2
  - Other

If "Virtual event only" is checked, the physical address is not required.

ryelle:
Adds two new meta values to WordCamp post types: a checkbox for if this is only a virtual event, and a dropdown for a streaming service account to be used. The streaming accounts are hardcoded, but as a function on the Event class to allow the possibility that meetups might use this, and that WordCamps or Meetups might use different accounts. Lastly, this removes the "required" label on physical address, if the event is online-only.

Fixes #398 

### Screenshots

Screen Shot 2020-03-23 at 2 26 59 PM

1. Edit a WordCamp on central
2. There should be these new options in the Venue metabox
3. Saving these fields should work

### DIFF
--- a/public_html/wp-content/plugins/wcpt/javascript/wcpt-wordcamp/admin.js
+++ b/public_html/wp-content/plugins/wcpt/javascript/wcpt-wordcamp/admin.js
@@ -34,7 +34,7 @@ window.wordCampPostType.WcptWordCamp = ( function( $ ) {
 			self.initializeMentorPicker( $mentorUserName );
 		}
 
-		$virtualEventCheckbox.change( self.togglePhysicalAddrRequire );
+		$virtualEventCheckbox.change( self.togglePhysicalVenueFields );
 		$virtualEventCheckbox.trigger( 'change' );
 	};
 
@@ -43,12 +43,16 @@ window.wordCampPostType.WcptWordCamp = ( function( $ ) {
 	 *
 	 * @param {object} event
 	 */
-	self.togglePhysicalAddrRequire = function( event ) {
-		var $label = $( '#wcpt_physical_address' ).closest( '.inside' ).find( '.description' );
+	self.togglePhysicalVenueFields = function( event ) {
+		var $container = $( event.target ).closest( '.inside' ).parent();
+		var $items = $container
+			.find( '.inside' )
+			.not( '.field__wcpt_virtual_event_only,.field__wcpt_streaming_account_to_use' );
+		console.log( $items );
 		if ( $( event.target ).is( ':checked' ) ) {
-			$label.hide();
+			$items.hide();
 		} else {
-			$label.show();
+			$items.show();
 		}
 	}
 
--- a/public_html/wp-content/plugins/wcpt/javascript/wcpt-wordcamp/admin.js
+++ b/public_html/wp-content/plugins/wcpt/javascript/wcpt-wordcamp/admin.js
@@ -12,7 +12,8 @@ window.wordCampPostType.WcptWordCamp = ( function( $ ) {
 		var createSiteCheckbox = $( '#wcpt_create-site-in-network' ),
 			$mentorUserName = $( '#wcpt_mentor_wordpress_org_user_name' ),
 			hasContributor = $( '#wcpt_contributor_day' ),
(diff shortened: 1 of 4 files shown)


coreymckrill (approved):
👍 Worked testing locally.

Some thoughts, but none of these are blockers.

1. I wonder if it would be useful to add a "Description" text input that only appears if you select "Other" from the streaming account select.
1. Related, I can see a scenario where the values in the streaming services array change, and a WordCamp post has a value saved for that that is no longer in the array. But maybe we'd want to keep the value for historical/reporting purposes. Instead of deleting an invalid value, maybe we could set it to "Other" and put the invalid value in the Description field?
1. When the "Virtual Event only" box is checked, it seems like we could just hide/disable every other field in that metabox below the "Streaming account to use" select field.

ryelle:
Added "other" text input, & now toggle all fields based on "virtual" checkbox:

Screen Shot 2020-03-23 at 6 11 29 PM

Enough code changed that it should get another look, if you can 🙂

public_html/wp-content/plugins/wcpt/javascript/wcpt-wordcamp/admin.js
```diff
@@ -32,8 +34,38 @@ window.wordCampPostType.WcptWordCamp = ( function( $ ) {
...
+	 * Toggle the "required" label on the physical address field.
+	 *
+	 * @param {object} event
+	 */
+	self.togglePhysicalVenueFields = function( event ) {
+		var $container = $( event.target ).closest( '.inside' ).parent();
+		var $items = $container
+			.find( '.inside' )
+			.not( '.field__wcpt_virtual_event_only,.field__wcpt_streaming_account_to_use' );
+		console.log( $items );
```
coreymckrill:
```suggestion

```

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -523,6 +539,26 @@ public function metabox_save( $post_id, $post, $verify_nonce = true ) {
...
+							update_post_meta(
+								$post_id,
+								$key_other,
+								sanitize_text_field( $_POST[ $key_other ] )
+							);
+						}
+					} else {
+						// The value isn't in the allowed values (anymore?) so we should save it as "other".
+						update_post_meta( $post_id, $key, 'other' );
+						update_post_meta( $post_id, $key_other, $values[ $key ] );
```
coreymckrill, line 558:
```suggestion
						update_post_meta( $post_id, $key_other, sanitize_text_field( $values[ $key ] ) );
```

coreymckrill (approved):
Two tweaks, and maybe an alignment adjustment, and then I think this is good to go.

The "Other" field for streaming seems to be misaligned. Seeing this on both Firefox and Chrome:
field-alignment

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
ryelle: the values in `$values` are [already escaped by `esc_attr`,](https://github.com/WordPress/wordcamp.org/blob/bad61fae01bf264bad736b24ebbfcadd3de95c5e/public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php#L490) so we don't need to do this twice. I only had to do it above because I was using `$_POST` directly.

ryelle:
I'm not sure if the reason was a cached CSS file, but I updated the `admin.css` registration to break the CSS cache in any case— I don't see that misalignment you've got.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
coreymckrill: 👍

PR #401, opened 2020-03-25
Payments: Restrict status changes to be from specific statuses

ryelle:
Sponsor invoice statuses can be changed, but it only makes sense for those to happen to invoices in specific status — ex, only paid invoices can be refunded. This updates the logic for showing status change actions, and DRY-s it up a bit so that we can also add status change actions to "undo" those changes.

See feedback on #393

### Screenshots

Screen Shot 2020-03-25 at 12 31 26 PM

1. View the Sponsor Invoices for a site
2. You should be able to mark…
    - Paid -> Refunded
    - Refunded -> Paid
    - Sent -> Uncollectible
    - Uncollectible -> Sent


coreymckrill (approved):
👍

PR #404, opened 2020-03-26
Docker: Use a volume for the database storage

ryelle:
This is transparently managed by docker, not listed as a directory in the host filesystem. We don't need to expose the mysql internals, and this is still persistent. MySQL is still reachable by using the docker CLI.

_This is one of the changes I made when trying to get my local env working again, and while I'm not sure it's related, it does clean up the host filesystem._

⚠️  I imagine switching to this will probably blow away the existing database, so export or backup anything you want before doing so.

### DIFF
--- a/.docker/Dockerfile.php-fpm
+++ b/.docker/Dockerfile.php-fpm
@@ -6,6 +6,7 @@ RUN apt-get update \
         build-essential \
         git \
         less \
+        libicu-dev \
         libjpeg-dev \
         libpng-dev \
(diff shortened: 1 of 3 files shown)


iandunn (commented):
Seems fine at first glance, but I don't use the Docker env either, so no preference.

ryelle:
Closing this, I've stopped using this method, and I don't think it had anything to do with whatever my issue was in the first place.

PR #405, opened 2020-03-26
Virtual Embeds: Add blocks for virtual events

ryelle:
Adds a CrowdCast block to embed the crowdcast channel iframe, and a StreamText block to embed a caption stream (uses the streamtext shortcode behind the scenes). You need the channel/event name for each service, and it renders a very simple embed (no configuration supported). It does support wide & full width.

### Screenshots

Screen Shot 2020-03-26 at 1 10 41 PM

1. After checking out the branch, run `yarn` to install the workspace
2. Run `yarn workspace virtual-embeds build` to build the files
3. Add CrowdCast and/or StreamText blocks to a page
4. Add valid events (ex, `IHaveADream` for an active streamtext event, `wordcamp-san-antonio-2020-mktg` for a crowdcast)
5. It should save & correctly show up on the page
6. Try adding malformed or empty events, it shouldn't break

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
 	},
 	"workspaces": [
 		"public_html/wp-content/mu-plugins/blocks",
+		"public_html/wp-content/mu-plugins/virtual-embeds",
 		"public_html/wp-content/plugins/wordcamp-forms-to-drafts",
 		"public_html/wp-content/plugins/wordcamp-speaker-feedback"
 	],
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
 	},
 	"workspaces": [
 		"public_html/wp-content/mu-plugins/blocks",
+		"public_html/wp-content/mu-plugins/virtual-embeds",
```
coreymckrill, line 16: Why create a separate project if these are also blocks?

coreymckrill (approved):
👍 Worked testing locally

I noticed while using Twenty Nineteen that if I set either of the blocks to Full Width, the block expanded beyond the edges of the screen, and I wasn't able to access the toolbar to change it back 😬 

Not sure if that's a problem with these blocks in particular, or with Twenty Nineteen.

ryelle:
Pretty sure that's a Twenty Nineteen issue, I've seen that before.

package.json
ryelle: These seemed conceptually different enough, not dealing with the WC content or APIs or anything; and because of that I didn't need the whole file/classes infrastructure of the blocks project.

PR #406, opened 2020-03-26
Speaker Feedback: Check submissions for spam, duplicates

### LINKED ISSUE #346: Speaker Feedback: Run submissions through Akismet
This might happen by default since the submissions are stored as comments, but we need to ensure that all of the related meta data is included in the scan.

Related: #337

coreymckrill:
Our feedback submissions require a custom configuration for Akismet to check them correctly for spam, even though they are comments. This PR ensures that submissions get sent successfully to the Akismet API when processing requests to our REST API endpoint. It also ensures that if a feedback comment is later manually set to `spam` or `not spam`, the correct info is sent to the Akismet API so it can "learn". All of this happens through our custom list table UI such that the spam workflow is identical to that of normal comments.

Finally, this fixes a bug that was allowing duplicate feedback comments to successfully be submitted.

Fixes #346 

### Screenshot:

spam

### To test:

1. Add a functional [Akismet API key](https://akismet.com) to your local dev environment or sandbox using this:
    ```php
    // Akismet
    add_filter( 'akismet_get_api_key', function( $key ) {
    	return '[your key goes here]';
    } );
    ```
1. [Submit some feedback comments](https://github.com/WordPress/wordcamp.org/pull/353). Try to get Akismet to mark some of them as spam. If you're having trouble, you can use `akismet-guaranteed-spam@example.com` for the `author_email` parameter and it [should work](https://akismet.com/development/api/#comment-check). I found that putting `asdf` in the meta fields `q1` - `q3` also worked :D  Note that there won't be any indication in the response from our API endpoint that the feedback comment has been marked as spam, unless it's so egregious that Akismet thinks it can be immediately discarded.
1. Check our Feedback list table. You should see your feedback comments appear either as Pending or as Spam.
1. Try marking a Pending feedback as "Spam" using the row action. Afterward, if you view it in the Spam section of the list table, the row action area should have a note saying "Flagged as spam by [username]".
1. Try marking a Spammed feedback as "Not Spam" using the row action. Afterward, if you view it in the Pending section of the list table, the row action area should have a note saying "Un-spammed by [username]".
1. Try clicking the "History" row action for a feedback. The Comment History metabox should show what Akismet originally thought the comment was, plus any changes that were made manually by a user.
1. Try submitting the same feedback twice in a row. The second time you should get an error about it being a duplicate.
1. All the phpunit tests should pass.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
@@ -257,12 +257,7 @@ function enqueue_assets() {
  * @return bool
  */
 function user_can_edit_request( $post ) {
-	$editable_status = false;
-
-	if ( $post instanceof WP_Post ) {
-		$editable_status = in_array( $post->post_status, array( 'auto-draft', 'draft', 'wcb-incomplete' ), true );
-	}
-
+	$editable_status = in_array( $post->post_status, array( 'auto-draft', 'draft', 'wcb-incomplete' ), true );
 	return current_user_can( 'manage_network' ) || $editable_status;
 }
 
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
@@ -257,7 +257,12 @@ function enqueue_assets() {
  * @return bool
  */
 function user_can_edit_request( $post ) {
-	$editable_status = in_array( $post->post_status, array( 'auto-draft', 'draft', 'wcb-incomplete' ), true );
+	$editable_status = false;
+
+	if ( $post instanceof WP_Post ) {
+		$editable_status = in_array( $post->post_status, array( 'auto-draft', 'draft', 'wcb-incomplete' ), true );
+	}
+
 	return current_user_can( 'manage_network' ) || $editable_status;
 }
 
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
@@ -6,7 +6,7 @@ use WP_Error, WP_REST_Request, WP_REST_Response, WP_REST_Server;
 use WP_REST_Comments_Controller;
 use const WordCamp\SpeakerFeedback\Comment\COMMENT_TYPE;
 use function WordCamp\SpeakerFeedback\Comment\add_feedback;
-use function WordCamp\SpeakerFeedback\CommentMeta\validate_feedback_meta;
+use function WordCamp\SpeakerFeedback\CommentMeta\{ get_feedback_meta_field_schema, validate_feedback_meta };
 use function WordCamp\SpeakerFeedback\Spam\spam_check;
 
 defined( 'WPINC' ) || die();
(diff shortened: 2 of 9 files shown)


public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
```diff
@@ -257,7 +257,12 @@ function enqueue_assets() {
  * @return bool
  */
 function user_can_edit_request( $post ) {
-	$editable_status = in_array( $post->post_status, array( 'auto-draft', 'draft', 'wcb-incomplete' ), true );
+	$editable_status = false;
+
+	if ( $post instanceof WP_Post ) {
+		$editable_status = in_array( $post->post_status, array( 'auto-draft', 'draft', 'wcb-incomplete' ), true );
+	}
+
```
ryelle: Was this added in by mistake?

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
```diff
@@ -232,4 +236,86 @@ public function create_item_permissions_check( $request ) {
...
+		}
+
+		$schema = get_feedback_meta_field_schema();
+
+		foreach ( $comment_data['comment_meta'] as $key => $value ) {
+			$args['meta_query'][] = array(
+				'key'   => $key,
+				'value' => $value,
+				// This might need to change if we introduce a meta field with a type other than string or integer.
+				'type'  => ( 'string' === $schema[ $key ]['type'] ) ? 'CHAR' : 'NUMERIC',
```
ryelle, line 304: We could add a note in the schema function that only string & integer are supported? The only other case I could think of would be boolean, and that could easily be saved as string `true` or int `1`.

ryelle (commented):
Looks good & works as described — I tested a few variations of spam & duplicates. I did see this message when submitting feedback. I'm not sure if this is something you can/should set in `is_akismet_spam`, or if something's off in my local env.

```
Notice: Undefined index: HTTP_REFERER in /usr/src/public_html/wp-content/plugins/jetpack/modules/contact-form/grunion-contact-form.php on line 670
```

public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
coreymckrill: Bah, that was to hide a php error that was showing when you click the History row link for a feedback comment. But since it's otherwise unrelated, I should probably just commit that directly to production instead.

coreymckrill: 1124076b1f350e24fbb247cc59b9a521dfa15225

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
coreymckrill: Good idea. Added a comment: 8fcb7df5

coreymckrill:
> `Notice: Undefined index: HTTP_REFERER`

I ran into this as well. I think it probably is an edge case caused by using Insomnia. I went in and added a referrer header to my Insomnia requests and it fixed the error. Presumably when requests are submitted to the API endpoint from our front end form, a referrer header will be included.

This is happening in Jetpack's `prepare_for_akismet` method, so there's not a lot we can do on our side if we want to keep using it. Setting the value of a global `$_SERVER` parameter directly seems like a bad idea. Copying over the whole function to our codebase is also not great, though perhaps palatable.

I'm inclined to let it go for now, and see if it becomes a problem.

ryelle:
> I'm inclined to let it go for now, and see if it becomes a problem.

Sounds good - I'll pay attention when I can get time to update #369 to see if it happens from a real request 👍

ryelle (approved):

PR #409, opened 2020-03-30
WCPT: Work around "Physical Address" requirement for online events

### LINKED ISSUE #408: WCPT: Virtual events still require physical address
The status check for "WordCamp Scheduled" requires the Physical Address, even when "Virtual event only" is checked — but the address input is hidden.

If the event is virtual, the address should not be required here.

ryelle:
The requirement checks don't have any knowledge about the current post, so there's no (easy) way to add "Physical Address" to `get_required_fields` only when another value is selected. Rather than conditionally remove the requirement, this injects a value of "Online". When the online value is used, it will also skip the geocoding process & warning.

Fixes #408

1. Create or edit a WordCamp post on central.wordcamp.test with the "scheduled" status
2. With virtual event only unchecked, make sure "Physical Address" is empty
3. Check "virtual event only"
4. Save the event
5. It should save, and still be in "scheduled", with no error messages
6. You can toggle "virtual event only", and you should see "Online" in the Physical Address field

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
@@ -164,7 +164,11 @@ if ( ! class_exists( 'WordCamp_Admin' ) ) :
 			$new_address = $_POST[ wcpt_key_to_str( 'Physical Address', 'wcpt_' ) ];
 
 			// If there is no venue, shortcut out of the request.
-			if ( empty( $new_address ) ) {
+			if ( empty( $new_address ) || 'Online' === $new_address ) {
+				$geocode_reply = $this->parse_geocode_response( array() );
+				foreach ( array_keys( $geocode_reply ) as $keys ) {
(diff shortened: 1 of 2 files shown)


public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
```diff
@@ -164,7 +164,11 @@ public function update_venue_address( $post_id, $original_meta_values ) {
 			$new_address = $_POST[ wcpt_key_to_str( 'Physical Address', 'wcpt_' ) ];
 
 			// If there is no venue, shortcut out of the request.
-			if ( empty( $new_address ) ) {
+			if ( empty( $new_address ) || 'Online' === $new_address ) {
+				$geocode_reply = $this->parse_geocode_response( array() );
+				foreach ( array_keys( $geocode_reply ) as $keys ) {
```
coreymckrill:
```suggestion
				foreach ( array_keys( $geocode_reply ) as $key ) {
```

coreymckrill (approved):

Issue #400, opened 2020-03-23
Events Page: Add page for all "online" meetups & wordcamps


ryelle:
Like http://make.wordpress.org/community/events

We can search through our cached events for "online" events using Meetup's naming guideline:

> If the title start with [ONLINE], [EN LINEA], [EN LIGNE], [オンライン] or [ON-LINE], they’ll be considered online events

 ⚠️ Check if ^ is still applicable, I don't see that in [these docs.](https://help.meetup.com/hc/en-us/articles/360040609112-Hosting-an-online-event-on-Meetup)

Maybe add some indication of language, to allow for language filtering?

ryelle:
- Events were updated to sync "offline" status in [r9625,](https://meta.trac.wordpress.org/changeset/9625) using the "offline" venue ID
- [r9632](https://meta.trac.wordpress.org/changeset/9632) added the UTC offset back into the database (there's a col for it, we just didn't use it)
- [r9633](https://meta.trac.wordpress.org/changeset/9633) adds the shortcode to display online events - either tagged with the online venue, or using `[ONLINE]` in the title. No events were found using the other variations in brackets, and that's no longer the recommended method for marking as online.

The page itself is live at [make.wordpress.org/community/events/online/](https://make.wordpress.org/community/events/online/) 🎉

ryelle:
This page should also show online WordCamps.

- Make sure the virtual event value is exposed in the API
- Add the first scheduled session time for each WordCamp into the wcpt endpoint for a real start time
- Sync virtual WordCamps over with `location=online`? Or some other method?
- These events _should_ show up on the events page by default, but make sure 🤞🏻

ryelle:
- [r9661](https://meta.trac.wordpress.org/changeset/9661) syncs "Virtual event only" WordCamps as location = "online"
- [r9662](https://meta.trac.wordpress.org/changeset/9662) syncs over the first session time as start time (if it exists)

ryelle:
Virtual WordCamps were not syncing over because they have no latitude/longitude - I've added a(nother) workaround for online events, to sync with the lat/long 1,1 ([r9675](https://meta.trac.wordpress.org/changeset/9675)). I'll check back when the sync next runs ~45min from now.

None of these online WordCamps should show up on the core events widget, since 1,1 is >400km from any city. Getting those into the API response is TBD, will be in another issue once we know the criteria.

ryelle:
[...]

PR #410, opened 2020-03-31
WCPT: Add fields to wordcamp endpoint for online events

ryelle:
This adds two fields to the `/wp/v2/wordcamps` endpoint, `Virtual event only` & `session_start_time`. The first is a boolean which says whether this is an online-only event. The other, `session_start_time`, is added to all WordCamps, regardless of online status. This picks out the earliest published session, and uses the UTC timestamp of that session to mark the "start time" of the WordCamp.

See #400 — this is necessary for adding WordCamps to [the online events page.](https://make.wordpress.org/community/events/online/)

This does increase the time for the API response, but only by ~0.5s (on my sandbox), which seems acceptable.

1. Make a request to `https://central.wordcamp.org/wp-json/wp/v2/wordcamps`
2. You should see these two fields in the response
3. Pick a scheduled WordCamp that has published sessions, it should have that first session's start time as `session_start_time`
4. Toggle the "Virtual event only" in the WordCamp post on central; it should update the API response

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
@@ -404,6 +404,10 @@ class WordCamp_Loader extends Event_Loader {
 			'session_start_time',
 			array(
 				'get_callback' => function( $object, $field_name ) {
+					// Short out if the event is not scheduled.
+					if ( $object['status'] !== 'wcpt-scheduled' ) {
+						return 0;
+					}
 					$site_id = get_post_meta( $object['id'], '_site_id', true );
 					switch_to_blog( $site_id );
 					$sessions = get_posts( array(
--- a/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
@@ -403,6 +403,10 @@ class WordCamp_Loader extends Event_Loader {
 			'wordcamp',
 			'session_start_time',
 			array(
+				'schema'       => array(
+					'type'        => 'string',
+					'description' => __( 'The start time of the first session of WordCamp, when WordCamp content will begin.', 'wordcamporg' ),
+				),
 				'get_callback' => function( $object, $field_name ) {
 					// Short out if the event is not scheduled.
 					if ( 'wcpt-scheduled' !== $object['status'] ) {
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
```diff
@@ -397,6 +398,37 @@ public function register_rest_public_fields() {
 				)
 			);
 		}
+
+		register_rest_field(
+			'wordcamp',
+			'session_start_time',
+			array(
+				'get_callback' => function( $object, $field_name ) {
```
coreymckrill: It might be worth renaming this field to something like `first_session_start_time` and/or adding a `description` parameter to the field schema to better publicly document what this field is for.

coreymckrill (approved):
👍 Looks good from the sandbox.

I noticed that a couple of the "online only" events have Tulsa, Oklahoma in their `_venue_*` fields, even though they aren't related at all to Tulsa:

https://central.wordcamp.org/wp-json/wp/v2/wordcamps/3098548

For this one, however, the venue fields are blank:

https://central.wordcamp.org/wp-json/wp/v2/wordcamps/3095591

So perhaps this is just an artifact of having some events switch to online before [this commit](https://github.com/WordPress/wordcamp.org/commit/50b34c445624e5b4e82dad58d2c3af48d219b441) was added... ?

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
ryelle: Added a description in https://github.com/WordPress/wordcamp.org/pull/410/commits/a628240e37b7933864dbf578383a60b59b4eee13 — I'd prefer to keep the field name shorter, and I think it's still descriptive enough without `first`.

ryelle:
> I noticed that a couple of the "online only" events have Tulsa, Oklahoma in their _venue_* fields, even though they aren't related at all to Tulsa:

Yeah, @CdrMarks noticed that too, that WC Spain Online was showing up in the middle of the US. I think it should be fixed now, I re-saved the WordCamp post. I can check for others, too.

I think there's an outstanding question about how to locate events that are online, for the core events widget — language has come up, IMO timezone should be a factor… but we're not requiring any of this info for virtual WordCamps right now (meetups are a different case because they'll fall back to the group's location). Seems like this should be a community p2 discussion?

iandunn:
`.5s` feels significant to me, what about caching the callback results?

ryelle:
Considering the biggest use case for this API is basically as proxy to cache events, the jump from ~3.1s to ~3.6s didn't seem like a huge issue, especially compared to the complexity that would've been added by creating a cron job to cache the session time. If you think it's significant enough to need that, do you want to add it?

iandunn:
I prefer the stale-while-revalidate approach for better UX, but a transient would work too. Maybe in the future we can build a convenience function to make priming more convenient.

Were you testing production or your local env? I'm seeing ~350ms on prod. If that's accurate then I don't think it's a big deal.

PR #412, opened 2020-03-31
Speaker Feedback: Add check for whether a session is currently accepting feedback

### LINKED ISSUE #343: Speaker Feedback: Automatically close form after time limit
> The feedback form will be automatically closed after a set amount of time (most likely 2 weeks). We can send an automated reminder email to organisers 2 weeks after the event to remind them about sending the feedback to speakers.

What should this time limit be? 2 weeks? 4 weeks?

The automated email can go out to organizers as a reminder, but should not auto-send anything to speakers.

### LINKED ISSUE #394: Speaker Feedback: "Break" sessions don't need feedback
Sessions that are classified as "Break, Lunch, etc." currently show the feedback form when you add `/feedback` to the session URL. While someone might want to leave feedback on the lunch, we probably don't need to allow this 😆

It's probably easier to conditionally display the form, vs trying to conditionally add the endpoint.

coreymckrill:
This adds a function to comprehensively check whether a post can and is currently accepting feedback comments. The checks include whether it's the right type of session, and close the current time is to the session's start time, plus some other things. This is useful in both the REST API and front end templates for determining what kind of response or content to return.

Both the REST API controller and the feedback form renderer are also updated to use this function.

Fixes #343, Fixes #394

### Screenshots

free-lunch

1. Try viewing the feedback form for a session post (i.e. `site.wordcamp.test/session-slug/feedback`)
1. If the session type is set to regular and the session time is slightly before the current time, you should see the feedback form. Otherwise, you should see a message explaining why the feedback form isn't available. Try changing those parameters to get different messages.
1. Try submitting feedback for that same session post using the [REST API](https://github.com/WordPress/wordcamp.org/pull/353). The same error messages should be in the API responses when the session's conditions aren't met.
1. All phpunit tests should pass.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
@@ -6,6 +6,7 @@ use WP_Post, WP_Query;
 use function WordCamp\SpeakerFeedback\{ get_views_path, get_assets_url };
 use function WordCamp\SpeakerFeedback\Post\post_accepts_feedback;
 use const WordCamp\SpeakerFeedback\{ OPTION_KEY, QUERY_VAR };
+use const WordCamp\SpeakerFeedback\Post\ACCEPT_INTERVAL_IN_SECONDS;
 
 defined( 'WPINC' ) || die();
 
@@ -37,8 +38,7 @@ function has_feedback_form() {
 function render( $content ) {
 	global $post;
 
-	$feedback_window = WEEK_IN_SECONDS * 2;
-	$now             = date_create( time(), wp_timezone() );
+	$now = date_create( time(), wp_timezone() );
 
 	if ( has_feedback_form() ) {
 		$accepts_feedback = post_accepts_feedback( $post->ID );
@@ -62,7 +62,7 @@ function render( $content ) {
 		if ( $now->getTimestamp() < absint( $start_date ) ) {
 			$message = __( 'Feedback forms are not available until the event has started.', 'wordcamporg' );
 			$file    = 'form-not-available.php';
-		} elseif ( $now->getTimestamp() > absint( $end_date ) + $feedback_window ) {
+		} elseif ( $now->getTimestamp() > absint( $end_date ) + ACCEPT_INTERVAL_IN_SECONDS ) {
 			$message = __( 'Feedback forms are closed for this event.', 'wordcamporg' );
 			$file    = 'form-not-available.php';
 		} else {
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
@@ -38,7 +38,7 @@ function has_feedback_form() {
 function render( $content ) {
 	global $post;
 
-	$now = date_create( time(), wp_timezone() );
+	$now = date_create( 'now', wp_timezone() );
 
 	if ( has_feedback_form() ) {
 		$accepts_feedback = post_accepts_feedback( $post->ID );
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
@@ -55,9 +55,23 @@ function render( $content ) {
 
 		$content = $content . ob_get_clean(); // Append form to the normal content.
 	} elseif ( is_page( get_option( OPTION_KEY ) ) ) {
-		$wordcamp   = get_wordcamp_post();
-		$start_date = $wordcamp->meta['Start Date (YYYY-mm-dd)'][0] ?? $now->getTimestamp() + YEAR_IN_SECONDS;
-		$end_date   = $wordcamp->meta['End Date (YYYY-mm-dd)'][0] ?? $start_date + DAY_IN_SECONDS;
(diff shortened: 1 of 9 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
```diff
@@ -17,16 +27,52 @@ function has_feedback_form() {
...
+
 		ob_start();
-		require get_views_path() . 'form-feedback.php';
-		return $content . ob_get_clean();
+		require get_views_path() . $file;
+
+		$content = $content . ob_get_clean(); // Append form to the normal content.
 	} elseif ( is_page( get_option( OPTION_KEY ) ) ) {
+		$wordcamp   = get_wordcamp_post();
+		$start_date = $wordcamp->meta['Start Date (YYYY-mm-dd)'][0] ?? $now->getTimestamp() + YEAR_IN_SECONDS;
```
ryelle: Why default this to a year from now, so it's an arbitrary future date, or is the year important?

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/post.php
```diff
@@ -0,0 +1,66 @@
+<?php
```
ryelle, line 1: Why add this as a new file/namespace, instead of adding to `form.php`? It feels like an extra abstraction to me.  (no change requested, just curious)

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/post.php
```diff
@@ -0,0 +1,66 @@
...
+		return new WP_Error(
+			'speaker_feedback_post_not_supported',
+			__( 'This post does not support feedback.', 'wordcamporg' )
+		);
+	}
+
+	if ( 'publish' !== get_post_status( $post ) ) {
+		return new WP_Error(
+			'speaker_feedback_post_unavailable',
+			__( 'This post does is not available for feedback.', 'wordcamporg' )
```
ryelle:
```suggestion
			__( 'This post is not available for feedback.', 'wordcamporg' )
```

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
```diff
@@ -17,16 +27,52 @@ function has_feedback_form() {
...
+ * if more post type support is added.
+ *
+ * @global WP_Post $post
+ *
+ * @return string
  */
 function render( $content ) {
+	global $post;
+
+	$now = date_create( time(), wp_timezone() );
```
ryelle:
```suggestion
	$now = date_create( 'now', wp_timezone() );
```

This was `false` on my local env, the docs for [DateTime](https://www.php.net/manual/en/datetime.construct.php) say you should use `"now"` to get the current time with a timezone.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
ryelle: I think the fact that the start date is not the real UTC value will bite some of the UTC+ timezones — they might have morning sessions, but `/feedback` will say the event hasn't started. We could convert this value into the site's timezone, since that has a good chance of being correct. Something like

```php
$date_string = date( 'Y-m-d', $wordcamp->meta['Start Date (YYYY-mm-dd)'][0] );
$start_date  = date_create( $date_string, wp_timezone() )->getTimestamp();
```

coreymckrill: > Why default this to a year from now, so it's an arbitrary future date, or is the year important?

The thinking was that if the site doesn't have a meta value set for the start date, it probably hasn't been scheduled yet, in which case, the correct message would be `Feedback forms are not available until the event has started`. If the value defaulted to `0`, the message shown would be `Feedback forms are closed for this event`, which doesn't seem as accurate. The fact that it's a year ahead is arbitrary, it just needs to be far enough ahead that it won't fall into any window of time that we may set in the future.

coreymckrill: > I think the fact that the start date is not the real UTC value will bite some of the UTC+ timezones

Good call. I expanded on your suggested here: 223b0f61

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/post.php
coreymckrill: Two reasons:

1. The `post_accepts_feedback` function is also used in the REST API controller. Since it has multiple contexts, it seemed better to keep it separated from either context. That helps avoid the code making assumptions that are too specific to one particular context.
1. It seems possible that we will eventually have other functions that deal specifically with the posts that feedback comments are attached to. When I wrote this PR, I had at least one other in mind, but now I'm not thinking of it 😕

ryelle: Makes sense! thanks for answering my curiosity 🙂

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/form.php
ryelle: Looks good, and you were able to add comments which addressed my initial  🤔❓

ryelle (approved):
Looks good now 👍

PR #415, opened 2020-04-06
Blocks: Fix overeager sanitization

### LINKED ISSUE #349: SVG are stripped out of social block in speakers list block
There’s a problem with the speakers list block. The SVGs aren’t being loaded within the social links. They’re created in the individual speaker content, where they’re showing correctly. https://2020.geneve.wordcamp.org/orateurs-trices/

ryelle:
This PR removes the aggressive `wp_kses_post` calls from the render helper functions, and uses more context-aware sanitization & escaping functions — for example, esc_html on content that should not have HTML tags. `wp_kses_post` is still used on the title & content, but is added before `the_content` filter is run, so that it's only applied to user content, not rendered blocks.

Lastly this removes the `noscript` exception, since it's not necessary with the earlier sanitization.

Fixes #349

- Try viewing one of each block, there should be no visual changes.
- Replicate the issue in #349 by adding a social links block to a speaker, then in your speakers block, show the full content. When you view the post, you should see the social link icons.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/includes/content.php
+++ b/public_html/wp-content/mu-plugins/blocks/includes/content.php
@@ -31,6 +31,25 @@ function get_all_the_content( $post ) {
 }
 
 /**
+ * Get a trimmed excerpt from a post.
+ *
+ * @param int|WP_Post $post           Post ID or post object.
+ * @param int         $excerpt_length Number of words in excerpt.
+ *
+ * @return string The escaped post excerpt.
+ */
+function get_trimmed_content( $post, $excerpt_length = 55 ) {
+	$post = get_post( $post );
+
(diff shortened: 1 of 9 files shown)


iandunn:
~What are some of the cases that you ran into that this fixes?~

Er, nevermind, I missed the link to #349.

ryelle:
yeah, #349; there's also https://github.com/WordPress/wordcamp.org/pull/236 which was fixed with a workaround (that https://github.com/WordPress/wordcamp.org/pull/415/commits/840a9fdae7c12f634590c9c831ac393e63bd387a removes); and @coreymckrill mentioned something about `srcset` & `sizes` being stripped, but I can't actually replicate that one.

iandunn (approved):
👍

PR #416, opened 2020-04-06
Schedule block prereqs

iandunn:
These are miscellaneous small commits that I made while working on #104. I figured it'd be easier to review if they were split off into a separate PR.

To test the crash boundary, just introduce some fatal error into the live schedule block. You should see an error message on the front end, instead of the block being empty without explanation.

_(the linter failures are in old, unrelated code.)_

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,12 @@
 	"browserslist": [
 		"extends @wordpress/browserslist-config"
 	],
+	"eslintConfig:comment": {
+		"root": "Setting this to true is necessary to avoid merging WordCamp.org configs with any that exist in parent folders in the developer's environment. See https://github.com/eslint/eslint/commit/c99e9b621fe46b8535cf59fe489977dd36cb2a39."
+	},
+	"eslintConfig": {
+		"root": true
+	},
 	"eslintIgnore": [
 		"*.min.js"
 	]
--- a/public_html/wp-content/mu-plugins/blocks/readme.md
+++ b/public_html/wp-content/mu-plugins/blocks/readme.md
@@ -7,6 +7,21 @@
 2. Run `yarn build` to generate the built files.
 3. Run `yarn start`  while developing to continuously watch the files, this will automatically re-build the files when the source changes.
 
+
+## Scripts
+
+Check `package.json` for the scripts that are available around builds, testing, linting, etc.
+
+General notes:
+
+* When passing a path as an argument, it generally needs to start from the `mu-plugins/blocks` folder,
+	* e.g. `yarn workspace wordcamp-blocks lint:js source/blocks/speakers`
+	* Alternatively, `npm run lint:js source/blocks/speakers`
+* Arguments for `npm run` can be passed as you'd expect, but arguments for the proxied command (e.g., `eslint`) need to be separated by a `--`. That isn't needed for `yarn`, though.
+	* e.g., `npm run lint:js -- --fix`, `npm run test -- -h`.
+	* e.g., `yarn workspace wordcamp-blocks lint:js --fix`
+
+
 ## Testing
 
 We use Jest for testing, run `yarn test`. With Jest, you can create snapshot tests for components.
--- a/public_html/wp-content/mu-plugins/blocks/source/utils/render-frontend.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/utils/render-frontend.js
@@ -1,11 +1,63 @@
 /**
- * External dependencies
+ * WordPress dependencies
  */
-import { render } from 'react-dom';
+import { Component, createInterpolateElement, render } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
 
 const getPropsFallback = () => ( {
 	attributes: {},
 } );
+
+/*
+ * Catch errors thrown by children and display a friendly error message.
+ *
+ * Gutenberg does this in the back end via the internal `BlockCrashBoundary` class, but we need to replicate that
+ * for the front end.
+ */
+class FrontBlockCrashBoundary extends Component {
+	/**
+	 * Constructor.
+	 *
+	 * @param {Object} props
+	 */
+	constructor( props ) {
+		super( props );
+
+		this.state = { hasError: false };
+	}
+
+	/**
+	 * Catch errors thrown by children components and update state.
+	 *
+	 * @return {Object}
+	 */
+	static getDerivedStateFromError() {
+		return { hasError: true };
+	}
+
+	/**
+	 * Render an error message in place of the child component.
+	 *
+	 * @return {Element}
+	 */
+	render() {
+		const { hasError } = this.state;
+		let content;
+
+		if ( hasError ) {
+			content = createInterpolateElement(
+				__( 'There was an error trying to render this content, please try another browser or device. If that doesn\'t work, please <a>contact us</a>.', 'wordcamporg' ),
+				{
+					a: <a href={ 'https://central.wordcamp.org/contact-us/' } >#21441-gutenberg</a>,
+				}
+			);
+		} else {
+			content = this.props.block;
+		}
+
+		return content;
+	}
+}
 
 /**
  * Renders a block component in the place of a specified set of selectors.
@@ -29,7 +81,12 @@ export default ( selector, Block, getProps = getPropsFallback ) => {
 
 			element.classList.remove( 'is-loading' );
 
-			render( <Block { ...props } attributes={ attributes } />, element );
+			render(
+				<FrontBlockCrashBoundary
+					block={ <Block { ...props } attributes={ attributes } /> }
+				/>,
+				element
+			);
 		} );
 	}
 };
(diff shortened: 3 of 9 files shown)


public_html/wp-content/mu-plugins/blocks/.eslintrc.js
```diff
@@ -131,5 +131,11 @@ module.exports = {
 		 * @todo `@param` tags should align the variable name and description, just like in PHP.
 		 */
 		'jsdoc/require-returns-description': 'off',
+
+		/*
+		 * This has false positives with every translated string, due to the way `createInterpolateElement()`
+		 * is designed.
+		 */
+		'jsx-a11y/anchor-has-content': 'off',
```
ryelle: Since there's no way to disable this just for `createInterpolateElement`, this could let real errors through.

We can use the inline comment format to disable this at the source though, so it won't flag the error in `render-frontend`:

```
// eslint-disable-next-line jsx-a11y/anchor-has-content
```

public_html/wp-content/mu-plugins/blocks/source/utils/render-frontend.js
```diff
@@ -3,10 +3,67 @@
  */
 import { render } from 'react-dom';
 
+/**
+ * WordPress dependencies
+ */
+import { Component, createInterpolateElement } from '@wordpress/element';
```
ryelle: Like I mentioned in https://github.com/WordPress/wordcamp.org/pull/104/commits/22083319f57f3c4bb7f95a831733dc5a3e2a726f , you should pull in `render` from `@wordpress/element` since we're not actually bypassing the `wp-element` dependency.

package.json
```diff
@@ -20,6 +20,12 @@
 	"browserslist": [
 		"extends @wordpress/browserslist-config"
 	],
+	"eslintConfig:comment": {
+		"root": "Setting tnis to true is necessary to avoid merging WordCamp.org configs with any that exist in parent folders in the developer's environment. See https://github.com/eslint/eslint/commit/c99e9b621fe46b8535cf59fe489977dd36cb2a39."
```
ryelle:
```suggestion
		"root": "Setting this to true is necessary to avoid merging WordCamp.org configs with any that exist in parent folders in the developer's environment. See https://github.com/eslint/eslint/commit/c99e9b621fe46b8535cf59fe489977dd36cb2a39."
```

public_html/wp-content/mu-plugins/blocks/readme.md
```diff
@@ -7,6 +7,19 @@
...
 
+
+## Scripts
+
+Check `package.json` for the scripts that are available around builds, testing, linting, etc.
+
+General notes:
+
+* When passing a path as an argument, it generally needs to start from the `mu-plugins/blocks` folder,
+	* e.g. `npm run lint:js source/blocks/schedule`
```
ryelle: Since we recommend using yarn, our docs should be consistent:
```suggestion
	* e.g. `yarn workspace wordcamp-blocks lint:js source/blocks/schedule`
```

If you want to leave the npm command here, that's fine, but might be confusing to other devs.

public_html/wp-content/mu-plugins/blocks/readme.md
```diff
@@ -7,6 +7,19 @@
...
+* Arguments for `npm run` can be passed as you'd expect, but arguments for the proxied command (e.g., `eslint`) need to be separated by a `--`.
+	* e.g., `npm run lint:js -- --fix`, `npm run test -- -h`.
```
ryelle, line 21: You don't need the `--` in yarn, so if you remove the `npm` reference above you  can also remove this note.

ryelle (commented):
All the current blocks work as expected. I couldn't figure out how to test the error boundary, a fatal PHP error broke the whole site, and invalid JS wouldn't build. I'm happy to test again if you could give a step-by-step for that.

iandunn:
The boundary protects against fatal JS runtime errors, e.g., an uncaught exception, calling a function that isn't defined, etc.

public_html/wp-content/mu-plugins/blocks/.eslintrc.js
iandunn: I agree that it's not ideal, but it doesn't seem likely that we'd ever create an anchor without text, so I think the likelyhood of this causing tangible problems in practice is very low. It seems much better to me than the hassle and clutter of eslint comments every time we translate a link.

I opened https://github.com/WordPress/gutenberg/issues/21441 in case there are any upstream solutions in the future.

package.json
iandunn: Good catch, thanks.

public_html/wp-content/mu-plugins/blocks/readme.md
iandunn: I don't use yarn, so yarn commands are "confusing" to me 🤷‍♂ . Happy to add them if you'd like, though.

public_html/wp-content/mu-plugins/blocks/readme.md
iandunn: I'd prefer to keep the npm commands.

public_html/wp-content/mu-plugins/blocks/source/utils/render-frontend.js
iandunn: c9c32a0b274a21f01d47d55e790a3408da5a28ac

public_html/wp-content/mu-plugins/blocks/readme.md
ryelle: ok, I guess since `npm` does still work we can leave it in as an alternative.

ryelle: btw, I'm happy to write up the docs for yarn if you want.

public_html/wp-content/mu-plugins/blocks/.eslintrc.js
iandunn: One potential workaround that's cleaner would be doing something like this:

```js
a: <a href={ 'https://central.wordcamp.org/contact-us/' } >%s</a>,
```

The `%s` wouldn't be used anywhere, so it's potentially confusing for our future selves, but we could add a `todo` note to go back and remove them when/if WordPress/gutenberg#21441 is resolved.

ryelle: > it doesn't seem likely that we'd ever create an anchor without text

Sure, but the rule is there to catch any mistakes. Having to add the comment every time would also ensure that we _know_ that the interpolated component has content too.

> we could add a todo note …

I don't see how this is that different than adding the eslint comment, and it's more confusing to our future selves.

public_html/wp-content/mu-plugins/blocks/readme.md
iandunn: How does 3f8272749584cdb333a48e4b6d38507421729b74 look to you?

ryelle: Looks good, thanks!

public_html/wp-content/mu-plugins/blocks/.eslintrc.js
iandunn: I still think the downsides outweigh the benefits.

I wasn't picturing adding it every time, just on one of them. Or a github issue if you'd prefer.

iandunn: Maybe something like `a: <a href={ 'https://central.wordcamp.org/contact-us/' } >#21441-gutenberg</a>` would be more self-documenting.

ryelle: I don't want us to disable the rule, so as long as adding content into the `<a>` doesn't introduce side effects, we should go with that.

The `%s` style makes more sense to me, since that's basically how printf etc works, otherwise I think future-me would come back and wonder why we're outputting a gutenberg ticket (we're not, but that's what I read first).

And (unsurprisingly) I'd prefer this be surfaced in an issue (blocked by that GB one).

iandunn: I liked `%s` at first for that reason, but then it seemed like it'd be confusing, because in the future I might actually try to change it, like making it `%s %s` if I wanted to add another argument to the string.

Linking to the G issue is definitely weird, but I think that's helpful, because it makes it obvious that something odd is happening, and gives a curious dev all of the information they need to figure out why.

ryelle: But that's not how `createInterpolateElement` works, there's no way to do that in practice. It feels much less clear to me to use the github issue inline, but if we're at an impasse otherwise then it's fine.

iandunn: 👍 , 6f732edbbe2a0cc2002d37f9ad6778bf97bd2d59

ryelle (approved):
I was able to test the crash boundary, that's a nice improvement. Looks good to merge now 👍

PR #417, opened 2020-04-06
Tools: Try using `phpcs-changed` to limit php linting to changed lines

ryelle:
We have a lot of large files with a lot of errors, and while we can go through and commit fixes to these while working, it's not feasible to fix all errors this way. This introduces [phpcs-changed](https://github.com/sirbrillig/phpcs-changed) to only report on changed lines in a PR. By changing travis to only sniff changed lines, we could actually require that CI passes before merging ✅

By default, `phpcs-changed --git` works on staged changes (so you could use it locally pre-commit, I think); but it needs to be called manually to diff between branch changes. That's where `phpcs-branch.php` comes in, travis calls that script, which uses the diff between `TRAVIS_BRANCH` (the PR target) and `HEAD` (the branch) to get the changes in each modified file. It also runs phpcs directly for added files (a better dev-ex, otherwise the file is null).

You can run the phpcs-branch file directly, if you want, but you need to define the target branch: `export TRAVIS_BRANCH='production'; php .github/bin/phpcs-branch.php`

Otherwise the CI check will be the test. This should report the newly introduced errors (will fix those commits before merging, of course)

### DIFF
--- a/.github/bin/phpcs-branch.php
+++ b/.github/bin/phpcs-branch.php
@@ -1,0 +1,68 @@
+#!/usr/bin/php
+<?php
+/**
+ * Report on phpcs violations that are introduced in the current branch (not production).
+ * Runs `phpcs` as normal on new files, and `phpcs-changed` on modified files. `phpcs-changed` will only report
+ * on changed lines in each modified file.
+ *
+ * How to use: php .github/bin/phpcs-branch.php
+ */
+
+// phpcs:ignoreFile
+namespace WordCamp\Bin\PHPCS_Changed;
+
+function run_phpcs( $file, $bin_dir ) {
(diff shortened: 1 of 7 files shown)


ryelle:
ok, this didn't actually work, so I've cancelled my review requests until i fix it.

iandunn (approved):
Great idea 👍

ryelle:
Okay, the issue with exit status is fixed now— you can see it failed in [this job](https://travis-ci.com/github/WordPress/wordcamp.org/jobs/316233708), then when I fixed the issues in https://github.com/WordPress/wordcamp.org/pull/417/commits/fb5a0c90e4fee68983ba3ed3044b98b2427538b6, it passed.

iandunn (approved):

coreymckrill (approved):

Issue #342, opened 2020-01-27
Speaker Feedback: Track "helpfulness" of feedback


ryelle:
> We can add a “Was this feedback useful?” question for the speaker to submit as a way of tracking the value of the feedback.

Should this be per comment? Could use the same pattern as moderation?

![Speaker Feedback-help](https://user-images.githubusercontent.com/541093/73208354-520daa00-4114-11ea-8db1-7804c8522e5e.png)

Or should this be a specific question somewhere on the page?

coreymckrill:
> Should this be per comment? ... Or should this be a specific question somewhere on the page?

I think it depends on whether we want to assess the helpfulness of the speaker feedback functionality as a whole, or track the amount of feedback that is considered helpful vs unhelpful.

Going the more granular route seems best to me, but then I guess the risk is that speakers wouldn't take the time to rate the helpfulness of each individual comment. If they just marked the "unhelpful" ones for example, that might skew our data.

coreymckrill:
Maybe instead of having two options, we just have a "Helpful" button. Speakers would only click the button for things that were truly helpful to them. Then we could measure the _percentage_ of feedback that was deemed helpful on each individual session.

I've seen similar schemes on things like Yelp reviews. You can mark something as "Cool" or "Helpful", but there aren't any negative options.

ryelle:
> Then we could measure the percentage of feedback that was deemed helpful on each individual session.

I like that idea. There's probably a lot of overlap between actively unhelpful comments & spam (or otherwise moderated-away content).

Would we also want to track percentage of moderated comments, to keep an eye on negativity? that might need a separate status (since they'd be trashed now).

coreymckrill:
> Would we also want to track percentage of moderated comments, to keep an eye on negativity? that might need a separate status (since they'd be trashed now).

If we wanted to track comments that aren't spam, but still aren't worth showing to speakers, then a separate status would probably be the way to go. But I'm not sure I understand what we'd gain by tracking that specifically.

ryelle:
Whether people are being jerks, really… 

Is the "helpful" button only for the speaker(s)? I'm wondering if it would only be on the frontend view we're talking about in #347, or also in the wp-admin list.

coreymckrill:
> Is the "helpful" button only for the speaker(s)?

I was thinking that it was. I think it would be weird if the organizers could mark which feedback was "helpful" for the speakers.

coreymckrill:
[...]

PR #420, opened 2020-04-07
Speaker Feedback: Add REST endpoint for updating feedback with specific meta values

coreymckrill:
Adds the ability to mark a feedback comment as "helpful" via a new API endpoint. This is stored as comment meta, just like the other values that make up the original feedback submission. However, "helpful" can only be set during feedback update, not creation. This is achieved by introducing a "context" for each possible meta field.

This also makes some changes to the way request objects are processed by our endpoint controller to allow for the flexibility to consider the context of the request.

Refs #342 

1. To make is easier to send authenticated requests to the endpoint via something like [Insomnia](https://insomnia.rest/), you'll probably want to temporarily install the [Basic Auth plugin](https://github.com/WP-API/Basic-Auth).
1. Create some feedback comments on one or more session posts.
1. To mark one of the comments as helpful, you'll need the ID of the comment.
1. The endpoint URL will be `[your-test-site].wordcamp.org/wp-json/wordcamp-speaker-feedback/v1/feedback/[comment ID]`
1. Currently there is only valid parameter for the payload:
    * `meta[helpful]` (A boolean-ish value)
1. The response to a successful request to the endpoint should be a 201 status and an empty array.
1. Try sending a non-boolean value. You should get an error message.
1. Try including other existing meta values besides `helpful` (`rating` or `q1` for example). These should be ignored and not changed during a successful request.
1. Try including `meta[helpful]` when sending a request to [create a new feedback](https://github.com/WordPress/wordcamp.org/pull/353). It should also be ignored and not be present for the newly created comment.
1. Try authenticating as a non-admin/editor user who is also not a speaker on the session. You should get an error message.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
@@ -59,22 +59,28 @@ function register_meta_fields() {
  * WordCamp\SpeakerFeedback\REST_Feedback_Controller::duplicate_check() method will
  * need to be updated.
  *
- * @param string $key Optional. A specific key to get the schema for.
+ * @param string $context Optional. The context in which the field schema is being used.
+ *                        'all', 'create', or 'update'. Defaults to 'all'.
+ *                        Note that this does not map to the context values used in
+ *                        the REST API.
+ * @param string $key     Optional. A specific key to get the schema for.
  *
  * @return array
  */
-function get_feedback_meta_field_schema( $key = '' ) {
+function get_feedback_meta_field_schema( $context = 'all', $key = '' ) {
 	$schema = array(
 		'version' => array(
+			'sft_context'       => array( 'create' ),
 			'description'       => 'The version of the feedback questions.',
 			'type'              => 'integer',
 			'single'            => true,
 			'sanitize_callback' => 'absint',
 			'show_in_rest'      => false,
 			'default'           => META_VERSION,
-			'required'          => true,
+			'required'          => false,
 		),
 		'rating'  => array(
+			'sft_context'       => array( 'create' ),
 			'description'       => 'The rating. A number between 1 and 5.',
 			'type'              => 'integer',
 			'single'            => true,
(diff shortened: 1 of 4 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
```diff
@@ -124,8 +133,29 @@ function get_feedback_meta_field_schema( $key = '' ) {
...
 		),
+		'helpful' => array(
+			'sft_context'       => array( 'update' ),
+			'description'       => 'The speaker found this feedback helpful.',
+			'type'              => 'boolean',
+			'single'            => true,
+			'sanitize_callback' => 'wp_validate_boolean',
+			'show_in_rest'      => false,
+			'default'           => false,
+			'required'          => false,
```
ryelle, line 144: Should this be required? I can call the update endpoint with `meta[q1]=fake` and it returns 201, even though it doesn't update anything.

ryelle (approved):
Left one question, but it's not a blocker.

Another issue I noticed is that the speaker needs to be a user on the site for the cap check to work, my guess is that they don't have `read` caps. Once I added the user to the site, everything worked as expected. This can also be addressed in a followup issue/PR.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
coreymckrill: Ah, good point. I hesitate to make it `required` because semantically it implies that all feedback should have that value, when that is not the case. However, in 4bcda31e I made it so that if there's nothing left after meta validation callback strips out invalid keys (like `q1` in your case above, which is invalid in the update context), it will throw an error instead of returning 201. Do you think that's sufficient?

coreymckrill:
> Another issue I noticed is that the speaker needs to be a user on the site for the cap check to work, my guess is that they don't have read caps. Once I added the user to the site, everything worked as expected.

Thanks, that should definitely be addressed. I've added #425 to look into it further.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
ryelle: Yeah, that looks good 👍

PR #421, opened 2020-04-08
Speaker Feedback: Fix capability checks

coreymckrill:
When I previously created the meta caps for viewing feedback comments, I incorrectly assumed that the post meta value `_wcpt_speaker_id` that is attached to Session posts was the actual user ID of the speaker. In fact it is the _post ID_ of the Speaker post associated with the Session.

This introduces a function that can take a Session post ID and derive the actual user IDs of the speakers associated with the session, and uses that in the meta caps. It also updates existing tests and adds a couple of new ones.

1. All the phpunit tests should pass.
1. There isn't really a UI yet for speakers to view their feedback, so I'm not sure how else to test this right now.


ryelle (approved):
Looks good, tests reflect the real data now & pass 👍

Issue #347, opened 2020-01-28
Speaker Feedback: UI for speakers to review their feedback


coreymckrill:
The [scoping document](https://make.wordpress.org/community/2020/01/27/speaker-feedback-tool-final-outline/) says:

> An option will be added to allow this feedback view to be shared with the speaker. Speakers can be sent an email with a unique URL that can be used to view the feedback. We could also allow speakers to access the feedback directly if we have their WordPress.org username, and they are logged into the site.

And:

> Speakers will receive a link from the organisers to view their feedback. We can add a “Was this feedback useful?” question for the speaker to submit as a way of tracking the value of the feedback.

But this still leaves some questions:

* Will this view live on the frontend of the site, or the backend?
  * If the frontend, where would the content be displayed? How would it be styled?
    * A dedicated page that shows feedback for whichever wporg user is currently logged in?
    * On the session page, in a section that is only visible to the speaker, via login or access key query parameter?
  * If the backend, speakers would probably need to be added as users on the WordCamp site and be logged in in order to access their feedback.

---

### Update

After discussion on this issue, the current plan is:

* Speakers can view feedback on their session by going to the same URL where attendees go to leave feedback, i.e. `2020.city.wordcamp.org/my-cool-session/feedback`.
* They must be logged into the WordPress network. Then we can detect that the current user is one of the speakers on the current session, and replace the feedback form with a list of the feedback that has been moderated and approved by the organizers.
* This could also show the speaker if there is feedback that _hasn't_ been moderated yet, so that they can bug organizers about it.
* This is also where speakers would be able to mark specific feedback comments as "helpful" (see #342)

For a live example of the current feedback page, see [2017.testing.wordcamp.org/session/demo/feedback/](https://2017.testing.wordcamp.org/session/demo/feedback/)

ryelle:
I was under the impression this would be the backend, hence why we're doing #340 — the "UI for speakers to review feedback" is just the "my feedback" view of the comments list table in wp-admin. (340 is the issue I created to cover this bullet point from the scope doc)

> If the backend, speakers would probably need to be added as users on the WordCamp site and be logged in in order to access their feedback.

Yeah, I think that's part of #340 — working out the permissions to make this work. I thought we said somewhere that it would be OK that this is requiring a wporg login, since we technically "require" that of speakers already.

_I'm removing needs design until we come to a decision, just so no designers pick this up while we're still figuring it out._

hlashbrooke:
[...]

PR #424, opened 2020-04-09
Speaker Feedback: Add feedback rendering helper functions

ryelle:
This adds three functions that can be reused by the front & back end when viewing feedback, `render_rating_stars`, `render_feedback_rating`, and `render_feedback_comment`. The frontend doesn't exist yet, so this updates the list table view to use these new functions. 

There should be no visual changes, the only markup change is for screen reader interaction. The star container is now an "image", so that we can label it as one unit and prevent screen readers from accidentally interacting with the raw SVGs.

See #347 

(another PR for the frontend view is in progress 🙂  )

1. View the Sessions -> Feedback table
2. It should look the same as on `production`, content should have questions & answers, rating should have the star ratings

Optionally, check the accessibility:
- use browser dev tools to check the accessibility information of the star container & stars
- Or use a screen reader & move through the list table to hear how the ratings are labelled

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/view.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/view.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace WordCamp\SpeakerFeedback\Form;
+namespace WordCamp\SpeakerFeedback\View;
 
 use WP_Post, WP_Query;
 use function WordCamp\SpeakerFeedback\{ get_views_path, get_assets_url };
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/view.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/view.php
@@ -3,7 +3,8 @@
 namespace WordCamp\SpeakerFeedback\View;
 
 use WP_Post, WP_Query;
-use function WordCamp\SpeakerFeedback\{ get_views_path, get_assets_url };
+use function WordCamp\SpeakerFeedback\{ get_views_path, get_assets_url, get_assets_path };
+use function WordCamp\SpeakerFeedback\Comment\get_feedback_comment;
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/view.php
```diff
@@ -103,6 +104,94 @@ function( $key, $question ) {
...
+
+/**
+ * Render a rating of X out of Y stars. Used by `render_feedback_rating`.
+ *
+ * @param int $rating    The selected rating value (stars below this will be "filled").
+ * @param int $max_stars The total rating value (will render this many stars).
+ *
+ * @return string The feedback rating stars in HTML.
+ */
+function render_rating_stars( $rating, $max_stars = 5 ) {
```
coreymckrill, line 170: If we're going to show an [average rating](https://github.com/WordPress/wordcamp.org/issues/347#issuecomment-610594528), we might need to adapt this to accept `$rating` as a float and be able to show a half star or something...

coreymckrill (approved):
👍 Looks good testing locally. Grouping the stars as a "graphic" seems like a good way to go for a11y.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/view.php
ryelle: I've been rounding to an int on the "average", no half-stars. Do you think we need half-stars?

coreymckrill: I dunno. It seems like a common practice. The Plugin Directory does it for plugin reviews. But... maybe not a big deal?

ryelle: I'll leave all this as-is for now, and not worry about it — that'd be easy enough to change later if we want to.

Issue #425, opened 2020-04-09
Speaker Feedback: Allow logged-in speakers with no role on a site to view their feedback


coreymckrill:
Speakers who are logged into the WordCamp.org network should be able to view their feedback on a WordCamp site without the organizer needing to add them as a user. Right now our meta capabilities mapping specifies that if the current user is a speaker for the current session, the only capability they need is `read`, but this still implies that they have been added as a user on the site.

A couple of potential ways to fix this:

* Map to some universal capability instead of `read` (not sure if this actually exists)
* Assign a custom capability to the speaker at the same time that we map it.

ryelle:
Another issue I noticed today is that site admins can view speaker feedback too— I haven't tested enough to see if this is just a super-admin power, though. We'll want to make sure organizers see the form, not the feedback list.

coreymckrill:
🤔 Organizers have to see the feedback list so they can approve each comment before it's shown to speakers...

ryelle:
In wp-admin, right— but on the frontend (in the PR I haven't pushed yet 🤦🏼‍♀️) I'm checking `if ( current_user_can( 'read_post_' . COMMENT_TYPE, $post->ID ) ) {` to decide whether to show the feedback form or the feedback list. Maybe I should just check the speaker IDs agains the session directly?

coreymckrill:
Oh, gotcha. Yeah, I would just check against the session directly. All the meta caps assume that site admins can see stuff too.

PR #428, opened 2020-04-10
Speaker Feedback: Display feedback to session speakers

### LINKED ISSUE #347: Speaker Feedback: UI for speakers to review their feedback

ryelle:
When a session's speaker views the feedback URL, switch out the feedback form for a list of approved feedback. The speaker can sort by date & rating, and toggle only the helpful feedback. They can also mark each feedback as helpful (or un-mark it, by clicking the button again).

Fixes #347 

### Screenshots

On Twenty Twenty:
-ZDHUZlMAl

[See screenshots from other themes here](https://cloudup.com/c22cZShSide)

1. Make sure you have some feedback approved on a session
2. Make sure the session has an attached speaker, and that speaker has the wp.org profile filled in to a real user on the network
3. Logged in as that user, view `your-session/feedback`
4. You should see the list of approved feedback
5. You should be able to sort the feedback using the filters at the top of the page
6. You should be able to mark feedback as helpful (need to be a user on the site too, until #425 is fixed)
7. It should say how many approved feedback + how many pending, or "No feedback has been submitted yet."

Try it with different sessions— one with no feedback submitted, or only pending, or you're not a speaker on, etc.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
@@ -67,7 +67,7 @@ class Walker_Feedback extends Walker_Comment {
 		$comment_id = $comment->comment_ID;
 
 		?>
-		<div <?php comment_class( '', $comment ); ?>>
+		<div <?php comment_class( 'speaker-feedback__comment', $comment ); ?>>
 			<article id="speaker-feedback-<?php echo absint( $comment_id ); ?>" class="speaker-feedback__comment-body comment-body">
 				<header class="speaker-feedback__comment-meta comment-meta">
 					<div class="speaker-feedback__comment-author comment-author vcard">
@@ -95,7 +95,7 @@ class Walker_Feedback extends Walker_Comment {
 					<?php render_feedback_comment( $comment ); ?>
 				</div><!-- .comment-content -->
 
-				<footer class="speaker-feedback__comment-meta comment-meta">
+				<footer class="speaker-feedback__helpful">
 					<form>
 						<span id="sft-helpful-<?php echo absint( $comment_id ); ?>">
 							<?php esc_html_e( 'Was this feedback helpful?', 'wordcamporg' ); ?>
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
@@ -96,18 +96,17 @@ class Walker_Feedback extends Walker_Comment {
 				</div><!-- .comment-content -->
 
 				<footer class="speaker-feedback__helpful">
-					<form>
-						<span id="sft-helpful-<?php echo absint( $comment_id ); ?>">
-							<?php esc_html_e( 'Was this feedback helpful?', 'wordcamporg' ); ?>
-						</span>
-						<button
-							type="submit"
-							aria-pressed="<?php echo ( $comment->helpful ) ? 'true' : 'false'; ?>"
-							aria-describedby="sft-helpful-<?php echo absint( $comment_id ); ?>"
-						>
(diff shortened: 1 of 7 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/views/view-feedback.php
```diff
@@ -0,0 +1,98 @@
...
+			<div class="speaker-feedback__filter-helpful">
+				<label>
+					<input
+						name="helpful"
+						type="checkbox"
+						value="yes"
+						id="sft-filter-helpful"
+						<?php checked( $show_helpful ); ?>
+					/>
+					<?php esc_html_e( 'Only show comments marked as helpful', 'wordcamporg' ); ?>
```
coreymckrill:
```suggestion
					<?php esc_html_e( 'Only show feedback marked as helpful', 'wordcamporg' ); ?>
```

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
```diff
@@ -0,0 +1,114 @@
...
+						<?php esc_html_e( 'Was this feedback helpful?', 'wordcamporg' ); ?>
+					</span>
+					<button
+						type="button"
+						data-comment-id="<?php echo absint( $comment_id ); ?>"
+						aria-pressed="<?php echo ( $comment->helpful ) ? 'true' : 'false'; ?>"
+						aria-describedby="sft-helpful-<?php echo absint( $comment_id ); ?>"
+					>
+						<?php esc_html_e( 'Yes', 'wordcamporg' ); ?>
+					</button>
```
coreymckrill, line 109: It seems like it might not be obvious that you can click the button again to _un_mark something as helpful. I'm wondering if it would be worth it to:

* When the button is focused/hovered and the feedback is already marked as helpful, change the icon from a checkmark to an 'X'
* Modify `aria-describedby` or something so that the screen reader makes it clear that clicking the button again will unmark it.

coreymckrill (approved):
👍 Worked great in local testing. A couple of comments, but no blockers.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
ryelle: Good ideas — I'm going to merge this & start that in a new PR  🙂

Issue #422, opened 2020-04-08
Speaker Feedback: Inform speakers when they have approved feedback comments


coreymckrill:
[The scope](https://docs.google.com/document/d/1-qeTPpRTRV4zXtEaMNjcyfDN_ks-USjT6h8NgZJZlLA/edit) and #411 both mention that organizers should somehow notify speakers manually when they have feedback comments available.

> After feedback has been received, organisers will send a unique link to speakers for them to view their feedback.

Manually doing this for each speaker sounds like a hassle, and even automating the process so that it's one click of a button on a screen somewhere still sounds like something that a lot of organizers won't end up doing, especially in the flurry of other event wrap up tasks.

I think it would be better to completely automate this. I can think of a couple of slightly different approaches:

1. Run a cron job on the site _once per day_ that looks for newly approved feedback comments. Email each speaker who has new feedback with link(s) to where they can [read](https://github.com/WordPress/wordcamp.org/issues/347) these comments (there was previous discussion of sending an email for each new comment, but that would be a bit much). In the days immediately after the event that might mean that a speaker gets one email per day for a while (if the organizers are on top of it). Or it might mean that they just get one on the day when the organizers go through and moderate all the feedback at once after [they get a reminder](https://github.com/WordPress/wordcamp.org/issues/411). (This would be my preferred approach.)
    1a. Instead of notifying about approved feedback once a day, notify about newly submitted feedback once a day, regardless of whether an organizer has approved it yet. Then speakers might be motivated to ask the organizers to go in and approve it.
1. Regardless of how much has been approved, run a one-time cron job to send _one_ email to each speaker after the two week window for leaving feedback has closed.

Note: with either of these approaches, or just the manual way, there should be a way for speakers to **opt out** of receiving notifications about feedback.

hlashbrooke:
I like the daily cron job idea that sends an email if there are new comments to read. I imagine the 1a option might just cause frustration for speakers if they keep getting emails and organisers aren't moderating the comment - I get the intent and I like where it's headed, but I don't know if that's a good practical option.

We could even couple that with option 2 as a final "you have feedback and we won't contact you about it again" email.

PR #429, opened 2020-04-10
Speaker Feedback: Remove the session author from getting session comment notifications

ryelle:
Currently, since the speaker is set as the session author, approved comments trigger email notifications to that account. Speakers shouldn't be getting these notifications, since we'll be sending custom notifications (see #422). This PR removes the emails for feedback comments, so no emails will be sent.

1. Open MailCatcher, and mark a comment approved
2. You should not get an email.


coreymckrill (approved):
👍 Good catch!

PR #431, opened 2020-04-13
Speaker Feedback: Improve flow when marking/unmarking feedback helpfulness

### LINKED ISSUE #430: Speaker Feedback: Indicate that feedback helpfulness can be un-marked
From https://github.com/WordPress/wordcamp.org/pull/428#discussion_r406959820

>It seems like it might not be obvious that you can click the button again to _un_mark something as helpful. I'm wondering if it would be worth it to:
>
> - When the button is focused/hovered and the feedback is already marked as helpful, change the icon from a checkmark to an 'X'
> - Modify aria-describedby or something so that the screen reader makes it clear that clicking the button again will unmark it.

ryelle:
The feedback UI implemented in #428 was a little confusing when it comes to toggling helpfulness of feedback. We only have "helpful" and "neutral" states, so we don't want to imply it's possible to mark comments as unhelpful. Now feedback submissions show different action labels reflecting whether they're helpful or neutral, and clicking the button makes it more obvious what will happen in either state. This also adds a `speak` notification, so that screen reader users will hear that a change has happened.

Fixes #430 

### Screenshots

Default/neutral feedback gets the actionable question:

neutral

Helpful feedback gets the status label + "Remove" button:

helpful

Screen reader users will hear status updates when the API request finishes:

```
Feedback marked as helpful.
```

```
Feedback unmarked as helpful
```

@enriquesanchez This is a bit different than what you designed in #347, so I'd like your 👍 / 👎  Also open to different copy for labels & screen reader content. 🙂 

0. Build the branch with `yarn workspace wordcamp-speaker-feedback build`
1. Make sure you have some feedback approved on a session
2. Make sure the session has an attached speaker, and that speaker has the wp.org profile filled in to a real user on the network
3. Logged in as that user, view your-session/feedback
4. You should see the list of approved feedback
5. Mark & unmark some feedback submissions
6. The labels should update shown above
7. Reload the page, and each marked feedback should display correctly
8. Try with a screen reader, you should hear "Feedback marked as helpful"/"Feedback unmarked as helpful".

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -74,8 +74,8 @@
 
 	function onHelpfulClick( event ) {
 		event.preventDefault();
-		var button = event.target;
-		var isHelpful = 'true' === button.getAttribute( 'aria-pressed' );
+		var button = $( event.target ).closest( 'button' ).get( 0 );
+		var isHelpful = 'is-helpful' === button.className;
 
 		wp.apiFetch( {
 			path: '/wordcamp-speaker-feedback/v1/feedback/' + button.dataset.commentId,
@@ -85,7 +85,16 @@
 			},
 		} )
 			.then( function() {
-				button.setAttribute( 'aria-pressed', isHelpful ? 'false' : 'true' );
+				$( button ).toggleClass( 'is-helpful' );
+				var $label = $( '#' + button.getAttribute( 'aria-describedby' ) );
+				if ( isHelpful ) {
+					// Previous state was helpful, has been un-marked, label should flip back to "mark as helpful".
+					wp.a11y.speak( SpeakerFeedbackData.messages.unmarkedHelpful, 'polite' );
+					$label.text( SpeakerFeedbackData.messages.markHelpful );
+				} else {
+					wp.a11y.speak( SpeakerFeedbackData.messages.markedHelpful, 'polite' );
+					$label.text( SpeakerFeedbackData.messages.unmarkHelpful );
+				}
 			} );
 	}
 
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -73,27 +73,24 @@
 	}
 
 	function onHelpfulClick( event ) {
-		event.preventDefault();
-		var button = $( event.target ).closest( 'button' ).get( 0 );
-		var isHelpful = 'is-helpful' === button.className;
+		var $container = $( event.target ).closest( 'footer' );
+		var input = $container.find( 'input[type="checkbox"]' ).get( 0 );
+		var isHelpful = !! input.checked;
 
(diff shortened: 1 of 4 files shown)


enriquesanchez:
>This is a bit different than what you designed in #347, so I'd like your 👍 / 👎 Also open to different copy for labels & screen reader content. 🙂

Hi @ryelle! 👋

I think that because both states look very similar, it wouldn't be as easy to identify which one has been marked and which one hasn't.

In the mockups explored in #347, the difference in state is established by the button's style:

Screen Shot 2020-04-13 at 17 46 44
Screen Shot 2020-04-13 at 17 46 40

What I think the toggle button accomplishes is both show state (marked or unmarked) and because it is a toggle, I don't think implies that it's possible to mark comments as unhelpful.

If we do want to be more explicit with labels, then I think we'd need to have a different style for each state.

Let me know and I can explore other ideas.

ryelle:
@enriquesanchez I'm not actually sure which state each of those are in. Is the darker blue button meant to be "this is already marked helpful" or "click this to mark it helpful"? 

We also can't easily use color for this, because every WordCamp site customizes the theme, so each button style is different. I took a couple screenshots across other themes, [you can see them here.](https://cloudup.com/c22cZShSide) If we need to, we can override these button styles, but that'll make it harder for the few organizers who make highly customized sites (like WCEU, or WCUS).

enriquesanchez:
>I'm not actually sure which state each of those are in.

@ryelle Oh totally, I can refine the toggle if that is the case. I still think the semantics of radio buttons/toggles make sense for this case. It may be that I need to work more on refining the design?

That being said, I'm happy to try another idea if you think that's what we should do.

ryelle:
> I still think the semantics of radio buttons/toggles make sense for this case.

Yeah, this morning I thought about how this is basically a checkbox 🙂  If you could do another iteration on the design, that'd be great.

enriquesanchez:
Hi @ryelle!

Here's an iteration on this component:

Screen Shot 2020-04-15 at 12 13 07

Each column is using different theme colors for comparison, and there are two versions of each. Both using the checkbox, but the second option (bottom) adds a stroke to the checked version to create a clearer visual indication and make it easier to see which ones have been marked as helpful. 

Contrast and clarity feel good to me but I'd love to hear what you think.

coreymckrill (approved):
👍 Worked in local testing. I realize there are still some visual pending changes, but the functionality and code look good to go.

Just to note: when I tried Voiceover, I was expecting that it would tell me the result of pushing the button, but it doesn't (Firefox). However, after thinking about it, I think this is probably fine because it is still explaining the button when you first focus it.

ryelle:
Hm, VoiceOver doesn't usually play well with anything other than Safari, but it should have said something after the button is pushed. I'll check again when I make the button changes.

ryelle:
@enriquesanchez  Latest update — since we can't directly style the checkbox (we can only inherit from the theme), I'm using a browser-default checkbox. It'll also be up to the camps to style the background & border color if they want. What do you think?

WCUS
wcus

Twenty Twenty
twenty

WCEU
wceu

enriquesanchez:
@ryelle Would it be possible to center the vertical aligment of both the checkbox and the 'Yes' label? Other than that, these look good to me 👍

Issue #432, opened 2020-04-14
Speaker Feedback: "Comments" sidebar menu incorrectly gets a "pending comment count" badge when unapproving a feedback using a row action


coreymckrill:
In the Feedback list table, when unapproving a feedback that has already been approved, the view counts at the top of the table are correctly updated, but then the "Comments" sidebar menu item also gets a new count, shown as a red badge, because the list table script is incorrectly assuming that the feedback is a normal comment. Refreshing the screen removes this erroneous badge.

The culprit is the JS list table functionality found in **wp-admin/js/edit-comments.js**, which augments **wp-includes/js/wp-lists.js**. This script is necessary for the list table row actions to work asynchronously via Ajax, but unfortunately does not accommodate non-standard comment implementations very well.

PR #433, opened 2020-04-14
Speaker Feedback: Add "inappropriate" comment status for feedback that is not spam but should not be sent to the speaker

### LINKED ISSUE #414: Speaker Feedback: Add a comment status for feedback that has been reviewed, but won't be sent to the speaker
Originally the idea was that if a feedback moderator reviewed a comment that was inappropriate to send on to the session speaker (i.e. not spam, but still mean-spirited or off-topic), they would just trash it. However, there are multiple reasons a comment could be in the trash, so it's too ambiguous if we want to be able to track what percentage of all feedback is inappropriate and/or be able to see if specific people are consistently submitting inappropriate feedback.

Initial proposal: Add a custom comment status `inappropriate` and allow moderators to apply it to feedback comments from within the Feedback list table.

coreymckrill:
This shoehorns in a custom comment status, despite WP Core's best efforts to prevent it. Feedback comments can be marked as `inappropriate` via bulk editing in the list table. A new list table view shows just feedback marked with this new status, along with a total count.

At this point it does not include row actions for asynchronously marking/unmarking feedbacks as inappropriate, because I haven't been able to find a way to make it work with the existing **edit-comments.js** implementation (somewhat related: #432). @ryelle maybe you'd be able to take a crack at this? Could be in a subsequent PR.

It also adds a list table view for feedbacks that the speaker has marked as "helpful".

Fixes #414 

### Screenshots

Marking a feedback as "inappropriate" via bulk edit

bulk-edit-inappropriate

Viewing all feedbacks that have been marked as "inappropriate"

view-inappropriate

1. Create some feedbacks.
1. On the Feedbacks list table, check the box for one or more feedbacks and select "Mark as Inappropriate" from the bulk edit select. Click Apply.
1. The screen will refresh. The selected feedback(s) should now have an orange left border (instead of the red used for unapproved comments). The view links and the comment bubbles should have updated counts.
1. Click on the "Inappropriate" view link. You should see a list table containing only "inappropriate" feedbacks.
1. Mark a few feedbacks as "helpful".
1. Back on the list table screen, click the "Helpful" view link. You should see a list table containing only "helpful" feedbacks.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
@@ -378,15 +378,17 @@ function count_helpful_feedback( $post_id = 0 ) {
 		$post__in[] = absint( $post_id );
 	}
 
-	$meta_query = array(
-		array(
-			'key'   => 'helpful',
-			'value' => 1,
-			'type'  => 'NUMERIC',
+	$args = array(
+		'meta_query' => array(
+			array(
+				'key'   => 'helpful',
+				'value' => 1,
+				'type'  => 'NUMERIC',
+			),
 		),
 	);
 
-	$feedbacks = get_feedback( $post__in, array( 'hold', 'approve' ), $meta_query );
+	$feedbacks = get_feedback( $post__in, array( 'hold', 'approve' ), $args );
 
 	$helpful_count = 0;
 
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
@@ -3,6 +3,7 @@
 namespace WordCamp\SpeakerFeedback\CommentMeta;
 
 use WP_Error;
+use function WordCamp\SpeakerFeedback\Comment\get_feedback;
 use const WordCamp\SpeakerFeedback\Comment\COMMENT_TYPE;
 
 defined( 'WPINC' ) || die();
@@ -362,4 +363,38 @@ function get_feedback_questions( $version = META_VERSION ) {
 	}
 
 	return $questions;
+}
+
+/**
+ * Count feedback comments marked as helpful for a specific post or overall.
+ *
+ * @param int $post_id
+ *
+ * @return int
+ */
+function count_helpful_feedback( $post_id = 0 ) {
+	$post__in = array();
+	if ( $post_id ) {
+		$post__in[] = absint( $post_id );
+	}
+
+	$meta_query = array(
+		array(
+			'key'   => 'helpful',
+			'value' => 1,
+			'type'  => 'NUMERIC',
+		),
+	);
+
+	$feedbacks = get_feedback( $post__in, array( 'hold', 'approve' ), $meta_query );
+
+	$helpful_count = 0;
+
+	foreach ( $feedbacks as $feedback ) {
+		if ( $feedback->helpful ) {
+			$helpful_count ++;
+		}
+	}
+
+	return $helpful_count;
 }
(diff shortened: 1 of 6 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
```diff
@@ -363,3 +364,37 @@ function get_feedback_questions( $version = META_VERSION ) {
...
+
+	$meta_query = array(
+		array(
+			'key'   => 'helpful',
+			'value' => 1,
+			'type'  => 'NUMERIC',
+		),
+	);
+
+	$feedbacks = get_feedback( $post__in, array( 'hold', 'approve' ), $meta_query );
```
ryelle: This function changed in the feedback UI PR to [accept all WP_Query args,](https://github.com/WordPress/wordcamp.org/blob/f977142e506a58b8a2561751b83d1793ac33ec03/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php#L183) so this last param should be `array( 'meta_query' => $meta_query )`

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/query.php
```diff
@@ -38,3 +39,24 @@ function( $var ) {
...
+ */
+function comments_clauses( $clauses, $query ) {
+	$status = $query->query_vars['status'];
+	$status = (array) $status; // $query is passed by reference, don't want to alter it.
+
+	if ( in_array( 'all', $status, true ) || in_array( '', $status, true ) ) {
+		$search  = "( comment_approved = '0' OR comment_approved = '1' )";
+		$replace = "( comment_approved = '0' OR comment_approved = '1' OR comment_approved = 'inappropriate' )";
+		$clauses = str_replace( $search, $replace, $clauses );
+	}
```
ryelle: I don't think we want to show inappropriate in the "all" status— IMO it should behave like spam, and disappear, only appearing in the Inappropriate tab.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
```diff
@@ -74,6 +75,11 @@ function count_feedback( $post_id = 0 ) {
 				$feedback_count['total_comments'] += $row['total'];
 				$feedback_count['all']            += $row['total'];
 				break;
+			case 'inappropriate':
+				$feedback_count['inappropriate']   = $row['total'];
+				$feedback_count['total_comments'] += $row['total'];
+				$feedback_count['all']            += $row['total'];
```
ryelle: If we remove inappropriates from the "all" tab, you can also remove this count here.

ryelle (commented):
Good idea on the "helpful" tab. Overall this is working well. I have a comment about removing inappropriate from the all tab, since I think most people won't want to see the inappropriate content.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
coreymckrill: 👍 ddb753ce

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/query.php
coreymckrill: Yeah, I think you're right.

53853129

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment.php
coreymckrill: Updated in 53853129

coreymckrill: Addressed in 5385312

ryelle (approved):
looks good now 👍

PR #436, opened 2020-04-15
WC Post Type: Remove protected status for `_wcpt_` meta keys

### LINKED ISSUE #435: WC Post Types: Sessions can't be updated due to permissions error
When a non-admin tries to edit a session, they can't save it:

> Publishing failed. Sorry, you are not allowed to edit the _wcpt_session_time custom field.

The full API response:

```
{
    "code":"rest_cannot_update",
    "message":"Sorry, you are not allowed to edit the _wcpt_session_time custom field.",
    "data": {
        "key":"_wcpt_session_time",
        "status":403
    }
}
```

ryelle:
By default, meta that starts with `_` is considered "protected", and can't be edited by anyone but superadmins. Our wcpt meta should be editable by anyone who can create content, so it needs to be moved out of the protected status.

Fixes #435 

I've opted to allow all `_wcpt_` values vs creating an allowed-list, since this could come up again in future CPT improvements. Switching the meta to unprotected does not allow them to be viewed in the API by default- still only those we set `show_in_rest` for are exposed. The only impact this has is on the classic editor's Custom Fields section, and that'll be phased out as each CPT is gutenberg-ified. If there are edge cases I've missed, I'm open to making an allowed-list, though.

⚠️ There's a bug that seems to happen when there are duplicate meta values. Happened to me on the 2014.seattle test data, but I spot checked a few production sites and didn't see it (including 2014.seattle). I'm assuming it's a local import bug, but something to watch for when testing.

```
Updating failed. Could not update the meta value of _wcpt_session_type in database.
```

1. Create a new session, fill out some or all of the meta
2. It should publish
3. Try editing an existing post, you should be able to update the meta & save.
4. Repeat as a different user level: `contributor`, `author`, `editor`, `admin`

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -14,6 +14,7 @@ require_once 'favorite-schedule-shortcode.php';
 
 add_action( 'init', __NAMESPACE__ . '\register_sponsor_post_meta' );
 add_action( 'init', __NAMESPACE__ . '\register_session_post_meta' );
+add_action( 'is_protected_meta', __NAMESPACE__ . '\unprotect_wcpt_meta', 10, 2 );
 
 /**
  * Registers post meta to the Sponsor post type.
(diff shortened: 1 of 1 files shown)


coreymckrill (approved):
👍 Worked in local testing.

I looked through the code to try and review all of the `_wcpt_` values we're using right now, and I didn't find any that I thought needed to stay "protected".

I ran into the same bug as you with the local test data. Removing the duplicates fixed it. There is at least one meta key that is intended to have multiple different values for the same post, `_wcpt_speaker_id`, but it does not seem to be impacted by the bug.

ryelle:
> I ran into the same bug as you with the local test data. Removing the duplicates fixed it. There is at least one meta key that is intended to have multiple different values for the same post, _wcpt_speaker_id, but it does not seem to be impacted by the bug.

Yeah, it's just a bug where `register_rest_field` specifies `single = true`.

PR #437, opened 2020-04-15
Add block for embedding YouTube live chat

iandunn:
This will help organizers who want to use YouTube as their streaming platform.

Testing is pretty straight-forward. `youtube-live-chat.js` has some notes at the top about examples of good URLs to test with.

Screen Shot 2020-04-15 at 17 46 08

PR #438, opened 2020-04-16
Tools: Use single linter config for all workspaces

ryelle:
As we spin up more workspaces, we're also adding more linter configs. This PR pulls out all the linter config files for eslint, stylelint, and prettier, and sets all workspaces to pull from those root files. It also updates the travis file to run stylelint & eslint on all the workspaces.

See #437 

This does not introduce any new rules. The prettier config is only moved up, you still do not need to use it if you don't want (I've been using it, but only on new files, not reformatting any existing files).

1. Travis should pass
2. Pull down the changes and try running lint:js & lint:css on any of the workspaces


iandunn (approved):
Looks good, thanks!

coreymckrill (approved):

PR #439, opened 2020-04-16
Speaker Feedback: Add custom cap to speakers for viewing their own sessions

### LINKED ISSUE #425: Speaker Feedback: Allow logged-in speakers with no role on a site to view their feedback

coreymckrill:
Instead of requiring the `read` capability for speakers when they are attempting to view the feedback on their sessions, require a custom capability instead. Then conditionally add that capability to the speaker user via a filter when the right conditions are met. This way the speaker user doesn't have to be a member of the site, they just need to be logged into the network.

Fixes #425 

1. Set up the following data:
  * A speaker post with a user ID of a user that is in the network, but not added to the site
  * A session post linking to that speaker
  * Add some feedback comments to the session and approve them
1. Log in to the network as the speaker user, and navigate to the feedback URL for their session
1. The feedback comments should be visible instead of the feedback form
1. Marking a feedback comment as "helpful" should succeed (this is what wasn't happening before)

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/capabilities.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/capabilities.php
@@ -2,6 +2,7 @@
 
 namespace WordCamp\SpeakerFeedback\Capabilities;
 
+use WP_User;
 use function WordCamp\SpeakerFeedback\Comment\get_feedback_comment;
 use function WordCamp\SpeakerFeedback\Post\get_session_speaker_user_ids;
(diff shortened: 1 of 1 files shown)


ryelle (approved):
🎉  working great!

PR #441, opened 2020-04-17
Speaker Feedback: Add a daily cron job to notify speakers of new feedback

### LINKED ISSUE #422: Speaker Feedback: Inform speakers when they have approved feedback comments

coreymckrill:
This creates a system that gathers feedback comments that have been approved but have not yet been included in a notification (using a `speaker_notified` comment meta value). From those comments it derives a list of the related session posts, and from that derives a list of speaker user accounts. Each speaker is then sent one email with a list of each of their sessions that has newly approved feedback on it. The cron job is set up to ensure this doesn't happen more than once per day. 6pm in the WordCamp site's local timezone was chosen for the send time so that organizers may have time to approve some feedback comments the same day that the event occurred and still have notifications go out.

This also includes a user meta value, `sft_speaker_notifications_opt_out` that will skip sending an email notification if it is set for a user. However, it does not include any UI for toggling the value. I've left that for a separate ticket/PR since it will need some design work.

Fixes #422 

1. Set up the data:
    * Two speaker posts with usernames that exist on the network.
    * Three or more session posts. One for each of the speakers, and one with both of the speakers. The session times should be a day or two before the current time to ensure they fall within the window when notification emails will be sent out.
    * Add some feedback comments to each session, but leave them unapproved.
1. Open up MailCatcher so you can see when email notifications get sent out.
1. You can trigger the cron job manually with wp-cli. On the docker environment it will be something like
    `wp --allow-root --url=https://2014.seattle.wordcamp.org cron event run sft_notify_speakers_approved_feedback`
1. When you trigger it before approving any feedback, no emails should be sent out.
1. Approve a feedback comment for each session, and then trigger the cron job again. This time there should be emails in MailCatcher. Notice that there should be one email per speaker, not per session. The email should list each session that has new feedback.
1. If you then immediately trigger the cron job again, no new emails should be sent out, because the approved feedback comments have all been marked as `speaker_notified`. If you unapprove comments and then reapprove them, it should not reset the notified status.
1. Try changing the timestamp of the sessions to get outside the window when email notifications will be sent (earliest session + 3 months). You'll need to use wp-cli to delete the `sft_speaker_notification_start_time` transient in order to test this.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
@@ -143,6 +143,16 @@ function get_feedback_meta_field_schema( $context = 'all', $key = '' ) {
 			'default'           => false,
 			'required'          => false,
 		),
+		'speaker_notified' => array(
+			'sft_context'       => array( 'update' ),
+			'description'       => 'The speaker has been notified about this feedback.',
+			'type'              => 'boolean',
+			'single'            => true,
+			'sanitize_callback' => 'wp_validate_boolean',
+			'show_in_rest'      => false,
+			'default'           => false,
+			'required'          => false,
+		),
 	);
 
 	if ( 'all' !== $context ) {
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
@@ -144,7 +144,7 @@ function get_feedback_meta_field_schema( $context = 'all', $key = '' ) {
 			'required'          => false,
 		),
 		'speaker_notified' => array(
-			'sft_context'       => array( 'update' ),
+			'sft_context'       => array( 'internal' ),
 			'description'       => 'The speaker has been notified about this feedback.',
 			'type'              => 'boolean',
 			'single'            => true,
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
@@ -76,7 +76,7 @@ function notify_speakers_approved_feedback() {
 				'feedback_ids' => $associated_feedbacks,
 				'title'        => get_the_title( $session_id ),
 				'count'        => count( $associated_feedbacks ),
-				'link'         => get_session_feedback_url( $session_id ),
+				'link'         => wp_login_url( get_session_feedback_url( $session_id ) ),
 			);
 		}
 
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
@@ -35,7 +35,11 @@ function notify_speakers_approved_feedback() {
 	$wordcamp_name = get_wordcamp_name( get_current_blog_id() );
 
 	$feedbacks       = get_unnotified_feedback();
-	$post_ids        = wp_list_pluck( $feedbacks, 'comment_post_ID', 'comment_ID' );
+	$post_ids        = array_combine(
+		// Using the 'index_key' parameter doesn't work here, maybe because the values are numeric?
+		wp_list_pluck( $feedbacks, 'comment_ID' ),
+		wp_list_pluck( $feedbacks, 'comment_post_ID' )
+	);
 	$unique_post_ids = array_unique( array_values( $post_ids ) );
 
 	$speaker_session_ids = array_reduce(
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
@@ -3,7 +3,7 @@
 namespace WordCamp\SpeakerFeedback\Cron;
 
 use function WordCamp\SpeakerFeedback\Comment\{ get_feedback, update_feedback };
-use function WordCamp\SpeakerFeedback\Post\get_session_speaker_user_ids;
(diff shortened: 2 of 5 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
```diff
@@ -0,0 +1,203 @@
...
+
+			return $carry;
+		},
+		array()
+	);
+
+	foreach ( $speaker_session_ids as $user_id => $session_ids ) {
+		$speaker = get_user_by( 'id', $user_id );
+		// Todo: Implement a UI for toggling this user meta value.
+		if ( true === $speaker->sft_speaker_notifications_opt_out ) {
```
ryelle: This will opt the user out for all WordCamps in the future too, right? since it's a per-user setting

I _think_ that's fine, we'll just want to make note of that whatever UI lets users control this

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
```diff
@@ -0,0 +1,203 @@
...
+			$feedback_info[] = array(
+				'feedback_ids' => $associated_feedbacks,
+				'title'        => get_the_title( $session_id ),
+				'count'        => count( $associated_feedbacks ),
+				'link'         => wp_login_url( get_session_feedback_url( $session_id ) ),
+			);
+		}
+
+		$speaker_locale = get_user_locale( $speaker );
+		switch_to_locale( $speaker_locale );
```
ryelle, line 84: 💯

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
ryelle: Tried testing this with `wp user meta set 4 sft_speaker_notifications_opt_out true`, and it set the value as a string. I'm not sure if setting meta directly will preserve type, but you might want to switch this to `wp_validate_boolean( … )`

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
```diff
@@ -143,6 +143,16 @@ function get_feedback_meta_field_schema( $context = 'all', $key = '' ) {
 			'default'           => false,
 			'required'          => false,
 		),
+		'speaker_notified' => array(
+			'sft_context'       => array( 'update' ),
```
ryelle: This meta value is editable by the speaker via the API, like `helpful` — did you intend that? (I'm not sure if it matters, but it seemed unexpected)

ryelle (approved):
Works great! Have a few questions, but nothing blocking merge.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
coreymckrill: I was aware of that when I added it, but it seems unlikely that it would ever be set through the API in a way that we don't intend. I guess we could add another value for `sft_context`. Maybe `internal` or something that indicates it's only used programmatically?

ryelle: `internal` sounds good 👍

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
coreymckrill: > This will opt the user out for all WordCamps in the future too, right?

Yeah, I believe this will be a value stored in the usermeta table that is shared across the entire WordPress/WordCamp network. So if the speaker opts out of notifications on one site, they will be opted out on other future WordCamp sites as well. But they'd also be able to opt back in from any site.

> you might want to switch this to wp_validate_boolean( … )

Good idea. ecbd74fd

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/comment-meta.php
coreymckrill: I guess another option would be to just not include it in the field schema, since, as mentioned, it's only for internal usage...

coreymckrill: Decided to go with `internal`: 6c6207b5

PR #443, opened 2020-04-17
Speaker Feedback: Prevent feedback actions from triggering comment UI

### LINKED ISSUE #432: Speaker Feedback: "Comments" sidebar menu incorrectly gets a "pending comment count" badge when unapproving a feedback using a row action

ryelle:
When moderating feedback and comments, the list table uses an ajax request to perform the status change. When the request succeeds, a callback is called that triggers various UI changes. We want some of those, but not all — moderating feedback should not change the Comments bubble, but it should update the views above the table.

This is handled by a callback called `dimAfter`. We can override this, and call only the changes we want. Rather than attempting to re-calculate the counts, I'm just reloading the fragment of the page that contains the view links & counts.

Fixes #432 

1. Moderate some feedback, it should still change status
2. The bubble next to Comments in the menu should not change
3. The bubble in the admin bar should not change
4. Moderate some comments, make sure they still work as expected

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/admin.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/admin.js
@@ -1,5 +1,5 @@
 /**
- * Remove the `dimAfter` callback for feedback replies. This function is called after a comment is unapproved or
+ * Update the `dimAfter` callback for feedback replies. This function is called after a comment is unapproved or
  * approved. In the comments table, it triggers the HTML updates in the admin menu & admin bar (among others).
  * We can override this by passing in our own function, which only handles the top view + count links.
  *
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/admin.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/admin.js
@@ -15,11 +15,10 @@ jQuery( document ).ready( function( $ ) {
 
 	// We can assume window.theList exists because we're after `edit-comments.js`
 	// And if we know that wpList is an object, we can assume the structure of it is stable.
-	if ( window.theList[0] && 'object' === typeof window.theList[0].wpList ) {
-		window.theList[0].wpList.settings.dimAfter = function() {
+	if ( window.theList[ 0 ] && 'object' === typeof window.theList[ 0 ].wpList ) {
+		window.theList[ 0 ].wpList.settings.dimAfter = function() {
 			// Reload the page, but only the top view + count links.
 			$container.load( location.href + ' .subsubsub' );
 		};
 	}
 } );
-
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/admin.js
```diff
@@ -0,0 +1,25 @@
...
+
+jQuery( document ).ready( function( $ ) {
+	// The `.load()` function can't load into itself, so we need to wrap the views with a container.
+	var $container = $( '<div>' ).html( $( '.subsubsub' ).get( 0 ).outerHTML );
+	$( '.subsubsub' ).replaceWith( $container );
+
+	// We can assume window.theList exists because we're after `edit-comments.js`
+	// And if we know that wpList is an object, we can assume the structure of it is stable.
+	if ( window.theList[0] && 'object' === typeof window.theList[0].wpList ) {
+		window.theList[0].wpList.settings.dimAfter = function() {
```
coreymckrill: Dang, I tried this, but I wasn't specifying the settings property correctly.

public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/admin.js
```diff
@@ -0,0 +1,25 @@
...
+			// Reload the page, but only the top view + count links.
+			$container.load( location.href + ' .subsubsub' );
```
coreymckrill, line 21: TIL that `$.fn.load` can do selective page fragments. 😮

coreymckrill: Could this also reload a fragment for the comment bubbles in the "In Response To" column?

comment-bubble

coreymckrill (approved):

public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/admin.js
ryelle: hm, this doesn't seem easy to do without doing a request for every part that needs replacing. I tried writing JS for this, but it felt too fragile… So I think a better solution is to just remove the bubble itself, which is what 0496a0fa358821595f39ba1f161334b7e45909fb does. (I also linked to the session's feedback page, rather than the session page).
 
Screen Shot 2020-04-22 at 3 26 43 PM

coreymckrill: 🤔 I guess the bubble's not really necessary, it just makes it match the normal Comments list table a bit better.

ryelle: yeah… but i'm not sure that it's important enough functionality-wise to justify spending more time on it. We can always revisit later.

ryelle:
I've rebased (to get the `get_session_feedback_url` function), and will merge this today.

PR #444, opened 2020-04-20
WCPT: Update build process, use React from `@wordpress/element`

### LINKED ISSUE #317: Application Tracker: JS should use `@wordpress/*` packages to prevent conflicts
The application tracker does not use core-provided packages (guessing they didn't exist then 🙂), but now that we're adding more JS to various pages, we're going to run into conflicts.

For example, originally reported as [an issue in slack,](https://wordpress.slack.com/archives/C037W5S7X/p1579001986080400?thread_ts=1579001986080400&cid=C037W5S7X) the application tracker adds `babel-polyfill` to its build, while the wordcamp blocks add `wp-polyfill` as a dependency. Trying to load both creates a JS error. This specific issue was fixed by disabling the relevant blocks on central https://github.com/WordPress/wordcamp.org/commit/8a0138516642414e99357c8c2a3c5b43195cff2a — but this is only a temporary fix.

The better solution would be to update the application tracker to use `@wordpress/element` instead of React (and so on), and `wp-scripts` to build so that core can manage dependencies.

ryelle:
Convert build process to use `wp-scripts`, and pull React dependencies from `@wordpress/element`. This will allow core to manage the vendor dependencies, enabling better caching, fewer conflicts, & a smaller overall filesize for `applications.min.js` (from 246K down to 9.3K). There should be no visible changes.

This also removes the one special-case project from our repo, everything uses the yarn workspaces now.

Fixes #317 

Note: This changeset is messy because each file had to be renamed from `*.jsx`, since wp-scripts doesn't understand how to process .jsx files. The basic overview of changes are:

- Create yarn workspace `wcpt`
- Rename files: `*.jsx -> *.js`
- Update PropTypes definitions to use the 3rd-party package ([removed from React core v15.5](https://reactjs.org/docs/typechecking-with-proptypes.html))
- Switch from `React.createClass` (also deprecated in v15.5) to native classes, extend Component from `@wordpress/element`
- Update [`createRef` usage,](https://reactjs.org/docs/refs-and-the-dom.html) previous version deprecated in v16.3 (or earlier?)
- Fix linter issues in JS & CSS
- Remove class properties ([this option](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties) is not supported in wp-scripts' process)
- In a few cases, moved from class components to functional components

1. Run `yarn workspace wcpt build`
2. Check that the application trackers for WordCamps & Meetups both still work
3. Tables are sortable & searchable, and show all expected content


coreymckrill (approved):
👍 Tested on sandbox

Issue #448, opened 2020-04-24
Migrate skip feature flag data from each site's options table to the blogmeta table


coreymckrill:
Currently we add skip feature flags to an array that lives at the `wordcamp_skip_features` key in a site's options table. Then we use `wcorg_skip_feature( $flag )` as a conditional in the context of the current site that the function is executing on to determine whether or not to skip a particular chunk of code. This works fine for the purpose actually skipping a feature on a specific site, but it makes it really hard to see the forest for the trees if you want to know which sites in the network have a particular flag. You have to run a separate query on each site to check its options table for the value.

I propose that we move these arrays of flags to the blogmeta table. This location is just as accessible from within a site as the options table (using the confusingly named `get_site_meta()` function, which gets values from the blogmeta table instead of the sitemeta table). Further, I think we should consider storing each flag as a separate meta value instead of storing all of them as one array. They would still end up being in an array when you retrieved them with `get_site_meta()` as long as the `$single` parameter is set to `false`.

The improvement from these two changes would be that we could efficiently query for a particular flag and get a list of all the sites that have it--or don't have it. Admittedly that's not something we have to do very often, but it would be one less place where we have to iterate through a 1000+ item array of blog IDs and `switch_to_blog()` / `restore_current_blog()` for each one.

_There are lots of other things we could move to the blogmeta table as well, but I'm currently thinking about this in the context of #370._

ryelle:
Related (but broader scope): https://github.com/WordPress/wordcamp.org/issues/287

Overall agree, I think this is a good idea 👍

ryelle:
There was also [this ticket (closed)](https://meta.trac.wordpress.org/ticket/1403) that mentioned updating the places we check against min site IDs, if feature flags are being updated maybe updating these checks could also get rolled in? It could be a good devex thing to clean up.

From ticket:

> I think this is a complete list of the old filters:
>
>```
> wcpt_require_complete_meta_min_site_id
> wcpt_speaker_post_avatar_min_site_id
> wcpt_session_post_speaker_info_min_site_id
> wcpt_speaker_post_session_info_min_site_id
> wcb_back_compat_max_site_id
> wcb_back_compat_site_id_exceptions
>```

Issue #287, opened 2019-11-22
Network: Use site meta for flagging sites as beta or testing


ryelle:
I saw the improvements to [the network sites screen in 5.3,](https://make.wordpress.org/core/2019/10/15/enhancements-to-the-network-sites-screen-in-wordpress-5-3/) which lead me to [this addition of site meta in 5.1…](https://make.wordpress.org/core/2019/01/28/multisite-support-for-site-metadata-in-5-1/) which all got me thinking about how we flag testing & beta sites. Right now we use functions with lists of sites, like [`get_excluded_site_ids`](https://github.com/WordPress/wordcamp.org/blob/8cc71f2a5c3037b55e0daebf57454645fb8fecaa/public_html/wp-content/plugins/wordcamp-reports/index.php#L355),  [`$blocks_test_sites`](https://github.com/WordPress/wordcamp.org/blob/dbacdfb5f452d74034fb7875aa9e507d16fbc3dd/public_html/wp-content/mu-plugins/blocks/blocks.php#L36) etc. This requires keeping multiple lists of site IDs, and knowing that you need to edit these if you create a new site.

What if instead, we add site meta to each test site, and use that to get the excluded sites? We can use those 5.3 filters add a "test" label to each test site, so we'll be sure which sites are excluded or not. This would probably still be something we set via CLI, but _could_ be a UI setting if needed. Thinking a boolean `is_test` or something.

For beta sites, we could do the same, but also add a UI for organizers to opt in to beta features themselves (or maybe not - I can see arguments for & against this).

coreymckrill:
I think this idea is worth exploring. There are a lot of possibilities for us with site meta. For example, we could also sync values from the WordCamp posts' postmeta on WC Central (e.g. `Start Date`) to the site meta for the site, and then we could eliminate a bunch of lines of code where we have to use `switch_to_blog` to go to Central and retrieve certain data.

coreymckrill:
> also add a UI for organizers to opt in to beta features themselves

I think I would prefer to skip the UI for now. When someone wants to opt in to a beta, we can just add the appropriate blogmeta value via wp-cli.

If we have a feature that we want to be broadly available but we're still calling it a beta, that's when I think we might build a UI for opting in. But then again, assuming it's a new plugin, we could just make the plugin available to organizers for activation and add `(Beta)` to the plugin title.

ryelle:
[...]

PR #449, opened 2020-04-25
Helpers: Add functions to check test sites, get lists of blog IDs

coreymckrill:
After adding a `wordcamp_test_site` key to the blogmeta for our test sites (which I've already done in production), this allows us to check if a site is a test site with a simple `is_wordcamp_test_site()` call.

Additionally, this adds a helper function that queries the blogmeta table directly and returns an array of blog IDs based on the specified meta key, and optionally an associated meta value. This is useful for situations where we want to get a list of all sites with a specific meta entry, like all test sites, or all sites that have opted into a specific beta. It could also be used to address #448.

This PR also updates some places in the codebase to use these new functions.

Refs #287 

1. The most direct way to test the new functions is probably to run this PR on your sandbox and use the wp-cli shell.
1. I've already set the three existing test sites on the network to have the `wordcamp_test_site` blogmeta, so once in shell, try this:
`get_wordcamp_blog_ids_from_meta( 'wordcamp_test_site', 1 )`
1. You should get an array back with three integers. Then try inputting one of the integers into this function:
`is_wordcamp_test_site( $my_integer )`
1. You should get back a boolean `true`. Using a different arbitrary integer should give you a `false`.
1. Check that the **Search engine visibility** input in Settings > Reading for one of the test sites is checked, and cannot be saved to an unchecked state.

### DIFF
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -29,6 +29,49 @@ function get_wordcamp_environment() {
 }
 
 /**
+ * Check if a WordCamp site is for testing.
+ *
+ * Currently the `wordcamp_test_site` blogmeta key needs to be set manually via wp-cli.
+ *
+ * @param int|null $blog_id Optional. The blog ID to check. Defaults to current blog ID.
+ *
+ * @return bool
+ */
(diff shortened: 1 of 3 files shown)


ryelle (approved):
Working well in testing. Flagging sites is much easier this way 👍

PR #450, opened 2020-04-27
Speaker Feedback: Add "Inappropriate" action to comments list table

### LINKED ISSUE #447: Speaker Feedback: Add row actions for marking/unmarking feedback as inappropriate
In #433 the `inappropriate` custom comment status was introduced. It can be applied to feedback comments via the Bulk Actions select in the Feedback list table, but there currently aren't any row actions for it, because I couldn't figure out how to get around the limitations of the normal comments row actions script. However, the approach in #443 might help, so we should look at this again.

ryelle:
Fixes #447 — Adds "Inappropriate" status action to the row actions on each comment. This should allow users to toggle the inappropriate status on a single feedback. In the Inappropriate view, the status can be set back to Pending (`hold`, technically, but for UI consistency Pending seemed a better string).

### Screenshots

Screen Shot 2020-04-27 at 2 51 18 PM

1. Try toggling the status of various feedbacks (both to/from inappropriate, but also trash/spam/approved)
2. This should also work with JS disabled (it uses the bulk edit action in that case).

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
@@ -503,8 +503,7 @@ function feedback_extra_actions( $actions, $comment ) {
 			'action' => 'unmark-inappropriate',
 			'_wpnonce' => wp_create_nonce( "inappropriate-comment_{ $comment_id }" ),
 			'bulk_edit_nonce' => wp_create_nonce( 'bulk_edit_' . COMMENT_TYPE ),
-		),
-		get_subpage_url( get_post_type( $comment->comment_post_ID ) )
+		)
 	);
 
 	$inappropriate_url = add_query_arg(
@@ -513,8 +512,7 @@ function feedback_extra_actions( $actions, $comment ) {
 			'action' => 'mark-inappropriate',
 			'_wpnonce' => wp_create_nonce( "inappropriate-comment_{ $comment_id }" ),
 			'bulk_edit_nonce' => wp_create_nonce( 'bulk_edit_' . COMMENT_TYPE ),
-		),
-		get_subpage_url( get_post_type( $comment->comment_post_ID ) )
+		)
 	);
 
 	if ( 'inappropriate' === $comment->comment_approved ) {
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
@@ -528,13 +528,21 @@ function feedback_extra_actions( $actions, $comment ) {
 			__( 'Move to Pending', 'wordcamporg' )
 		);
 	} else {
-		$actions['inappropriate'] = sprintf(
-			'<a href="%1$s" data-wp-lists="%2$s" class="vim-a vim-destructive aria-button-if-js" aria-label="%3$s">%4$s</a>',
-			esc_url( $inappropriate_url ),
-			"delete:the-comment-list:comment-{$comment_id}:e7e7d3:action=dim-comment&amp;new=inappropriate",
-			esc_attr__( 'Mark as Inappropriate', 'wordcamporg' ),
-			__( 'Inappropriate', 'wordcamporg' )
+		$action_inappropriate = array(
+			'inappropriate' => sprintf(
+				'<a href="%1$s" data-wp-lists="%2$s" class="vim-a vim-destructive aria-button-if-js" aria-label="%3$s">%4$s</a>',
+				esc_url( $inappropriate_url ),
+				"delete:the-comment-list:comment-{$comment_id}:e7e7d3:action=dim-comment&amp;new=inappropriate",
+				esc_attr__( 'Mark as Inappropriate', 'wordcamporg' ),
+				__( 'Inappropriate', 'wordcamporg' )
+			),
 		);
+
+		$key_indexes = array_keys( $actions );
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
```diff
@@ -478,6 +482,95 @@ function handle_bulk_edit_actions( $action, $nonce ) {
...
+		$actions['unapprove'] = sprintf(
+			'<a href="%1$s" data-wp-lists="%2$s" class="vim-u aria-button-if-js" aria-label="%3$s">%4$s</a>',
+			esc_url( $unapprove_url ),
+			// This is meta information for wpList, to trigger an action when clicked. Most of it can be ignored,
+			// the relevant part is the last: query params to pass to admin-ajax.php. See wp_ajax_mark_inappropriate
+			// for how these are used.
+			// See https://core.trac.wordpress.org/browser/trunk/src/js/_enqueues/lib/lists.js?rev=46800#L217.
+			"delete:the-comment-list:comment-{$comment_id}:e7e7d3:action=dim-comment&amp;new=uninappropriate",
+			esc_attr__( 'Move this comment to pending', 'wordcamporg' ),
+			__( 'Pending', 'wordcamporg' )
```
coreymckrill: While the `Inappropriate` label seems pretty self-explanatory and doesn't need a verb, I think it would be a bit more intuitive on this one if the label was `Move to Pending` or something, so that it's clear where it's going.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
```diff
@@ -478,6 +482,95 @@ function handle_bulk_edit_actions( $action, $nonce ) {
...
+		$actions['inappropriate'] = sprintf(
+			'<a href="%1$s" data-wp-lists="%2$s" class="vim-a vim-destructive aria-button-if-js" aria-label="%3$s">%4$s</a>',
+			esc_url( $inappropriate_url ),
+			"delete:the-comment-list:comment-{$comment_id}:e7e7d3:action=dim-comment&amp;new=inappropriate",
+			esc_attr__( 'Mark as Inappropriate', 'wordcamporg' ),
+			__( 'Inappropriate', 'wordcamporg' )
+		);
+	}
+
+	return $actions;
```
coreymckrill, line 548: I think the action of marking/unmarking a feedback as Inappropriate would be more discoverable for organizers if it was next to the Approve/Unapprove action, rather than at the far end of the row action list.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
```diff
@@ -478,6 +482,95 @@ function handle_bulk_edit_actions( $action, $nonce ) {
...
+	}
+
+	$unapprove_url = add_query_arg(
+		array(
+			'bulk_edit[]' => $comment_id,
+			'action' => 'unmark-inappropriate',
+			'_wpnonce' => wp_create_nonce( "inappropriate-comment_{ $comment_id }" ),
+			'bulk_edit_nonce' => wp_create_nonce( 'bulk_edit_' . COMMENT_TYPE ),
+		),
+		get_subpage_url( get_post_type( $comment->comment_post_ID ) )
```
coreymckrill: 💯 Good idea falling back to bulk edit. Here's an edge case that might be worth considering, although almost everyone will have JS enabled. If the user is on a view other than "All" (e.g. looking at just Approved or Inappropriate feedbacks), or they're on a subsequent page, or they've sorted the table by a non-default column, those choices are represented as additional parameters in the URL query string, and they will be lost when they click this link.

coreymckrill (approved):
👍 This is working great! No blockers, but a couple things that might be worth discussing.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
ryelle: Yeah, I guess that's unexpected. There's still an edge case because we're not re-calculating pages (so you can end up in a "no comments found" state), but that happens for bulk edit too… We can think about fixing that in another PR though.

ryelle: (I'll make the change to keep the user on the same page in this PR)

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
ryelle: Sure, that works— I was thinking about "Spam" and "Trash" as nouns, so I missed that they're all actually verbs, ha. Will change it.

ryelle: Changed to  "Move to Pending" in 17f0c0d9443308ea58662bfe1bcb331fda93f062

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
ryelle: Moved (and made red to match other "destructive" actions) in d756cd1515d6f3ecc923950acb9050cdac7b65ba

Screen Shot 2020-04-28 at 11 42 03 AM

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
ryelle: Done in bea41e69a2534d29b69c9bf974423f265ec1cd59

PR #451, opened 2020-04-27
Speaker Feedback: Consistently use sessions to determine start/end times of the event

coreymckrill:
Currently we are using different values in different places to determine when the WordCamp event as a whole starts and ends. Some places we are using the start time of the earliest session as a start time, some places we are using `Start Date (YYYY-MM-DD)` from the WordCamp post. For the ending time, we are mostly extrapolating from whatever we're using as the start time.

This creates new functions for getting a timestamp for the earliest session and a timestamp for the end of the latest session, which I think are the most relevant numbers to use in the context of this plugin, which is all about the sessions.

This also makes it easier for us to test the plugin on a production site. We can use a test site that doesn't have an associated WordCamp post, as long as the session times are set up correctly.

1. Set up some sessions so that the earliest published session is 2 days ago. Then visit the Leave Feedback page on the site. It should be available.
1. Try adjusting the session times so that the event falls outside of the window for leaving feedback (either the earliest session hasn't happened yet, or the latest session is more than two weeks in the past). Then you should no longer get the session selection form on the Leave Feedback page.
1. The above tests are regardless of what the `Start Date (YYYY-MM-DD)` value is in the WordCamp post for your test site.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
@@ -3,7 +3,10 @@
 namespace WordCamp\SpeakerFeedback\Cron;
 
 use function WordCamp\SpeakerFeedback\Comment\{ get_feedback, update_feedback };
-use function WordCamp\SpeakerFeedback\Post\{ get_session_speaker_user_ids, get_session_feedback_url };
+use function WordCamp\SpeakerFeedback\Post\{
+	get_earliest_session_timestamp, get_latest_session_ending_timestamp,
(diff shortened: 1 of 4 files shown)


ryelle (approved):
Works as described 👍 

I think `get_earliest_session_timestamp` & `get_latest_session_ending_timestamp` could be useful outside of SFT, maybe in the WC Post Types plugin… not required for merge, just thinking out loud.

coreymckrill:
> I think get_earliest_session_timestamp & get_latest_session_ending_timestamp could be useful outside of SFT, maybe in the WC Post Types plugin

Good idea. I'm going to keep them where they are for now, so I can keep moving on SFT, but definitely something to consider down the road.

PR #452, opened 2020-04-28
Speaker Feedback: Fix feedback table on small screens

ryelle:
Update styles in wp-admin for the smaller-screen view of the feedback list. This better matches the comments screen behavior by switching over to a mostly "single column" view for each item in the table, and fixes the expanded view.

### Screenshots

Before:

Screen Shot 2020-04-28 at 12 55 57 PM

After:

Screen Shot 2020-04-28 at 12 49 55 PM

1. Build the styles
2. Resize your screen to <782px
3. It should be readable, no overlapping text or awkwardly small columns

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/scss/style.scss
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/scss/style.scss
@@ -301,11 +301,12 @@
 }
 
 .wp-list-table.fixed {
-	.is-expanded .column-name {
+	.is-expanded .column-name,
+	.is-expanded .column-rating {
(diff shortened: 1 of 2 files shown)


coreymckrill (approved):
👍 Works great locally. The icon to expand the info in the row is a bit too subtle for me. But I guess that's a Core pattern, not specific to our UI.

ryelle:
> The icon to expand the info in the row is a bit too subtle for me

One thing we could do is move the rating into the main section too, so more info is exposed without needing to expand:

Screen Shot 2020-04-28 at 2 48 32 PM

In that, the only info shown when expanded is date & session. What do you think?

coreymckrill:
💯 I like that!

PR #453, opened 2020-04-28
Speaker Feedback: Append a feedback link to single session content

### LINKED ISSUE #440: Speaker Feedback: Should single session posts show a link/button to access the feedback form when applicable?
Right now an attendee can access the feedback form for a session by either going to the Feedback page where they can select a session in a form and redirect to it, or by manually adding `/feedback` to the end of the single session URL. It seems like it would be intuitive to have a "Leave feedback" button at the bottom of the content of a single session, that only appears when `post_accepts_feedback()` returns true.

coreymckrill:
This makes leaving feedback on a session more discoverable, as an alternative to navigating to the Leave Feedback page and selecting a session from the dropdown. The link is only appended during the time window when feedback is being accepted.

Fixes #440

### Screenshots

leave-feedback

1. Add a session and make sure the session time is a day or two in the past.
1. View the session on the front end. The link should be appended to the bottom of the content.
1. Make sure the link goes to the right place 😄


ryelle (approved):
Looks good 👍

PR #467, opened 2020-04-30
Speaker Feedback: Give better information on invalid fields

### LINKED ISSUE #460: Speaker Feedback: Give more explicit information about the bounds of the form fields
If someone fills in one of the feedback question fields and exceeds the character limit, the error message just says "This answer is too long", without providing a specific number. We should probably do one or both of these things:

* Update the error message to say something like "This answer is too long. It must be between 0 and 5,000 characters"
* Add help text near the field saying what the limits are.

We could also add a live updating character counter for each field, but that might be better as a v2 feature.

ryelle:
Return more descriptive errors when validating field values, and add a character counter to each textarea to indicate max length of content.

Fixes #460

### Screenshots

Screen Shot 2020-04-30 at 4 27 40 PM

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -72,6 +72,17 @@
 			} );
 	}
 
+	function characterCounter( event ) {
+		var len = event.target.value.length;
+		var maxLen = Number( event.target.dataset.maxlength );
+		if ( len > maxLen ) {
+			$( event.target ).addClass( 'has-error' );
+		} else {
+			$( event.target ).removeClass( 'has-error' );
+		}
+		$( event.target ).siblings( '.speaker-feedback__field-help' ).text( len + '/' + maxLen );
+	}
+
 	function onHelpfulClick( event ) {
 		var $container = $( event.target ).closest( 'footer' );
 		var input = $container.find( 'input[type="checkbox"]' ).get( 0 );
@@ -103,6 +114,7 @@
 	var feedbackForm = document.getElementById( 'sft-feedback' );
 	if ( feedbackForm ) {
 		feedbackForm.addEventListener( 'submit', onFormSubmit, true );
+		$( feedbackForm ).on( 'keyup', 'textarea[data-maxlength]', lodash.debounce( characterCounter, 250 ) );
 	}
 
 	var helpfulButtons = document.querySelectorAll( '.speaker-feedback__helpful input' );
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -73,7 +73,14 @@
 	}
 
 	function characterCounter( event ) {
-		var len = event.target.value.length;
+		// Some characters (like 🖖) are represented by a pair of code points, which JS counts as 2 separate
+		// characters. In PHP, we use `mb_strlen`, which correctly counts this as 1 character. For the same result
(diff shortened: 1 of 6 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
```diff
@@ -72,6 +72,17 @@
 			} );
 	}
 
+	function characterCounter( event ) {
+		var len = event.target.value.length;
```
coreymckrill: This doesn't appear to handle multibyte characters, like emoji, the way I would expect. "💩" counts as two characters instead of one. On the PHP side we're using `mb_strlen` in the `validate_meta_string` callback to ensure that multibytes are only counted as one character, but I'm not sure how best to do that on the JS side.

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
```diff
@@ -110,8 +115,14 @@
 		<textarea
 			id="sft-<?php echo esc_attr( $key ); ?>"
 			name="sft-<?php echo esc_attr( $key ); ?>"
+			<?php echo $maxlength_attr; // phpcs:ignore ?>
 			<?php echo $schema[ $key ]['required'] ? 'required' : ''; ?>
 		></textarea>
+		<?php if ( $maxlength_attr ) : ?>
+			<div class="speaker-feedback__field-help">
+				0/<?php echo absint( $schema[ $key ]['attributes']['maxlength'] ); ?>
+			</div>
```
coreymckrill: Should this be `aria-hidden`?

public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
ryelle: ah, of course it couldn't be this simple. i'll take another look in the morning, but if it still seems too complex I might pull this part out.

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
ryelle: hm… no, because we want to expose that content to screen readers, but it might need more context. i'll take a look at what other char counters do for a11y.

public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
coreymckrill: Looks like there might be a solution [here](https://dmitripavlutin.com/what-every-javascript-developer-should-know-about-unicode/#33-string-length), although it might only be compatible with ECMAScript 2015.

Something like

```js
var len = [...event.target.value].length;
```

coreymckrill: Or there's this:
https://mathiasbynens.be/notes/javascript-unicode#accounting-for-astral-symbols

(Sorry, I got intrigued)

ryelle: Done in 34596f713ef7b03d937519a1ab473a93c9178a75

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-feedback.php
ryelle: Added a describedby in 6cd11e9d1174bce4f706432801048c81e10b6fc9 — the counter stays available to all users (since some sighted users use screen readers), and now has more context with the extra description.

coreymckrill (approved):

PR #469, opened 2020-04-30
Speaker Feedback: Show both form and approved feedback to organizers on the front end

### LINKED ISSUE #457: Speaker Feedback: Allow organizers to access the feedback view on the front end as well as the form
Currently only speakers can access the front end view of approved feedback comments, and only for their own sessions. Organizers (i.e. site admins) just see the feedback submission form like everyone else, because they may want to submit feedback as well.

Suggestion from @dd32:

> My experience is that organisers need to see something before they send the link / direct speakers to a page, if they can’t see what it looks like, they can’t trust it. Also styling it is a challenge if they can’t view it.

So perhaps we should let organizers see _both_ the feedback form and the approved comments. Maybe just append the comments below the form? (I still think only speakers should be able to mark a feedback as "helpful" though.)

coreymckrill:
Organizers need to be able to see what a speaker would see when viewing their approved feedback, especially so the view can be styled to match the rest of the site. But organizers should also be able to submit feedback about a session through the form, just like everyone
else. So this makes it so everyone except the session speaker sees the form, and both the session speaker and organizers see the approved feedback list.

This also hides the "helpful" button on approved feedbacks from organizers. Only the speakers should be able to mark something as helpful.

Fixes #457

1. View the feedback URL for a session while logged out. Only the feedback form should be visible.
1. View the feedback URL for a session while logged in as a session speaker. Only the list of approved feedback should be visible.
1. View the feedback URL for a session while logged in as a site admin. Both the feedback form and the list of approved feedback should be visible. However, the "Was this feedback helpful?" section should be omitted from each approved feedback.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
@@ -97,22 +97,22 @@ class Walker_Feedback extends Walker_Comment {
 					<?php render_feedback_comment( $comment ); ?>
 				</div><!-- .comment-content -->
 
-				<?php if ( in_array( get_current_user_id(), $session_speakers, true ) ) : ?>
-					<footer class="speaker-feedback__helpful <?php echo ( $comment->helpful ) ? 'is-helpful' : ''; ?>">
-						<span id="sft-helpful-<?php echo absint( $comment_id ); ?>">
-							<?php esc_html_e( 'Was this feedback helpful?', 'wordcamporg' ); ?>
-						</span>
-						<label>
+				<footer class="speaker-feedback__helpful <?php echo ( $comment->helpful ) ? 'is-helpful' : ''; ?>">
+					<span id="sft-helpful-<?php echo absint( $comment_id ); ?>">
+						<?php esc_html_e( 'Was this feedback helpful?', 'wordcamporg' ); ?>
+					</span>
+					<label>
+						<?php if ( in_array( get_current_user_id(), $session_speakers, true ) ) : ?>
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
```diff
@@ -96,20 +97,22 @@ protected function render_feedback( $comment, $depth, $args ) {
...
-						<input
-							type="checkbox"
-							data-comment-id="<?php echo absint( $comment_id ); ?>"
-							aria-describedby="sft-helpful-<?php echo absint( $comment_id ); ?>"
-							<?php checked( $comment->helpful ); ?>
-						/>
-						<?php esc_html_e( 'Yes', 'wordcamporg' ); ?>
-					</button>
-				</footer>
+				<?php if ( in_array( get_current_user_id(), $session_speakers, true ) ) : ?>
```
ryelle: I think this should still be displayed, but it should be un-clickable, just in case anyone wants to style it (also so organizers know it should be there). I think if you remove the input when not a speaker, that should prevent the submission.

maybe with a note … somewhere… that only speakers can mark things as helpful?

ryelle (commented):
Working great, just the one comment about the helpful UI.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-walker-feedback.php
coreymckrill: Fair enough. Fixed here: c16c8302

ryelle (approved):

PR #470, opened 2020-05-01
Speaker Feedback: Don't encode post titles in plaintext emails

### LINKED ISSUE #468: Speaker Feedback: Character encoding issues in email notifications
HTML entities should be converted to plaintext characters in the email, which is text-only.

image-3-2

coreymckrill:
Using `get_the_title` to retrieve a session title runs several filters, including `wptexturize` and `convert_chars`, both of which convert plaintext characters into HTML entities. We don't want this in the email notifications, because they are sent in plaintext only, and thus won't render the entities correctly.

Fixes #468

1. Test as you would for #441, but make sure the session title has special characters in it. The title from the original issue works: `If at first you don’t succeed…` (the ellipses is one character). In MailCatcher, the notification should contain the original title, with no substituted HTML entities.


ryelle (approved):
💯

PR #475, opened 2020-05-04
Speaker Feedback: Fix toggle action when feedback is marked helpful

ryelle:
While testing #469, I noticed that the helpful toggle wasn't behaving correctly — it was sending the opposite status for helpfulness. I think it's due to the browser checking the checkbox & running our click callback at the same time, so that `input.checked` is already in the opposite state. Instead of relying on something the browser changes, this now checks against the `is-helpful` class, which we intentionally toggle on/off.

1. Log in as a speaker for a session
2. Try toggling a feedback's helpfulness, the helpful state should visibly toggle
3. Reload the page, the feedback(s) you just marked should stay in the same state

Note about testing: The JS assumes the request did the right thing, so it _appears_ to work on production, but if you reload the page, everything's opposite.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/js/script.js
@@ -92,6 +92,11 @@
 
 	function onHelpfulClick( event ) {
 		var $container = $( event.target ).closest( 'footer' );
+		if ( $container.hasClass( 'is-inflight' ) ) {
+			return;
+		}
+		$container.addClass( 'is-inflight' );
+
 		var input = $container.find( 'input[type="checkbox"]' ).get( 0 );
 		var isHelpful = $container.hasClass( 'is-helpful' );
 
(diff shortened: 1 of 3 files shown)


coreymckrill (approved):
👍 Worked in local testing. Good catch.

One thing I've noticed is that there is a significant pause between clicking the Yes button and having its visual state change, because of the background API request. I wonder if it would be worth showing some kind of "working" state on the button while it's waiting for the response. Or, at least ignoring subsequent clicks until the first click is resolved, so people don't undo their first click thinking it didn't work...

ryelle:
Good ideas— I added both, a check for a current in-flight request https://github.com/WordPress/wordcamp.org/pull/475/commits/d38ac545dfa0c0a2216497679be1735d334d2c58 & a really simple "wait" cursor onto the button https://github.com/WordPress/wordcamp.org/pull/475/commits/a557478d877606479b259cab733dfdfb3d304f6c. Further clicks won't do anything until the request completes & removes the class.

Issue #461, opened 2020-04-29
Speaker Feedback: The "organizer note" content on the Leave Feedback page is visible in meta data used by crawlers


coreymckrill:
The "Leave Feedback" page gets created automatically by the plugin when it is activated on a site. The content added to the page is intended only as a note to organizers:

> This page is a placeholder for the Speaker Feedback form. The content here will not be shown on the site.

However, since that content is still stored in the database, it's still used in other contexts, so it shows up in places like Slack when you post a link to the page, and presumably in search engine results.

One idea would be to change this behavior so that the content _does_ show on the page, and the form is just appended to the bottom. In that case, we would seed the page with an explanatory blurb about leaving feedback for speakers, and the organizers could edit it as they see fit.

ryelle:
Rather than changing the page content, we could override the `og:description` with the `jetpack_open_graph_tags` filter.

We need a blurb for social embeds, since this is what currently displays:
![Screen Shot 2020-05-01 at 1 45 21 PM](https://user-images.githubusercontent.com/541093/80827481-0a366d00-8bb2-11ea-9131-dab66008a26e.png)

Also suggested was injecting "a short blurb that linked to a handbook page somewhere with advice on giving good feedback" — which could be added to `views/form-select-sessions.php`, so that we can push out updates as needed (at least for a v1, this could be editable later). So we need that content & handbook link.

ryelle:
Should that "how to give feedback" blurb also be on the individual session pages? Not everyone will come to the form by way of the main `/feedback/` page.

coreymckrill:
> Should that "how to give feedback" blurb also be on the individual session pages? Not everyone will come to the form by way of the main /feedback/ page.

This is a good point, and it makes me think that we should actually put the blurb at the top of the feedback form on individual sessions _instead_ of on the Leave Feedback page. That way it's closer to the source. We should probably open a separate issue for this, though.

> Rather than changing the page content, we could override the og:description with the jetpack_open_graph_tags filter.

Generally I'm not a fan of having "invisible" content that is not editable showing up in social embeds. But perhaps that's the right approach for this v1, given our time constraints.

ryelle:
Added the `og:description` in #484 

If a camp doesn't have Jetpack sharing active, the embed should read the content, which right now is a bit sparse:

[...]

PR #484, opened 2020-05-05
Speaker Feedback: Set an `og:description` for the main feedback page

ryelle:
This prevents the placeholder content from showing up on social share embeds.

See #461 

The content used was pulled from the new handbook page, it says

> You can show your appreciation and contribute back to the community by leaving constructive feedback. This not only helps speakers know what worked in their presentation and what didn’t, but it helps organizers get a sense of how successful the event was as a whole.

happy to change this if there's feedback (:pundog:) ^

Requires active Jetpack— since Jetpack pulls the post content for the description, this isn't a problem on non-Jetpack sites.

1. Try embedding the `/feedback/` page on another post on the site, the embed should use the new content
2. Or just view source, the `og:description` should be the new content, no "placeholder"


coreymckrill (approved):
👍 Looks good locally. I did have to enable the Sharing module in Jetpack before it worked, but I think most WordCamp sites will have that on by default.

PR #486, opened 2020-05-05
 Speaker Feedback: Display unapproved feedbacks in admin menu

### LINKED ISSUE #478: Speaker Feedback: Jetpack's "Feedback" menu item for contact form submissions is a naming collision, creates confusion
If an organizer is trying to find the speaker feedback submissions screen, the first thing they will see in the Dashboard is a top-level item in the left menu called "Feedback" That is the natural place to go, but it's actually submissions from Jetpack's Contact Form module, not the speaker feedback tool.

The actual list table for speaker feedback is buried under the Sessions item in the left menu. I believe this is the right place for it, especially since we may eventually want to accept feedback submissions on other post types. Each supporting post type will have its own feedback screen. Also, I'm not excited about adding yet another top-level menu item, because clutter.

To mitigate this problem here are a couple of ideas:

* Use a filter to rename Jetpack's menu item to something more accurate, like "Contact Forms" or just "Forms"
* Add a red badge with the count of unapproved feedbacks to the Sessions menu item, so that organizers will know that something there needs their attention.

ryelle:
Add a count-bubble with the unapproved feedbacks count to the Sessions menu item. This moves to the Feedback item when the Sessions menu is opened (when on Sessions list, tracks, etc). When a feedback is approved/unapproved, the bubble count should be updated using the same JS action as the top row counts.

Fixes #478

### Screenshots

Screen Shot 2020-05-05 at 5 56 52 PM

Screen Shot 2020-05-05 at 5 56 34 PM

1. Have some unapproved feedbacks
2. From the dashboard, or posts, etc, see the count on Sessions
3. Click into Sessions (or any sub-item)
4. The count should move to Feedback
5. Go to Sessions -> Feedback
6. Approve some feedbacks
7. The bubble count should update to reflect the real count

_Note:_ I'm really not happy with this JS solution, as it's loading the page twice now behind the scenes… I don't think we're at risk of DoSing ourselves, since an org still needs to physically click a link to trigger it, but… 🤷‍♀️  I wish this was better?

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
@@ -20,6 +20,7 @@ foreach ( SUPPORTED_POST_TYPES as $supported_post_type ) {
 }
 
 add_action( 'admin_menu', __NAMESPACE__ . '\add_subpages' );
+add_action( 'current_screen', __NAMESPACE__ . '\add_feedback_bubble' );
 add_filter( 'admin_enqueue_scripts', __NAMESPACE__ . '\enqueue_assets' );
 add_filter( 'set-screen-option', __NAMESPACE__ . '\set_screen_options', 10, 3 );
 add_filter( 'wp_count_comments', __NAMESPACE__ . '\adjust_comment_counts', 10, 2 );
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
```diff
@@ -206,6 +207,54 @@ function() {
...
+		return;
+	}
+	if ( ! current_user_can( 'moderate_' . COMMENT_TYPE ) ) {
+		return;
+	}
+
+	$pending_count = get_feedback( array(), array( 'hold' ), array( 'count' => true ) );
+	if ( $pending_count <= 0 ) {
+		return;
+	}
```
coreymckrill: It seems like it would be more consistent with the rest of the plugin to use `count_feedback` here instead...

coreymckrill (approved):
> 🤷‍♀️ I wish this was better?

My reaction to everything about comments in WP Core.

One item to discuss, but otherwise looks good 👍

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/admin.php
ryelle: I don't think `count_feedback` returns unapproved comments, but maybe I read over it too quickly

ryelle: Oh, is `moderated` the count of unapproved/pending? 3 terms for the same thing? 🙃

coreymckrill: Yep, see previous comment about how much I love comments in Core 😄

PR #490, opened 2020-05-06
Speaker Feedback: Add UI for speakers to opt out of email notifications about approved feedback

### LINKED ISSUE #488: Speaker Feedback: Add UI for speakers to opt out of email notifications about approved feedback
In #441 I introduced a usermeta value to check whether a speaker has opted out of email notifications before sending the email about new approved feedbacks. However, that PR didn't include any UI additions where a speaker could actually set that value.

Storing the value in usermeta means that the opt out will apply for _all_ WordCamps. This way the speaker doesn't need to re-opt out if they speak again at subsequent events.

I'm unsure where this UI addition should go, but here are a couple of rough ideas that are maybe not mutually exclusive:

* A checkbox on the front end feedback view. The speaker would need to receive at least one email notification, and then when they follow the link, they could check the box to prevent further notifications. They could also then uncheck the box later if they wanted to turn the notifications back on.
* A checkbox somewhere on the Edit Speaker screen for the Speaker post type. This would most likely be toggled by an organizer, at the request of the speaker.

coreymckrill:
Building on the usermeta value introduced in #441, this adds a toggle to the frontend Session Feedback view where speakers and organizers can disable and re-enable email notifications about newly approved feedback submissions. The toggle updates the usermeta value, which carries over across all WordCamp sites in the network.

To accomplish this, I created a new endpoint on our custom REST API route for feedback. The standard `users` endpoint doesn't work for updating this usermeta value because it requires that the user be a member on the current site. Most speakers are not added as members on the WordCamp sites that they speak at. The custom endpoint allows us to circumvent that.

Note that this does **not** include a metabox or other UI on the Speaker CPT screen for toggling this value. Adding a metabox that would be functional in the block editor turned out to be overly-complex, given the necessary interaction with the existing `WordPress.org Username` metabox field. Building a custom sidebar plugin is beyond the scope of this PR, but it might be simpler, and doable as a separate PR, once the other metabox fields for the Speaker CPT have been converted to sidebar panels, as was [recently done](https://github.com/WordPress/wordcamp.org/pull/303) for the Sessions CPT. In the mean time, organizers can toggle this value for speakers upon request by going to the frontend feedback view, where they will see the same UI that speakers do, but with a separate toggle for each speaker in the session.

Fixes #488 

### Screenshots

Speaker UI, notifications enabled (default)

speaker-view-enabled

Speaker UI, notifications disabled

speaker-view-disabled

Organizer UI

admin-view

1. Rebuild the stylesheet to get the style updates.
1. With the same testing setup as for #441, open the feedback moderation screen and approve a new feedback submission.
1. Run the cron job and note that an email is sent in MailCatcher.
1. In a separate tab, navigate to the frontend feedback view for a session while logged in as a speaker. Click the toggle to disable notifications.
1. Approve another feedback submission and run the cron job. No new email should be sent.
1. Re-enable notifications for the speaker, and run the cron job again. Now the email should get sent.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
@@ -2,13 +2,14 @@
 
 namespace WordCamp\SpeakerFeedback;
 
-use WP_Error, WP_REST_Request, WP_REST_Response, WP_REST_Server;
+use WP_Error, WP_REST_Request, WP_REST_Response, WP_REST_Server, WP_User;
 use WP_REST_Comments_Controller;
-use const WordCamp\SpeakerFeedback\Comment\COMMENT_TYPE;
 use function WordCamp\SpeakerFeedback\Comment\{ add_feedback, is_feedback, update_feedback };
 use function WordCamp\SpeakerFeedback\CommentMeta\{ get_feedback_meta_field_schema, validate_feedback_meta };
 use function WordCamp\SpeakerFeedback\Post\post_accepts_feedback;
 use function WordCamp\SpeakerFeedback\Spam\spam_check;
+use const WordCamp\SpeakerFeedback\Comment\COMMENT_TYPE;
+use const WordCamp\SpeakerFeedback\Cron\SPEAKER_OPT_OUT_KEY;
 
 defined( 'WPINC' ) || die();
 
@@ -64,6 +65,36 @@ class REST_Feedback_Controller extends WP_REST_Comments_Controller {
 					'callback'            => array( $this, 'update_item' ),
 					'permission_callback' => array( $this, 'update_item_permissions_check' ),
 					'args'                => $this->get_endpoint_args_for_item_schema( WP_REST_Server::EDITABLE ),
+				),
+			)
+		);
+
+		// The Core users endpoint does not allow updating a user that is not a member of the site,
+		// so we have to create our own endpoint.
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/notifications/(?P<id>[\d]+)',
+			array(
+				'args' => array(
+					'id' => array(
+						'description' => __( 'Unique identifier for the user.', 'wordcamporg' ),
+						'type'        => 'integer',
+					),
+				),
+				array(
+					'methods'             => WP_REST_Server::EDITABLE,
+					'callback'            => array( $this, 'update_notifications' ),
+					'permission_callback' => array( $this, 'update_notifications_permissions_check' ),
+					'args'                => array(
+						'speaker_opt_out'           => array(
+							'description' => __( 'Speaker has opted out of notifications about feedback.', 'wordcamporg' ),
+							'type'        => 'boolean',
+							'required'    => false,
+							'arg_options' => array(
+								'sanitize_callback' => array( $this, 'wp_validate_boolean' ),
+							),
+						),
+					),
 				),
 			)
 		);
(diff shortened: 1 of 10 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/views/view-feedback.php
```diff
@@ -41,6 +44,51 @@
...
+		if ( $is_session_speaker && get_current_user_id() !== $session_speaker ) {
+			continue;
+		}
+		$user = new WP_User( $session_speaker );
+		$notifications_opt_out = get_user_meta( $user->ID, SPEAKER_OPT_OUT_KEY, true );
+		?>
+		<div class="speaker-feedback__notifications <?php echo ( $notifications_opt_out ) ? 'is-disabled' : ''; ?>">
+			<span>
+				<?php if ( ! $is_session_speaker ) : ?>
+					<span class="sft-notifications-speaker-name">
```
ryelle: I know it's long, but these should follow the class name patterns: `speaker-feedback__notifications-speaker-name`, `speaker-feedback__notifications-label-text`, etc. (`sft-` is only used on IDs/names)

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/view-feedback.php
```diff
@@ -41,6 +44,51 @@
...
+					<?php else : ?>
+						<?php esc_html_e( 'Feedback notifications are enabled.', 'wordcamporg' ); ?>
+					<?php endif; ?>
+				</span>
+			</span>
+			<label>
+				<input
+					type="checkbox"
+					data-user-id="<?php echo absint( $user->ID ); ?>"
+					<?php checked( $notifications_opt_out ); ?>
```
ryelle, line 97: This needs an `aria-describedby` pointing to the content in the visual label, otherwise it's not totally clear to a screen reader user what you're enabling.

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
```diff
@@ -67,6 +68,36 @@ public function register_routes() {
 				),
 			)
 		);
+
+		// The Core users endpoint does not allow updating a user that is not a member of the site,
+		// so we have to create our own endpoint.
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/notifications/(?P<id>[\d]+)',
```
ryelle: I think this endpoint needs a separate controller, `class-rest-notifications-controller.php`, so that it can be its own base: `/wordcamp-speaker-feedback/v1/notifications/…`

It would also be nice to see tests for this endpoint (that doesn't have to block merge though)

ryelle (commented):
It works as described, toggling off notifs skipped that user in the automated emails. It's a little unclear with a screen reader, but that should be a simple fix (noted below).

I feel like this needs some help text, maybe above the box. Thinking for speakers, something like:
> You will receive up to an email a day if there are new feedback submissions. Disabling this will turn off notifications for all feedback.

And organizers:
> Manage speaker notifications. Speakers will receive up to an email a day if there are new feedback submissions.

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/view-feedback.php
coreymckrill: Fixed with fe3c145

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/view-feedback.php
coreymckrill: Fixed with 4fe2df0

public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/class-rest-feedback-controller.php
coreymckrill: Sigh. You're right.

Moved to separate controller here: 40b5d8f

Tests added here: e836d56

coreymckrill:
Helpful feedback on this, thanks. I've added help text blurbs. Didn't go with exactly what you had in mind, the wording is here: b150ad7

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/view-feedback.php
```diff
@@ -41,6 +44,69 @@
...
+
+	<?php foreach ( $session_speakers as $session_speaker ) :
+		if ( $is_session_speaker && get_current_user_id() !== $session_speaker ) {
+			continue;
+		}
+		$user = new WP_User( $session_speaker );
+		$notifications_opt_out = get_user_meta( $user->ID, SPEAKER_OPT_OUT_KEY, true );
+		?>
+		<div class="speaker-feedback__notifications <?php echo ( $notifications_opt_out ) ? 'is-disabled' : ''; ?>">
+			<span id="sft-notifications-description">
```
ryelle: This isn't unique in the organizer's view if there's more than one speaker. appending the speaker ID should be good to make it unique.

ryelle (approved):
One minor note that should get fixed before merging, but otherwise it's working great 👍

public_html/wp-content/plugins/wordcamp-speaker-feedback/views/view-feedback.php
coreymckrill: Thanks, good catch. 0b4d5a90

PR #495, opened 2020-05-13
Speaker Feedback: Add context to the sessions list

### LINKED ISSUE #489: Speaker Feedback: Add more context to the session select on the Leave Feedback page
Some feedback:

> * It’d be useful if the name of the speaker would be next to the title of the talk, because it’s eaiser (at least for me) to remember the name of the speaker that the title of the talk.
> * I was expecting to see the talks ordered by the order of the program, but it was alphabetically.

Sorting by time without any context might make it look random. But if we also prepended the day/time of the session to each select option, it would then be clear how it was sorted.

So here's a first draft of what a complete select option might look like:

session-select

ryelle:
This adds session time & speaker names to the session list on `/feedback`. It also updates the sorting from alphabetical by title to by time, so sessions are in the order they happened.

Fixes #489

### Screenshots

The full list:

Screen Shot 2020-05-13 at 12 12 43 PM

You can search for a speaker by name now, due to how selectWoo filters:

Screen Shot 2020-05-13 at 12 12 50 PM

1. Make sure to rebuild the CSS
2. Visit the `/feedback` page
3. Select a session by searching for any part
4. It should direct you to the feedback form/page for that session.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-select-sessions.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-select-sessions.php
@@ -5,11 +5,16 @@ namespace WordCamp\SpeakerFeedback\View;
 $session_args = array(
 	'post_type'      => 'wcb_session',
 	'posts_per_page' => -1,
-	'orderby'        => 'title',
+	'orderby'        => 'meta_value_num',
 	'order'          => 'asc',
+	'meta_key'       => '_wcpt_session_time',
 	// get only sessions, no breaks.
-	'meta_key'       => '_wcpt_session_type',
-	'meta_value'     => 'session',
+	'meta_query'     => array(
+		array(
+			'key'   => '_wcpt_session_type',
+			'value' => 'session',
+		),
+	),
 );
 
 $sessions = new \WP_Query( $session_args );
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-select-sessions.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-select-sessions.php
@@ -30,11 +30,26 @@ if ( $sessions->have_posts() ) : ?>
 					$sessions->the_post();
 					$session_title = get_the_title();
 					$session_time  = absint( get_post_meta( get_the_ID(), '_wcpt_session_time', true ) );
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/views/form-select-sessions.php
```diff
@@ -22,10 +28,18 @@
 			<select name="sft_session" id="sft-session">
 				<?php while ( $sessions->have_posts() ) :
 					$sessions->the_post();
+					$session_title = get_the_title();
+					$session_time  = absint( get_post_meta( get_the_ID(), '_wcpt_session_time', true ) );
+					$option_text   = sprintf( '%1$s: %2$s', wp_date( 'M d, g:ia', $session_time ), $session_title );
+
+					$speakers = get_session_speaker_names( get_the_ID() );
+					if ( ! empty( $speakers ) ) {
+						$option_text .= ' — ' . implode( ', ', $speakers );
```
coreymckrill: There are several locales that use alternative characters for list item separators, rather than a comma. Here's an example of the standard way we're making this sort of thing translatable:
https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/addons/privacy.php#L152-L153

coreymckrill: Also, I'm not sure if the `' - '` part needs to be translatable. It might be best to bring the entire `$option_text` assignment down into the conditional and make the whole thing one translation:

```php
if ( ! empty( $speakers ) ) {
	__( '%1$s: %2$s - %3$s' )
} else  {
	__( '%1$s: %2$s' )
}
```

coreymckrill: Also, the date format in the `$option_text` should probably be localizable. Either that or grab the site's preferred format from the options table.

ryelle: Since we want a short combo of date & time, I think we should keep this unique — but I've added a translation wrapper to it.

ryelle: (and to everything else 🙂  )

coreymckrill (approved):

PR #502, opened 2020-05-15
Add beta version of Schedule block.

### LINKED ISSUE #309: Schedule block inspector controls
Make the sidebar work and dynamically affect the content preview. Depends on #104.

### LINKED ISSUE #310: Schedule block: Build front-end renderer
Explore re-using JS rather than building duplicate PHP view.

iandunn:
Supersedes #104
Fixes #309
Fixes #310 

There's still some minor polishing left to do, but I don't want to make that a bottleneck for getting the beta testing started. I'd like to focus this review on major architectural pieces and things that have back-compat implications, and then circle back once the beta is running to review the things that aren't foundational.

There's some sample data in the `test/` folder if you want to test locally. Or, you can build it on your w.org sandbox and check out the following pages:

* https://2013.sf.wordcamp.org/wp-admin/post.php?post=1404690&action=edit
* https://2014.seattle.wordcamp.org/wp-admin/post.php?post=915536&action=edit
* https://2016.kansascity.wordcamp.org/wp-admin/post.php?post=2349&action=edit
* https://2017.us.wordcamp.org/wp-admin/post.php?post=12186&action=edit
* https://2018.montreal.wordcamp.org/wp-admin/post.php?post=3361&action=edit
* https://2019.europe.wordcamp.org/wp-admin/post.php?post=14070&action=edit
* https://2017.testing.wordcamp.org/wp-admin/post.php?post=13245&action=edit

Some of those have a few issues, but those are mostly because of incorrect data in the session posts. I didn't want to modify production data, though, beyond just setting the duration.

There's also a few instances where the layout differs from the shortcode, because the shortcode is buggy. Or cases where the `single column` block style should be used rather than the grid style, because it fits the content better.


public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/controller.php
```diff
@@ -0,0 +1,323 @@
...
+		'editor_style'  => 'wordcamp-blocks',
+
+		/*
+		 * @todo This gets loaded on all front-end pages unnecessarily. See:
+		 *
+		 * https://github.com/WordPress/gutenberg/issues/15256
+		 * https://github.com/WordPress/gutenberg/issues/21838
+		 *
+		 * There isn't a simple workaround, though, since the block needs to be registered before the global $post
+		 * is available. We can probably just wait until that bug is fixed upstream. An alternative option might be:
```
ryelle: Does the render_callback method (the second link above) not work in this case? Register the scripts, then enqueue them in `render` (I haven't tried this, but it seems to work for other folks)

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/front-end.js
```diff
@@ -0,0 +1,108 @@
...
+/*
+ * todo
+ *
+ * this file is being loading in the editor, but should only load on the front end.
+ * probably a similar problem mentioned in init(), so maybe conditionally register it if on the front end?
+ * `wp_using_themes() && ! wcorg_is_rest_api_request()` ? - https://wordpress.stackexchange.com/a/360401/3898
+ *
+ * the front-end build file is really big
+ * still seeing a bunch of dependencies being added to it, like classnames, lodash, memoize, @babel and @emotion,
+ * etc.
```
ryelle: Over half of it is from moment-timezone. I see classnames, which makes sense since it's not a WP-provided package. lodash is correctly pulled out into the assets manifest, and I don't see these others.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Internal dependencies
+ */
+import { useScheduleData } from './data';
+import { ScheduleGrid } from './schedule-grid';
+import InspectorControls from './inspector-controls';
+import { NoContent } from '../../components/';
+import './edit.scss';
+
+export const ScheduleGridContext = createContext();
```
ryelle: If you pull this out into its own file, you can import it here & in front-end without also importing all the rest of the edit-related dependencies into front-end. Right now I think this is causing a bunch of things to get imported into the front-end build— you can see that `block-editor` is in the asset manifest, and that shouldn't be necessary on the front end.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/front-end.js
```diff
@@ -0,0 +1,108 @@
...
+ * also contains stuff from other custom blocks, like components/image/avatar.js, even though not used in this
+ * block.
```
ryelle: That's from importing from the main `/components` export — this is generally fine in the editor, since we usually want all those at some point, but for the front-end file you'll want to be explicit about imports (pull in directly from the component file).

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/front-end.js
```diff
@@ -0,0 +1,108 @@
...
+ * some of those may be expected, but others should definitely not be bundled.
+ * maybe just need to explicitly add the @wordpress/* components that we're using to package.json `dependencies`?
+ * then the above packages will be registered as externals automatically?
```
ryelle: No, the dependencies are automatically pulled out by webpack, we don't need to do anything (except make sure we're always importing from `@wordpress/*`).

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/front-end.scss
```diff
@@ -0,0 +1,11 @@
+/*
+ * todo
+ *
+ * does this file get bundled into blocks.min.css? if so, that could mess up the order b/c it might be
+ * re-including these after edit.scss does.
+ * 		maybe need to exclude front-end.scss, or just only include edit.scss
```
ryelle: blocks.min.css is generated from the blocks entry point, and the only place this is imported is front-end.js, so no — it should not be included in the blocks.min.css.

ryelle (commented):
Partway through a review, but ending my day— here's what I've got so far, will pick up again next week :)

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/controller.php
iandunn: I didn't look at that new one closely since I've been focused on the beta, but I just tested it and it looks like it works 👍 

0a0ba2da

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/edit.js
iandunn: d934e474 only had a marginal impact, but moving it makes sense logically, regardless 👍 

It looks like this accounted for a lot of the size:

`import { NoContent } from '../../components/';`

Fixed in 951a0a09. The front end file is down to 25k now, not counting `moment-timezone` :)

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/front-end.js
iandunn: 👍 951a0a0

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/front-end.js
iandunn: > Over half of it is from moment-timezone.

That's what i'm seeing too. Do you see a better way to work around that Gutenberg bug?

> classnames [...] makes sense since it's not a WP-provided package.

It's bundled with Core, though. If every plugin bundles their own copy of it, with different versions, wouldn't there be conflicts and unnecessary performance hits?

ryelle: > Do you see a better way to work around that Gutenberg bug?

Nope, aside from fixing the bug :) This is fine for now.

> classnames

It's 1KB, and shouldn't have any issues with conflicts. I imagine having to do an extra HTTP request would be more of a perf-hit than it's worth.

iandunn: >  shouldn't have any issues with conflicts

Why's that (now, and in the future)?

iandunn: > having to do an extra HTTP request would be more of a perf-hit than it's worth

Wouldn't it be `cat`'d with other scripts by `load-scripts.php`? I guess it also depends on HTTP1 vs HTTP2 also.

But I guess it doesn't really matter, since it doesn't look like Core exposes a discrete `classnames` script, the way it does with `moment` and `lodash`.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/front-end.scss
iandunn: 👍

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/front-end.js
ryelle: > Why's that (now, and in the future)?

Each script that bundles it bundles its own local version, and it doesn't populate any globals.

> But I guess it doesn't really matter, since it doesn't look like Core exposes a discrete classnames script, the way it does with moment and lodash.

Yeah, that's what I said originally. I thought you were asking why core doesn't expose it. (which my guess is "it's small enough to not matter")

iandunn: > doesn't populate any globals.

Ah, you're right, that's the difference.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -0,0 +1,354 @@
...
+	startEndTimes.sort(); // Put them in chronological order.
+
+	const timeList = startEndTimes.reduce( ( accumulatingTimes, time ) => {
+		const formattedTime = dateI18n( 'Hi', time );
+
+		return accumulatingTimes += `[time-${ formattedTime }] auto `;
+	}, '' );
+
+	const templateRows = `
+		/* Organizers: Set these to \`1fr\` to make the row height relative to the time duration of the session. */
```
ryelle, line 405: Why would someone want to do this, and should it be the default?

If it's a "nice alternative", could we add a functional CSS snippet to the handbook that organizers can copy-paste into the customizer?

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/session.js
```diff
@@ -0,0 +1,281 @@
...
+			<h4 className="wordcamp-schedule__session-title">
+				<a href={ titleLinkUrl } target={ titleLinkTarget } rel="noopener noreferrer">
+					{ title.rendered }
+
+					{ /* todo need to decode entities, e.g., "Doors Open &amp; Check-in". maybe other places too,
+					     like speaker & track names, etc */ }
+				</a>
+			</h4>
+
+			{ speakers && renderSpeakers( speakers, renderEnvironment ) }
```
ryelle: an empty array is truthy, so it's still rendering the speakers markup even if no speakers are set - you can check `speakers.length` instead.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/session.scss
```diff
@@ -0,0 +1,88 @@
...
+.wordcamp-schedule__session-tracks {
+
+	@include breakpoint-grid-layout {
+		display: none; /* Once the column headers are visible, this is no longer needed. */
+	}
+}
+
+.wordcamp-schedule__session-category {
+	font-style: italic;
+	color: #a8a8a8;
```
ryelle: This is much too light, try `#707070` so it keeps a valid contrast against the light grey background.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.scss
```diff
@@ -0,0 +1,116 @@
...
+ * todo
+ *
+ * time column is wider than track columns, but it should be equal or maybe a bit smaller
+ *
+ * alignfull causes horizontal scrolling, at least with some themes
+ *
+ * gutenberg's built-in "tablet" and "mobile" doesn't recognize the breakpoints
+ *
+ * setup vars for other colors. use convention from gutenberg if one is documented in future:
+ * https://github.com/WordPress/gutenberg/issues/19352
```
ryelle, line 14: Giving you a basic answer to that issue here: Generally, the `$dark-gray-900`, `$dark-gray-800` (vars with numbers) are from the WP (core) color palette, and that's the set of colors you can work with (well, the grays and blues you can use). The numbers correspond to how light/dark they are (higher = darker). The descriptively-named colors _ideally_ would inherit from those, so you'd never create a new grey (though that's not how it works in practice, because there isn't an enforcement of this). Those don't have a naming scheme AFAIK, but we might create one … soon-eventually.

We'd want to use the core colors so that our block's UI will match GB, and so that we don't need to involve a designer just to pick a color. In the past I've cherry-picked some of the base colors down into [a local variables file](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wordcamp-speaker-feedback/assets/scss/_variables.scss) for project use.

ryelle (commented):
🎉  Very exciting to see this in action! I clicked around a few of the demo page above, and it's working pretty well! Some screen reader issues, but otherwise good work 🙂  My more general feedback:

Something's off on the WCEU schedule, the last session ends at "Invalid date"

Screen Shot 2020-05-18 at 5 49 28 PM

It would be nice to hide the "Show Categories" & "Choose specific tracks" toggles if I don't have categories or tracks (if you agree, I can create an issue for it).

The screen reader experience is a bit weird. In [the original demo,](https://codepen.io/mrwweb/pen/ZaONLW) each time was announced first, and was just the start time. Then in each session, the time is displayed. This is better for screen reader users, since it gives the session time in context. IMO it's also better for everyone, since someone looking at this grid would have to think about it to see that Session 5 is 11:30-1pm.

The reading order also doesn't match visual order, it sort of jumps around the page a bit. Sighted users also use screen readers, so ideally the reading order would match how a person would scan the content.

screen-reader-flow

The column view isn't working correctly, it shows Session 5 as 11:30-12pm — I think this could also be fixed by swapping to just start time & adding the times into the session?

Screen Shot 2020-05-18 at 5 22 44 PM

The warning about overlapping sessions makes the grid pretty hard to read. Has this been looked at by a designer? maybe there's a better way for this warning.

Screen Shot 2020-05-18 at 3 25 40 PM

iandunn:
> invalid date

I think that's just because those sessions pre-date the `duration` field:

https://2019.europe.wordcamp.org/wp-json/wp/v2/sessions/6062

That won't happen in real world scenarios, because it'll always be set when post is saved. The beta announcement has a note letting people know they'll need to populate that field for the block to work.

iandunn:
> The screen reader experience is a bit weird. In the original demo, each time was announced first, and was just the start time.

Adding the end time was something @melchoyce [asked for](https://make.wordpress.org/community/2018/10/26/wordcamp-block-schedule/#comment-27668).

> The column view isn't working correctly, it shows Session 5 as 11:30-12pm — I think this could also be fixed by swapping to just start time & adding the times into the session?

Ah, yeah, that worked until we added the end time, but doesn't now, good catch.

@melchoyce, do you still want the end time there?

iandunn:
> It would be nice to hide the "Show Categories" & "Choose specific tracks" toggles if I don't have categories or tracks (if you agree, I can create an issue for it).

I personally dislike UI that appears and disappears without direct user action, since it's easy to overlook why it's not there, and be confused. Sometimes there are cases where someone might want to set it in anticipation of future data. Knowing that there's an option to display categories might also be a reason for someone to create them, when they otherwise wouldn't.

Feel free to start an issue to discuss future, though.

iandunn:
> The warning about overlapping sessions makes the grid pretty hard to read. Has this been looked at by a designer? maybe there's a better way for this warning.

Nope, that's just the best I could do without spending a ton of time on it. The fundamental problem is that the session times are overlapping, so the visual warnings are also overlapping. 

@melchoyce , do you have any ideas?

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
iandunn: It's just a different perspective, I don't think it should be the default. I think people are normally more likely to see something like this in their browser inspector than the docs, but in this case that probably isn't true.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/session.js
iandunn: 👍 7986949d

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/session.scss
iandunn: 👍 662d94cd

iandunn:
re: [time slots / sorting](https://github.com/WordPress/wordcamp.org/pull/502#issuecomment-630499895)

70d35cbc is a rough wip that fixes the mobile view, but breaks the grid view. fixing the grid view is probably possible, but i wanted to get some opinions before going deeper.

Screen Shot 2020-05-19 at 2 23 53 PM

Sessions 1 and 2 are sorted according to their tracks. Session 3 appears before sessions 1/2, but that makes sense to me b/c they're in a different time slot, and one that ends sooner. Combining 1-4 into the same slot seems misleading, because those sessions have different end times. People don't read, so when they scan the page it'll appear as if they have the same end times.

Adding the start & end times to each session cell is a different way to go, but that feels like it'd add a relatively large amount of clutter.

It seems like there's a fundamental difference between how the grid view and single column view order sessions, so it's not really possible for them to have the same order (without the side-effects above).

cc @mrwweb in case you have any thoughts.

ryelle:
> [Invalid date] won't happen in real world scenarios

k, let's keep an eye out in beta though, just in case.

---

I gave adding the full times to each session (mocked up in browser), and I don't think it looks too cluttered:

Screen Shot 2020-05-19 at 6 01 41 PM

Even in the mobile view, I think this is okay.

Screen Shot 2020-05-19 at 6 06 50 PM

I think you can use "people don't read" as an argument for why they should be grouped at start time too— I might see the the second grouping and assume it's a session block after the first. Linear is just not a great format for multiple tracks like this 🙂   So I think this needs to be a designer's call (sorry @melchoyce)

melchoyce:
> > The warning about overlapping sessions makes the grid pretty hard to read. Has this been looked at by a designer? maybe there's a better way for this warning.
> 
> Nope, that's just the best I could do without spending a ton of time on it. The fundamental problem is that the session times are overlapping, so the visual warnings are also overlapping.
> 
> @melchoyce , do you have any ideas?

Maybe something like this, where any sessions with errors are highlighted, and the error shows up immediately before the block?

image

@michaelarestad would love a gut-check from you here ^

melchoyce:
re: session times, could we do something more like this?

image

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.scss
iandunn: I don't think this is a blocker for beta, but I appreciate the background info 👍

ryelle: I'm not requesting any changes, was just answering your question.

iandunn:
> where any sessions with errors are highlighted, and the error shows up immediately before the block?

That seems worth exploring 👍

iandunn:
2f06f26a moves the `overlapping` error message to the start of the day.

iandunn:
>  I might see the the second grouping and assume it's a session block after the first.

That's a fair point.

There's a good amount of spacing between the elements, but it does feel cluttered to me in the sense that there's just a lot of information in each cell. That's probably a worthwhile tradeoff, though 👍

iandunn:
> It would be nice to hide the "Show Categories" & "Choose specific tracks" toggles if I don't have categories or tracks (if you agree, I can create an issue for it).

Ah, good point about how it's unintuitive when no dates/tracks exist, though. 929dd4f9 replaces the blank space with an explanation.

iandunn:
b994488 changes back to grouping by start time, irrespective of end time

MichaelArestad:
> @MichaelArestad would love a gut-check from you here 


Sorry for the delay. That's seems like a good start. Alternatively, we could do something like Gcal and other calendars do where the two conflicting items are "next to each other" in the same cell.

image

mrwweb:
Sorry I'm late to this discussion, but it seems like you all ended up in a good place. Congrats on the beta launch everyone!

A quick note on another reason for grouping by start time (which is what I did in the original demo, fwiw): I think the primary use case for browsing the schedule by start time is, "Where do I want to go at {start time}?" Even if a session ends 30 minutes before another, I can't go to the beginning of both. I can only walk through one door at 8:00am.

The end time is _definitely_ a useful secondary consideration. That's particularly true for workshops that might span multiple time "slots", but that's a less-frequent scenario. I like keeping the time span visible detail after session title on mobile for that exact reason, but end time feel distinctively secondary in importance to start time.

PR #504, opened 2020-05-19
Speaker Feedback: Add a periodic email notification for organizers when there are unapproved feedback submissions

### LINKED ISSUE #454: Speaker Feedback: Daily notification for organizers when there's unapproved feedback?
Similar to the [daily notification to speakers](https://github.com/WordPress/wordcamp.org/pull/441) when there's new feedback, what if we had a daily notification get sent to the WordCamp's Speaker Organizer's email, with a fallback to the main `city@wordcamp.org` address, that just tells how many unapproved feedback comments there are with a link? (And obviously don't send the email when there aren't any.)

Currently the only notification that organizers will get is a [one-time notification](https://github.com/WordPress/wordcamp.org/issues/411) from WC Central two weeks after the event has ended. This would be a way to prompt organizers to moderate feedback sooner, so speakers can get it sooner, without deluging anyone with notifications (one per day seems reasonable).

cc @dd32 who reminded me about this idea

coreymckrill:
Similar to #441, this adds a daily cron job that will send an email to the `city@wordcamp.org` email address when there are unapproved feedback submissions that need attention. It is timed to send one hour before the cron job for speakers, so that there is the possibility that speakers can get notified about feedback that was just approved.

The email that goes to organizers gives a snapshot of how many feedback submissions are unapproved, approved, and spammed. This provides context as to whether new feedbacks are continuing to come in or not. It also nudges them to check the spammed feedbacks to make sure there aren't any false positives.

Fixes #454 

1. Add some feedback comments to a test site. Have a mix of approved, unapproved, and spammed comments.
1. Similar to the testing instructions for #441, trigger the cron job manually with wp-cli. In this case the command would be something like
  `wp --allow-root --url=https://2014.seattle.wordcamp.org cron event run sft_notify_organizers_unapproved_feedback`
1. The email that comes into MailCatcher should be sent to the `city@wordcamp.org` address for the test site. It should have a link to the feedback list table for the site. It should have a list of counts for unapproved, approved, and spammed feedback.
1. Try adjusting those counts so that they are `1`, to make sure the single/plural translation functions are set up correctly.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
@@ -2,7 +2,8 @@
 
 namespace WordCamp\SpeakerFeedback\Cron;
 
-use function WordCamp\SpeakerFeedback\Comment\{ get_feedback, update_feedback };
+use function WordCamp\SpeakerFeedback\Admin\get_subpage_url;
+use function WordCamp\SpeakerFeedback\Comment\{ count_feedback, get_feedback, update_feedback };
 use function WordCamp\SpeakerFeedback\Post\{
 	get_earliest_session_timestamp, get_latest_session_ending_timestamp,
 	get_session_speaker_user_ids, get_session_feedback_url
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Working great, the email is sent to my test organizer email, with the correct strings 👍 

The email itself feels very transactional… a greeting could help, even just "Hi [WC] Organizers" or something.

coreymckrill:
> The email itself feels very transactional… a greeting could help

That's a good point. Added here: 6b5d3873

PR #506, opened 2020-05-26
Speaker Feedback: Use the event name as the sender in notif emails

### LINKED ISSUE #479: Speaker Feedback: Change the sender name in notification emails to be the name of the WordCamp event
Currently, the sender is just "WordPress". It seems like it would be more helpful, and more likely to be opened if the email came from the specific WordCamp.

coreymckrill:
This provides a bit more context for the recipients of the email notifications, instead of just coming from "WordPress".

Fixes #479

1. Try testing the email notifications for [organizers](https://github.com/WordPress/wordcamp.org/pull/504) or for [speakers](https://github.com/WordPress/wordcamp.org/pull/441). The `From:` header in the email notification should be something like `WordCamp Test 2020 <wordpress@2020.test.wordcamp.org>`. Without this patch, the `From:` header would be `WordPress <wordpress@2020.test.wordcamp.org>`.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
@@ -120,7 +120,9 @@ function notify_organizers_unapproved_feedback() {
 	$message .= "\n\n";
 	$message .= get_subpage_url( 'wcb_session' );
 
+	add_filter( 'wp_mail_from_name', __NAMESPACE__ . '\filter_notification_from_name' );
 	wp_mail( $to, $subject, $message );
+	remove_filter( 'wp_mail_from_name', __NAMESPACE__ . '\filter_notification_from_name' );
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Works as described 👍   (once I figured out where to look in mailcatcher 🙂  )

Do you think this would be useful for all emails sent from WordCamp sites? (no need to change anything here, just thinking aloud)

coreymckrill:
> Do you think this would be useful for all emails sent from WordCamp sites?

I was thinking about that. I think there are some other places in the codebase where we're also modifying the sender name, but they don't all use the same method. Examples: [1](https://github.com/WordPress/wordcamp.org/blob/b26ed9ab986cf0dfe82f1a9a01d5cfe1da478ac3/public_html/wp-content/plugins/camptix/camptix.php#L7877-L7906), [2](https://github.com/WordPress/wordcamp.org/blob/b26ed9ab986cf0dfe82f1a9a01d5cfe1da478ac3/public_html/wp-content/plugins/camptix-invoices/includes/class-camptix-addon-invoices.php#L341-L364), [3](https://github.com/WordPress/wordcamp.org/blob/b26ed9ab986cf0dfe82f1a9a01d5cfe1da478ac3/public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php#L417-L425)

But those (and a few others) would be pretty easy to clean up if we decided to use filters to standardize the sender name and email. I guess the real question is, are there any cases where we _wouldn't_ want the email to come from `WordCamp Name <city@wordcamp.org>`?

ryelle:
You could keep the "if WordPress" check, and only override the header if it's the default value - that would prevent interactions with any of those customizations, at least.

I can't think of a time we wouldn't want to use the camp name, it just seems more clear overall.

coreymckrill:
Created a new issue to continue the sender name discussion: https://github.com/WordPress/wordcamp.org/issues/510

PR #507, opened 2020-05-26
Speaker Feedback: Add blurb about wporg account in speaker email notification

### LINKED ISSUE #493: Speaker Feedback: Explain wp.org login in speaker notification email
Based on some feedback:

> The email should mention that they need to log in using their wordpress.org credentials, because it’s not mentioned anywhere

The link takes the speaker directly to a login screen, where there isn't any explanation of which login to use or where they're going. This info should go into the email above the links.

coreymckrill:
Since the link takes the speaker directly to a login screen, where there isn't any explanation of which login to use or where they're going, we should include it in the notification email instead.

Fixes #493

1. [Send a speaker notification.](https://github.com/WordPress/wordcamp.org/pull/441) The changed wording should correctly be singular or plural depending on how many sessions have new feedback submissions.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/cron.php
@@ -200,7 +200,14 @@ function notify_speakers_approved_feedback() {
 
 		$message  = sprintf( esc_html__( 'Hi %s,', 'wordcamporg' ), $speaker->display_name );
 		$message .= "\n\n";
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Works as expected, with one & two sessions 👍 

The wording feels a little clunky to me… but I don't have a better suggestion 🤷🏼‍♀️

coreymckrill:
Let's try this wording for now, and we can iterate again later if it's still causing confusion for speakers.

PR #509, opened 2020-05-28
Speaker Feedback: Show feedback forms to organizers even before event

### LINKED ISSUE #505: Speaker Feedback: Allow access to front end views prior to the event for theme styling
The front end views, including the "select session" form on the `/feedback` page, the feedback forms on individual sessions, and the approved feedback shown to speakers are currently not accessible, even to organizers who are trying to style the site's theme, until the event has started. This is obviously too late for theming efforts.

One potential solution would be to show these elements when in preview mode for the page/sessions in question. In the case of the Leave Feedback page, though, that would require unpublishing the page first (since it gets published upon creation when the plugin is activated) and then previewing the draft.

A simpler solution would be to just always show the elements to organizers, regardless of proximity to the event date. One potential downside might be that organizers wouldn't necessarily realize that site visitors aren't seeing the same things that they are, but there are probably ways around that with some explanatory text in notice boxes.

coreymckrill:
Organizers need to be able to access the feedback forms for testing and styling purposes before the event date. This overrides the date and session time checks if the current user has the right capes.

It also adds explanatory notices so that organizers understand that they aren't necessarily seeing the same view that their site visitors will.

Fixes #505

### Screenshots

Notice on the `/feedback` page:

session-list-form

Notice on a session feedback form:

feedback-form

1. Configure your session times so that you're not within the window when feedback submissions are open. (i.e. times in the future, or times more than two weeks in the past)
1. Visit the `/feedback` page while logged out and confirm that the session select form is not available. Visit the feedback URL for a single session and confirm the same there.
1. Then visit those URLs while logged in as an admin. You should see the forms, along with the notices in the screenshots above.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/post.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/post.php
@@ -4,6 +4,7 @@ namespace WordCamp\SpeakerFeedback\Post;
 
 use WP_Error;
 use WordCamp_Post_Types_Plugin;
+use const WordCamp\SpeakerFeedback\Comment\COMMENT_TYPE;
 
 defined( 'WPINC' ) || die();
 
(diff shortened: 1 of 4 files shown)


ryelle (approved):
Looks good, and 👍  for pulling out `event_accepts_feedback`, the render function is much easier to follow now.

PR #511, opened 2020-06-01
Tooling: Remove `no-descending-specificity` stylelint rule

ryelle:
We've been disabling this rule ad-hoc across projects. While descending specificity is generally a code smell (because the earlier, higher-specificity rule will override the later rule), this is only true if they're both setting the same property. In some cases, this is flagging places where we're overriding core styles, and it isn't possible or doesn't make sense to refactor just to satisfy this rule.

1. Travis should pass


iandunn (approved):
👍

PR #512, opened 2020-06-02
Post Types: Add gutenberg support on speaker & organizer post types

ryelle:
Started in #303, this switches the metaboxes on Speaker and Organizer posts to use native gutenberg panels. It also refactors the sessions controls due to introducing a new hook, `usePostMeta`. Since these panels are introduced, this also restricts Speaker & Organizer to the gutenberg editor only.

### Screenshots

Organizer:
organizer

Speaker:
speaker

The usernames are validated when saving, so invalid usernames will be cleared from the input. On save, there's a hook that runs to fetch the user ID and fill in `_wcpt_user_id`. This runs for both organizers & speakers, since both use the same meta key.

1. Try adding or editing a speaker
2. Fill out the email & username
3. Save— the values should save correctly (check either by reloading, or using wp-cli's `post meta list`, etc)
4. Repeat for an organizer
5. Try with invalid usernames, they should not save (it would be a better UX for these to not fail silently, but this replicates the current flow)


public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -32,6 +34,49 @@ function register_sponsor_post_meta() {
...
+				return $wporg_user->user_login;
+			},
+		)
+	);
+	register_post_meta(
+		'wcb_speaker',
+		'_wcb_speaker_email',
+		array(
+			'type'         => 'string',
+			'show_in_rest' => true,
```
coreymckrill: It looks like this exposes the speaker's email address in responses to unauthenticated requests to the `speakers` endpoint. Limiting the context to `edit` seems to avoid this:

```suggestion
			'show_in_rest' => array(
				'schema' => array(
					'context' => array( 'edit' ),
				),
			),
```

public_html/wp-content/plugins/wc-post-types/js/src/speaker/panel-info.js
```diff
@@ -0,0 +1,36 @@
...
+export default function SpeakerInfoPanel() {
+	const [ email, setEmail ] = usePostMeta( '_wcb_speaker_email', '' );
+	const [ username, setUsername ] = usePostMeta( '_wcpt_user_name', '' );
+
+	return (
+		<PluginDocumentSettingPanel
+			name="wordcamp/speaker-info"
+			className="wc-panel-speaker-info"
+			title={ __( 'Speaker Info', 'wordcamporg' ) }
+		>
```
coreymckrill, line 23:
```suggestion
			title={ __( 'Speaker Info', 'wordcamporg' ) }
			initialOpen={ true }
		>
```

I'm not actually sure off the top of my head if `PluginDocumentSettingPanel` supports this parameter, but I think it would be useful to have the panel start open so it's more discoverable, especially for users with prior experience with the old meta boxes.

public_html/wp-content/plugins/wc-post-types/js/src/organizer/panel-info.js
```diff
@@ -0,0 +1,29 @@
...
+
+export default function OrganizerInfoPanel() {
+	const [ username, setUsername ] = usePostMeta( '_wcpt_user_name', '' );
+
+	return (
+		<PluginDocumentSettingPanel
+			name="wordcamp/organizer-info"
+			className="wc-panel-organizer-info"
+			title={ __( 'Organizer Info', 'wordcamporg' ) }
+		>
```
coreymckrill, line 21: Same as other comment, I think it would help for this panel to start "open".

coreymckrill (changes requested):
👍 Works in local testing. The one blocker is exposing the speaker email address, everything else is just non-blocking questions/comments.

> it would be a better UX for these to not fail silently, but this replicates the current flow

Since we're removing the old metaboxes here anyway, do we need to fully replicate the current flow? I agree that the UX is not great. Could we at least show an error notice adjacent to the field?

public_html/wp-content/plugins/wc-post-types/js/src/speaker/panel-info.js
ryelle: It doesn't, unfortunately — whether the panel is open or not is a user setting (persisted from the last state).

ryelle:
> Since we're removing the old metaboxes here anyway, do we need to fully replicate the current flow? I agree that the UX is not great. Could we at least show an error notice adjacent to the field?

Not… easily 😅   I was aiming to not add any extra API endpoints/data stores for this, which we'd need to fire off a validation check on this field. I'll double check, but I'm pretty sure there's no way to communicate invalid data back from the meta saving request.

ryelle:
Switching back to in-progress so I can rebase & double check the email permissions.

ryelle:
note: check that this fixes [#4353-meta](https://meta.trac.wordpress.org/ticket/4353)

ryelle:
I've rebased this on #549 so that the tests pass, but can drop those commits once that's merged.

Re invalid usernames & showing some error message… I've added a simple `validation` endpoint, which does nothing but runs the provided validation on the parameters. I left it a bit vague in case we want to add any other data validation, but it could also be easily removed if we ever get a multisite-capable users endpoint (which could then be used as a dropdown field, instead of the free text currently).

Anyway, when an invalid username is entered, once the user leaves the text field (on blur), the username value is checked against the API. If it's invalid, a message is shown:

Screen Shot 2020-08-14 at 5 07 09 PM

coreymckrill (approved):
👍 Works in local testing

I think this validation endpoint is a good first iteration. It might be worth it later to abstract it a bit to use the `type` and/or `sanitize_callback` properties of a registered meta field to do the validation, instead of having to define another validation function for each key.

Another idea for a future iteration: Use a delay after the last key up to run the validation request instead of waiting for blur, and also show some kind of "waiting" message/animation while running the request. The validation took long enough, at least locally, that I wasn't sure anything was happening.

ryelle:
> Use a delay after the last key up to run the validation request instead of waiting for blur, and also show some kind of "waiting" message/animation while running the request. The validation took long enough, at least locally, that I wasn't sure anything was happening.

Yeah, the UX leaves something to be desired - I started down this path but it was quickly getting over-engineered. Do you think it's worth creating a followup issue now, or wait to see if anyone else reports it?

coreymckrill:
Probably fine to wait for now 🤷‍♂️

PR #514, opened 2020-06-04
Docker Docs: Add note about noglob zsh

0aveRyan:
The asterisk/wildcard operator triggers character expansion and creation of the key files fails without noglob in zsh. As this is the default for new Macs, this seems worth highlighting.


ryelle (approved):
Good catch, thanks for the PR 👍

PR #518, opened 2020-06-05
Post Types: Use correct capability checks for editing post meta

### LINKED ISSUE #516: Post Types: Meta (session duration, etc) doesn't save when custom field panel is open
Sometimes it seems like session meta doesn't save. I think the meta fields in the panel are overriding our custom inputs.

To reproduce:

1. Edit a session
2. In the kebab menu, click Options and turn on Custom fields
3. Page will reload, and custom fields panel will display
4. Edit the session date, time, etc. Save.
5. It appears to have saved, but reload the page
5. (if this is the first time saving these values, it'll work. try again.)
7. Meta is reset back to previous value 😞

ryelle:
Rather than hacking around the "protected" check, which opens up other issues by exposing the values in custom fields, we should use the auth_callback property to intentionally allow access to `_wcpt` meta.

Fixes #516 

Check the permissions work as expected:

1. Sign in as a non-admin user— author, editor
2. Edit a session
3. Set the session time, duration, add a value for slides URL, etc
4. It should save correctly

Check that it fixes #516:

1. Edit the session again
2. In the kebab menu, click Options and turn on Custom fields
3. Page will reload, and custom fields panel will display
4. No `_wcpt_*` values should display
5. Edit the session date, time, etc.
6. It should save correctly (reload page to check values).

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -14,7 +14,6 @@ require_once 'favorite-schedule-shortcode.php';
 
 add_action( 'init', __NAMESPACE__ . '\register_sponsor_post_meta' );
 add_action( 'init', __NAMESPACE__ . '\register_session_post_meta' );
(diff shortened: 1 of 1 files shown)


iandunn (approved):
That looks good to me 👍

PR #521, opened 2020-06-11
Camptix: change payment gateways account selector label

### LINKED ISSUE #520: Camptix: Clarify Payment Gateway Account Settings
_[Original ticket in Trac.](https://meta.trac.wordpress.org/ticket/4243)_

The option "None" for Tickets >> Setup >> Predefined Account is unclear - this seems like there would be no payment gateway if that setting is chosen, which makes it unclear how you can access the custom gateway settings.

Changing this label to "Custom Account" or "Local Payment Account" would be clearer.

timiwahalahti:
Make Paypal and Stripe gateway account selector label clearer.

Fixes #520 

1. Navigate to Camptix setup and Payments tab

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/payment-paypal.php
+++ b/public_html/wp-content/plugins/camptix/addons/payment-paypal.php
@@ -59,7 +59,7 @@ class CampTix_Payment_Method_PayPal extends CampTix_Payment_Method {
 	function payment_settings_fields() {
 		// Allow pre-defined accounts if any are defined by plugins.
 		if ( count( $this->get_predefined_accounts() ) > 0 ) {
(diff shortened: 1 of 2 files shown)


ryelle:
I think the ticket is actually asking to change this "None" option to "Custom Account" — once you save that, custom fields appear to enter the account info.

Screen Shot 2020-06-15 at 4 07 04 PM

timiwahalahti:
Oh yes! Fixed the PR, tho not quite as described in the original ticket:

* Label is "Account" instead of "Predefined Account" because, in my opinion, it doesn't make sense to use "Predefined Account" if there is an option to use one which is not predefined ;)
* "Custom" instead of "Custom Account" to remove repeating terms, predefined options doesn't have "Account" in names neither

ryelle:
Apparently I didn't do that exactly right for github, but this was merged in https://github.com/WordPress/wordcamp.org/commit/17b8b95597620789740725dfd7ea539b172b46c9 — thanks!

PR #523, opened 2020-06-12
Central: Require login before submitting application forms and pre-fill wp.org username field

timiwahalahti:
As discussed on #333, here comes the separate PR for requiring login before users can submit WordCamp or Meetup applications.

If the user is not logged in, we wrap the form into a wall and show a login prompt. Also on submit, login status is checked and the error message returned if the user is not logged in.

When user is logged in, WordPress.org username fields in both form do get pre-filled.

Maybe not the most prettiest way and does not uniform the way how these forms and `wordcamp-forms-to-drafts` plugin does the login requirement, but these forms are build in quite a different way so standardising the functionality is hard. Styles could be loaded from one and same place tho, for consistent experience - but I'll do that after this has been merged as now it's more critical to get the wall in place to mitigate spam.

See also #321 

### Screenshots

**When user is not logged in**

**After user has logged in**

1. Visit [Meetup application page](https://central.wordcamp.test/meetup-organizer-application/) when not logged in (private browsing). You should see a login prompt.
2. With inspector remove the prompt, fill the form and submit. You should see an error message telling that WordPress.org login is required.
3. Now visit the Meetup application page when you are logged in. You should not see the login prompt and WordPress.org username field should be filled.

### DIFF
--- a/public_html/wp-content/plugins/wcpt/css/applications/common.css
+++ b/public_html/wp-content/plugins/wcpt/css/applications/common.css
@@ -71,6 +71,7 @@
 	opacity: 0.5;
 }
 
+/* stylelint-disable-next-line selector-class-pattern */
 .wcfd-disabled-form .PDF_pageInner {
 	width: auto;
 }
--- a/public_html/wp-content/plugins/wcpt/css/applications/common.css
+++ b/public_html/wp-content/plugins/wcpt/css/applications/common.css
@@ -25,3 +25,58 @@
 	border-left-color: #dc3232;
 	color: #dc3232;
 }
+
+.wcfd-disabled-form {
+	position: relative;
+}
+
+.wcfd-disabled-form #wcorg-login-message {
+	position: absolute;
+	z-index: 10;
+	margin: 2.4em;
+	padding: 1.8em;
+	background-color: #c62828;
+	border-radius: 0.25em;
+	margin-bottom: 2em;
+}
+
+.wcfd-disabled-form #wcorg-login-message p {
+	color: #fff !important /* !important to prevent theme from unintentionally overriding */;
+}
+
+.wcfd-disabled-form #wcorg-login-message p:first-child {
+	margin-top: 0;
+}
+
+.wcfd-disabled-form #wcorg-login-message p:last-child {
+	margin-bottom: 0;
+}
+
+.wcfd-disabled-form #wcorg-login-message a {
+	color: #fff;
+}
+
+.wcfd-disabled-form #wcorg-login-message a:hover,
+.wcfd-disabled-form #wcorg-login-message a:focus,
+.wcfd-disabled-form #wcorg-login-message a:active {
+	color: #fff;
+}
+
+.wcfd-overlay {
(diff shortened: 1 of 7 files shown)


public_html/wp-content/wp-cache-config.php
```diff
@@ -24,13 +24,13 @@
 $wp_cache_disable_utf8         = 0;
```
ryelle: Can you remove the changes to this file?

public_html/wp-content/plugins/wcpt/css/applications/common.css
```diff
@@ -25,3 +25,58 @@
...
+.wcfd-overlay {
+	position: absolute;
+	z-index: 5;
+	width: 100%;
+	height: 100%;
+	background-color: #000;
+	opacity: 0.5;
+}
+
+.wcfd-disabled-form .PDF_pageInner {
```
ryelle, line 75: The underscore here is causing issues with stylelint, but it's valid. You can disable it by adding `/* stylelint-disable-next-line selector-class-pattern */` on the line above the rule.

ryelle (approved):
This is looking good, especially as a step to stopping all the spam. And ➕   for adding in the pre-filled form fields, good idea.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-application.php
```diff
@@ -67,7 +67,26 @@ public function render_application_shortcode() {
...
 			$countries = wcorg_get_countries();
-			$this->render_application_form( $countries );
+
+			$wporg_username = '';
+			if ( is_user_logged_in() ) {
+				$current_user = wp_get_current_user();
+				$prefilled_fields = array(
+					'wporg_name'			=> $current_user->display_name,
+					'wporg_username'	=> $current_user->user_login,
+					'wporg_email'			=> $current_user->user_email,
```
ryelle:
```suggestion
					'wporg_name'     => $current_user->display_name,
					'wporg_username' => $current_user->user_login,
					'wporg_email'    => $current_user->user_email,
```

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-application.php
```diff
@@ -67,7 +67,26 @@ public function render_application_shortcode() {
...
+				);
+			}
+
+			if ( ! is_user_logged_in() ) {
+				echo '<div class="wcfd-disabled-form">' . wcorg_login_message( '', get_permalink() ) . '<div class="wcfd-overlay"></div><div inert>';
```
ryelle, line 83:
```suggestion
				// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
				echo '<div class="wcfd-disabled-form">' . wcorg_login_message( '', get_permalink() ) . '<div class="wcfd-overlay"></div><div inert>';
```

This "security issue" can be ignored, we can add better escaping into #524.

ryelle (approved):

PR #524, opened 2020-06-12
Update wcorg-login-message.php plugin

### LINKED ISSUE #522: Update wcorg_login_message function
Currently `wcorg_login_message` returns old forums address for login. That address has been deprecated and it redirects to login.wordpress.org. We should update the address for login.

At the same time, the message could be tweaked a little bit.

The function should also return a link to the login itself, as it is used also outside of wp-login.php view (for example submission forms and soon to come application forms).

timiwahalahti:
`wcorg_login_message` returned old WordPress.org login link that referred to forums and logic to create that link was complex. Now that WordPress.org has one centralised "SSO" login at login.wordpress.org, we should direct users there instead of extra redirect hop from old forums login address.

The same message is also used outside `wp-login.php` view, but it didn't contain any link to the place where user can actually do the login and return to the same page. Added this link to the message and made it hidden on `wp-login.php` view, where it is not needed.

At the same time changed the function name from where the wp.org URL is get and deprecated the old function.

Fixes #522 

### Screenshots

**Login prompt on forms with new message and link to wp.org login which will redirect back to the current page.**

**wp-login.php on WordCamp.org stays as is.**

1. Navigate to [Call for Speakers post](https://2017.fyre.wordcamp.test/2020/01/26/call-for-speakers/) when not logged in. You should see in message link to do the login.


ryelle:
The CSS to hide this on the wp-login page isn't working for me. What if you took a different approach, and updated `wcorg_login_message`? Add a parameter for a custom message— so that both the speaker form & application can pass in a string, and it'll replace that first line (which would also fix [the string-replace workaround here](https://github.com/WordPress/wordcamp.org/blob/f31a2ec0d8616ffcb317955750a918acb301aa1c/public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php#L105-L109)).

timiwahalahti:
IIRC there is some other place that trusts that the message exists there. But I can be wrong, will check and refresh this PR tomorrow.

timiwahalahti:
PR updated:

* Fixed the styling issue, wp.org login link should be now hidden on the wp-login screen
* As @ryelle suggested, updated `wcorg_login_message` to actually use the `$message` argument. The function will fallback to default message if none provided.
* Updated `camptix-tweaks` and `wordcamp-forms-to-drafts` to use the `$message` argument instead of `str_replaced`. I did search on `mu-plugins` and `plugins` directories, looks like those were only places where string-replacements were made.

ryelle:
I've added some commits:
- Fix the linter issues
- Remove the deprecated function— since this code is only used on WordCamp.org, and we know that it isn't used elsewhere, we can remove the function.
- I removed the login instructions, since it's redundant now that we can create custom messages, and now we don't need to conditionally hide anything. 
- I merged in production, which brings in #523, so I also added the custom message to that overlay.

@timiwahalahti Are you okay with my changes here?

timiwahalahti:
@ryelle 🙏👍Thanks!

PR #525, opened 2020-06-15
Disable Jetpack's automatic spam deletion if the WordCamp is in future or has ended less than month ago

### LINKED ISSUE #464: Jetpack: Don't delete old contact form "spam" until a camp is over
_[Original ticket in Trac](https://meta.trac.wordpress.org/ticket/3524)_

 Jetpack contact forms use Akismet to catch spam, and then delete the spam periodically. Sometimes there are false positives, though, which can mean that sponsor requests, etc get deleted before the organizers manually check the spam folder.

We could solve that by adding something to the jetpack-tweaks mu-plugin that would check if the camp has happened yet (via get_wordcamp_post()->meta['Start Date (YYYY-MM-DD']), and if it hasn't, then unhook grunion_delete_old_spam.

timiwahalahti:
Jetpack deletes normally spam submissions after 15 days. Sometimes there are false positives and organisers do miss important messages because those get deleted before team manually checks spam folder.

Keep the spam submissions until a month has passed from the start of WordCamp, just in case of some.

Fixes #464

Probably the easiest way:

1. Visit WordCamp site which is still in future and check that action calls for `grunion_delete_old_spam()` and `Grunion_Contact_Form_Plugin->daily_akismet_meta_cleanup()` are **not** there (with Query Monitor for example).
2. Visit WordCamp site that has ended months ago and check that action calls for those functions are there.

### DIFF
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
@@ -133,8 +133,7 @@ add_filter( 'admin_init', __NAMESPACE__ . '\disable_jetpack_blog_follow_emails' 
  *
  * Jetpack deletes normally spam submissions after 15 days. Sometimes there are false positivies and
  * organisers do miss important messages because those get deleted before team manually checks spam folder.
- * Keep the spam submissions until month has passed from the start of WordCamp, just in case of some
- * @return [type] [description]
+ * Keep the spam submissions until month has passed from the start of WordCamp, just in case of some.
  */
 function disable_jetpack_spam_delete() {
 	$wordcamp = get_wordcamp_post();
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
@@ -138,7 +138,7 @@ add_filter( 'admin_init', __NAMESPACE__ . '\disable_jetpack_blog_follow_emails' 
  */
 function disable_jetpack_spam_delete() {
 	$wordcamp = get_wordcamp_post();
-	$wc_start = $wordcamp->meta['Start Date (YYYY-mm-dd)'][0];
+	$wc_start = $wordcamp->meta['Start Date (YYYY-mm-dd)'][0] ?? '';
 
 	/**
 	 * Bail if WordCamp start date has not been set.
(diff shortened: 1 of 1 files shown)


public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
```diff
@@ -127,3 +127,40 @@ function disable_jetpack_blog_follow_emails() {
...
+ * Disable Jetpack's automatic spam deletion if the WordCamp is in future or has ended less than month ago.
+ *
+ * Jetpack deletes normally spam submissions after 15 days. Sometimes there are false positivies and
+ * organisers do miss important messages because those get deleted before team manually checks spam folder.
+ * Keep the spam submissions until month has passed from the start of WordCamp, just in case of something
+ * is needed from submissions after WordCamp has ended.
+ */
+function disable_jetpack_spam_delete() {
+	$wordcamp = get_wordcamp_post();
+	$wc_start = $wordcamp->meta['Start Date (YYYY-mm-dd)'][0];
```
coreymckrill: I think this would trigger an undefined index notice if the start date has not been set yet. One way to avoid this would be to use the null coalescing operator to set an empty value if it doesn't exist.

```suggestion
	$wc_start = $wordcamp->meta['Start Date (YYYY-mm-dd)'][0] ?? '';
```

public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
```diff
@@ -127,3 +127,40 @@ function disable_jetpack_blog_follow_emails() {
...
+	if ( absint( $wc_start ) < strtotime( '-30 days' ) ) {
+		return;
+	}
+
+	/**
+	 * Remove spam deletion actions.
+	 * One for the actual submit and second for Akismet metadata attached to each submission.
+	 */
+  remove_action( 'grunion_scheduled_delete', 'grunion_delete_old_spam' );
+  remove_action( 'wp_scheduled_delete', array( \Grunion_Contact_Form_Plugin::init(), 'daily_akismet_meta_cleanup' ) );
```
coreymckrill:
```suggestion
	remove_action( 'grunion_scheduled_delete', 'grunion_delete_old_spam' );
	remove_action( 'wp_scheduled_delete', array( \Grunion_Contact_Form_Plugin::init(), 'daily_akismet_meta_cleanup' ) );
```

coreymckrill (approved):
👍 A couple of code suggestions, but otherwise looks good to me! Thanks!

coreymckrill (approved):

PR #526, opened 2020-06-17
Helpers: Add function to check for "beta test" status

### LINKED ISSUE #287: Network: Use site meta for flagging sites as beta or testing

ryelle:
This uses a site meta value, which can be set via WP-CLI, to check if a given site has opted in to a beta test. Development environments are always considered beta testers.

Fixes #287

1. Set a beta value using wp-cli, ex: `wp site meta set [site ID] wordcamp_beta_test true`
2. Check that `is_wordcamp_beta( 'test' )` returns true
3. Remove the meta, or set it to false
4. Check that `is_wordcamp_beta( 'test' )` returns false

### DIFF
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -46,6 +46,30 @@ function is_wordcamp_test_site( $blog_id = null ) {
 }
 
 /**
+ * Check if a WordCamp site is participating in a beta test of a feature. All development environments are
+ * considered beta testers.
+ *
+ * Currently `wordcamp_beta_` blogmeta keys need to be set manually via wp-cli:
+ *    wp site meta set [site ID] wordcamp_beta_{$beta} true
+ *
+ * @param string   $beta The slug ID of the beta feature.
+ * @param int|null $blog_id Optional. The blog ID to check. Defaults to current blog ID.
+ *
+ * @return bool
+ */
+function is_wordcamp_beta( $beta, $blog_id = null ) {
+	if ( 'production' !== get_wordcamp_environment() ) {
+		return true;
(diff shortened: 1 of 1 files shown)


coreymckrill (approved):
This is so funny, I was just about to create this same PR. LGTM.

ryelle:
hah! well, it is your code originally, but I needed it for another PR in progress 😆

PR #529, opened 2020-06-18
Speaker Feedback: Handle activate/deactivate routines across network

### LINKED ISSUE #370: Speaker Feedback: Network enable/disable doesn't handle the activation/deactivation hooks correctly
When the activation state of this plugin is changed network-wide, the related activation/deactivation routines that are registered do not appear to be firing, because the Leave Feedback page is not created upon activation, and it's not deleted upon deactivation.

This was partially addressed [here](https://github.com/WordPress/wordcamp.org/pull/357/commits/b999a6a5be08a32bcaae1dbd31f0c9807274c642) by also checking during `init` and creating the page if necessary, but it doesn't address the deactivation part.

I have a theory (untested, just wanted to get this idea out in an issue) that these routines _are_ actually running, but because they're being run from the network admin, they're not running in the proper context to do what is intended.

So I think what might be needed is for the `activate` and `deactivate` callbacks to check the optional boolean parameter that's supplied during the relevant action hook that says whether or not this is a network context. If it is, the routine will need to run through a `switch_to_blog`/`restore_current_blog` loop for each site in the network in order to add/remove the feedback page and related option.

If this is the solution, the other question is, do we actually want to do that on a network with 1k+ sites? That would potentially be an intensive operation. Obviously we could skip sites with the [skip feature flag](https://github.com/WordPress/wordcamp.org/pull/363). We also wouldn't run these routines very often (`deactivate` maybe never on production).

coreymckrill:
This ensures that the plugin activate/deactivate routines are fired from the root WordCamp site where network plugin operations happen, even though the plugin itself won't be initialized there. It also ensures that the routines run correctly on all valid sites in the network (valid = has `wc-post-types` plugin activated and doesn't have the skip flag).

Being able to run these routines on the right sites without cycling through every site in the network was made possible by #528.

Fixes #370 

1. On your local dev environment, make sure you have multiple WordCamp sites set up that aren't the root site or WC Central.
1. Try network activating the plugin. The "Leave Feedback" page should get generated on every site in the network.
1. Network deactivate the plugin, and the page should get deleted from every site in the network.
1. Set the `speaker_feedback` skip flag for one of the WordCamp sites, plus the root site and WC Central. Example command:
  `wp --allow-root wc-misc skip-feature-flag set speaker_feedback 46`
1. Now try network activating the plugin. The "Leave Feedback" page should only be added on sites without the skip flag, and the speaker feedback UI should only be visible on those sites.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
@@ -21,18 +21,21 @@ const OPTION_KEY           = 'sft_feedback_page';
 const QUERY_VAR            = 'sft_feedback';
 const SUPPORTED_POST_TYPES = array( 'wcb_session' );
 
-// Only add actions to sites without the skip flag, and only if WC Post Types exist.
-if ( ! wcorg_skip_feature( 'speaker_feedback' ) && class_exists( 'WordCamp_Post_Types_Plugin' ) ) {
-	register_activation_hook( __FILE__, __NAMESPACE__ . '\activate' );
-	register_deactivation_hook( __FILE__, __NAMESPACE__ . '\deactivate' );
+register_activation_hook( __FILE__, __NAMESPACE__ . '\activate' );
+register_deactivation_hook( __FILE__, __NAMESPACE__ . '\deactivate' );
 
-	add_action( 'plugins_loaded', __NAMESPACE__ . '\load' );
-	add_action( 'init', __NAMESPACE__ . '\add_support', 99 );
-	add_action( 'init', __NAMESPACE__ . '\add_page_endpoint' );
-	add_action( 'rest_api_init', __NAMESPACE__ . '\register_rest_routes', 100 );
+add_action( 'plugins_loaded', __NAMESPACE__ . '\load' );
 
-	// Check if the page exists, and add it if not.
-	add_action( 'init', __NAMESPACE__ . '\add_feedback_page' );
+/**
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
```diff
@@ -55,16 +62,52 @@ function load() {
...
+		restore_current_blog();
+	}
+}
+
+/**
+ * The activation routine for a single site.
+ *
+ * @return void
+ */
+function activate_on_current_site() {
```
ryelle, line 110: I think this should also check for `can_load` or at least the `has_dependency` check — the Leave Feedback page was created on central in my testing.

ryelle (approved):
Looks good. One suggestion before merging, but otherwise it's working well 👍

public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
coreymckrill: I think there was a reason I didn't do the check here, but I can't remember it now...

🤔 Did you add the skip flag to Central before network activating? Theoretically that should prevent the page from getting created there...

ryelle: I didn't think to do that, since I assumed the lack of WC Post Types would prevent it — but if you'll set it on central on production, then it's fine.

PR #530, opened 2020-06-24
WCPT: Add application question, is event IRL/online

### LINKED ISSUE #498: WCPT: Add input to WordCamp application form that can prepopulate the "online only" setting
At this point most aspiring WordCamp organizers will know if their event is going to be "online only" when they fill out the application. Adding a checkbox to the application form that pre-populates the existing "Virtual event only" checkbox on the WordCamp post type helps streamline the process for both application vetters and for organizers once they are in pre-planning.

coreymckrill:
Adds a question, "Would this WordCamp be an in-person event or streamed online?" to the WordCamp application form, above the question about possible venues.

If an answer is not chosen that includes IRL, the possible venues question is hidden.

If the answer is chosen that the event will be online only, the "Virtual event only" postmeta value is set when the WordCamp application post is created.

Fixes #498 

### Screenshots

event-irl

event-online

1. In your local dev environment, go to the [WordCamp Application form](https://central.wordcamp.test/wordcamp-organizer-application/).
1. Fill out the form (most fields aren't required) and choose "It would be an online event" for the new question. The resulting WordCamp post should have the "Virtual event only" checkbox checked already.
1. Fill out the form again and choose one of the other options for the new question. The resulting WordCamp post should not have that checkbox checked.

### DIFF
--- a/public_html/wp-content/plugins/wcpt/views/applications/wordcamp/shortcode-application.php
+++ b/public_html/wp-content/plugins/wcpt/views/applications/wordcamp/shortcode-application.php
@@ -1321,6 +1321,46 @@ function render_wordcamp_application_form( $countries ) {
 
 				<div class="PDF_questionDivide" id="pd-divider-29"></div>
 
+				<div class="PDF_question" id="pd-question-35">
+					<div class="qNumber">
+						Q.35
+					</div>
+
+					<div class="qContent">
+						<div class="qText">
+							Would this WordCamp be an in-person event or streamed online?
+						</div>
+
+						<div class="PDF_QT400">
+							<ul>
+								<li>
+									<input type="radio" name="q_in_person_online"
+									       value="It would be an in-person event" id="q_in_person_online_1" />
+
+									<label for="q_in_person_online_1">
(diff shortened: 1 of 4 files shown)


public_html/wp-content/plugins/wcpt/views/applications/wordcamp/shortcode-application.php
```diff
@@ -1321,6 +1321,46 @@ class="survey-email required" autocomplete="off" required/>
...
 
+				<div class="PDF_question" id="pd-question-35">
+					<div class="qNumber">
+						Q.35
+					</div>
+
+					<div class="qContent">
+						<div class="qText">
+							Would this WordCamp be an in-person event or streamed online?
+						</div>
```
ryelle, line 1332: no change requested, but refactoring these forms to be not crowdsignal-format would also let us fix these accessibility issues… (none of these use labels, this should be a fieldset & legend, etc)

ryelle (approved):
looks good. would  be nice to take some time eventually to refactor these forms for better a11y & UX, maybe if a central redesign happens that could be a focus.

PR #532, opened 2020-06-25
CampTix Tweaks: Invalidate attendees shortcode cache on privacy change

coreymckrill:
When the "Hide from public attendees list" setting is changed for an attendee (either by them editing their ticket or by an admin editing the attendee post), all of the cached versions of the camptix_attendees shortcode content should be immediately invalidated so that someone doesn't have to wait up to 24 hours for their avatar to be hidden or shown in the attendees list.

Fixes https://meta.trac.wordpress.org/ticket/4370

1. Load the front end Attendees page on a test site to make sure its contents are cached. You can use something like Sequel Pro to check the options table in the database to make sure there is at least one `_transient_camptix-attendees-*` entry.
1. Edit an attendee post and toggle the checkbox for "Hide from public attendees list". After saving, the database entries should be gone.
1. Reload the Attendees page to regenerate the cache.
1. Now try editing an attendee via its Edit Link. Toggle the option for "Do you want to be listed on the public Attendees page?" and save. As before, the database entries should now be gone.

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/addons/privacy.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/addons/privacy.php
@@ -45,16 +45,21 @@ class Privacy_Field extends CampTix_Addon {
 			'no'  => _x( 'No', 'ticket registration option', 'wordcamporg' ),
 		);
 
-		// Registration field
+		// Registration field.
 		add_action( 'camptix_attendee_form_after_questions', array( $this, 'render_registration_field' ), 10, 2 );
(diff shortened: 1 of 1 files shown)


iandunn (approved):
LGTM.

I think we'll need to clear Super Cache too. Here's one way I've done that programmatically in the past:

https://github.com/iandunn/regolith/blob/10096e8/bin/wp-cli/multiple-use/purge-wp-super-cache.php#L17-L33

coreymckrill:
Ah good call. Added clearing for WP Super Cache here: 3ba949c2

PR #534, opened 2020-06-28
Better UX for payment requests

### LINKED ISSUE #533: Vendor payment request form: use select2 for all country fields
_[Original ticket in Trac](https://meta.trac.wordpress.org/ticket/4303)_

In vendor payment request form there are few country fields that are set up to use select2 for better UX (select2 has a search). There are also country select fields that do not use select2. Select2 should be enabled to all country select fields.

timiwahalahti:
Payment requests; vendor payment and reimbursement requests do have select2 for some select fields but not for all. PR fixes that by adding select2 for all select field in payment request forms. At the same time, the width of all fields is normalised - some fields were wider than others previously. These improve the UX on forms.

Fixes #533 

### Screenshots

**Old vendor payment view**

1. Log in to some WC site and go to the reimbursement request
2. Change the payment method selection and see that country has always select2
3. Add second expense and see that category has select2
4. Go to create a new vendor payment and enjoy form fields that are aligned

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/css/wordcamp-budgets.css
+++ b/public_html/wp-content/plugins/wordcamp-payments/css/wordcamp-budgets.css
@@ -71,14 +71,6 @@ body.post-type-wcb_sponsor_invoice #save-action {
 	color: #dc3232;
 }
 
-/* Make room for the required field indicator (*) */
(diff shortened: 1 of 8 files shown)


ryelle (approved):
Looks good 👍

PR #535, opened 2020-06-28
Ignore Query Monitor plugin

timiwahalahti:
At least I like to use [Query Monitor](https://wordpress.org/plugins/query-monitor/) during development, but having it on production isn't necessary or even recommended. Adding it to `.gitignore` prevents the unintentional commits with QM included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,11 @@ public_html/wp-content/debug.log
 
 .docker/test_suite
 
+#
+# Dev tools
+#
+public_html/wp-content/plugins/query-monitor/
+
 
 #
 # Plugins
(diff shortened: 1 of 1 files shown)


ryelle:
Adding the ignore rule makes sense to me, but pinging @iandunn or @coreymckrill in case you have other thoughts.

coreymckrill:
The only concern I have is that by putting it in the middle of the list of plugins we're ignoring because they are third party and installed via composer, it kind of implies that it is also getting installed by composer. Maybe add a separate section to the gitignore file for "Dev tools" and add it there?

timiwahalahti:
Created a new section and moved Query Monitor ignore line there.

coreymckrill:
Looks good to me!

Issue #372, opened 2020-02-18
Event Applications: Review fields & metabox display for usefulness


ryelle:
WordCamp & Meetup applications create new posts in wp-admin when the form is submitted. These posts expose some of the application fields as inputs in the editor, but not all of them. For example, partial screenshots:

**WordCamp**
![Screen Shot 2020-02-18 at 5 46 53 PM](https://user-images.githubusercontent.com/541093/74784801-bdabe880-5276-11ea-9d65-3dffd1a7992c.png)

**Meetup**
![Screen Shot 2020-02-18 at 5 47 58 PM](https://user-images.githubusercontent.com/541093/74784841-d61c0300-5276-11ea-9bf3-53973628fa1d.png)

But application reviewers know this is is a subset of data, so they use the "Original Application" data dump:
 
![Screen Shot 2020-02-18 at 5 50 55 PM](https://user-images.githubusercontent.com/541093/74785020-488ce300-5277-11ea-9fa2-50978e58ba18.png)

We could make this a nicer experience:

- Better formatting of the data dump?
- Updating the fields in the other metaboxes to show all the full set of application data?

Would need to know which fields are not currently used, which fields should be populating from the application, and which fields are missing (in the application data, but not currently in the form UI).

timiwahalahti:
> Updating the fields in the other metaboxes to show all the full set of application data?

Most of the data in applications aren't needed anywhere else than in the vetting process, so it does not really make sense to fill other metaboxes with one-use-only fields.

> Better formatting of the data dump?

👍👍👍

> Would need to know which fields are not currently used, which fields should be populating from the application, and which fields are missing (in the application data, but not currently in the form UI).

I can make a list of these for Community Team deputies to review and share their opinions. After we have agreed about fields, I could also work on PR.

ryelle:
> I can make a list of these for Community Team deputies to review and share their opinions

That would be great!

timiwahalahti:
Fixed in #536 so closing.

PR #536, opened 2020-06-28
Application tracker improvements

### LINKED ISSUE #519: Application tracker improvements
Some improvements to application tracker could be made. From [this P2 discussion](https://make.wordpress.org/community/2020/04/04/request-for-feedback-revising-application-tracker-back-end/), at least:

- Improving the “Original application” section and making it friendlier to read
- Moving the “Add private note” text field from sidebar or adding a second one in the main area to make writing long notes, for example, vetting notes easier
- Removing the “Already a meetup” field from “Application information” metabox on Meetup applications, because same data is visible on the original application and all other fields in the metabox are for tracking the steps for application
- Removing the “Swag notes” text field because swag related information can be added as a private note, which is better because those have always a timestamp and trace of the user who made the note. Existing swag notes would be migrated to log.

It would also make sense to fix  [4028](https://meta.trac.wordpress.org/ticket/4082) at the same time.

timiwahalahti:
Improve WordCamp and Meetup application tracker UX. 

Fixes #519, #372 

### Remove Meetup Swag notes metabox

Swag related information belongs to log instead of its own text field, because over time Meetup can get more swag related notes (as we send those out at request regularly). The log is better also because there the note has always a timestamp and trace of the user who made the note.

**NOTE! Before merging this PR to production, all existing swag notes need to be migrated automatically to part of the log.** I can help with tooling for that if needed.

### "Already a meetup" field

Remove "Already a meetup field" from Meetup applications, because same data is visible on the original application and all other fields in the metabox are for tracking the steps for application.

### Move note metabox to the main column and make it first

Sometimes, especially when doing a vetting for application, the deputy needs to write a long note. Having a small text field on sidebar does not make it easy, so move the metabox and text field to main column. Because sometimes there's need to just quickly add a short note, make the metabox as a first in view.

### Improving the "Original application" section

The original application is now displayed as a table and question key is sanitized to more human readable format.

1. Log in to Central
2. Visit some Meetup application and confirm changes described above
3. Visit some WordCamp application and confirm changes described above


timiwahalahti:
### Improving the "Original application" section

Showing basically a `var_dump` of all fields sent in the application form isn't the friendliest way to present all the information. Especially in WordCamp applications, which do contain lots of information.

_But_, how the presentation could be improved? Currently, there's no place that maps the field keys to their labels and this has been done manually in some places (like in [`wcpt-wordcamp/class-wordcamp-application.php`](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/class-wordcamp-application.php#L223-L226)).

Doing some kind of place for mapping and then changing every place to use that mapping would sound like cleanest way, but it's probably also the hardest and most complex way. It's a nice idea, but maybe something for the time when application forms are changed and unified on a larger scale?

So, should we do a manual mapping in the metabox for now? I'd argue that applications forms do not change that often, so it isn't much of a task to add a new field to metabox if something gets changed.

Other and laziest solution would be just using the field key/name as a label (same as now), but displaying the fields in the table to make it easier to perceive the content. This would leave us with field keys/names that aren't so descriptive.

Thoughts? Ideas?

ePascalC:
@timiwahalahti I have no view on how it looks today, but I would first create an array with the parts you know that holds the arraykey, the label and the type of field, i.e.:
"fname","First Name","input"
"lname","Last Name","input"
"bio","Biography","textbox"

Looping the above array, you check what keys are available in the data array and show them on screen.
After that you loop the data array and show the ones you do not have defined in the above array and show those in a standard input box.

timiwahalahti:
@ePascalC Yeah, that was my first thought - making an array that would define the display both on the application form and tracker side. But the way how forms and their handling/saving is made now, would mean many things to be re-written on a larger scale.

During last bug scrub, the team [pointed out](https://wordpress.slack.com/archives/C08M59V3P/p1593104939469300) that forms in their current state are quite... horrible. Those do work, but at some point (maybe when central redesign happens), re-coding those should be considered.

Maybe creating a field array for this need would be a good starting point for that at some point? Tho we don't know how the re-coded forms would be done, so the array might be unusable there.

timiwahalahti:
[Related Slack discussion](https://wordpress.slack.com/archives/C08M59V3P/p1594316723088500)

timiwahalahti:
The original application is now displayed as a table and question key is sanitized to more human readable format in a way that @iandunn suggested.

ryelle (approved):
Looks good to me 👍 

I wrote up a quick script to move existing meetup swag notes to the Log, with the prefix "[Migrated from Swag Notes]", I'll run that right after merging.

PR #538, opened 2020-07-01
Speaker Feedback: Tweak rewrite handling during plugin activation

coreymckrill:
Calling `flush_rewrite_rules` within the context of `switch_to_blog` is problematic because it's not using an accurate site configuration, since it doesn't change which plugins are activated, etc. Instead, if we just delete the current rewrite rules from the database, WP will automatically regenerate them the next time there is a request to the site.


iandunn (approved):
LGTM 👍

PR #544, opened 2020-08-06
Coming Soon Page: If WordCamp cancelled, replace the date with "Cancelled" text and remove Jetpack subscription form

### LINKED ISSUE #543: Hide WordCamp dates and Jetpack subscription in Coming Soon page if WordCamp is Cancelled [COVID-19 response]
[Discussion in Slack](https://wordpress.slack.com/archives/C037W5S7X/p1596578955116500).

With a handful of WordCamps cancelled that had sites created, they still have active landing pages since they are set in “Coming Soon” mode. Community Team does not want to take those sites offline for now.

The first thing you see when you go to the site is the date - I’d (Cami Kaos) like to see that gone as it’s not uncommon for folx to not read past the super large print of the date at the top. That also makes sense. or maybe the date becomes “Canceled” since it is such large text.

While we are at it, Jetpack subscription form should be removed as well because Cancelled camp isn't going to send any updates.

timiwahalahti:
COVID-19 situation cancelled many WordCamps that were already in pre-planning and had a site in Coming Soon mode. Now those pages aren't really reflecting the state of the planning anymore. Community Team would like to those sites exists for a while at least, but for that content needs to be updated.

To reduce the amount of work that needs to be done for those sites pragmatically hiding the event dates and replacing it with "Cancelled" text string and removing Jetpack subscription module based on Central `wordcamp` CPT helps a lot. After that only the introduction text needs to be manually changed and possibly the page with contact form reverted to draft (which hides the form from Coming Soon page).

Fixes #543

Props @brandondove, @camikaos, courtneypk

1. Change some existing WordCamp status to "Cancelled" on Central tracker
2. Enable "Coming Soon" mode on that WordCamp
3. Go to the WordCamp website

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
+++ b/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
@@ -279,6 +279,10 @@ class WordCamp_Coming_Soon_Page {
 	 * @return string|false
 	 */
 	public function get_dates() {
+		if ( 'wcpt-cancelled' === $this->get_status() ) {
+			return esc_html__( 'Cancelled', 'wordcamporg' );
+		}
+
 		$dates         = false;
 		$wordcamp_post = get_wordcamp_post();
 
--- a/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
+++ b/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
@@ -173,14 +173,14 @@ class WordCamp_Coming_Soon_Page {
 	 */
 	public function get_template_variables() {
 		$variables = array(
-			'image_url'								=> $this->get_image_url(),
-			'background_url'					=> $this->get_bg_image_url(),
-			'dates'										=> $this->get_dates(),
-			'active_modules'					=> Jetpack::$instance->get_active_modules(),
-			'contact_form_shortcode'	=> $this->get_contact_form_shortcode(),
-			'colors'									=> $this->get_colors(),
-			'introduction'						=> $this->get_introduction(),
-			'status'									=> $this->get_status(),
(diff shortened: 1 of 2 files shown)


iandunn:
> To reduce the amount of work that needs to be done for those sites pragmatically hiding the event dates and replacing it with "Cancelled" text string and removing Jetpack subscription module based on Central wordcamp CPT helps a lot. After that only the introduction text needs to be manually changed and possibly the page with contact form reverted to draft (which hides the form from Coming Soon page).

What do you think about having two template parts for the page instead?

1. One for camps that aren't cancelled, which shows everything it normally does today.
1. A new one for camps that _are_ cancelled, which only show a message that the camp has been cancelled, and maybe links to Central to find another nearby camp.

That way nobody will have to update the custom messages, draft the contact page, etc. The visitor would just automatically get a totally different page, which only has content that's designed for that situation.

brandondove:
@iandunn I like that approach. The only potential downside is that the message to the camp's attendees is less personalized.

iandunn:
Ah, ok, I didn't realize deputies were writing personalized cancellation messages for each camp. We could keep that part in both templates if y'all want.

brandondove:
It wouldn't be terrible to have a standard message here by default but allow it to be changeable if the org puts in some content.

iandunn:
Yeah, that wouldn't be too hard. We could add a new `cancellation message` field in the customizer, but default to a general one if it's empty.

timiwahalahti:
> A new one for camps that are cancelled, which only show a message that the camp has been cancelled, and maybe links to Central to find another nearby camp.

The reasoning behind why I didn't go that way is the custom messages as Brandon already pointed out, but as well that in some situations organizers might want to leave the contact form in place. Especially when looking at situations other than current COVID-19 response.

All WordCamps that are cancelled because of COVID-19 have been already manually modified so that the contact form is not visible and default content should have been replaced as well.

We could definitely add a new field for custom cancellation message and show default message if there's no content. Thought we would need to go all currently cancelled camps thru again and copy the few custom messages into that field.

iandunn:
Ah, those are good points 👍

PR #548, opened 2020-08-07
Coming Soon mode: use edit_theme_options as required capability instead of administrator role

### LINKED ISSUE #547: Trusted deputy not able to see Coming Soon mode settings
When the user is a trusted deputy, they are not able to see or modify Coming Soon mode settings.

Settings aren’t visible when using wordpress.org proxy (which gives trusted deputy capabilities) but turning off the proxy fixes that. [See Slack discussion](https://wordpress.slack.com/archives/C037W5S7X/p1596825593350200?thread_ts=1596581360.144100&cid=C037W5S7X).

**Expected behavior**

Trusted deputy should be able to see Coming Soon mode settings in the customizer.

timiwahalahti:
When the user is a trusted deputy, they are not able to see or modify Coming Soon mode settings because capability check was using `administrator` role instead of actual capability. Change that to `edit_theme_options` so trusted deputies can access the settings without adding themselves to the site with administrator role.

Fixes #547

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wccsp-settings.php
+++ b/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wccsp-settings.php
@@ -8,7 +8,7 @@
 
 class WCCSP_Settings {
 	protected $settings;
-	const REQUIRED_CAPABILITY = 'administrator';
+	const REQUIRED_CAPABILITY = 'edit_theme_options';
 
 	/**
 	 * Constructor.


iandunn:
Huh, trusted deputies should already have all of the admin caps on every individual site. Maybe there's a bug in that?

https://github.com/WordPress/wordcamp.org/blob/4fdfee4/public_html/wp-content/mu-plugins/trusted-deputy-capabilities.php#L6-L9

https://github.com/WordPress/wordcamp.org/blob/4fdfee4/public_html/wp-content/mu-plugins/trusted-deputy-capabilities.php#L35

timiwahalahti:
I'm able to access all other pages that do need administrator capabilities, expect the coming soon one 🤔

ryelle:
I think the `administrator` cap is added dynamically, not as part of the `administrator` role by default - so when `$allcaps` is [overwritten by the administrator role directly,](https://github.com/WordPress/wordcamp.org/blob/4fdfee4/public_html/wp-content/mu-plugins/trusted-deputy-capabilities.php#L35) it's missing these capabilities. The last 3 are obviously added from function on the `user_has_caps` hook, I'm not totally sure where administrator comes from.

```
administrator
install_languages
resume_plugins
resume_themes
```

ryelle (approved):
I think it's fine to make this change so things can work, but the [root of the problem](https://github.com/WordPress/wordcamp.org/pull/548#issuecomment-675037830) should be addressed, so either #547 should be left open, or another issue opened for following up.

PR #549, opened 2020-08-14
Plugins: Add `permission_callback` to API registration

ryelle:
As of WP 5.5, API endpoints must set a permission_callback. If the endpoint is public, this can be `__return_true`. In other cases, the endpoint was being checked in the main callback, or before being registered, so that functionality was moved into the permission_callback.

1. Enable WP_DEBUG
2. Try requesting posts: `/wp-json/wp/v2/posts` (or editing a post in GB)
3. On the production branch, this triggers errors
4. This branch, no errors
5. Try requesting each endpoint changed, you should see no errors. Use [basic-auth plugin](https://github.com/WP-API/Basic-Auth) to test logged in access.

```
POST [city site]/wp-json/wc-post-types/v1/email-fav-sessions # public
GET [city site]/wp-json/wordcamp-site-cloner/v1/sites # private, admin-only

GET https://wordcamp.test/wp-json/wordcamp-letsencrypt/v1/domains-dehydrated # public, key check happens in callback
GET/POST https://wordcamp.test/wp-json/wordcamp-qbo/v1/expense # private, checks header authorization
GET/POST https://wordcamp.test/wp-json/wordcamp-qbo/v1/invoice
GET https://wordcamp.test/wp-json/wordcamp-qbo/v1/invoice_pdf
GET https://wordcamp.test/wp-json/wordcamp-qbo/v1/paid_invoices
```

Searching the codebase for `register_rest_route` also brings up [wordcamp-reports,](https://github.com/WordPress/wordcamp.org/blob/6f777b8613ec270f013d1b8273304ab65d38b5c8/public_html/wp-content/plugins/wordcamp-reports/index.php#L337) but it seems none of the reports register an endpoint.

### DIFF
--- a/public_html/wp-content/mu-plugins/lets-encrypt-helper.php
+++ b/public_html/wp-content/mu-plugins/lets-encrypt-helper.php
@@ -27,8 +27,9 @@ class WordCamp_Lets_Encrypt_Helper {
 			'wordcamp-letsencrypt/v1',
 			'/domains-dehydrated',
 			array(
-				'methods'  => 'GET',
-				'callback' => array( __CLASS__, 'rest_callback_domains_dehydrated' ),
+				'methods'             => 'GET',
+				'callback'            => array( __CLASS__, 'rest_callback_domains_dehydrated' ),
+				'permission_callback' => '__return_true',
 			)
 		);
 	}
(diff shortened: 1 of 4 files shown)


coreymckrill (approved):
👍 Looks good to me!

I tested some of the endpoints locally using Insomnia. Then to test a QBO endpoint I had to go through the UI flow, but it seems to be working.

Just noting that REST API responses in Insomnia are still showing a bunch of notices related to the newly required `permission_callback`, but they're all related to Jetpack, so as soon as we get that updated, those should go away.

Issue #553, opened 2020-08-20
WordCamp.org has a mix of Canceled and Cancelled


timiwahalahti:
_[Original Track ticket](https://meta.trac.wordpress.org/ticket/4925)_

While translating I found instances of both Canceled and Cancelled in the wordcamp.org translations. As en_US is the base locale these should all be Canceled.

PR #554, opened 2020-08-20
Changes occurrences of Cancelled to Canceled

CdrMarks:
Changes occurrences of Cancelled to Canceled, regardless case.

Fixes 553.

After implementing this, it will be necessary to update existing WordCamp statuses from `wcpt-cancelled` to `wcpt-canceled`.  Public statuses may also need to be changed.


CdrMarks:
Looks like I'm unable to apply any labels such as Components here.  Can I be added as a contributor, please?

ryelle:
I think we wanted this to only change the instances a user will see, so that we don't need to migrate statuses or worry about other edge cases. Can you update the PR to so that it only changes "cancelled" in user-facing strings?

(and you should have gotten an invite to be a contributor, let me know here or on slack if you have any questions!)

CdrMarks:
@ryelle I've update the PR.  Let me know if you see any other changes that need to be made.

timiwahalahti:
Fixes #553

pkevan:
It still seems like there are some code strings which are being altered, we should fix those before merging to avoid unnecessary breakage.

PR #555, opened 2020-08-21
Disable Jetpack Simple Payments feature

timiwahalahti:
Jetpack has a Simple Payments feature that registers a block and widget for adding PayPal/Stripe payment button. This feature isn't really needed on WordCamp.org as we sell only tickets with Camptix.

The feature is causing some issues when saving posts, so the easiest thing is just to disable it completely.

See https://wordpress.slack.com/archives/C08M59V3P/p1597998977062100

1. Go to block editor and try to search "Payments" block - shouldn't be there anymore

### DIFF
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
@@ -8,7 +8,7 @@ defined( 'WPINC' ) || die();
 // Allow Photon to fetch images that are served via HTTPS.
 add_filter( 'jetpack_photon_reject_https',    '__return_false' );
 
-// Disable Simple Payments feature
+// Disable Simple Payments feature.
 add_filter( 'jetpack_disable_simple_payments', '__return_true' );
 
 /**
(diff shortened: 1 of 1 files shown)


ryelle:
I still see the Payments block. I think this one is coming from the recurring payments feature, and there doesn't seem to be a way to disable that block/feature.

timiwahalahti:
Mmmh. This must be one of the features that can be tested only when Jetpack is really connected and/or have activated with a right plan because I can't see nor test the block on dev env.

The error message posted on Slack was saying `_spay_email` and the only instance where I can [find references](https://github.com/Automattic/jetpack/search?q=spay_email&unscoped_q=spay_email) for that are in simple pay module.

Though I'm a bit curious why the error message contains that meta field key at the first place, because the page definitely didn't contain any Jetpack block. And you're right that "Payments" block is recurring payments, not the simple payments module - which I can't actually see even on production. I'm bit confused right now, to be honest 😅

Maybe it was just a glitch in a matrix or something.

coreymckrill:
I haven't confirmed this, but I recall seeing `_spay_email` as a field in REST responses for posts. I wonder if Jetpack is adding that field regardless of whether Simple Payments is being used...

timiwahalahti:
Going to close this PR as no new reports of the issue caused by `_spay_email` field reported.

PR #556, opened 2020-08-21
Fix speaker meta save

timiwahalahti:
[See Slack discussion for full bug description](https://wordpress.slack.com/archives/C08M59V3P/p1597998977062100).

Looks like some `register_post_meta` calls were missing `auth_callback`. That caused issues with saving speaker username and email. Additionally, the `meta_auth_callback` just checked if the meta key contains `_wcpt_` but some new meta fields are prefixed with `_wcb_` and thus weren't authorized to edit by organizers.

Props nukaga

1. Without this patch, try to update the speaker WordPress.org username. You should end up with error similar to "Updating failed. Sorry, you are not allowed to edit the _wcpt_user_name custom field."
2. Apply patch and try to update the speaker WordPress.org username again. This time it should work.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -45,11 +45,12 @@ function register_speaker_post_meta() {
 		'wcb_speaker',
 		'_wcpt_user_id',
 		array(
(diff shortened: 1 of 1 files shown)


ryelle (approved):
Thanks for finding this!

PR #568, opened 2020-09-08
Format local dates instead of converting a second time

### LINKED ISSUE #559: Schedule block shows dates 1 day in the past
**Describe the bug**

We're using the [schedule block on our site](https://colombia.wordcamp.org/2020/agenda/), but dates inside this block appear to be 1 day in the past. For instance, our sessions are on Sept 10, 11 and 12, but the schedule block displays them as being on Sept 09, 10 and 11.
The site timezone is correctly configured.

**To reproduce**

1. Create a session post on our site with a given date.
2. Create a page using the schedule block.
3. The session will appear as if it was scheduled 1 day in the past.

**Expected behavior**

Session time in session posts should match what the schedule block displays.

**Screenshots**

- Session time meta for one of our sessions:
  Screen Shot 2020-08-22 at 12 22 13
- Session as it appears on the schedule page:
  Screen Shot 2020-08-22 at 12 21 50

**WordCamp**

https://colombia.wordcamp.org/2020/

 - Device: Laptop
 - OS: macOS 10.15.6
 - Browser: Chrome
 - Version: 84.0.4147.135

iandunn:
_This shouldn't be merged until we install the version of Gutenberg that includes both https://github.com/WordPress/gutenberg/pull/25083 and https://github.com/WordPress/gutenberg/pull/22866._

---

Fixes #559

The dates are _already_ in the site timezone, so `dateI18n()` et al add the UTC offset a _second_ time. In some cases that would cross a date boundary, and the wrong date would be displayed.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
@@ -1,7 +1,7 @@
 /**
  * WordPress dependencies
  */
-import { dateI18n } from '@wordpress/date';
+import { dateI18n, format } from '@wordpress/date';
 import { __, _x } from '@wordpress/i18n';
 import { createContext, useContext } from '@wordpress/element';
 import { decodeEntities } from '@wordpress/html-entities';
@@ -49,7 +49,7 @@ export function ScheduleGrid( { sessions } ) {
 		scheduleDays.push(
 			<ScheduleDay
 				key={ date }
-				date={ date }
+				localDate={ date }
 				sessions={ sessionsGroup }
 			/>
 		);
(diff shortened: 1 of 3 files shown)


public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
```diff
@@ -94,17 +94,17 @@ function groupSessionsByDate( sessions ) {
  * Render the schedule for a specific day.
  *
  * @param {Object} props
- * @param {string} props.date     In a format acceptable by `wp.date.dateI18n()`.
- * @param {Array}  props.sessions
+ * @param {string} props.localDate The date being displayed (in the site's timezone).
```
ryelle:
```suggestion
 * @param {string} props.localDate The day being displayed in Ymd format (in the site's timezone).
```

Just to clarify that this isn't a timestamp.

ryelle (approved):
This looks good 👍 

One docs suggestion, otherwise good to merge.

public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
iandunn: 👍🏻

PR #570, opened 2020-09-09
Participation: Sync badge status when the user ID field changes 

### LINKED ISSUE #569: Participation Notifier: Organizers and Speakers aren't getting badges
**Describe the bug**

The participation notifier is what tells profiles.w.org to give badges to speakers & organizers. It watches these posts in WordCamp sites, and when they're updated, it sends a request to update the user profile. After #512, Speakers & Organizers were updated to use Gutenberg, and this code can't find those user IDs anymore.

The code used to add badges to organizers & speakers expects the user ID to be in a $_POST variable, but this it not set in Gutenberg. https://github.com/WordPress/wordcamp.org/commit/e406a0a523735f2414005ba98ce0f5d835c3457b / https://github.com/WordPress/wordcamp.org/pull/125 started to fix this, but still relies on that POST value.

ryelle:
Instead of relying on post updates, which work differently in Gutenberg (and don't provide the new values in $_POST), this watches the user ID meta field directly for add, update, and delete actions, and syncs with the user's profile (and previous user's profile) to add or remove badges.

Fixes #569 

1. Install [Log HTTP Requests](https://wordpress.org/plugins/log-http-requests/) to make sure the requests are sent to `https://profiles.wordpress.org/wp-admin/admin-ajax.php`
2. Add or update a speaker with a valid user on your site
3. After saving, you should see at least one request in the HTTP log
    If you added a new speaker, you'll see two requests, one with `body.action: wporg_handle_activity`, and the other should match this:
    
    ```
    body: {
    	action: wporg_handle_association
    	command: add
    	user_id: [the speaker's ID]
    	association: wordcamp-speaker
    }
    ```

    If you change an existing speaker user, you'll see up to 3, the above 2 and a third that should match

    ```
    body: {
    	action: wporg_handle_association
    	command: remove
    	user_id: [the previous speaker's ID]
    	association: wordcamp-speaker
    }
    ```
4. Check the same with organizers, this should have the same outcome as above with `wordcamp-organizer` as the association.
5. Try deleting the username on a post, it should trigger a single `remove` request.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -41,6 +41,7 @@ class WordCamp_Post_Types_Plugin {
 
 		add_filter( 'updated_post_meta', array( $this, 'update_wcorg_user_id' ), 10, 4 );
 		add_filter( 'added_post_meta', array( $this, 'update_wcorg_user_id' ), 10, 4 );
+		add_filter( 'deleted_post_meta', array( $this, 'delete_wcorg_user_id' ), 10, 3 );
 
 		add_filter( 'manage_wcb_speaker_posts_columns', array( $this, 'manage_post_types_columns' ) );
 		add_filter( 'manage_wcb_session_posts_columns', array( $this, 'manage_post_types_columns' ) );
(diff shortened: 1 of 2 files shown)


iandunn (approved):
LGTM 👍🏻

PR #666, opened 2020-10-14
Fix special chars in "favorite sessions" emails

### LINKED ISSUE #665: Text rendering issues in "Favorite sessions" emails when special characters are present
Hi there,
I am a member of the organizing team for the upcoming WordCamp Italy. We have discovered an issue regarding the "Favorite sessions" feature.
In the "Favorite sessions" emails, there's a text rendering issue occurring when session titles contain special characters like `'`, `-`, `&` because they are translated to the corresponding HTML entities and look ugly and unreadable in the (plain-text) emails.

**To reproduce**

1. Create a session that includes special character like `'`, `-`, `&`
2. In the frontend schedule page select the session and request to receive the list via email
3. Check the email received. The characters were converted to HTML entities, but since the email is in plain text format, they are not rendered.

**Expected behavior**

Any special character should be converted back to plain text since these emails are plain text and not HTML.

**Suggested solution**

An easy fix could be to use the `htmlspecialchars_decode` function when printing the session title.
Alternatively, it would be possible to avoid the application of the `the_title` filter to the session title in the function `generate_plaintext_fav_sessions`.
At the moment we adopted the workaround of avoiding such characters whenever possible to prevent the replacement made by `wptexturize` via the `the_title` filter (i.e. using `’` in place of `'`), but it's not ideal.

**Screenshots**

image

**WordCamp**

This seems to be a general issue. Our page is here https://2020.italia.wordcamp.org/schedule (not yet published).

 - Device: any
 - OS: any
 - Browser: any
 - Version: any

marcochiesi:
This change applies `html_entity_decode` to the session titles in the "Favorite sessions" email messages.

Fixes #665 

Props marcochiesi

1. Create a session that includes special character like `'`, `-`, `&`
2. In the frontend schedule page select the session and request to receive the list via email
3. Check the email received. The HTML entities, which previously were not rendered, should now be ok.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/favorite-schedule-shortcode.php
@@ -337,7 +337,7 @@ function generate_plaintext_fav_sessions( $sessions_rev, $fav_sessions_lookup ) 
 			}
 
 			$session              = get_post( $session_id );
-			$session_title        = apply_filters( 'the_title', $session->post_title );
+			$session_title        = html_entity_decode( apply_filters( 'the_title', $session->post_title ) );
 			$session_tracks       = get_the_terms( $session_id, 'wcb_track' );
 			$session_track_titles = is_array( $session_tracks ) ? implode( ', ', wp_list_pluck( $session_tracks, 'name' ) ) : '';
 


ryelle (approved):
Works in my testing, thanks for the PR!

PR #667, opened 2020-10-23
Use specific CSS class for slides link container in session posts

enrico-sorcinelli:
This simple PR aims to use a specific class for session's slides link container. Currently the same `session-video` class is used both for slides and  video links in session pages.
 

Props enrico-sorcinelli

1. Create a session
2. Add a slide link to session info metabox 
3. See new CSS class name in generated HTML page.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -859,7 +859,7 @@ class WordCamp_Post_Types_Plugin {
 			return $content;
 		}
 
-		$session_slides_html  = '<div class="session-video">';
+		$session_slides_html  = '<div class="session-slides">';
 		$session_slides_html .= sprintf( __( '<a href="%s" target="_blank">View Session Slides</a>', 'wordcamporg' ), esc_url( $session_slides ) );
 		$session_slides_html .= '</div>';
 


public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -859,7 +859,7 @@ public function add_slides_info_to_session_posts( $content ) {
 			return $content;
 		}
 
-		$session_slides_html  = '<div class="session-video">';
+		$session_slides_html  = '<div class="session-slides">';
```
iandunn, line 862: 🤔 , changing this breaks backwards compatibility. A few options:

1) Do it anyway, since using `video` for the slides was a bug.
2) Add additional `slides` class to the slides link, leaving the `video` class. Organizers could use something like `:not( '.session-slides' )` to target the `video` class only.
3) Add additional classes to both links, e.g., `session-slides-container` and `session-video-container`. Then add HTML comments letting organizers know the old ones are deprecated.

I don't think `#2` is good, since it's unintuitive, and extra work for organizers. `#1` would require manually reviewing the CSS for all past camps, and updating if necessary, or letting those styles break. So, I think `#3` is the best option.

iandunn (changes requested):

pkevan:
I think @iandunn's review on this is still valid, so if we fix this it should be with additional CSS classes added, but since there has been no updates in over 3 years, lets close it - it can be reopened with changes, should that be done.

PR #673, opened 2020-12-04
Blocks: Fix compatibility issues with WordPress 5.6/Gutenberg 9.5

ryelle:
I was testing Gutenberg 9.5 on wordcamp.org earlier today and found some date & api bugs.

- Dates: Moment had a wider range of permitted date formats, but now that wp.date uses date-fns, these unsupported dates fail and crash the block. Seen on the frontend, a page with the schedule block.
- API: At some point, the requests made by the schedule block stopped being understood by `getEntities`/`getEntityRecords`, and now always return null. Seen in the editor, the schedule block never stops "loading" even though the requests finish.

The fix for the first is straightforward, I don't think it has any unexpected side-effects. The second issue is "fixed" for now by removing `_fields`, in the interest of getting something working. I'll continue investigating the root issue, but this could be merged if we need a quick fix.

Update gutenberg to 9.5 (for the date issue). The API issue is visible on production currently (WP 5.6).

1. Have sessions on your site
2. Add a schedule block to a page
2. View the block in the editor, it should work.
3. Publish the page, the schedule should appear correct on the frontend
4. The dates & times should appear correct

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/data.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/data.js
@@ -93,29 +93,10 @@ const fetchScheduleData = ( select, editorContext ) => {
 	}
 
 	// These must be kept in sync with `get_all_sessions()`.
-	const sessionArgs = {
-		/*
-		 * This doesn't include `session_cats_rendered` because we already need the category id/slug in other
-		 * places, so it's simpler to have single source of truth.
-		 */
-		_fields: [
-			'id',
-			'link',
-			'meta._wcpt_session_time',
-			'meta._wcpt_session_duration',
-			'meta._wcpt_session_type',
-			'session_category',
-			'session_speakers',
-			'session_track',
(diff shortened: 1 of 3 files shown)


iandunn:
> Moment had a wider range of permitted date formats, but now that wp.date uses date-fns, these unsupported dates fail and crash the block

That seems like a back-compat break w/ `wp.date` to me. What do you think about fixing that upstream instead of here?

ryelle:
Looks like the date issue will be fixed in GB 9.6, since they've reverted the moment/date-fns switch: https://github.com/WordPress/gutenberg/pull/27398

I spent some time Friday trying to figure out what exactly the fields issue is- turns out it is a core bug, so I've got a PR for that up here: https://github.com/WordPress/gutenberg/pull/27526 — this workaround can go in for now, and if/when that's merged we can bring back the fields subset.

I'm going to merge this so we can jump up to 9.5, and since the `_fields` issue is happening on production already. When 9.6 comes out, I'll check dates again and update.

PR #674, opened 2020-12-07
Tools: Move from travis to github actions

ryelle:
This also fixes the node version to 12.x, until we can update packages for 14.x compatibility

1. The tests on this PR should run and pass


CdrMarks:
Is this PR related to how the code is tested once in GitHub?  Will this change how our local environments function?

ryelle:
@CdrMarks This only matters for github, it will not change anything for local environments.

PR #676, opened 2021-01-28
Add report class for getting meetup group attendee counts

### LINKED ISSUE #675: Meetup events: add attendance count to results
I'm working on gathering information about past Meetups for annual reports. [2017 annual report](https://wordpressfoundation.org/2017-annual-report/) had an attendee count included in the numbers reported and I think that's a really nice number to have and compare.

Currently, there's no way to get that number from reports tool. Meetup.com do have their own Pro dashboard from where you can get these numbers out surely, but it would be neat to have the attendance count (attended members) included on our reports as Meetup.com API do return that information.

timiwahalahti:
Adds a new class to reporting tools for getting the number of attendees by country and group. Does not contain public reports.

Fixes #675

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-meetup-attendees.php
+++ b/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-meetup-attendees.php
@@ -1,0 +1,570 @@
+<?php
+/**
+ * Meetup Groups.
+ *
+ * @package WordCamp\Reports
+ */
+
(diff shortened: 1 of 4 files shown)


timiwahalahti:
Since some parts need to communicate with meetup.com API have been moved to `wporg-mu-plugins`, which does not get included in the Docker at the moment (see #769), can't test and refresh this PR.

Since the meetup.com switched to GraphQL API after this PR was originally opened, some work to reflect those changes are probably needed.

PR #695, opened 2021-09-08
Blocks: Add a new "Avatar" block to use in Query Loop

### LINKED ISSUE #180: Post Types: Allow organizers to override speaker avatars
Anecdotally this has come up a few times - speakers either don't have a gravatar account, or want to use a different headshot for the WordCamp site. Currently, organizers don't have a good way of overriding that.

One option could be enabling the featured image, but themes might use that on the page in general, causing awkward just unexpected layouts, where maybe the old gravatar is on the page along with the new featured image.

Another idea could be a new custom field to upload an image, which would ensure there's no theme compat issues, but a bit more work up front.

ryelle:
The Query Loop block was introduced in WordPress 5.8. This block renders a list of posts (of any type), and gives the editor fine control over the layout and details. The WC post types (ex, speakers) can be rendered in this block, but it misses some of the CPT-specific info; there is no block for the user's avatar.

This PR adds a new block for Avatar, so that wordcamp organizers can create speaker and organizer pages using the Query Loop block, for finer control than the existing Speakers & Organizers blocks. This block is modeled after the featured image block, allowing users to pick from a set of avatar sizes, and supports the duotone feature.

Additionally, I added featured image support to speakers & organizers, and if a featured image is found, it will override the gravatar (fixes #180). This does not put any extra work on organizers unless they want it, since gravatars are still the default (see `WordCamp\Post_Types\Utilities\get_avatar_or_image`).

I plan to add a few other meta-specific blocks, for speaker sessions, session dates, etc, but this is by far the most involved one. Eventually we could also set some patterns that use the query block, which could go into the initial query placeholder.

### Screenshots

A demo of the Speakers in a query block, with my photo replaced by a featured image.

<details>
<summary>Query block code</summary>

```html
<!-- wp:query {"queryId":7,"query":{"perPage":"12","pages":0,"offset":0,"postType":"wcb_speaker","categoryIds":[],"tagIds":[],"order":"desc","orderBy":"date","author":"","search":"","exclude":[],"sticky":"","inherit":false},"displayLayout":{"type":"flex","columns":3}} -->
<div class="wp-block-query"><!-- wp:post-template -->
<!-- wp:wordcamp/avatar {"size":128,"align":"center","style":{"color":{"duotone":["#000000","#D1E4DD"]}}} /-->

<!-- wp:post-title {"textAlign":"center","isLink":true,"fontSize":"normal"} /-->
<!-- /wp:post-template --></div>
<!-- /wp:query -->
```
</details>

Query block in the editor:

block-query-loop

Avatar block options (size, link to post):

block-avatar-inspector

Toggling duotone on/off:

block-duotone

On the frontend, 128px avatars, duotone to match the TT1 theme:

Screen Shot 2021-09-08 at 1 43 09 PM

Frontend with Organizers in the query block, "Voxel Cat" uses my user ID, but shows the featured image instead:

Screen Shot 2021-09-08 at 1 44 34 PM

If set, the featured image is also used in the Speakers & Organizers blocks…

Screen Shot 2021-09-08 at 1 45 45 PM

And on the individual Speaker page – the featured image itself is disabled on the speaker page, to prevent duplicate images.

Screen Shot 2021-09-08 at 1 46 12 PM

1. Create some speakers with different gravatar emails, and/or different organizers
2. Optionally set some featured images
3. Add a query block, change the post type to "Speaker" or "Organizer" and click into the template container to edit them
4. Add an Avatar block, play around with the settings
5. It should behave as expected, and show correctly on the frontend

Note: There is a bug in TT1 where center alignment in the editor looks off, but is correct on the frontend; seems to be the case with the featured image block too.


ryelle:
@coreymckrill @iandunn I pinged you both for review, but don't feel like you need to line-by-line review this if you don't have time, I wanted to check in more than anything.

public_html/wp-content/plugins/wc-post-types/inc/utilities.php
```diff
@@ -0,0 +1,70 @@
...
+		$src = get_avatar_url(
+			0,
+			array(
+				'size' => $size,
+				'force_default' => true,
+			)
+		);
+	}
+
+	$avatar = sprintf( '<img src="%1$s" alt="%2$s" />', esc_url( $src ), $alt );
```
coreymckrill:
```suggestion
	$avatar = sprintf( '<img src="%1$s" alt="%2$s" />', esc_attr( $src ), esc_attr( $alt ) );
```

public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -1029,6 +1035,31 @@ public function add_session_info_to_speaker_posts( $content ) {
 		return $content . $sessions_html;
 	}
 
+	/**
+	 * Prevent featured images from being displayed on organizer & speaker pages
+	 * by shortcutting the thumbnail lookup.
```
coreymckrill: I think it would be helpful here to explain why as well.

coreymckrill (approved):
👍 Looks good. One spot that I think needs more escaping, and one minor comment.

public_html/wp-content/plugins/wc-post-types/inc/utilities.php
ryelle: Good catch, thanks!

iandunn: Is `esc_attr()` better for the `src` in this context? IIRC `esc_url()` is does both of these:

* URL-encodes the value
* makes it safe to use an an attribute

...while `esc_attr()` only does the latter function.

coreymckrill: > makes it safe to use an an attribute

🤔 Does it? It doesn't appear that `esc_url` calls either of the sanitizing functions that `esc_attr` does.

I'm also not clear on when `esc_url` is actually necessary. The two use cases seem to be for if you are displaying the URL directly in HTML (not the case when it's in an attribute) and when you are going to store it in the database. It's definitely possible that I'm missing something and there's more to it than that, though...

iandunn: I haven't dug into it myself, but it seems like trustworthy folks have concluded `esc_attr( esc_url())` is unnecessary:

- https://developer.wordpress.com/themes/escaping/
- https://developer.wordpress.org/reference/functions/esc_attr/#comment-2459

I'm not sure if it's _harmful_ to do _both_, but I wouldn't _only_ use `esc_attr()` on a URL.

coreymckrill: 👍 Ok, I'd trust those sources.

PR #696, opened 2021-09-14
Blocks: Improve date display for schedule & sessions across timezones

### LINKED ISSUE #492: Sessions Block & Schedule Block: Clarify session times with timezones
Now that there are more online events, you might see a session list or schedule that lists times, but not have the context for knowing which timezone a camp is in. For example, [the WCUS 2019 sessions](https://2019.us.wordcamp.org/sessions/) show 9:15am, but you need to know that St Louis is in Central US time, whether it's DST, before you know what time that is "for real".

Some ideas
- Add a timezone stamp to each time, `9:15am CST`
- Use something like the time shortcode from wporg, automatically show time in the client's local time (maybe with the link to timeanddate.com?)

The same applies to the schedule block.

related(ish): https://github.com/WordPress/wordcamp.org/issues/442

### LINKED ISSUE #562: Sessions in schedule block overlap when viewing from certain locations/timezones
**Describe the bug**

This might be related to #559. When visiting [our WC site](https://colombia.wordcamp.org/2020/agenda/) from certain locations (for instance, London (UK) or Barcelona (Spain)), some sessions start to overlap making the whole thing unreadable.
These sessions seem to be the ones that fall near the time the day changes (in that timezone), but I'm not sure if that's the real reason.

It is worth mentioning that to work around #559, our site's timezone was changed to UTC (from the previously set "America / Bogota" or UTC-5). I can confirm that the sessions look fine when in UTC-5 or UTC-3, so maybe the the problem only affects UTC+something timezones?

Is there a something we can do from our side besides manually creating a schedule page (not using the block)? Thanks!

**To reproduce**

1. Set time zone to "Bogotá - Colombia" or "Buenos Aires - Argentina".
   On macOS: Go to System Preferences > Date & Time > Time Zone.
1. Go to https://colombia.wordcamp.org/2020/agenda/.
1. Confirm that the whole schedule looks fine.
1. Change time zone to "London - United Kingdom" (British Summer Time) or "Barcelona - Spain" (Central European Summer Time).
1. Go to https://colombia.wordcamp.org/2020/agenda/.
1. See how some session items look wrong, either overlapping or times being displayed over borders. See screenshots below for examples.

**Expected behavior**

Schedule should look correct in all timezones.

**Screenshots**

Screen Shot 2020-08-25 at 16 01 26
Screen Shot 2020-08-25 at 18 07 24

**WordCamp**

https://2020.colombia.wordcamp.org/

 - Device: Laptop
 - OS: macOS 10.15.6
 - Browser: Chrome
 - Version: 84.0.4147.135

### LINKED ISSUE #565: Agenda: Local Time / UTC Time / Timezones?
I have been looking at the agenda of several WordCamps and depending on the time zone you are in it is a bit complex to know what the "real" time of the event is.

In some cases the time is that of the place where it takes place (country, usually) and it is usually indicated as being in "Local Time / UTC-9", in others it indicates that it is in "your local time" which means "my time".

In the first case, everyone "looks for life" to know what time it is local, although it is simple because you have the UTC+something as a reference. The problem comes when you have your own local time because the agenda is reorganized and if there are changes in the day the system goes a bit crazy.

Would it be possible that, in the table, there is a drop-down menu where you can select, for example, "local time of the event", in addition to "your local time" and, already set, a list of "timezones" or "UTC-12 to UTC+12, or simply UTC?

+info (example/problem) at [Slack meta-wordcamp](https://wordpress.slack.com/archives/C08M59V3P/p1599214770064000).

ryelle:
This PR iterates on how dates & times are displayed in the blocks.

1. Update the schedule block to use `format`, which will use the client's timezone to do all session-grouping, layout, and display. This ensures the schedule is always showing the viewers timezone, which makes more sense for online and regional WordCamps.
2. Prefix the grid rules with the numeric date. This fixes the issue in #562, which happened because the last session ended at 12:10 the following day, while the first session started at 12:10 -- this created 2 rules in the `grid-template-rows` for `time-0010`, but only the first was valid. The session spanned from that first start time to it's real start time, functionally the entire "day".
3. Add timezone strings to the sessions in the session and speaker blocks - this is more of a workaround, since these are rendered by PHP and we don't know the client timezone, but it's a small improvement.

Fixes #562, fixes #565, fixes #492.

~Note: I'm not sure if we could consider this as fixing 565 - once this PR is applied, it should be the case that session times are always either your (viewer's) time, or has a timezone identifier.~ Now that the schedule time also has a TZ displayed, it should always be clear what timezone you're seeing.

### Screenshots

Schedule block use the viewer's timezone. My browser timezone is EDT, the site's timezone is PDT.  The sessions start at 8:00 am PDT. On the left is Safari, which shows production, showing the ambiguous "8:00am" start time. On the right is Firefox showing my sandbox, correctly showing "11:00am". (it's still ambiguous, but most people will assume local time. Organizers can add a note above the schedule)

Screen Shot 2021-09-14 at 12 47 41 PM

Session overlapping (#562):

schedule-overlap@0 5x

Sessions with timezones

Screen Shot 2021-09-14 at 7 24 05 PM

1. Set up sessions and a schedule, pay attention to your site's timezone
2. View the schedule from various timezones, the schedule should reflect your current TZ.
3. View the speakers & sessions blocks with dates showing, they should include the site's timezone

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/data.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/data.js
@@ -2,7 +2,7 @@
  * WordPress dependencies
  */
 import { useSelect } from '@wordpress/data';
-import { dateI18n } from '@wordpress/date';
+import { format } from '@wordpress/date';
 
 /**
  * Internal dependencies
@@ -447,7 +447,7 @@ function filterSessionsByChosenDays( sessions, chosenDays ) {
 	}
 
 	return sessions.filter( ( session ) => {
-		const date = dateI18n( DATE_SLUG_FORMAT, session.derived.startTime );
+		const date = format( DATE_SLUG_FORMAT, session.derived.startTime );
 		return chosenDays.includes( date );
 	} );
 
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/inspector-controls.js
@@ -3,7 +3,7 @@
  */
 import { InspectorControls } from '@wordpress/block-editor';
 import { CheckboxControl, PanelBody, ToggleControl } from '@wordpress/components';
-import { dateI18n, format } from '@wordpress/date';
+import { format } from '@wordpress/date';
 import { createInterpolateElement } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { decodeEntities } from '@wordpress/html-entities';
@@ -68,7 +68,7 @@ export default function ScheduleInspectorControls(
  */
 function getDisplayedDays( sessions ) {
 	let uniqueDays = sessions.reduce( ( accumulatingDays, session ) => {
-		accumulatingDays[ dateI18n( DATE_SLUG_FORMAT, session.derived.startTime ) ] = true;
+		accumulatingDays[ format( DATE_SLUG_FORMAT, session.derived.startTime ) ] = true;
 
 		return accumulatingDays;
 	}, {} );
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/schedule/schedule-grid.js
@@ -386,7 +386,7 @@ function renderGridTemplateRows( startEndTimes ) {
 	startEndTimes.sort(); // Put them in chronological order.
 
 	const timeList = startEndTimes.reduce( ( accumulatingTimes, time ) => {
-		const formattedTime = format( 'Hi', time );
+		const formattedTime = format( 'dHi', time );
 
 		return accumulatingTimes += `[time-${ formattedTime }] auto `;
 	}, '' );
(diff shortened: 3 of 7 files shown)


iandunn (commented):
LGTM overall 👍🏻 

I think the Schedule block should have the TZID too, though; otherwise I worry the ambiguity will confuse some folks.

* folks who are out of town could setup calendar reminders before they travel, and end up putting the incorrect times, and miss sessions.
* some contributors are used to the times being local, and will assume that they still are unless they see something to indicate otherwise

I think the TZID on all the blocks should also link to something like timeanddate.com, so people can easily convert between zones when needed.

ryelle:
Yeah, that makes sense. I've added it to the row heading, but I think putting it on the times in the session cells would be too verbose, so it looks like this now: 

Screen Shot 2021-09-16 at 2 03 09 PM

As for linking to a 3rd party time conversion site, I think that might also get visually overwhelming - to have so many links in the schedule. If you feel strongly about it, can you open another issue for that so we can discuss it there?

iandunn:
> putting it on the times in the session cells would be too verbose

👍🏻 

> linking to a 3rd party time conversion site [...] might also get visually overwhelming

👍🏻  - https://github.com/WordPress/wordcamp.org/issues/698

PR #703, opened 2021-10-19
Update the Remote CSS example repos to be clear it's an example, and not available to be forked.

dd32:
Update the Remote CSS example repos to be clear it's an example, and not available to be forked.

1. Review diff
2. Apply diff
3. Look at Remote CSS.


iandunn:
Thanks!

Issue #697, opened 2021-09-16
Migrate to Meetup.com GraphQL endpoints


iandunn:
Meetup.com is deprecating their API again, deadline is roughly the end of 2021

- [ ] Add unit/e2e tests to cover existing critical paths
- [x] Take a quick look to check if there's a new-ish 3rd party client we can trust on production
- [x] Verify won't need to change oauth flow
- [x] Migrate `send_paginated_request()`
- [x] Migrate `send_total_count_request()`
- [ ] Test `wcpt-meetup`
- [ ] Test `wordcamp-reports`
- [ ] Test `wordcamp-reports`
- [x] Test `official-wordpress-events` (w.org)

coreymckrill:
> Take a quick look to check if there's a new-ish 3rd party client we can trust on production

One of my criteria for this would be that the 3rd party client is very lightweight. Most of the clients I've seen so far do a ton of stuff we don't need (since all we really need is sending GET requests and handling responses) and as a result have a fair amount of complexity, which makes it harder to audit them for production use.

If we end up rolling our own client, I think it should probably extend the `\WordCamp\Utilities\API_Client` class, the way the current one does, to keep things simple.

tellyworth:
Meetup.com has a migration guide: https://www.meetup.com/api/migrating

PR #712, opened 2021-12-22
Migrate WordCamp.org to the new Meetup GraphQL API

dd32:
See #697


iandunn (approved):
I didn't dig deep, but this LGTM 👍🏻

PR #715, opened 2021-12-24
CamptTx Attendee List: Fix margins

### LINKED ISSUE #713: CampTix Attendee List: Twenty Twenty theme prevents seeing 3 columns
**Describe the bug**

Without adding custom CSS, it was not possible to use `[camptix_attendees]` and see 3 columns of attendees.

Slack thread with solution: https://wordpress.slack.com/archives/C08M59V3P/p1640210242065600

**To reproduce**

1. As a WordCamp site administrator, set your site's theme to Twenty Twenty.
2. Have at least 3 attendees visible on the site's Attendee page.
3. Browse to https://city.wordcamp.org/year/attendees/
Assumes the shortcode in places is `[camptix_attendees]`.
4. See only two columns of attendees displayed.

**Expected behavior**

See three columns of attendees.

**Workaround**
Adding the following CSS using the Customizer > Additional CSS:
```
#tix-attendees .tix-columns-3 li {
	margin-left: 0;
	margin-right: 0;
}
```

**Screenshots**

Before the fix:
image
FYI, I used the Customizer to temporarily undo the CSS fix and capture this screenshot.

After the fix:
image

**WordCamp**
Birmingham 2022
Likely any using the Twenty Twenty theme.

 - Device: Laptop
 - OS: Mac 11.6.2
 - Browser Chrome
 - Version 96.0.4664.110 (Official Build) (x86_64)

CdrMarks:
This PR sets the margin-left and margin-right to 0px on li elements inside the `tix-attendees` id.

The width for a 2, 3, 4, or 5 column wide attendee list needs to take up 50%, 33%, 25%, and 20% respectively.  

The CSS in some of the themes available to WordCamp sites have a margin defined on all li elements.   For example, this [margin](https://github.com/WordPress/twentytwenty/blob/a9e73ab1b6db33a57bd6317ce9f319fbf37b23d1/style.css#L495) is applied on all li elements in the Twenty Twenty theme.  
`li { margin: 0.5rem 0 0 2rem; }` 

This was causing the attendee list margin to be narrower than needed to display as intended.

Fixes #713

Props @dd32 

### Screenshots
Before the fix:
image

After the fix:
image

1. As a WordCamp site administrator, set your site's theme to Twenty Twenty.
2. Have at least 3 attendees visible on the site's Attendee page.
3. Browse to https://city.wordcamp.org/year/attendees/
Assumes the shortcode in places is `[camptix_attendees]`.
4. See three columns of attendees displayed.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.css
+++ b/public_html/wp-content/plugins/camptix/camptix.css
@@ -71,6 +71,8 @@ body.admin-bar #tix {
 #tix-attendees li {
 	float: left;
 	width: 50%;
+	margin-left: 0px;
+	margin-right: 0px;
 	margin-bottom: 20px;
 	height: 80px;
 }


iandunn:
My only concern here is if there are past camps that relied on the margin being there for their custom CSS.

One way to mitigate that would be to narrow the selectors so that the new styles only apply to `body.theme-twenty-twenty #tix-attendees li`. Core doesn't output that by default, though, so we'd have to add it (similar to `wcorg_content_slugs_to_body_tag()`).

There may be a better way, though; that's just what first comes to mind. I also haven't tested any older themes, it may not be an issue in practice. It's hard to really be confident w/ ~1300 old sites though.

pkevan:
Given there have been a few changes to the CSS in the interim, and looking at the latest issue: https://github.com/WordPress/wordcamp.org/issues/1210 around attendees, I think we should close this PR, and reopen, or restart if still necessary.

PR #719, opened 2022-01-07
Add new default Jetpack modules

### LINKED ISSUE #494: Jetpack: Which modules should be activated by default for new WordCamp sites?
I recently saw a report that the Sharing module was off on a new WordCamp site. That seems like one that should definitely be on for new sites. It made me wonder what the default states are for other modules.

In **mu-plugins/jetpack-tweaks/new-sites.php** we've got the `default_jetpack_modules()` function hooked to `jetpack_get_default_modules`, which modifies the set that Jetpack turns on by default. It looks like we're removing Widget Visibility from the defaults, and then adding the Contact Form, Shortcodes, Custom CSS, and Subscriptions.

Ideally, I think, we'd specify the state of every module (known at the time) so that we could be sure every site had a consistent set.

The modules, as of Jetpack 8.3, and the state they currently start in on WordCamp sites:

* Asset CDN
* Beautiful Math
* Carousel
* Comment Likes
* Comments
* **Contact Form - Activated**
* Copy Post
* **Custom CSS - Activated**
* Custom content types
* **Enhanced Distribution - Activated**
* **Extra Sidebar Widgets - Activated**
* Gravatar Hovercards
* Image CDN
* Infinite Scroll
* **JSON API - Activated**
* Lazy Images
* Likes
* Markdown
* Monitor
* **Notifications - Activated**
* Post by email
* **Protect - Activated**
* Publicize
* Related posts
* Secure Sign On
* Sharing
* **Shortcode Embeds - Activated**
* **Site Stats - Activated**
* **Site verification - Activated**
* Sitemaps
* **Subscriptions - Activated**
* Tiled Galleries
* WP.me Shortlinks
* Widget Visibility
* WordPress.com Toolbar

I would probably also enable these by default:

* Copy Post
* Publicize
* Sharing
* Sitemaps
* Tiled Galleries

I'm also not sure why we turn off Widget Visibility. It seems useful for event sites, unless there's some kind of conflict with other functionality.

timiwahalahti:
In #494 it has been discussed what new Jetpack modules should be enabled by default when creating new WordCamp sites.

Based on that discussion, this PR added the following modules to the list of default enabled ones. At the same time, did minor improvements to the readability of the `default_jetpack_modules` function.

* Sharing (`sharedaddy`) - needed for og: tags and social shares looking nice, usually forgotten to put on by organisers
* Copy Post - useful little tool on the admin side, no effect on performace
* Sitemap - doesn't hurt on SEO side
* Site Accelerator (`photon` and `photon-cdn`) - I don't see any reasons why not to make use of someone else's server resources and save ours 😅

Fixes #494 

Props @iandunn, @coreymckrill, @andreamiddleton

1. Create a new site
2. In Jetpack settings, check that mentioned modules are enabled


timiwahalahti:
Took a look at this PR as well.

Instead of `photon`, there's an `image-cdn` module that does the image CDN part.

Also, since [Core generates basic sitemaps since version 5.5](https://make.wordpress.org/core/2020/07/22/new-xml-sitemaps-functionality-in-wordpress-5-5/) I don't see a reason to use Jetpack generated ones, so disabled that module.

Ready for review!

iandunn (approved):
LGTM 👍🏻

PR #721, opened 2022-01-08
Fix mailcatcher install command on Docker build

timiwahalahti:
On Docker build, mailcatcher is installed from gem. Previously there were options `--no-ri --no-rdoc` given, but those are deprecated and the build process didn't work as `gem install` failed. Changed the options to `--no-document` which seemed to do the same thing.

1. Try to run Docker build with `RUN gem install mailcatcher --no-ri --no-rdoc` > fails
2. Now run Docker build with `RUN gem install mailcatcher --no-document` > works


ryelle (approved):
This fixes the image build 👍🏻

PR #722, opened 2022-01-08
Add default empty select on Camptix country field

### LINKED ISSUE #434: Camptix: Require a selection in the country dropdown field
In our team we've detected a problem with the country selection dropdown field when registering a ticket. Since "Afghanistan" appears as the first option, even though users don't select any country, they submit the form with that country selected, and that isn't really their country. This could be solved by adding as first option something like "--Select a country--", and that this option cannot be selected. As it's a required field, then they cannot submit the form until they select their real country.

Live example: https://2020.spain.wordcamp.org/entradas/?tix_action=attendee_info#tix

Thanks!
screenshot

timiwahalahti:
Adds option "-- Select --" with empty value to the country select field on Camptix. The option is disabled always and selected if the attendee hasn't selected any other value during the purchase. This forces the attendees to select a proper country if the field is required, thus making the field values more reliable.

Fixes #434

Props @NobNob

### Screenshots

image

1. Add tickets and configure Camptix to allow purchases
2. Add the Country field to the ticket questions without making it required
3. Purchase a ticket without choosing any option on the country field
4. Check the attendee data either on the dashboard or via attendee information edit link -> field should be in select state
5. Purchase a ticket and choose some country on the field
6. Check the attendee data either on the dashboard or via attendee information edit link -> field should have selected country

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/field-country.php
+++ b/public_html/wp-content/plugins/camptix/addons/field-country.php
@@ -38,6 +38,10 @@ class CampTix_Addon_Country_Field extends CampTix_Addon {
 			name="<?php echo esc_attr( $name ); ?>"
 			<?php if ( $required ) echo 'required'; ?>
 		>
+			<option value="" disabled <?php selected( '', $user_value ); ?>>
+				-- <?php esc_html_e( 'Select', 'wordcamporg' ); ?> --
+			</option>
+
 			<?php foreach ( $countries as $country ) : ?>
 				<option value="<?php echo esc_attr( $country ); ?>" <?php selected( $country, $user_value ); ?>>
 					<?php echo esc_html( $country ); ?>


public_html/wp-content/plugins/camptix/addons/field-country.php
```diff
@@ -38,6 +38,10 @@ function question_field_country( $name, $user_value, $question, $required = fals
 			name="<?php echo esc_attr( $name ); ?>"
 			<?php if ( $required ) echo 'required'; ?>
 		>
+			<option value="" disabled <?php selected( '', $user_value ); ?>>
```
ryelle, line 41: This works in Chrome, but Firefox doesn't default to this option, I think because it's `disabled`:
| Chrome | Firefox |
|------|------|
| Screen Shot 2022-03-30 at 4 56 53 PM | Screen Shot 2022-03-30 at 4 55 39 PM |

timiwahalahti: Hmm, I tested with Firefox Developer Edition 115.0b6, Firefox 115.0.2, Safari 16.4.1 and Chrome 114.0.5735.198, and all defaulted to select. Can you test again, in FF has changed something?

pkevan (approved):
Tested in Firefox and works well 👍

PR #725, opened 2022-01-08
Add banner to previous year events redirecting visitors to the latest event

### LINKED ISSUE #688: Add "global banner" to previous year events redirecting visitors to the current year event
Often when searching for a WordCamp, it's not unexpected to end up on a previous year event site. This experience can be not-too-great for the average attendee, as there's often no link or banner to the current years site.

This can happen from not-search as well, for example, a company linking to a previous years site that they attended/sponsored/etc and that link may never be updated.

For example, https://europe.wordcamp.org/2020/ offers no suggestion that there's a 2021 or 2022 event.

Unfortunately, many sites, including europe/2020 use fixed positioning for the header element, so inserting a banner at the top of the screen like a toolbar is not always going to be possible, so it might need to be a "popover" or something similar, depending on what a designer can come up with.

I raised this on Slack which got some good support: https://wordpress.slack.com/archives/C08M59V3P/p1625206043052300

timiwahalahti:
Add notification if a user visits WordCamp site that has a newer site for the same location. Keeps the canonical link functionality for seo purposes.

Also renamed the file to follow better naming, as it made sense to put this functionality into the same file that already contained hints about the latest site for search engines.

This was a bit tricky to test because of the differences between the Docker environment and production. Also, the development database is lacking proper real-world examples. So needs proper thorough testing, mentioning @iandunn here as they had written the seo code on which this PR is based.

Fixes #688

Props @dd32 

### Screenshots

image

1. Create new WordCamp entry on tracker, based on some other existing event but change the year
2. Make sure the new WordCamp has a site
3. Visit the WordCamp site from which you based the new event, notification should be visible


public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -0,0 +1,216 @@
...
+	// Hook in before `WordPressdotorg\SEO\Canonical::rel_canonical_link()`, so that callback can be removed.
+	add_action( 'wp_head', __NAMESPACE__ . '\canonical_link_past_home_pages_to_current_year', 9 );
+
+	/**
+	 * Show notification about newer WordCamp. Logged in users most probably know that already,
+	 * so no need to bother them (and we also misuse the body.admin-bar class).
+	 */
+	if ( ! is_user_logged_in() ) {
+		add_filter( 'body_class',   __NAMESPACE__ . '\add_notification_classes_to_body' );
+		add_action( 'wp_head',      __NAMESPACE__ . '\add_notification_styles' );
```
iandunn: I'm seeing this fire when I'm logged in, and not seeing it when logged out. I don't see `show_notification_about_latest_site()` fire when logged out either.

Maybe it's something wrong in my env? I can dig deeper if this is still working in yours.

public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -0,0 +1,216 @@
...
+
+	printf(
+		'<link rel="canonical" href="%s" />' . "\n",
+		esc_url( set_url_scheme( trailingslashit( '//' . $latest_site->domain . $latest_site->path ) ) )
+	);
+}
+
+/**
+ * If notification is shown, misuse body.admin-bar class as most of the WordCamp themes already that that into
+ * account. This makes its a whole lot easier to place the notification on top of contens.
```
iandunn:
```suggestion
 * account. This makes its a whole lot easier to place the notification on top of content.
```

public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -0,0 +1,216 @@
...
+	return $classes;
+}
+
+/**
+ * Simple styles for the notification.
+ */
+function add_notification_styles() { ?>
+	<style type="text/css">
+		html {
+			margin-top: 35px !important;
```
iandunn:
```suggestion
			margin-top: var(--wp-admin--admin-bar--height) !important;
```

That was recently added to Core, and will automatically update from `32px` to `46px` at the various breakpoints.

public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		.wordcamp-latest-site-notify {
+			background: #1d2327;
+			text-align: center;
+			padding-top: 5px;
+			font-size: 15px;
+			height: 35px;
```
iandunn: This might also need to use `--wp-admin--admin-bar--height`, if it's intended to match the `html margin-top`

public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -0,0 +1,216 @@
...
+	remove_action( 'wp_head', 'WordPressdotorg\SEO\Canonical\rel_canonical_link' );
+
+	printf(
+		'<link rel="canonical" href="%s" />' . "\n",
+		esc_url( set_url_scheme( trailingslashit( '//' . $latest_site->domain . $latest_site->path ) ) )
+	);
+}
+
+/**
+ * If notification is shown, misuse body.admin-bar class as most of the WordCamp themes already that that into
```
iandunn: That's a clever idea, kudos 👍🏻

public_html/wp-content/mu-plugins/tests/bootstrap.php
```diff
@@ -27,7 +27,7 @@ function manually_load_plugins() {
 
 	require_once dirname( __DIR__ ) . '/0-error-handling.php';
 	require_once dirname( __DIR__ ) . '/lets-encrypt-helper.php';
-	require_once dirname( __DIR__ ) . '/seo.php';
+	require_once dirname( __DIR__ ) . '/latest-site-hints.php';
```
iandunn: It looks like the [tests are failing](https://github.com/WordPress/wordcamp.org/runs/4746639950?check_suite_focus=true) because the cases need to be updated to reflect the changes in `get_latest_site_id()`. 

`test:watch -- --group=latest-site-hints` makes it convenient. [The Docker readme](https://github.com/WordPress/wordcamp.org/blob/f58e6fc89a5cb755574576b5e0ab935c57449e51/.docker/readme.md) has more details.

public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -0,0 +1,216 @@
...
+
+	$latest_site_id = get_latest_site_id( $current_blog->domain, $current_blog->path );
+
+	if ( ! $latest_site_id ) {
+		return false;
+	}
+
+	$latest_site = get_site( $latest_site_id );
+
+	if ( $latest_site->domain === $current_blog->domain ) {
```
iandunn: This needs to check the `path` as well, since the year is now there. It's probably easier to compare the `blog_id` though.

public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		.wordcamp-latest-site-notify {
+			background: #1d2327;
+			text-align: center;
+			padding-top: 5px;
+			font-size: 15px;
+			height: 35px;
+			position: fixed;
+			top: 0;
```
iandunn: Why are are the positioning styles needed if we're reusing the admin bar classes? I'd assume it'd inherit those. Maybe enqueuing the stylesheet would add too much unnecessary stuff or something?

iandunn (commented):
Sorry for the delay here :(

I went ahead and merged the file renaming bits, to make it easier to review the actual changes.

I think this is good overall, just need to figure out a possible bug and a few details.

public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -0,0 +1,216 @@
...
+	add_action( 'wp_head', __NAMESPACE__ . '\canonical_link_past_home_pages_to_current_year', 9 );
+
+	/**
+	 * Show notification about newer WordCamp. Logged in users most probably know that already,
+	 * so no need to bother them (and we also misuse the body.admin-bar class).
+	 */
+	if ( ! is_user_logged_in() ) {
+		add_filter( 'body_class',   __NAMESPACE__ . '\add_notification_classes_to_body' );
+		add_action( 'wp_head',      __NAMESPACE__ . '\add_notification_styles' );
+		add_action( 'wp_body_open', __NAMESPACE__ . '\show_notification_about_latest_site' );
```
ryelle: This hook is only called if the theme uses [the `wp_body_open` function](https://make.wordpress.org/core/2019/04/24/miscellaneous-developer-updates-in-5-2/), and CampSite 2017 doesn't — if there isn't another suitable hook, the theme can be updated.

public_html/wp-content/mu-plugins/latest-site-hints.php
timiwahalahti: I think the original idea was to avoid loading the admin-bar stylesheet and make things lighter since not all admin bar styles work well with this notification and the admin bar stylesheet `html` tag `scroll-padding-top` does not seem to work at least with Firefox.

public_html/wp-content/mu-plugins/latest-site-hints.php
timiwahalahti: The idea is good, although the `--wp-admin--admin-bar--height` is defined in `admin-bar.css`, which isn't enqueued in the public site when user is logged out. If we end up loading that stylesheet, then the use of that variable is definitely wise.

public_html/wp-content/mu-plugins/latest-site-hints.php
timiwahalahti: Hmm, I couldn't reproduce this. Non-logged-in users do see the the banner and get inline styles, logged-in users do not. Of course, the `admin-bar` body class is there for logged-in users via core.

public_html/wp-content/mu-plugins/latest-site-hints.php
timiwahalahti: Oh, good catch! Changed the message output to `wp_footer`, but I'm not too thrilled about that because of accessibility. 🤔

timiwahalahti:
Sorry for the delay! Refreshed the code and made changes based on the review comments. I also made the message translatable and simplified it a bit by removing the site name of the next edition - I think that caused unnecessary queries since we already have the URL for the next edition.

Was thinking of extracting the year from URL, now that it's a path, but that started feeling a bit more fragile - and the lack of year does not even make message much different.

Also removed the europe.wordcamp.org exception since that seems to be covered upper in the same code.

Oh and URL structure change to year as a path seemed to actually broke some checks, since some functions were returning different url/domain than the one we checked against.

StevenDufresne:
@timiwahalahti Is there anything specific left that you'd like a review on? It would be nice to get this PR in ahead of WCUS!

timiwahalahti:
Nothing special, just re-review since there are changes made after @iandunn and @ryelle last time reviewed :)

StevenDufresne:
> Nothing special, just re-review since there are changes made after @iandunn and @ryelle last time reviewed :)

There seems to be a merge conflict, do you mind resolving?

public_html/wp-content/mu-plugins/latest-site-hints.php
dd32: > This hook is only called if the theme uses [the wp_body_open function](https://make.wordpress.org/core/2019/04/24/miscellaneous-developer-updates-in-5-2/), and CampSite 2017 doesn't

Regardless, we should probably add that to the theme just for future compatibility.

timiwahalahti: Agreed. I will open a PR on the theme, and let's switch to `wp_body_open` when that is in production.

timiwahalahti:
Sure thing, done!

ryelle (approved):
This is looking good! I do think this notification should appear for logged in users too, so I've just pushed an update to this PR to do that and tweak some of the CSS to better support that.

It's a little janky on small screens, but once it's moved to `wp_body_open` that should be fixable. I think this is good to merge 👍🏻

@renintw I know you were tagged here, are you able to merge and deploy this?

outdoor2kode:
Looks great code-wise and the UI displays well!

It's been deployed. But a minor issue was spotted with the admin bar dropdown. On some sites, it gets covered by the banner. (Originally, I thought only the admin account would get this problem. Because locally, when logging in with a non-admin, the dropdown isn't covered by the admin bar. However, in the production env, it gets covered)

Screen Shot 2023-08-10 at 10 25 24 PM
Screen Shot 2023-08-10 at 10 29 47 PM
Screen Shot 2023-08-10 at 10 24 12 PM

I have proceeded with the merge and deployment for this PR, and I'll look into the issue. Seems related to the theme used.

outdoor2kode:
@ryelle After testing, I've found that the issue occurs with most themes (2016-2023 at least) except for "CampSite 2017". On sites using themes other than "CampSite 2017", the `div#wpadminbar` is loaded earlier, causing the dropdown to be covered by the later loaded `div.wordcamp-latest-site-notify`.

I've tried the following two solutions and both seem to fix the issue:

1. Adjusting the z-index [here](https://github.com/WordPress/wordcamp.org/pull/725/files#diff-22b591d047311fc622c1be6e8dc4e422ce30d0f0375183b876e2da7d0fdee1d0R84) to 99998.
2. Changing wp_footer to wp_head [here](https://github.com/WordPress/wordcamp.org/pull/725/files#diff-22b591d047311fc622c1be6e8dc4e422ce30d0f0375183b876e2da7d0fdee1d0R30).

But neither option seems like best practice, so I thought I'd get your opinion on how we usually handle this.

StevenDufresne:
> Adjusting the z-index [here](https://github.com/WordPress/wordcamp.org/pull/725/files#diff-22b591d047311fc622c1be6e8dc4e422ce30d0f0375183b876e2da7d0fdee1d0R84) to 99998.

That approach seems fine to me, unless there's another reason why its exactly set to the same `z-index` as the adminbar.

outdoor2kode:
@StevenDufresne Thanks for the input 🙂 The issue was resolved after deploying the [commit](https://github.com/WordPress/wordcamp.org/commit/d5de714a7396fd2e944db477b34a5af4102775d8).

PR #727, opened 2022-02-03
Don't suffix the year to the WordCamp name if the year exists within the WordCamp name already.

### LINKED ISSUE #638: Profile activity: WordCamp year shown twice
Imported from https://meta.trac.wordpress.org/ticket/4675 
Created by @ePascalC:

As per picture, there is no need to add the WordCamp year if it is already present in the name.

dd32:
Fixes #638.


coreymckrill (approved):
👍

Issue #374, opened 2020-02-20
CampTix: Issues with the [camptix_attendees] shortcode


coreymckrill:
Coming out of [a Slack discussion](https://wordpress.slack.com/archives/C08M59V3P/p1582224263294100) in #meta-wordcamp, this issue is to collect various problems with `[camptix_attendees]` from Meta Trac ([meta-2282](https://meta.trac.wordpress.org/ticket/2282), [meta-2413](https://meta.trac.wordpress.org/ticket/2413)) and elsewhere.

Some problems are:

* For larger WordCamps (200+), having all of the attendees on the page at once is unwieldy, and it is not easy to browse or find specific people.
* The columns parameter for the shortcode is buggy, and doesn't work correctly in some situations such as when there are less attendees than there are columns.
* Lazy loading the attendee gravatars improves performance, but it may be using an outdated approach, given that a native solution [may be coming to WP Core in 5.4](https://make.wordpress.org/core/2020/01/29/lazy-loading-images-in-wordpress-core/).
* The HTML markup and CSS to output the columns is not using a modern approach (floated divs instead of flexbox or css-grid).

One proposed solution is to render the attendees list via JS instead of PHP, which would make it easier to add dynamic filtering/searching. On the back end, we could also convert the shortcode into a block, which would improve the UX for configuring options and make it more self-documenting (and anyway, ideally all of WordCamp's shortcodes will eventually become blocks).

timiwahalahti:
Related from Trac: [#2940 WordCamp Attendee JSON](https://meta.trac.wordpress.org/ticket/2940)

ryelle:
[meta-3217](https://meta.trac.wordpress.org/ticket/3217) is another we could add to this batch, to allow organizers to assign a default gravatar image (the potential attendee block could have an image upload attached).

PR #730, opened 2022-02-03
Camptix: Remove legacy lazy-loading for attendee avatars

coreymckrill:
Browser-based lazy loading for images has been built into WP since 5.4. We no longer need to use JavaScript and various jQuery scripts to accomplish this for the attendee avatars output by the Attendees shortcode.

This removes all the scripts and markup for our old homegrown lazy-loading scheme and just outputs the avatars via WP's `get_avatar` function. Core takes care of the rest.

This also changes the expiration for the cached attendee list markup to refresh every 3 hours instead of 24, as there have been a few complaints about that in the past when people wanted to see themselves show up on the attendees page.

(This refs #374, but only solves one of the bullet points therein.)

1. Upload the changed files to your wporg sandbox and then visit the attendees page on any WordCamp site.
2. You may need to also modify the `$force_refresh` value in the `get_attendees_shortcode_content` method to be `true` so that the new markup will load instead of the old markup that has avatar placeholders.

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/shortcodes.php
+++ b/public_html/wp-content/plugins/camptix/addons/shortcodes.php
@@ -120,18 +120,6 @@ class CampTix_Addon_Shortcodes extends CampTix_Addon {
 	 * Callback for the [camptix_attendees] shortcode.
 	 */
 	public function shortcode_attendees( $attr ) {
-		// Required scripts.
-		wp_enqueue_script( 'wp-util' ); // For wp.template().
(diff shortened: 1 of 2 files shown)


ryelle (approved):
This looks good 👍🏻 

The only issue I can see is that for up to 3hrs, while the cached markup is valid, all the attendee pages will break. Maybe it's worth updating the cache key in `generate_attendees_cache_key`?

coreymckrill:
Ah, good idea!

PR #738, opened 2022-03-04
Automate updating of sponsors from Multi Event Sponsors

### LINKED ISSUE #736: Automate updating Multi Event Sponsor posts on camp sites
_#735 should be done before this, so that we're not adding technical debt_

-------------------------------

When changes are made to Multi Event Sponsor posts, they have to be manually synced to active camps. Automating would save time.

* add a checkbox when editing MES posts. if checked, then all `wcb_sponsor` posts will be updates when the `mes` post is saved
* only update `scheduled` and `pre-planning` camps
* only update if existing post found, don't create new post. `_mes_id` can be used to link `wcb_sponsor` posts to their `mes` post.
* skip regional camps, b/c they don't use global sponsors

`create_post_stubs()` is where the `wcb_sponsor` posts originally get created, and `maybe_push_mes()` is where new ones get added to existing sites.

iandunn:
Fixes #736 

This adds a heart dashicon to Gutenbergs toolbar on Multi Event Sponsor pages (because that's the icon we use to represent sponsors). When you click it, a sidebar opens. From there you can click a button to push the current MES post out to active camps. 

That triggers an API request, which switches to each camp and updates the post if it hasn't been edited by the organizers.

The sidebar is not a very good UX for this, but it seems like the least-bad option, given [the lack of a proper slot in Gutenberg](https://github.com/WordPress/gutenberg/issues/16988#issuecomment-1055738184).

### DIFF
--- a/public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
+++ b/public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
@@ -19,6 +19,7 @@ class MES_Sponsor {
 	 */
 	public function __construct() {
 		add_action( 'init',                  array( $this, 'create_post_type' ) );
+		add_action( 'rest_api_init',         array( $this, 'register_routes'  ) );
 		add_action( 'admin_init',            array( $this, 'add_meta_boxes'   ) );
 		add_action( 'admin_enqueue_scripts', array( $this, 'admin_enqueue'    ), 20 );
 		add_action( 'save_post',             array( $this, 'save_post'        ), 10, 2 );
@@ -63,11 +64,127 @@ class MES_Sponsor {
 				'with_front' => false,
 			),
 			'query_var'       => true,
+			'menu_icon'       => 'dashicons-heart',
 			'supports'        => array( 'title', 'editor', 'author', 'excerpt', 'revisions', 'thumbnail' ),
 			'taxonomies'      => array( MES_Region::TAXONOMY_SLUG ),
 		);
 
 		register_post_type( self::POST_TYPE_SLUG, $post_type_params );
+	}
+
+	/**
+	 * Register endpoints for the REST API.
+	 */
+	public function register_routes() : void {
+		register_rest_route(
+			'multi-event-sponsors/v1',
+			'/push-to-active-camps',
+			array(
+				array(
+					'methods'             => WP_REST_Server::CREATABLE,
+					'callback'            => array( $this, 'push_to_active_camps' ),
+					'permission_callback' => function() {
+						return current_user_can( 'manage_network' );
+					},
+				),
+			)
+		);
+	}
+
+	/**
+	 * Copy the contents of MES posts to their forked posts on active camp sites.
+	 */
+	public function push_to_active_camps( WP_REST_Request $request ) : array {
+		$edited_posts = array();
+		$source_post  = get_post( $request->get_param( 'sponsorId' ) );
+		$wordcamps    = get_wordcamps( array(
+			'post_status' => array_merge(
+				// Any status where the camp might have a site created, and the event isn't over.
+				WordCamp_Loader::get_pre_planning_post_statuses(),
+				WordCamp_Loader::get_active_wordcamp_statuses(),
+			),
+		) );
+
+		foreach ( $wordcamps as $wordcamp ) {
+			if ( ! $wordcamp->meta['_site_id'][0] ) {
+				continue;
+			}
+
+			switch_to_blog( $wordcamp->meta['_site_id'][0] );
+
+			$fork_post = get_posts( array(
+				'post_type'      => 'wcb_sponsor',
+				'meta_key'       => '_mes_id',
+				'meta_value'     => $source_post->ID,
+				'meta_compare'   => '=',
+			) );
+			$fork_post = array_pop( $fork_post );
+
+			if ( ! $fork_post ) {
+				restore_current_blog();
+				continue;
+			}
+
+			// Organizers might have made changes that are specific to this camp, so don't overwrite them.
+			if ( self::post_has_been_edited( $fork_post->ID ) ) {
+				$edited_posts[] = array(
+					// Can't use `get_post_edit_link()` because the `wcb_sponsor` post type isn't registered on Central or while switching blogs.
+					'edit_url'  => admin_url( "post.php?post={$fork_post->ID}&action=edit" ),
+					'site_name' => get_wordcamp_name(),
+				);
+				restore_current_blog();
+				continue;
+			}
+
+			wp_update_post( array(
+				'ID'           => $fork_post->ID,
+				'post_title'   => $source_post->post_title,
+				'post_content' => $source_post->post_content,
+
+				// This triggers a database query for every item, so it's a good optimization candidate if needed.
+				'meta_input'   => WordCamp_New_Site::get_stub_me_sponsors_meta( $source_post ),
+			) );
+
+			// Reset date so that post_has_been_edited() will continue detecting that it hasn't been edited (by a user).
+			self::reset_post_modified_date( $fork_post );
+
+			// Default to Gutenberg since all the source posts now use it.
+			if ( is_callable( array( 'Classic_Editor', 'remember_block_editor' ) ) ) {
+				Classic_Editor::remember_block_editor( array(), $fork_post );
+			}
+
+			restore_current_blog();
+		}
+
+		return array(
+			'success'      => true,
+			'edited_posts' => $edited_posts,
+		);
+	}
+
+	/**
+	 * Check if the post has been edited since it was created.
+	 *
+	 * WP will update modified date when saving posts, even if only post meta is changed and the title/content are untouched.
+	 */
+	public static function post_has_been_edited( int $post_id ) : bool {
+		return get_the_modified_time( 'U', $post_id ) !== get_the_time( 'U', $post_id );
+	}
+
+	/**
+	 * Revert the post modified date back to the creation date.
+	 */
+	public static function reset_post_modified_date( WP_Post $post ) : void {
+		global $wpdb;
+
+		$wpdb->update(
+			$wpdb->posts,
+			array(
+				'post_modified'     => $post->post_date,
+				'post_modified_gmt' => $post->post_date_gmt,
+			),
+			array( 'ID' => $post->ID )
+		);
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


ryelle:
> The sidebar is not a very good UX for this, but it seems like the least-bad option

What about the [`PluginDocumentSettingPanel` slot](https://developer.wordpress.org/block-editor/reference-guides/slotfills/plugin-document-setting-panel/)? It's not necessarily at the top of the page, but is a little more inline than a separate sidebar.

Screen Shot 2022-03-04 at 1 32 24 PM

<details>
<summary>(my hacked together code to see exactly where it rendered)</summary>

```
const { registerPlugin } = wp.plugins;
const { PluginDocumentSettingPanel } = wp.editPost;
const { createElement } = wp.element;

function PluginDocumentSettingPanelDemo() {
	return createElement(
		PluginDocumentSettingPanel,
		{
			name: "custom-panel",
			title: "Push to Active WordCamps",
			className: "custom-panel",
		},
		"Button here…"
	);
}

registerPlugin("plugin-document-setting-panel-demo", {
	render: PluginDocumentSettingPanelDemo,
	icon: "heart",
});
```

</details>

iandunn:
It felt like the things in the inspector are more related to the post content, while the things in the toolbar are actions that don't necessarily impact `post_content` or post meta. That might be subjective, though? 🤷🏻‍♂️

harishankerr:
This is great! While I have not tested the PR, looking at the mockups, this looks fantastic. I'm agnostic about the positioning of the button, as long as the feature is available. Looking forward to seeing it in action! 💥

public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
```diff
@@ -63,13 +64,129 @@ public function create_post_type() {
...
+				continue;
+			}
+
+			// Organizers might have made changes that are specific to this camp, so don't overwrite them.
+			if ( self::post_has_been_edited( $fork_post->ID ) ) {
+				$edited_posts[] = array(
+					// Can't use `get_post_edit_link()` because the `wcb_sponsor` post type isn't registered on Central or while switching blogs.
+					'edit_url'  => admin_url( "post.php?post={$fork_post->ID}&action=edit" ),
+					'site_name' => get_wordcamp_name(),
+				);
```
ryelle, line 134: Could this variable use a different name? Maybe `unchanged_posts`? I keep interpreting "edited_posts" as "successfully updated posts".

public_html/wp-content/plugins/multi-event-sponsors/javascript/index.js
```diff
@@ -0,0 +1,177 @@
...
+				{ sprintf(
+					// translators: Title of the post
+					__( 'This will copy the title/content/etc of this post to all of its corresponding %s posts on active WordCamp sites (except regional camps).', 'wordcamporg' ),
+					decodeEntities( stripTags( sourcePost.title ) )
+				) }
+			</p>
+
+			<p className="disclaimer">
+				{ createInterpolateElement(
+					__( 'This wont push to sites that were created before this post; for that please <a>edit their WordCamp post</a> and <code>Push new sponsors to site</code>.', 'wordcamporg' ),
```
ryelle:
```suggestion
					__( "This won't push to sites that were created before this post; for that please <a>edit their WordCamp post</a> and <code>Push new sponsors to site</code>.", 'wordcamporg' ),
```

public_html/wp-content/plugins/multi-event-sponsors/javascript/index.js
```diff
@@ -0,0 +1,177 @@
...
+		<PluginSidebar
+			name="push-to-active-camps"
+			className="push-to-active-camps"
+			title="Push to Active Camps"
+			icon={ <Dashicon icon="heart" /> }
+		>
+			<p>
+				{ sprintf(
+					// translators: Title of the post
+					__( 'This will copy the title/content/etc of this post to all of its corresponding %s posts on active WordCamp sites (except regional camps).', 'wordcamporg' ),
```
ryelle, line 52: How does it know not to push to regional camps?

public_html/wp-content/plugins/multi-event-sponsors/javascript/index.js
```diff
@@ -0,0 +1,177 @@
...
+	if ( isDirty && result !== null ) {
+		// Make sure the UI doesn't show results from a previous push once the current changes are saved.
+		setResult( null );
+	}
+
+	return (
+		<PluginSidebar
+			name="push-to-active-camps"
+			className="push-to-active-camps"
+			title="Push to Active Camps"
```
ryelle:
```suggestion
			title={ __( 'Push to Active Camps', 'wordcamporg' ) }
```

ryelle (commented):
The path to this project needs to be added to `workspaces` in [the root package.json](https://github.com/WordPress/wordcamp.org/blob/6821bdcaf68e9d22292f5f510fe8645849412646/package.json#L14), otherwise yarn ignores it.

Everything's testing out well for me, I had a few comments but once those are resolved this should be good to merge.

public_html/wp-content/plugins/multi-event-sponsors/package.json
```diff
@@ -0,0 +1,18 @@
...
+	},
+	"scripts": {
+		"build": "wp-scripts build javascript/index.js --output-path=build",
+		"start": "wp-scripts start javascript/index.js --output-path=build --hot",
+		"lint:js": "wp-scripts lint-js",
+		"packages-update": "wp-scripts packages-update"
+	},
+	"devDependencies": {
+		"@wordpress/icons": "^7.0.1",
+		"@wordpress/scripts": "^22.0.1"
```
ryelle, line 16: Oh, I also realized that you'll need to add a `lint:css` script once you add the package to the workspaces list (the [github action script requires it](https://github.com/WordPress/wordcamp.org/blob/6821bdcaf68e9d22292f5f510fe8645849412646/.github/workflows/linter.yml#L46)). When I tried it, I got an error because this version of wp-scripts doesn't match the rest, I guess we need to update the packages. For now, you can pin this to `"@wordpress/scripts": "19.2.3"` and use `"lint:css": "wp-scripts lint-style css"`.

public_html/wp-content/plugins/multi-event-sponsors/javascript/index.js
iandunn: Good catch 👍🏻

public_html/wp-content/plugins/multi-event-sponsors/javascript/index.js
iandunn: AFAIK we don't have a way to identify regional camps programmatically. I'm assuming that it doesn't need to, though, since regional camps will either delete the posts after the site is created, or edit them. That's a good assumption to check and document, though.

@harishankerr , is that correct? or do regional camps sometimes leave the global sponsor posts unedited?

iandunn:
I'm gonna merge this since they're waiting on it, then circle back to the other stuff

public_html/wp-content/plugins/multi-event-sponsors/javascript/index.js
iandunn: It also only pulls published posts, so it'd skip ones they moved to draft

ryelle: Okay, so it assumes regional camps have edited posts, not that it does any special exclusion for regional camps? If a regional camp hasn't edited a sponsor, it will push out the update? Overall I think that's fine, I would just update the message to make the difference clear:

> This will copy the title/content/etc of this post to all of its corresponding %s posts on active WordCamp sites, unless they have edited the post.

public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
iandunn: 🤔 , I see the problem, but it seems like `changed_posts` would have the same one? Maybe `edited_forks`, since it's the _forked post_ that's been edited?

Or did you literally mean `unchanged_posts`, implying a refactor of the logic?

ryelle: No I meant `unchanged_posts`, because these posts have not been changed by this code. `changed_posts` would have the same issue as `edited_posts` does. I don't think that requires a logic refactor, unless I'm really misunderstanding the code here.

Actually, `forked_posts` has a similar problem, because to me forked implies changed too, but the `forked_posts` are not necessarily changed… but that's not as confusing in context.

iandunn: Ah, you meant changed by the _code_, while I meant changed by the _user_. So maybe `$forks_edited_by_user[]`, to clarify both: a) which post type; and b) who edited it? 

The forking I'm referring to is when the `wcb_sponsor` posts are created on new sites, based on the canonical `mes` posts, similar to forking a Git repo. To me forking doesn't imply changing, just the _option_ to change. Maybe it's best to not use a metaphor anyway, though, and instead do something like `$wcb_sponsor_posts[]`. That feels vaguely inelegant to me, but maybe less so? 🤷🏻‍♂️

ryelle: Yeah, the `fork` is vague enough that while it took a second to understand your meaning, it's fine since it's consistent (I understood the git reference, maybe I should have said "to me fork implies it's likely to have been changed"). I don't think you need to worry about changing that word.

What about `skipped_posts`? If you don't like that, I'm okay with the longer `forks_edited_by_user`.

public_html/wp-content/plugins/multi-event-sponsors/package.json
iandunn: `20.0.0` is required for `--hot`, which seems like a bigger benefit to me than linting. We could add it when we update the other packages though.

ryelle: okay, but you'll still need to add a lint:css script, otherwise the lint action will fail. `"lint:css": "echo \"No linter\""` or something should work

public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
iandunn: `skipped_posts` sounds good to me 👍🏻

PR #746, opened 2022-03-29
Jetpack Tweaks: Decode child selectors `>` in Custom CSS

### LINKED ISSUE #678: Custom CSS outputting HTML entities instead of correct characters
**Describe the bug**

I compared to a fresh WordPress install to verify, it seems like it's only happening on WordCamp sites. The `>` symbol is being output as `&gt;` in the customizer CSS output.

**To reproduce**

1. Log into a WordCamp site with admin access.
2. Click on Customize > Custom CSS.
3. Enter CSS with `>` selector. 
My example uses:
`#subscribe-submit > button {
	background: gray;
}` 
4. Notice preview updates.
5. Publish & look at Customizer output in inspector, `>` is replaced with `&gt;`

Might be related to #574?

**Expected behavior**

I would expect the output to have `>` since it's a proper CSS selector. Or if it must be sanitized out, showing an error message or something to indicate that it's not an allowed character would be nice.

**Screenshots**

I made a [screen recording](https://www.loom.com/share/210e3268becf43a7a535659b35b5e9fd) comparing the WordCamp site with a fresh WP install.

**WordCamp**

Noticed on [NEO](https://neo.wordcamp.org/2021/) but I'm guessing other sites have the same issue.

 - Device: MacBook
 - OS: MacOS
 - Browser: Firefox
 - Version: 86.0.1

ryelle:
Fixes #678 — The `>` in CSS is encoded as `&gt;` when output on WordCamp sites, which is invalid CSS. This fix is very simple, just replace the `&gt;` with `>` again on output. The sanitization already strips out any HTML-like tags, so this should not introduce any security issues.

Props @megmorsie, @CdrMarks.

1. Open the Additional CSS panel in the Customizer
2. Add CSS with an angle bracket, like
	```css
	body > div {
		background: lightcoral;
	}
	```
3. It should apply correctly in the customizer, so any divs that are direct children of body should have a pink background now.
4. Save and exit the customizer, view the site
5. It should still have the background on the live site
6. Alternately, view `https://yoursite.wordcamp.org/?custom-css` and the CSS should be there, with the `>` not encoded.

I also added two simple tests, you can see it fail if you disable the filter, and pass again after the filter is enabled.

### DIFF
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
@@ -12,6 +12,7 @@ add_action( 'csstidy_optimize_postparse', __NAMESPACE__ . '\sanitize_csstidy_par
 add_action( 'admin_notices',              __NAMESPACE__ . '\notify_import_rules_stripped'  );
 add_action( 'csstidy_optimize_subvalue',  __NAMESPACE__ . '\sanitize_csstidy_subvalues'    );
 add_action( 'safecss_parse_pre',          __NAMESPACE__ . '\update_csstidy_safelist', 0    );
+add_filter( 'wp_get_custom_css',          __NAMESPACE__ . '\filter_custom_css' );
 
 /**
  * Sanitize CSS saved through the Core/Jetpack editor inside the Customizer
@@ -221,4 +222,17 @@ function update_csstidy_safelist() {
 	if ( ! empty( $GLOBALS['csstidy']['all_properties'] ) ) {
 		$GLOBALS['csstidy']['all_properties'] = array_merge( $GLOBALS['csstidy']['all_properties'], $properties_for_csstidy );
 	}
+}
+
+/**
+ * Replace encoded &gt; in CSS with a > so that CSS rules are valid.
+ *
+ * Remove this once the root issue is fixed in https://github.com/Automattic/jetpack/issues/21603.
+ *
+ * @param string $css CSS pulled in from the Custom CSS post type.
+ *
+ * @return string
+ */
+function filter_custom_css( $css ) {
+	return str_replace( '&gt;', '>', $css );
 }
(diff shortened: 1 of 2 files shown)


public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
```diff
@@ -222,3 +223,16 @@ function update_csstidy_safelist() {
...
+ * Replace encoded &gt; in CSS with a > so that CSS rules are valid.
+ *
+ * Remove this once the root issue is fixed in https://github.com/Automattic/jetpack/issues/21603.
+ *
+ * @param string $css CSS pulled in from the Custom CSS post type.
+ *
+ * @return string
+ */
+function filter_custom_css( $css ) {
+	return str_replace( '&gt;', '>', $css );
```
coreymckrill, line 237: Are there any other characters this could happen to? Would it make sense to run something like `html_entity_decode()` here instead of a simple string replace?

coreymckrill (approved):
👍 Works on the sandbox. Took a lot of effort to get the tests to run locally 😅 but that looks good too. One minor comment, but not a blocker.

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
ryelle: I was hesitant to do that at first because I'm not totally sure why it's encoded, if there would be a security implication to decoding everything. Looking at [the different selectors available](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors), no others are encoded, but it would be possible to add anything to an attribute selector, which could get encoded (like `a[href*='a=1&b=2]`, would become `a[href*='a=1&amp;b=2]`). I'm not sure if supporting that edge case is worth any security issue (but… maybe there is no issue here?)

I updated the test to check against almost all the selectors available.

ryelle: I'm going to merge this as-is, if other selectors become a problem we can make the switch then.

PR #747, opened 2022-03-29
Jetpack tweaks: Update to CSSTidy allowed properties

ryelle:
- Jetpack added the `mask` properties as part of https://github.com/Automattic/jetpack/pull/15044, so they can be removed from the overrides.
- Add the `gap` and `row-gap` properties (`column-gap` is already allowed)
- Add [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) for spacing based on `writing-mode` (also added).

I also updated the `unit_value` and `shorthand` configuration objects. The properties in `unit_value` will be given a unit (`px`) if they're saved without, because unitless values are invalid for those properties. It looks like `optimise_shorthands` is off, but I added the new values & their shorthands just in case.

1. Try using any of the new properties
2. Run the tests by following [the phpunit directions](https://github.com/WordPress/wordcamp.org/tree/production/.docker#testing-with-phpunit) and running `phpunit --group=remote-css`


iandunn (approved):
LGTM 👍🏻

PR #763, opened 2022-05-08
Change unencrypted Git protocol to https

CdrMarks:
On March 15, 2022, GitHub [removed](https://github.blog/changelog/2022-03-15-removed-unencrypted-git-protocol-and-certain-ssh-keys/) the unencrypted Git protocol.  This PR changes the protocol to https.

### The error when accessing using the removed git protcol
```
$ git clone git@github.com:WordPress/wordcamp.org.git wordcamp.test
Cloning into 'wordcamp.test'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

1.  On your local machine, clone using this command:
> git clone https://github.com/WordPress/wordcamp.org.git wordcamp.test

### DIFF
--- a/.docker/readme.md
+++ b/.docker/readme.md
@@ -6,7 +6,7 @@ Follow these steps to setup a local WordCamp.org environment using [Docker](http
 
 1. Clone the repo:
     ```bash
-    git clone git@github.com:WordPress/wordcamp.org.git wordcamp.test
+    git clone https://github.com/WordPress/wordcamp.org.git wordcamp.test
     cd wordcamp.test
     ```
 


ryelle:
I think now that they've deprecated the unencrypted protocol & less secure keys, you need to be using [a valid SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/about-ssh) to use that URL. I've updated [the directions](https://github.com/WordPress/wordcamp.org/tree/production/.docker#initial-setup) with some help for if you run into that error, how does that look?

CdrMarks:
The updated directions look good.  

I just realized that I the git protocol that GitHub deprecated was `git://github.com/<org>/<repo>.git` and not `git@github.com/<org>/<repo>.git`.  Looks like I misinterpreted the error.

I'm going to close this PR.

PR #774, opened 2022-05-25
CampTix: Ensure camptix scripts are loaded in the footer

### LINKED ISSUE #773: [Registration Form] Excessively strict validation and no error messages on the "Company or personal website" field
**Describe the bug**

The required formatting for the `Company or personal website` field is very strict (a simple example.com does not work it has to include the https://) but the problem is made worse by the fact that no error notices are shown (the page just scrolls back to that point) which can be very confusing. I know it took me a couple of attempts to figure it out. I am sharing a recording to illustrate it.

**To reproduce**

1. Go to https://europe.wordcamp.org/2022/tickets/
2. Click on "Register" to access the registration form
3. Fill in the form and enter an incorrect or incomplete URL on the `Company or personal website` such as example.com (without the `https://` part)
4. Scroll to the bottom and click "Submit".
5. The form will not be submitted, but not error notice is being shown either.

**Expected behavior**

I expected the form to recognise `example.com` as a valid URL. Not recognising example.com as a valid URL feels unnecessarily strict.

I also expected to see error messages next to the form fields that were incorrect. The page scrolls back to the field that needs reviewing but there are no hints on what needs to corrected.

**Screenshots**

https://user-images.githubusercontent.com/4452464/170238814-6efbe9e0-52d0-4396-a2aa-8c1294b2d6c0.mp4

**WordCamp**

I noticed this on the WCUE registration form, but I suspect the form is probably the same one on other WordCamp websites.

ryelle:
Currently, the payment options script is not set to load in the footer, so when it's added to the page, it pulls `camptix.js` up into the header. The `camptix.js` script expects to be called in the footer though, and some code will silently fail, like the URL formatting.

Moving this down also affected the `camptix-invoice` script, so that's been refactored to check the status of the checkbox, rather than `toggle` the default state (because with this change, it ended up in the opposite state).

Fixes #773.
Props @mrfoxtalbot.

⚠️  I've tried testing this, but I don't know if there will be other cascade effects for other camptix addons.

### Screenshots

URL validation

https://user-images.githubusercontent.com/541093/170320109-08c3cba2-078b-40f6-9b7e-9c0ee96e1a8c.mp4

If you add a company, when you click out it becomes obvious it should be a URL:

https://user-images.githubusercontent.com/541093/170320501-f162a9f5-4215-4c8e-b56b-5b7a370f7717.mp4

_If you have a WordCamp sandbox, sync this to your sandbox and try using the WordCamp Europe ticket form, start from step 5._

1. Enable up payments (can be sandboxed)
2. Create a ticket
3. Add a website/URL question to the ticket
4. Set a price on the ticket
5. Go to the tickets page and buy a ticket
6. Try adding any value to the website field
7. When you click out of it, it should prepend `http://` to the value
8. Delete it, try adding another value with `http://` or `https://` already set
9. When you click out of it, it should leave the value alone
10. Click through the rest of the purchase process

Enable the CampTix Invoices plugin, turn invoices on in Tickets > Setup > Invoices.

1. Try buying a ticket
2. Toggle on & off the invoice checkbox
3. The fields should display correctly, and are required when they're visible

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/addons/class-payment-options.php
@@ -34,7 +34,8 @@ class Payment_Options extends CampTix_Addon {
 			'payment_options',
 			plugins_url( 'js/payment-options.js', __FILE__ ),
 			array( 'stripe-checkout', 'camptix' ),
-			filemtime( __DIR__ . '/js/payment-options.js' )
+			filemtime( __DIR__ . '/js/payment-options.js' ),
+			true
 		);
 
 		wp_register_style(
--- a/public_html/wp-content/plugins/camptix-invoices/admin/js/camptix-invoices.js
+++ b/public_html/wp-content/plugins/camptix-invoices/admin/js/camptix-invoices.js
@@ -1,12 +1,22 @@
-jQuery( document ).ready( function ($) {
+/* global jQuery */
 
-	$(document).on( 'change', '#camptix-need-invoice', toggleInvoiceDetailsForm );
-	function toggleInvoiceDetailsForm() {
-		var $camptixInvoiceDetailsForm = $( '.camptix-invoice-details' );
-		$camptixInvoiceDetailsForm.toggle();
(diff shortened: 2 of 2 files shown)


iandunn (approved):
LGTM 👍🏻

PR #787, opened 2022-07-23
Add safelisted properties to Custom CSS sanitizer

0aveRyan:
In building the WCUS 2022 website, there were a number of CSS properties I was hoping to use for broad-audience (`text-stroke`) and progressive-enhancement (`overscroll-behavior`).


public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
```diff
@@ -244,12 +244,19 @@ function get_custom_css_properties_safelist() {
 		array(
 			'aspect-ratio',
 			'background-blend-mode',
+			'box-orient',
```
ryelle, line 251: This is [a nonstandard property](https://developer.mozilla.org/en-US/docs/Web/CSS/box-orient) from a previous flexbox spec, is there a standard one that corresponds to it? `flex-direction` maybe?

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
```diff
@@ -244,12 +244,19 @@ function get_custom_css_properties_safelist() {
...
 			'isolation',
+			'line-clamp',
```
ryelle, line 253: It looks like this [only works with the prefixed version](https://caniuse.com/css-line-clamp), which would still be stripped out. Should `-webkit-line-clamp` be added too?

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
```diff
@@ -244,12 +244,19 @@ function get_custom_css_properties_safelist() {
...
 			'mask',
 			'mix-blend-mode',
+			'overscroll-behavior',
+			'place-content',
 			'shape-image-threshold',
 			'shape-outside',
+			'text-decoration-thickness',
 			'text-orientation',
+			'text-stroke',
+			'text-fill',
```
ryelle, line 263: These are [also non-standard](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-stroke), and appear to be supported only with a prefix. Should only the prefixed properties be added?

ryelle (commented):
I have a few questions on some of the properties, but the rest are okay.

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
0aveRyan: I was hoping to use this for a progressive-enhancement for supported browsers, in combination with `-webkit-line-clamp`. Unfortunately `flex-direction` and others aren't a substitute. This totally isn't essential, more nice-to-have.

This little combo gives the ability to limit excerpts, use a nicer ellipsis display and get the Read More buttons all in the same position
excerpt-limiting

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
0aveRyan: Ah the magic of devtools, missed that! I've pushed a commit with the `-webkit-*` prefix for the three properties that'll need 'em.

public_html/wp-content/mu-plugins/jetpack-tweaks/css-sanitization.php
0aveRyan: Ah the magic of devtools, missed that! I've pushed a commit with the `-webkit-*` prefix for the three properties that'll need 'em.

0aveRyan:
@ryelle thanks for taking a look at this! I commented and pushed ~a commit~ two commits with the `-webkit-line-clamp`, `-webkit-text-stroke`, `webkit-text-fill` and `webkit-box-orient` properties in the safelist as well.

PR #789, opened 2022-08-08
WIP - Add budget info to WordCamp Details report

iandunn:
This adds some budget information to the WordCamp Details report.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-wordcamp-details.php
+++ b/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-wordcamp-details.php
@@ -110,7 +110,19 @@ class WordCamp_Details extends Base_Details {
 			[
 				'Name',
 			],
-			WordCamp_Loader::get_public_meta_keys()
+			WordCamp_Loader::get_public_meta_keys(),
+			self::get_public_custom_data_keys()
+		);
+	}
+
+	// document that these are things that need to be added w/ custom code, not as simple as just a meta value
+	public static function get_public_custom_data_keys() : array {
+		return array(
+			// document that should be transparent, see https://make.wordpress.org/community/handbook/wordcamp-organizer/first-steps/budget-and-finances/#transparency
+			'Total Approved Expenses',
+			'Total Working Expenses',
+			'Approved Budget Surplus/Deficit',
+			'Working Budget Surplus/Deficit',
 		);
 	}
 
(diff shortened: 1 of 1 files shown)


iandunn:
The WordCamp Details report is only setup to pull in `wordcamp` post type meta data, but the budget data is in `wp_options`. The plugin has a lot of abstractions, so I'm trying to figure out the best way to fit the new data into it. The stubs in 43a87d88 seem like a good approach to me. @coreymckrill, do you have any thoughts on that architecture?

iandunn:
Another complication here is that many of the income/expenses are based on other fields. For example, the venue cost could be per day, or the food cost could be per attendee.

When they are, getting the "real" amount of each entry involves linking those pieces of data together.

https://github.com/WordPress/wordcamp.org/blob/941d5f9b16eb1731ab7c2bd5548d56e346b7506f/public_html/wp-content/plugins/wordcamp-payments/javascript/budget-tool.js#L40-L48

https://github.com/WordPress/wordcamp.org/blob/941d5f9b16eb1731ab7c2bd5548d56e346b7506f/public_html/wp-content/plugins/wordcamp-payments/javascript/budget-tool.js#L402-L425

So, that might mean duplicating all the JS calculations in PHP. That might be the simplest option, but adds technical debt and risks the reports being silently inaccurate if the JS side is updated without syncing to the PHP side.

Another idea would be refactoring so that the JS side makes a API request to PHP for the calculations. That would be a good amount of work, and would slow down the UI.

A third option would be for the JS side to save the calculated summary amounts in the database along with the raw data. This is even simpler than duplicating the calculations in PHP, but would require back-filling the past ~7 years of budget data. Doing that manually would take a long time and be very tedious, and automating it could be complicated & time consuming.

I'm leaning towards `#3`, but am hoping to find a good way to script it. 🤔

coreymckrill:
> do you have any thoughts on that architecture?

In retrospect, I regret the extra level of class inheritance that [got added](https://github.com/WordPress/wordcamp.org/commit/217b4dc3be343eec037460b5b536550f5e291482). That's probably making this harder. That said, the stub looks like a reasonable approach to me.

> Another complication here is that many of the income/expenses are based on other fields.
> ...
> Another idea would be refactoring so that the JS side makes a API request to PHP for the calculations. That would be a good amount of work, and would slow down the UI.

This does sound like the most work-intensive option. But, probably the most future-proof since it would put the source-of-truth for all the budget numbers in one server-side place. I doubt the performance hit to the UI would be problematic. That being said, a change like this may not be worth the amount of required effort and time right now, unless there are other corollary benefits for budgets and/or reports, so I'd certainly understand if you picked one of the other options.

iandunn:
🤔 Yeah, I'm going back and forth. They both have aspects I like and those I don't. I'm gonna check w/ the team on priorities, because I agree that it may not be worth it at this stage.

pkevan:

PR #800, opened 2022-10-24
Clarify "No comma" in payment amount description

naokomc:
In some languages, commas are used for decimal separators. We recently received a payment request omitting the comma, so the result request was 100 times more amount than intended (800,00 was changed to 80000, instead of 800.00). I think we should ask not to include thousands separators or currency symbols, and use a period for the decimal separator.


iandunn:
Thanks!

PR #801, opened 2022-11-08
Keep the path component when redirecting from the root-domain to a sub-path.

dd32:
While testing some login changes, I ran into a slight issue - `asia.wordcamp.org/tickets` will redirect to `/2023/tickets` when linked to from asia.wordcamp.org, but not when linked to from outside.

This was intentional, links to `/tickets` should only be present on older sites which were migrated from `{year}.{city}.wordcamp.org` domains, but having the entire path of a request such as `asia.wordcamp.org/tickets` be lost during the redirect to `asia.wordcamp.org/2023/` feels slightly wrong. It would be better if it could retain that path component.

The above example is a contrived example, in my case I ran into it with `brisbane.wordcamp.org/wp-login.php` redirecting to `brisbane.wordcamp.org/2020/` and causing the remote-login code I was working on to fail. In my case, I'll fix this properly in a different manner, but brought up that supporting the above use-case would be worthwhile IMHO.

One downside of this would be that a link to `city.wordcamp.org/a-page-that-doesnt-exist` would end up on a 404 of `city.wordcamp.org/year/a-page-that-doesnt-exist` but I don't think that's a major concern.

1. Apply.
2. Request URL that redirects to current year
3. Observe "lost" path components.

### DIFF
--- a/public_html/wp-content/sunrise.php
+++ b/public_html/wp-content/sunrise.php
@@ -111,7 +111,7 @@ function main() {
 	/*
 	 * This has to run before `get_canonical_year_url()`, because that function will redirect these requests to
 	 * the latest site instead of the intended one, and would strip out the request URI.
-	 * 
+	 *
 	 * This branch specifically only processes root-relative urls which are linked to from the same domain.
 	 */
 	if ( ! $redirect ) {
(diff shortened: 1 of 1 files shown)


iandunn (approved):
LGTM 👍🏻

PR #803, opened 2022-11-29
Add support for Jetpack 11.6

dd32:
See https://github.com/Automattic/jetpack/pull/27187 for the rename/migration.

PR #805, opened 2022-12-05
Blocks: Return error when showing sessions inside session.

iandunn:
If the Session block is added to a `wcb_session` post, it leads to an infinite loop when the current session is in the array returned by `get_session_posts()`.

I can't think of any use cases for doing that, I think it was just a misunderstanding of what the block does. Given that, it's probably best to prevent the block from being inserted into a Session post.

I didn't see a good way to do that via Gutenberg, though. `unregisterBlockType()` feels kind of clunky, and there may be cases where we need to leave it registered (like a loop of several posts where some can have the block and others can't). `hideBlockTypes()` would add it to the user's preferences permanently.

I ended up just returning an error message, but am open to a better way.

See https://wordpress.slack.com/archives/C08M59V3P/p1670236017716039


iandunn:
The test failure is related to https://github.com/WordPress/wordcamp.org/pull/803

PR #806, opened 2022-12-08
Disable certain Jetpack modules which do not operate correctly on WordCamp.org

dd32:
This disables the Jetpack Downtime monitor, and the Firewall that's in Beta.

Jetpack Monitor alerts WordCamp site owners, and the WordCamp community team, but neither can do anything to resolve downtime issues. Our general infrastructure checks validate that WordCamp.org is operational.

Jetpack Firewall/WAF requires the ability to write to local files, which is not possible on the WordCamp infrastructure. We also have other server-level security measures in place.

### DIFF
--- a/public_html/wp-content/mu-plugins/disable-modules.php
+++ b/public_html/wp-content/mu-plugins/disable-modules.php
@@ -1,0 +1,15 @@
+<?php
+namespace WordCamp\Jetpack_Tweaks\DisabledModules;
+
+defined( 'WPINC' ) || die();
+
+add_filter( 'jetpack_get_available_modules', __NAMESPACE__ . '\disable_modules' );
+function disable_modules( $modules ) {
+	// WordCamp infrastructure has monitoring in place which alerts those who can resolve downtime issues.
+	unset( $modules['monitor'] );
+
+	// Not supported on the WordCamp infrastructure.
+	unset( $modules['waf'] );
+
+	return $modules;
+}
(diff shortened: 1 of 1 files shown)


CdrMarks (approved):

public_html/wp-content/mu-plugins/disable-modules.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
iandunn, line 1: I think this would be more consistent in `mu-plugins/jetpack-tweaks/miscellaneous.php` or `jetpack-tweaks/new-sites.php`. wordcamp.org tends to organize mu-plugins into categories, rather than make lots of small ones like w.org. That feels less cluttered IMO.

`new-sites.php` isn't an exact fit, but we already have a `jetpack_get_default_modules` callback there, and having these two callbacks side by side would make it more obvious that both exist.

https://github.com/WordPress/wordcamp.org/blob/f8d1ba751855f389984077af1f10a464af91d916/public_html/wp-content/mu-plugins/jetpack-tweaks/new-sites.php#L13-L21

Alternatively, moving it inside the `jetpack-tweaks/` folder would help keep it organized. Not a blocker, though.

CdrMarks: I had forgotten about the `jetpack-tweaks/` folder.  That location seems ideal.

iandunn: Ah, looks like it was already moved in fa2f6e4e73af8dc0bbed0831b7a2f6f3bd90b250 👍🏻

PR #817, opened 2023-01-17
Fix Jetpack fatals on update

pkevan:
Since Jetpack removed the 4.7 version in JP 11.6, updating Jetpack causes fatals due to these hardcoded includes no longer being in the plugin.

See: https://wordpress.slack.com/archives/GDDSW0WNS/p1673447878210639


iandunn (approved):
looks good, thanks for catching this 👍🏻 

related #803

Issue #771, opened 2022-05-20
Migrate caching to Surge?


iandunn:
Our own @kovshenin released [a lean caching plugin](https://wordpress.org/plugins/surge/) that looks really good. It'd be simpler and cleaner than WPSC, so I think we should consider migrating.

We could [use nginx microcache](https://konstantin.blog/2021/page-caching-filesystem/) if want to go even further, but that may not be worth it.

Any thoughts @ryelle @dd32 @coreymckrill?

dd32:
What's the intention behind it? Something simpler and with less options?

I'm unaware of any of the existing shortcomings or issues that the current plugin has though, so I tend to err on the side of don't touch it if it's working.

WordPress.org pretty much exclusively uses nginx caching, which doesn't have purge support, but is only for logged out users and short cache ttl's.

kovshenin:
Thanks for considering this! I'll be happy to help with the setup (there isn't much, really), and any custom configuration you might need, like invalidating things on certain custom events.

iandunn:
> I tend to err on the side of don't touch it if it's working

Yeah, I think that's a good instinct in general.

> What's the intention

The aren't any tangible problems at the moment, but in the past we've run into a few:

* https://github.com/WordPress/wordcamp.org/commit/35956d178aa7afb2c91dc2854352eb57f95eb49e
* https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/wp-super-cache-plugins/http-https-redirect.php
* `wp-cache-config.php` requires [periodic maintenance](https://github.com/WordPress/wordcamp.org/commits/production/public_html/wp-content/wp-cache-config.php) as well. The old SVN repo has a handful of additional entries as well.

The simplicity of Surge is appealing as a way to minimize the chances of that stuff in the future, and to remove some of the customizations we've added with something that "just works".

It'd also give us caching for `wp-json` endpoints, and would be a devex improvement IMO.

There's always the chance that it could introduce new problems, though. [The lack of preloading](https://github.com/kovshenin/surge/issues/21) would probably be a performance hit, too. We could mitigate it to some extent w/ a longer TTL.

I think it'd be nice, but don't feel strongly.

dd32:
> It'd also give us caching for `wp-json` endpoints, and would be a devex improvement IMO.

That alone would make it worth investigating IMHO. Especially given some current (and many future) things are likely built on top of it.

timiwahalahti:
[...]

PR #818, opened 2023-01-19
Reduce amount of results in dropdowns

### LINKED ISSUE #816: High memory usage in various /report pages
**Describe the bug**

Alerts are consistently occurring when visiting various report pages e.g. https://central.wordcamp.org/reports/payment-activity-report/

**To reproduce**

1. Go to https://central.wordcamp.org/reports/payment-activity-report/
2. Observe memory usage through a plugin such as query monitor

**Expected behavior**

Less memory used :)

pkevan:
Fixes #816

By reducing the amount of results by default (100), and narrowing results based on year input, the memory usage on the page is dramatically reduced.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
+++ b/public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
@@ -47,7 +47,14 @@ use WordCamp\Reports\Report;
 
 		<div class="field_wordcamp-id">
 			<label for="wordcamp-id">WordCamp <span>(optional)</span></label>
-			<?php echo get_wordcamp_dropdown( 'wordcamp-id', array(), $wordcamp_id ); ?>
+			<?php if ( ! isset( $_GET[ 'action' ] ) ) {
+				// Show only latest 100 WordCamps by default
+				echo get_wordcamp_dropdown( 'wordcamp-id', array( 'posts_per_page' => 100, 'order' => 'DESC' ), $wordcamp_id );
+			} else {
+				// If something is being searched, return WordCamps for that year, or the default latest year, to reduce memory usage
+				$wc_year = ( isset( $years ) ) ? $years : date( 'Y' );
+				echo get_wordcamp_dropdown( 'wordcamp-id', array( 'date_query' => array( 'year' => $year, 'compare' => '=' ) ), $wordcamp_id ); 
+			} ?>
 		</div>
 
 		<div class="submit_show-results">
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "63a0644303c0f977d3bebceafbb0496f",
+    "content-hash": "58dd76c84ca37d77ba5781dec61d1989",
     "packages": [
         {
             "name": "adhocore/jwt",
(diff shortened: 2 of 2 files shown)


public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
```diff
@@ -47,7 +47,14 @@
 
 		<div class="field_wordcamp-id">
 			<label for="wordcamp-id">WordCamp <span>(optional)</span></label>
-			<?php echo get_wordcamp_dropdown( 'wordcamp-id', array(), $wordcamp_id ); ?>
+			<?php if ( ! isset( $_GET[ 'action' ] ) ) {
+				// Show only latest 100 WordCamps by default
```
iandunn, line 51: Pre-COVID, we had around 150 camps per year, so I'd lean towards something more like `300` or `400`, which should still be totally fine for memory. 

We only recently started noticing the memory warnings, and it's not running out of memory yet, so I'm hesitant to make it harder for folks to access the data. I'm guessing we could do something like `800` safely, and it may be better to make the minimal necessary impact.

public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
```diff
@@ -47,7 +47,14 @@
...
+				echo get_wordcamp_dropdown( 'wordcamp-id', array( 'posts_per_page' => 100, 'order' => 'DESC' ), $wordcamp_id );
+			} else {
+				// If something is being searched, return WordCamps for that year, or the default latest year, to reduce memory usage
+				$wc_year = ( isset( $years ) ) ? $years : date( 'Y' );
+				echo get_wordcamp_dropdown( 'wordcamp-id', array( 'date_query' => array( 'year' => $year, 'compare' => '=' ) ), $wordcamp_id ); 
+			} ?>
```
iandunn, line 57: So if someone wanted to view WCUS 2018 info, would they have to do this?

1. Select `2018`
2. Select `Entire Year`
3. Press `Show Results`
4. Select the camp from the dropdown now that it's there
5. Press `Show Results` again

I'm guessing most folks won't realize they need to do that. We might need to add some instructions, or consider a more intuitive flow.

If [caching](https://github.com/WordPress/wordcamp.org/issues/816#issuecomment-1397354654) would solve the problem without an UI/UX changes, I think that'd be preferable.

public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
pkevan: > only recently started noticing the memory warnings, and it's not running out of memory yet

that's fair, although i'm always a little nervous when it almost reaches the limit, and could be knocked over by another ~100 WordCamps :)

public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
pkevan: > If https://github.com/WordPress/wordcamp.org/issues/816#issuecomment-1397354654 would solve the problem

let me re-look at a caching option and see what that looks like.

I'm unsure what the normal workflow here is (presumably it's in stats somewhere), although it does suggest that the dropdown is optional, and also if you search by a single year, you do get different year results - presumably based on when the site was created?

iandunn:
I just remembered that the reports are already using transient caching via `maybe_get_cached_data()`. If you were referring to static page caching, we use WP Super Cache for that (related #771). But these reports are also accessed in `wp-admin`, so those wouldn't get static page caching.

public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
```diff
@@ -47,7 +47,14 @@
 
 		<div class="field_wordcamp-id">
 			<label for="wordcamp-id">WordCamp <span>(optional)</span></label>
-			<?php echo get_wordcamp_dropdown( 'wordcamp-id', array(), $wordcamp_id ); ?>
+			<?php if ( ! isset( $_GET[ 'action' ] ) ) {
+				// Show only latest 100 WordCamps by default
+				echo get_wordcamp_dropdown( 'wordcamp-id', array( 'posts_per_page' => 100, 'order' => 'DESC' ), $wordcamp_id );
```
iandunn, line 52: 🤔 Are there other reports that we should make this change to? Why would `payment-activity` be the only one needed when they all call `get_wordcamp_dropdown()` with the default `$query_options`? Ah, searching for `in:#dotorg-wordcamp-logs peak memory usage` does show others having issues.

Maybe the change should be deeper, in `get_wordcamp_dropdown()` itself? Or more DRY, like by having all the `wordcamp-reports/` call it with a `WordCamp\Reports\Report\Base::get_dropdown_options()` argument?

It might also be worth considering reverting e058b4bb109c134fb006d8f1c6380a05171910e1 , so that the logs aren't cluttered with false positives. I enabled that after `95`% was cleaned up, just to see if there were others getting close, but `90` may be too low in practice.

public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
iandunn: 👍🏻 

If caching doesn't help, another option would be to refactor the flow. instead of populating the dropdown during the initial page load, we could do it async via XHR every time the year or time period changes. Then the XHR could only fetch the camps during that year.

Actually, just fetching the data via XHR without any year filtering might be enough for now, since the data needed to compile the page would be split across two processes. We may eventually reach a point where we can't show all camps, but I'd be surprised if we're anywhere close to that with only `1400` sites. 

🤔 That makes me think that [reducing the data](https://github.com/WordPress/wordcamp.org/issues/816#issuecomment-1398465282) is worth another look, since it seems like the best option if it works.

public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
iandunn: Yeah, that's fair, it'd be good to do something now so that it doesn't get worse 👍🏻 

Right now it seems like some of the other approaches are better (reducing the data, caching, fetching async), but if we have to then limiting the # of camps could work.

public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
pkevan: > 🤔 Are there other reports that we should make this change to?

Yes, I was mainly limiting the impact of any change being suggested here - should have said that in the original issue

pkevan:
Never got around to doing much with this, so going to close but keep in mind.

PR #820, opened 2023-01-20
Fix/816 2

### LINKED ISSUE #816: High memory usage in various /report pages

pkevan:
Second attempt to temporarily fix #816

### DIFF
--- a/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
+++ b/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
@@ -287,11 +287,11 @@ function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(),
 				$wordcamps_cached[] = (object) array(
 					'ID' => $wordcamp->ID,
 					'post_title' => $wordcamp->post_title,
-					'meta' => array( 
-						'Start Date (YYYY-mm-dd)' => array( 
-							$meta 
-						)
-					)
+					'meta' => array(
+						'Start Date (YYYY-mm-dd)' => array(
+							$meta,
+						),
+					),
 				);
 
 				?>
--- a/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
+++ b/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
@@ -252,7 +252,7 @@ function wcorg_get_wordcamp_duration( WP_Post $wordcamp ) {
 function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(), $selected = 0, $cached = false ) {
 	$cache_key = 'wc_dropdown_' . md5( serialize( $query_options ) );
 	$cached_data = get_transient( $cache_key );
-	
+
 	if ( $cache_data && $cached ) {
 		$wordcamps = $cached_data;
 	} else {
@@ -284,7 +284,15 @@ function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(),
 					echo ' ' . esc_html( gmdate( 'Y', $wordcamp->meta['Start Date (YYYY-mm-dd)'][0] ) );
 					$meta = $wordcamp->meta['Start Date (YYYY-mm-dd)'][0];
 				}
-				$wordcamps_cached[] = (object) array( 'ID' => $wordcamp->ID, 'post_title' => $wordcamp->post_title, 'meta' => array( 'Start Date (YYYY-mm-dd)' => array( $meta ) ) );
+				$wordcamps_cached[] = (object) array(
+					'ID' => $wordcamp->ID,
+					'post_title' => $wordcamp->post_title,
+					'meta' => array( 
+						'Start Date (YYYY-mm-dd)' => array( 
+							$meta 
+						)
+					)
+				);
 
 				?>
 			</option>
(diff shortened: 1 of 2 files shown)


public_html/wp-content/mu-plugins/4-helpers-wcpt.php
```diff
@@ -245,16 +245,25 @@ function wcorg_get_wordcamp_duration( WP_Post $wordcamp ) {
  * @param string $name          Optional. The `name` attribute for the `select` element. Defaults to `wordcamp_id`.
  * @param array  $query_options Optional. Extra arguments to pass to `get_posts()`. Defaults to the values in `get_wordcamps()`.
  * @param int    $selected      Optional. The list option to select. Defaults to not selecting any.
+ * @param bool   $cached        Optional. Cache the data. Default false.
  *
  * @return string The HTML for the <select> list.
  */
-function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(), $selected = 0 ) {
-	$wordcamps = get_wordcamps( $query_options );
+function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(), $selected = 0, $cached = false ) {
```
iandunn, line 252: What's the reason to default to `false`? I don't see the downside of defaulting to `true`, and it'd be faster.

Actually, I'm not sure there are any current use cases for passing `true`. Do you see one?

public_html/wp-content/plugins/wordcamp-reports/views/public/payment-activity.php
```diff
@@ -47,7 +47,7 @@
 
 		<div class="field_wordcamp-id">
 			<label for="wordcamp-id">WordCamp <span>(optional)</span></label>
-			<?php echo get_wordcamp_dropdown( 'wordcamp-id', array(), $wordcamp_id ); ?>
+			<?php echo get_wordcamp_dropdown( 'wordcamp-id', array(), $wordcamp_id, true ); // phpcs:ignore -- escaped in function. ?>
```
iandunn:
```suggestion
			<?php echo get_wordcamp_dropdown( 'wordcamp-id', array(), $wordcamp_id, true ); // phpcs:ignore -- escaped in calling function. ?>
```

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
```diff
@@ -245,16 +245,25 @@ function wcorg_get_wordcamp_duration( WP_Post $wordcamp ) {
...
+	$cache_key = 'wc_dropdown_' . md5( serialize( $query_options ) );
+	$cached_data = get_transient( $cache_key );
+
+	if ( $cache_data && $cached ) {
```
iandunn:
```suggestion
	if ( $cached_data && $cached ) {
```

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
```diff
@@ -245,16 +245,25 @@ function wcorg_get_wordcamp_duration( WP_Post $wordcamp ) {
...
+		$wordcamps = $cached_data;
+	} else {
+		$wordcamps = get_wordcamps( $query_options );
```
iandunn, line 259: Do we still risk running out of memory here, because we're initially loading the full results into RAM? It might be safer to only fetch the data we need in the first place.

The `fields` argument of `get_posts()` unfortunately only accepts `ids`, so we'd probably have to use a raw query, but that's fine IMO.

I think that'd also mean we could remove the code below that creates `$wordcamps_cached`.

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
```diff
@@ -270,9 +279,20 @@ function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(),
 				<?php
 
 				echo esc_html( $wordcamp->post_title );
+				$meta = '';
```
iandunn, line 282:
```suggestion
				$start_date = '';
```

That seems more descriptive to me. At first I was confused because I thought this was meant to represent the full array, and thought there might be a bug until I slowed down and read the code closely.

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
pkevan: only so it didn't break anything that was calling the function without the parameter, but I can see how setting it to `true` might be better.

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
pkevan: yeah - I did a prototype of calling `get_posts` with just the `ids` parameter, and what happened was less memory but 1900 queries! mainly due to how you access individual fields, and it has to call `get_post` each time - plus the lack of memcache causes high load since the query isn't ever in memory.

if we're not adverse to a custom query, that might be best.

pkevan: oh, I think one thing that was stopping me from a direct query was the `title` field, which if we do a direct query we are unable to apply any title filters that might exist, not sure how important that is?

iandunn: I think skipping the filters would be fine so that we can do a direct query, especially because we're already manually appending the year.

pkevan:
Given the feedback and discussions, going to close this, and continue in #826

PR #821, opened 2023-01-22
Give option to show description below text field

GautamMKGarg:
Earlier, there was option to show description below field_yesno but there was no option to show description below field_text and field_textarea. This code will allow to show description below these fields. Some payment gateways needs to show description below fields.

Fixes #, See #

Props <username>, <username>

1. 
`$this->add_settings_field_helper( 'title', 'Title', [ $this, 'field_text' ], 'Description' );`

Above code in `payment_settings_fields ` function of Payment gateway integration class will now show Description below the text field

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -1804,12 +1804,20 @@ class CampTix_Plugin {
 	function field_text( $args ) {
 		?>
 		<input type="text" name="<?php echo esc_attr( $args['name'] ); ?>" value="<?php echo esc_attr( $args['value'] ); ?>" class="regular-text" />
+
+		<?php if ( isset( $args['description'] ) ) : ?>
+		<p class="description"><?php echo wp_kses_data( $args['description'] ); ?></p>
+		<?php endif; ?>
 		<?php
 	}
 
 	function field_textarea( $args ) {
 		?>
 		<textarea class="large-text" rows="5" name="<?php echo esc_attr( $args['name'] ); ?>"><?php echo esc_textarea( $args['value'] ); ?></textarea>
+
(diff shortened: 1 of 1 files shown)


GautamMKGarg:
Hello @pkevan 
Could you please check and share your feedback on this?

dd32:
@GautamMKGarg Currently camptix is only intended on being used on ~WordPress.org~ WordCamp.org, and the available payment gateways are limited.

While I have no reason against adding this, it's not currently something that's needed.

You might as well also add it to the other `field_*` methods, even if you don't need it there.

GautamMKGarg:
@dd32 These 2 are reusable methods and are getting used not only in Payment Gateways but other settings also.
Same code is already there for field_yesno method but not for these 2 methods (field_text and field_textarea). This code is doing nothing much but enhancing experience for admins. By providing admins proper instruction below fields, it will become easier for admins to understand what to fill in those fields.

GautamMKGarg:
Regarding Payment Gateways, we have extended support for CampTix in Knit Pay WordPress Plugin. Knit Pay is a plugin, using which almost any payment gateway can be integrated with almost any WordPress Plugin or Theme. It means, using Knit Pay, WordCamp organisers will be able to integrate almost any payment gateway with CampTix supported in their country.
https://wordpress.org/plugins/knit-pay/

There are many payment gateways, which are currently not supported in CampTix. For example, CCAvenue, PayU, Paytm are very famous payment gateways in India, but these are not supported in Camptix currently. But with the help of Knit Pay, these will be supported. Knit Pay supports more than 200 payment gateways.

This could be another reason for merging this pull request. By showing description below text field, we will be able to guide admins, how to configure these settings.

timiwahalahti:
Going to close this, as there seems to be no feature request (issue) from the Community Team members for this change.

PR #831, opened 2023-02-17
Fix `composer/installers` not being an allowed plugin

2ndkauboy:
When running the Docker command to build the `php-fpm` container, it fails at the `composer install` step. Running the Docker command a second time would not even fix it, as the `composer install` would not even run again.

This change makes sure that the composer plugin is always allowed and the install works.


iandunn:
Looks good, thanks!

PR #843, opened 2023-03-06
Loop through queries in steps

pkevan:
Instead of calling up to 10,000 attendees, loop through in blocks of 1,000

Should hopefully reduce memory consumption which is alerting.


iandunn (approved):
LGTM 👍🏻

PR #853, opened 2023-03-20
Payments: Add field for PO/SCON number

### LINKED ISSUE #617: Vendor payment request form: add field for reference number
Imported from https://meta.trac.wordpress.org/ticket/4302 
Created by @timiwahalahti:

In the vendor payment request form, there should be a field for the reference number because multiple vendors do want their invoices paid with it to automate their bookkeeping. Now there's no field for it and reference number and note about using it needs to be filled into notes.

iandunn:
Fixes #617 

- Sponsor Invoices: Add a field for internal reference numbers.
- Payments Network: Link to handbook when there are extra instructions


timiwahalahti:
Probably fixes #617

PR #859, opened 2023-03-28
Coming Soon: Replace `jetpack_require_lib()` with it's replacement

dd32:
Jetpack 11.9 has removed `jetpack_require_lib()` in preference for direct file inclusions.

See https://github.com/Automattic/jetpack/pull/28866

This PR is **untested**.

I have deliberately **not** included `file_exists` checks and instead just relying upon the Jetpack Constant and a PHP Warning to be triggered if this file is ever removed / renamed.


pkevan:
Looks good, and tested with Jetpack 12.3

PR #863, opened 2023-04-03
Blocks: Fix styles for ItemSelect component

### LINKED ISSUE #855: Blocks: "Choose specific [posts]" interface CSS broken
The CSS for `react-select` is no longer applied to the select element.

I think this is because the editor is now inside an iframe, but the emotion css is still attached to the root page `head`.

**Screenshots**

Screenshot 2023-03-23 at 4 33 07 PM

If we want to keep using `react-select`, this is blocked by an upstream issue: https://github.com/WordPress/gutenberg/issues/38226

I'm going to try switching to ~FormTokenField~ SelectControl, which will also remove an external library.

ryelle:
Fixes #855 — Remove the `react-select` used in Organizers, Sessions, Speakers, and Sponsors blocks. This switches the behavior to simple multiple SelectControl.

The UX of react-select was nicer, when it was working, so a future iteration could bring back the token-style management of selected speakers.

### Screenshots

| Before | After |
|---|---|
| Screenshot 2023-03-23 at 4 33 07 PM |  |

The select section expands if there are a lot of options, maxing out at 40em:

1. Add one of these blocks: Organizers, Sessions, Speakers, or Sponsors
2. Select some values
3. It should be possible to select 1 or many items
4. If you select something from one group (ex, Organizer Teams) selecting from other groups is disabled
5. Click the Select button
6. The block should update to display those selected people/sessions/sponsors

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/package.json
+++ b/public_html/wp-content/mu-plugins/blocks/package.json
@@ -13,7 +13,6 @@
 	},
 	"dependencies": {
 		"classnames": "2.3.1",
-		"react-select": "5.2.2",
 		"rememo": "4.0.0"
 	},
 	"devDependencies": {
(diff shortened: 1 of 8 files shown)


StevenDufresne:
I got stuck getting the environment up on my new machine. 

If @iandunn doesn't get a chance to look at this, I'll try again to get it running.

iandunn (approved):
LGTM 👍🏻 

I agree that it could be nice to bring back the token-style management in the future.

Issue #739, opened 2022-03-09
Local custom fonts solution for WordCamp Sites


aurooba:
Pulling [this discussion](https://wordpress.slack.com/archives/C08M59V3P/p1646857677698749) out of the Make Slack into GitHub:

Now that Google Fonts has come under scrutiny from a privacy perspective, it would be really helpful if we had a way to upload/include custom fonts locally for WordCamp sites. There are [third-party plugins](https://wordpress.org/plugins/custom-fonts/) available that provide this capability, however, understandably, since this is a large MU network, whatever decision we arrive at needs to be given a lot of thought and consideration.

As we start planning for WordCamp US 2022, we want to be able to use custom fonts but still be as privacy-forward as we can be. I'm happy to help in whatever way I can to expedite this process. :)

ryelle:
Would [the webfonts work done in gutenberg](https://github.com/WordPress/gutenberg/pull/37140) help here at all?

timiwahalahti:
This might be a bigger issue, as we probably have multiple sites using Google hosted fonts because those are included [wc-fonts](https://github.com/WordPress/wordcamp.org/tree/production/public_html/wp-content/plugins/wc-fonts). In a perfect world, we'd switch Google hosted fonts on all existing sites to be hosted from somewhere else if we want to be extra cautious. What makes that _really_ hard, is the fact that wc-fonts for Google is just a textarea that allows @import's from fonts.googleapis.com 😅

Do I remember correctly @ryelle, that something related to Google fonts was done on the wordpress.org side?

aurooba:
> Would [the webfonts work done in gutenberg](https://github.com/WordPress/gutenberg/pull/37140) help here at all?

Perhaps! 

Although I had a simpler solution in mind: right now there is no way to upload custom font files into a site (.ttf, .woff, .woff2). If we had the ability to upload font file types, we could just link to them in an `@import` in `Additional CSS` or in `Remote CSS`.I know font files can pose a security risk, is there some way we can mitigate that?

ryelle:
I don't think we should force WordCamps to switch away from google fonts entirely. We can (probably) add the ability to load local fonts, and existing camps can opt in to using that instead.

> something related to Google fonts was done on the wordpress.org side?

Are you thinking of [the fonts mu-plugin](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/global-fonts)? This was added for use in wp.org themes, and only loads Inter & EB Garamond.

[...]

PR #868, opened 2023-04-13
WC Fonts: Register imported google fonts for use in the editor

ryelle:
Use the (experimental?) Fonts API to load selected Google Fonts in block-based themes.

When fonts are loaded via the WordCamp Fonts admin screen, there's no easy way to use them in the editor. The Typography settings don't know about these fonts, and without being able to update the theme.json, a user can't add them.

With this PR, if a site is using a block theme and the user has added a google font to Appearance > Fonts, the font is registered & enqueued using the Fonts API (rather than output in `wp_head`). It's also loaded into the editor, and listed in the available Typography > Font Family options. When selected, Gutenberg automatically applies style or class, and the font is used.

_While this does touch the use of google fonts, this PR does not address anything in #739._

**Tech details**

- This only works for Google Fonts because we can parse out the individual fonts.
- I've created a new font provider called "wordcamp-google" because by default only local fonts are supported. It's prefixed in case a core provider (or another plugin like Jetpack*) eventually adds Google support.
- It's necessary to enqueue all the fonts, core/GB doesn't currently have a mechanism to load only the fonts in use… but that's not any different than how `wp_head_google_web_fonts` already works.
- This works best with variable fonts, but as far as I can tell GB does not restrict the weight/style to just the registered values - for example, Rubik Pixels only has 400 weight, no italics, but you can still set it to "Black (900) Italic". The browser faux-italicizes & faux-bolds it. I think this is a core limitation, and not something we need to address. 

*&nbsp;Jetpack does have [a beta Google Fonts feature](https://jetpack.com/support/google-fonts/), but it's pulling from a small subset of fonts.

### Screenshots

This example uses the following settings to load Comfortaa (variable), Tilt Warp (default), Open Sans (normal & italic, 300 weight), & Rubik Pixels (default).

```css
@import url('https://fonts.googleapis.com/css2?family=Comfortaa:wght@400..700');
@import url('https://fonts.googleapis.com/css2?family=Tilt+Warp');
@import url('https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;1,300&family=Rubik+Pixels&family=Tilt+Warp&display=swap');
```

| | Editor | Front end |
|---|---|---|
| Site editor, setting the default font for all text | editor-site | frontend-sitewide |
| Post editor, setting fonts on a specific block | editor-post | frontend-post |

1. Go to google fonts and grab some `@import` URLs
2. Using a block theme (TT2 or TT3), put those in Appearance > Fonts
3. Load up the editor and try to change the typography of any element
4. The editor preview should match the front end
5. Switch to a non-block theme
6. The fonts are still loaded on the front end (but are not applied to anything without custom CSS)

There are checks in place to make sure this does not break if GB is accidentally disabled.

### DIFF
--- a/public_html/wp-content/plugins/wc-fonts/wc-fonts.php
+++ b/public_html/wp-content/plugins/wc-fonts/wc-fonts.php
@@ -1,8 +1,9 @@
 <?php
-
 /*
  * Plugin Name: WordCamp.org Fonts
  */
+
+require_once __DIR__ . '/wc-google-fonts-provider.php';
 
 class WordCamp_Fonts_Plugin {
 	protected $options;
@@ -19,6 +20,7 @@ class WordCamp_Fonts_Plugin {
 		add_action( 'wp_head',            array( $this, 'wp_head_google_web_fonts' ) );
 		add_action( 'wp_head',            array( $this, 'wp_head_font_awesome'     ) );
 		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_core_fonts'       ) );
+		add_action( 'init',               array( $this, 'register_fonts_for_editor' ) );
 	}
 
 	/**
@@ -49,8 +51,10 @@ class WordCamp_Fonts_Plugin {
 			return;
 		}
 
-		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		printf( '<style>%s</style>', $this->options['google-web-fonts'] );
+		if ( ! wp_is_block_theme() ) {
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			printf( '<style>%s</style>', $this->options['google-web-fonts'] );
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


pkevan:
Presumably once a google font is enable, it would load it irrelevant of whether it is needed on the page?

Not sure if this is an issue, but couldn't tell from the code if this was possible, or desirable?

ryelle:
> Presumably once a google font is enable, it would load it irrelevant of whether it is needed on the page?

Yep, and that's how the Fonts feature currently works. Once you add the `@import` to the settings, it loads on all pages. This will do the same thing, enqueuing the font on all pages.

Maybe as the Fonts API evolves it will be able to detect and load like it does for block assets, but right now we need to keep loading them on all pages.

pkevan:
Given the above, is it worth considering a limit for total number of fonts? Just thinking about extreme cases, or someone doing something unintentional like including every Google font :)

ryelle:
Is there something about this implementation that would make adding too many fonts worse than on production? I don't mind the idea of adding in some extra checks, but I think it could be done in another PR/issue.

pkevan:
chatted through with @ryelle as I was working under the assumption that adding Google fonts was a new thing, which it isn't 😄

pkevan (approved):

PR #874, opened 2023-04-25
Jetpack Tweaks: Inject fieldset & legend into contact form fields

ryelle:
WordCamps use the Jetpack contact form for all forms on the site, from plain contact forms to calls for speakers, organizers, etc. When testing the forms for WCUS, @alexstine noticed that the Single Choice (Radio) & Multiple Choice (Checkbox) fields were not marked up correctly for screen reader users.

The upstream issue is being tracked here: https://github.com/Automattic/jetpack/issues/16685

Since these forms are widely used, I wanted to ship out a hotfix while we wait for the root issue to be fixed. The fix here is a bit hacky, based on string replacements, but I tried to keep each step straightforward. First, it wraps the field in a `fieldset`, then wraps the first label (the question) in a legend, which we hide visually. The legend text is then duplicated in an `aria-hidden` `div`, to avoid the browser default styles on legends (I grabbed [this method from Adrian Roselli](https://adrianroselli.com/2022/07/use-legend-and-fieldset.html)).

This change is only added to the two relevant fields, and only if the form style is default. The "animated" and "outline" style put the labels at the end of the field HTML, so it would require more careful manipulation, and the majority of sites I spot-checked used the default form style.

Props @alexstine

**Markup example**

The previous markup for a Single Choice (Radio) field (style & class attributes removed for easier reading):

```html
<div>
	<label>Will you be able to attend WordCamp US 2023 in person?</label>
	<div>
		<label><input type="radio" name="g549" value="Yes" required="" aria-required="true"> Yes</label>
		<label><input type="radio" name="g549" value="No" required="" aria-required="true"> No</label>
		<label><input type="radio" name="g549" value="Maybe/Unsure" required="" aria-required="true"> Maybe/Unsure</label>
	</div>
</div>
```

The new markup is:

```html
<fieldset>
	<legend>Will you be able to attend WordCamp US 2023 in person?</legend>
	<div aria-hidden="true">Will you be able to attend WordCamp US 2023 in person?</div>
	<div>
		<label><input type="radio" name="g549" value="Yes" required="" aria-required="true"> Yes</label>
		<label><input type="radio" name="g549" value="No" required="" aria-required="true"> No</label>
		<label><input type="radio" name="g549" value="Maybe/Unsure" required="" aria-required="true"> Maybe/Unsure</label>
	</div>
</fieldset>
```

**Screenshots**

There _should_ be no visual changes, but some customizations might not apply to the new visual field (that's why the text is slightly larger in Montclair's screenshot). I don't think that's major enough to block this fix though.

| | Before | After |
|---|---|---|
| Form on a block-based theme | before-us | after-us |
| Form on a classic theme | before-montclair | after-montclair |
| Form using "outline" style, no change | before-sylhet | after-sylhet |

**How to test the changes in this Pull Request:**

1. Create a contact form
2. Add some fields, including radio & checkbox fields
3. They should display correctly
4. Try testing with a screen reader
5. When you hit the radio & multiple checkbox fields, each option should be associated with the question text

### DIFF
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php
@@ -188,3 +188,60 @@ function workaround_is_frontend( $is_frontend ) {
 	return $is_frontend;
 }
 add_filter( 'jetpack_is_frontend', __NAMESPACE__ . '\workaround_is_frontend' );
+
+/**
+ * Filter the contact form HTML to replace the wrapper with a fieldset & legend
+ * for radio & multi-checkbox fields. This is necessary for screen reader users
+ * to understand the form questions.
+ *
+ * The single checkbox field is okay as-is.
+ *
+ * Upstream issue: https://github.com/Automattic/jetpack/issues/16685.
+ * When that is fixed, this & `inject_css_for_fieldset` can be removed.
+ *
+ * @param string   $rendered_field Contact Form HTML output.
+ * @param string   $field_label    Field label.
+ * @param int|null $post_id        Post ID.
+ */
+function wrap_checkbox_radio_fieldset( $rendered_field, $field_label, $post_id ) {
+	if (
+		str_contains( $rendered_field, 'grunion-checkbox-multiple-options' ) ||
+		str_contains( $rendered_field, 'grunion-radio-options' )
+	) {
+		// remove any whitespace so the offsets below work.
+		$rendered_field = trim( $rendered_field );
+		// replace wrapper div.
+		$rendered_field = substr_replace( $rendered_field, '<fieldset ', 0, 5 );
+		$rendered_field = substr_replace( $rendered_field, '</fieldset> ', -6, 6 );
+		// replace only the first label, others are for the options.
+		$rendered_field = preg_replace( '/<label/', '<legend', $rendered_field, 1);
+		$rendered_field = preg_replace( '/<\/label>/', '</legend>', $rendered_field, 1);
+	}
+	return $rendered_field;
+}
+add_filter( 'grunion_contact_form_field_html', __NAMESPACE__ . '\wrap_checkbox_radio_fieldset', 10, 3 );
+
+/**
+ * Add styles for the injected fieldset & legend.
+ *
+ * This resets the spacing around the fieldset, and styles the legend to match the label.
+ *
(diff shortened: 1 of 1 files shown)


alexstine (approved):
@ryelle Oh, this is so hacky but I enjoy how you got to this solution. 👍 HTML looks good.

pkevan:
Given this is a hack to fix something that could be fixed in Jetpack forms, it might be good to get a PR ready there too, for the proposed changes - that way we have an way towards removing the hack in the future, since the issues noted seems to have been around for a while without much movement.

ryelle:
> Given this is a hack to fix something that could be fixed in Jetpack forms, it might be good to get a PR ready there too, for the proposed changes

I did start going down that route, but thought my time would be better spent getting this quick fix out & letting the JP folks fix the forms. At least the upstream issue has been triaged now, so maybe there will be more movement there soon. Do you think a lack of upstream PR should block this PR?

alexstine:
We should definitely get this one in before it is addressed upstream. This is a basic accessibility feature of forms.

pkevan:
> Do you think a lack of upstream PR should block this PR?

Not necessarily, but given we're solving the issue in our context, it would be nice to try and move along an issue which hasn't yet been resolved in nearly 3 years. 

Whether that is through a PR, additional up-to-date information, or references to the where it should be fixed in code, it feels like something where we should contribute an appropriate amount of effort upstream.

pkevan:
Still trying to figure out why, but this broke the WordCamp network when updating Jetpack from 11.9.1 to 12.2.1

```
Uncaught Error: Class 'Grunion_Contact_Form' not found in wp-content/mu-plugins/jetpack-tweaks/miscellaneous.php:209
URL
https://belfast.wordcamp.org/2018/contact/
```

pkevan:
narrowed it down to between 12.1.1 and 12.2 - still investigating.

pkevan:
tracked it down to a change in the default form handler in Jetpack: https://github.com/Automattic/jetpack-production/commit/2a2ff9de9c9e02af2ead01178b2576a1d88d0b01. Not sure if this does a better job with what this PR was trying to resolve, but could be worth investigating further.

ryelle:
Looks like that override stopped working (or some other bug in jetpack), I've updated this fieldset fix to work with the new form package: https://github.com/WordPress/wordcamp.org/pull/1074

sabernhardt:
https://github.com/Automattic/jetpack/pull/34147 has been merged, and that seems to have fixed the form in the plugin.

Can the tweaks be removed now?

PR #879, opened 2023-05-18
Blocks: Add query loop variations for WordCamp post types

ryelle:
This adds special query loop variations for the rest of the WordCamp post types (Organizers, Speakers, Sponsors), so that it's easier to use the query loop with our content types.

The Query approach is more flexible, allowing organizers to move around blocks and adjust the formatting on a per-block level. Eventually, these could replace the old list blocks in the new site content. Until then, they're kind of a hidden feature, discovered by searching the blocks list for "speakers", etc. (once this is merged, I'll add another issue to update the new content)

When added, these new blocks set up a default query (10 items per page, sorted A-Z), which can be updated to filter a specific taxonomy (or any other changes). It also drops in set of default blocks, different for each post type.

**Organizers**
- A row with organizer avatar & name
- Post excerpt
- A row with "Teams:" and the teams tagged to this organizer

**Speakers**
- A row with speaker avatar & name
- Post excerpt
- Speaker sessions list, showing session details and linking to each session

**Sponsors**
- Featured image, linked to sponsor post
- Post title, linked to sponsor post
- Post excerpt

This also updates the session list variation with a new icon, and updates the attributes on the group block.

Fixes #797, See #754

### Screenshots

| | Screenshot |
|---|---|
| organizers |  |
| sessions |  |
| sponsors |  |
| speakers |  |

These blocks all have custom icons to differentiate them from the old list blocks:

### How to test the changes in this Pull Request:

1. Have content set up on your site
2. Add any of these list blocks
3. It should display the content you chose
4. Changing the filters should apply to the content

PR #834, opened 2023-02-20
Fix: add nginx rules for subdir installs

2ndkauboy:
The nginx.conf was missing some configuration for backends of individual WordCamps to work then using the year as a subdirectory, which is the default now.

Fixes #782

Props @nullbytes

### How to test the changes in this Pull Request:

1. Install the environment
2. Change the domain from 2014.seattle.wordcamp.org to seattle.wordcamp.org/2014
3. Try to access https://seattle.wordcamp.org/2014/wp-admin

PR #883, opened 2023-06-01
Use new patterns for initial site content

ryelle:
This updates the stub content used when creating new WordCamp sites to use the patterns added in #760. Using block patterns works on classic and block themes, and when the pages are edited, the pattern is injected into the content so the organizers can fully customize the content.

The other blocks' syntax has also been updated to match the current syntax, and the organizer note on Sponsors was reformatted to a grey box like the other organizer notes.

See #754, #879, #760

This should be the last of this set of [FSE/block improvements](https://github.com/WordPress/wordcamp.org/projects/2). 

### Screenshots

By default, these won't show anything, since the content they would show doesn't exist yet. (The site I used to screenshot the content had organizers set up already.)

| organizers | sessions |
|---|---|
|  |  |
| speakers | sponsors |
|  |  |

When edited, the pattern expands out to the pattern content:

new-orgs-pattern

1. In your local dev environment, create or edit a WordCamp post, and check the box to create a new site
2. Go to the new site's dashboard, and view the newly created pages
3. The Organizers, Speakers, Sessions, and Sponsors pages should use the query loop variations

### DIFF
--- a/public_html/wp-content/plugins/wcpt/stubs/page/organizers.php
+++ b/public_html/wp-content/plugins/wcpt/stubs/page/organizers.php
@@ -2,4 +2,4 @@
 <p class="has-background" style="background-color:#eeeeee"><em>Organizers note:</em> You can enter content for this page in the Organizers menu item in the sidebar.</p>
 <!-- /wp:paragraph -->
 
-<!-- wp:wordcamp/organizers {"mode":"all"} /-->
+<!-- wp:pattern {"slug":"wordcamp/organizer-list-bio"} /-->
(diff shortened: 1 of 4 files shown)


StevenDufresne:
Apologies for the newb question, the testing says: `In your local dev environment`, but when i run this environment (docker) and try to create a new site, I get caught on production validation, specifically `{username} is not a valid WordPress.org username.` (after using a `city.wordcamp.org/year` site url which probably doesn't work either)

Does this need to be run in a sandbox? And if so, can I create test sites and delete them without causing a mess of the data?

ryelle:
IIRC, the validation `{username} is not a valid WordPress.org username` is checking based on users in the local network (since that's wp.org on production), so if you use `admin` you should get past it.

> after using a city.wordcamp.org/year site url which probably doesn't work either

Yeah, currently the year subfolder setup is broken — but #834 does fix that. I've been hacking my way around it though, by creating the city.wordcamp.org/year site, then going into the network admin and removing the year path, so then you have `city.wordcamp.org`, and that works.

pkevan (approved):
👍 looks good

PR #882, opened 2023-05-31
Payments: Add a button to download budget CSV

ryelle:
This adds a button to the budget actions area to download a CSV of the current budget view. It exports the fields that are displayed in the "Expenses" and "Income" tables on the budget page.

~This is in draft mode until I can confirm this is the expected solution.~

~Eventual reviewing note— the first commit is a coding-standards cleanup, no other changes were done there. The second has the functionality changes.~ This PR is based on #890, which contains the coding standards changes.

### Screenshots

Currently, the download button appears below the data tables, above the other action buttons for the budget (if available), and it downloads a CSV of the budget.

On an approved budget:
Screenshot 2023-05-31 at 2 08 26 PM

On a working budget:
Screenshot 2023-05-31 at 2 08 54 PM

On click, it kicks off a download of a CSV, named `budget_wordcamp-name_YYYY-MM-DD.csv`. This looks like:

Screenshot 2023-05-31 at 1 51 28 PM

### How to test the changes in this Pull Request:

View the budget of a WordCamp— I used `https://testing.wordcamp.org/2019/wp-admin/admin.php?page=wordcamp-budget` for testing, and try downloading the CSV. It should match the current tab (if you download on the approved tab, it should download the approved budget, and working should download the working budget).

PR #890, opened 2023-06-05
Coding standards: Fix or disable lint checks for budget files

ryelle:
This fixes the coding standards issues for the files updated in #882. There should be no functionality changes. I only fixed up the files edited by #882, as trying to update all of `wordcamp-payments` was going to be a rabbit hole (even after autofixing, there are hundreds of errors).

I decided to ignore the errors in `wordcamp-payments/views/budget-tool/main.php` because fixing them up would create more changes in the HTML formatting of the file. The issues in `budget-tool.js` are ignored because the eslint config expects React/modern JS.

(This PR is just to prepare these files, the actual budget tool changes are in #882).

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
@@ -1,13 +1,20 @@
 <?php
 class WordCamp_Budget_Tool {
-    public static function load() {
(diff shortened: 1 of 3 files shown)


pkevan (approved):
Reviewed via https://github.com/WordPress/wordcamp.org/pull/890/files?diff=unified&w=1, changes look good 👍

Issue #875, opened 2023-04-26
New Attendee Checkbox on the Ticket Registration Form


catederosia:
Add a new checkbox in the WordCamp ticket registration that asks if the person is a first-time attendee. 

- [x] #892 
- [ ] Track data based on event type
- [x] Add report

iandunn:
That sounds good, that'll be more accurate than the bin script we have 🎉 

We may need to integrate that data into the bin script, but continue to use the current name/email matching, so that we get stats on date ranges from before this launches. That may not be worth it, though, and we could just have the current script for the dates before, and the new data for the dates after.

iandunn:
The report was added in #935

StevenDufresne:
Is this complete? We still have "Track data based on event type" unchecked. What is that referencing specifically?

iandunn:
That refers to the 2nd question in https://github.com/WordPress/wordcamp.org/pull/892#issuecomment-1585154451

iandunn:
This is the last open issue in the milestone, and most of it is done. IIRC there's been some discussion on if we should do the remaining task or not, so I think it's fine to close this now, and then open a new issue for that once the community team decides on it.

PR #892, opened 2023-06-09
Add registration question for first-time attendees

iandunn:
See #875

WIP

Testing:
* Answer `yes`, `no`, `unsure`, and skip the question
* Check meta values in the CPT item, and also Summarize tool, attendee export

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/addons/first-time.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/addons/first-time.php
@@ -1,0 +1,446 @@
+<?php
+
+namespace WordCamp\CampTix_Tweaks;
+
+defined( 'WPINC' ) || die();
+
+use CampTix_Addon;
+use WP_Post;
+
+/**
+ * Add an required attendee field asking if they've attended a WordCamp before.
+ */
+class First_Time_Field extends CampTix_Addon {
+	const SLUG = 'first_time_attending_wp_event';
+
+	public $label    = '';
+	public $question = '';
+	public $options  = array();
+
+
+	/**
+	 * Hook into WordPress and CampTix.
+	 */
+	public function camptix_init() {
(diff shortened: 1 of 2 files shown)


iandunn:
Hey @catederosia 👋🏻 , I have a few questions:

1. Should this question be required, or optional? 

    I'm leaning towards optional because that'd reduce the burden on users when filling out the form. It'd also mean that we'd get less data, though, and that the data we would get may be biased in one direction or the other. First-time attendees may be more or less motivated to answer the question, so it wouldn't be a random sample.

1. What event type(s) should this ask about?

    Currently I have it saying:

    > Will this be your first time attending a WordPress event? 

    We could also have it insert the type of event being registered for, though, like:

    > Will this be your first time attending a WordCamp? 
    > Will this be your first time attending a Pilot Event? 
    > etc

    Tracking it for different types would add some complexity, but I'm sure not how important having the data separated is.

catederosia:
Apologies for the delay @iandunn. 

1. It's a small enough question, let's make it required.
2. If it's not too much extra work, it would be ideal to have it ask it insert the type of event being registered for. It would give us more specific data.

I am sway-able on both. I value your experience here.

iandunn:
Those sound good 👍🏻  I'll work on getting the specific data, and follow up if it turns out to be more complicated than expected.

iandunn:
I looked at asking the question based on the event type, but that's going to depend on how some other issues are done, so it's currently blocked. For now I just asked about WP events in general. I added a checklist item to #875 to track it.

StevenDufresne:
Circling back here, I don't know that the attendees will see a distinct line between the 2 types. Attendees would need to be very aware of the difference and I don't think we, as organizers are super clear on where the line is drawn. I say we don't go this route. We could revisit if it seems like it would provide significant insight. Is that okay with everyone?

Issue #550, opened 2020-08-17
Profile badges for WordCamp volunteers


timiwahalahti:
As per the conclusion of [this proposal on Community Team P2](https://make.wordpress.org/community/2020/07/03/proposal-recognition-for-event-volunteers-and-attendees-in-wordpress-org-profile/), Community Team would like to start handing out badges for volunteers at WordCamp.

Technical aspects are still wide open and need consideration about what is the best approach for this. From a dev time point of view, re-using speaker/organizer Custom Post Type and profile notify functionality build for those would probably be the best solution. But that's not the easiest one for organisers thought, especially for large events with tens of volunteers.

Could it work if we add a setting where organisers can define the discount code and/or ticket reservation pool handed out to volunteers. When registration is done from the pool or with the code draft gets added in "Volunteers" CPT with pre-filled information. Then organisers would just need to publish the post when volunteer is confirmed. This would allow also manually adding the volunteers if organisers decide so or for some reason don't use Camptix at all for them.

This would need asking the WordPress.org username (not mandatory) while purchasing the ticket, but I don't think that's a bad idea in general. That would allow multiple possible things in future, like linking to attendees WordPress.org profile on attendees page etc. Of course, this would need its own decision by Community Team - let's ask it if the approach that requires WordPress.org username during registration sounds like best way to help to achieve volunteer badges.

iandunn:
🤔 those are good questions.

If we use CampTix, we could maybe have some conditional logic where it would only ask for their username if they're using a volunteer coupon code.

Another approach would be to have a Call for Volunteers page, similar to speakers. Submitting that form could automatically create drafts of a Volunteers CPT.

timiwahalahti:
Call for Volunteers page 🙌

IIRC we actually have one already, but it's not asking WordPress.org username nor (clearly) prefilling any CPT. I'm actually unsure how many camps do use the form in that page or are they replacing it with something other. To determine this, would it be easy to run a query that checks how many 2019 events have "WordCamp Volunteer Application" feedback?

If most of the camps are already using the form provided, extending the functionality sounds like the best approach. _I'm also starting to dream about tool based on Volunteers CPT where shifts and tasks can be planned 😅_

CdrMarks:
I do like both of these suggestions:
> If we use CampTix, we could maybe have some conditional logic where it would only ask for their username if they're using a volunteer coupon code.

[...]

PR #895, opened 2023-06-12
Add Volunteer post type

### LINKED ISSUE #887: Add Volunteers post type
Related https://github.com/WordPress/wordcamp.org/issues/550

Pilot sites want to take volunteer stats into account when measuring success, so we'll need a volunteer post type for that. 

- [x] #895 
- [x] #898 
- [x] Add/update Handbook docs

iandunn:
See #887
See #550 

This adds a new custom post type for volunteers, similar to the post type for organizers and speakers. This will let us gather some stats on the success of pilot events, automatically grant profile badges, etc.

Automation with the `Call for Volunteers` form can be added in a subsequent PR (similar to how it works for speakers).

### DIFF
--- a/.githooks/prepare-commit-msg
+++ b/.githooks/prepare-commit-msg
@@ -43,7 +43,7 @@ function get_previous_components() {
 		);
 
 		if ( $short_descriptions ) {
-			$short_descriptions = trim( $short_descriptions );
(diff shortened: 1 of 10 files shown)


onemoreprince:
There is no block to pull the volunteer's data on any page. 
At least the WP Query loop block should detect the Volunteer post type which currently it doesn't.

Issue #906, opened 2023-06-16
Update WordCamp mu-plugins to work with Pilot Sites


iandunn:
Similar to #886, the `mu-plugins` and `mu-plugins-private` folders will need to be audited and updated to work with the Pilot network.

If there's time, some of the things in `mu-plugins-private` could be moved into the public `mu-plugins` folder, per #576 .

## general checks

- [x] #920
- [x] #936 
- [x] Move mu-plugins that should only run on the WordCamp network into `mu-plugins/wordcamp`, and add a `require` for them. Sometimes this may mean only moving certain functions from a file, rather than the entire file.
- [x] Check if any instances of `SITE_ID_CURRENT_SITE` should to be updated to use `WORDCAMP_NETWORK_ID` or `EVENTS_NETWORK_ID`. The latter are more explicit/descriptive if we want the code to only use one of the networks.
- [ ] #938 
- [ ] check all `(get|update|delete)_site_option` calls and see if we want to share that data across networks (based on context). if we do, then they can be updated to use `{action}_network_option( WORDCAMP_NETWORK_ID, ...)` instead. add a comment documenting the reason they're shared
- [ ] Test each plugin to make sure it works on both networks and doesn't have any unintended consequences

## individual plugins

- [x] #915 
- [x] Move Let's Encrypt helper in #921 
- [x] #922
- [x] Re-enable `latest-site-hints.php` when tested/updated
- [x] Re-enable `load-other-mu-plugins.php` when tested/updated
- [x] SSO and GDPR should be permanently enabled. related cb577f071fa313c235823b1703d34fc953e1b16b. GDPR done in 3f48d5169848a3a24b0c1f0b61dac02ef74b2c94
- [x] Once it's all done, reorganize `load-other-mu-plugins.php` if needed. Some things may be `require`d twice because of the temporary code.

outdoor2kode:
Is the 866 the issue you'd like to refer to here? As I'm a bit not sure how to connect with this issue, maybe I missed some stuff, or perhaps it's issue 886? (not linking them here to prevent them from being referenced in both tickets)

dd32:
[...]

Issue #886, opened 2023-06-02
Update WordCamp plugins to work with Pilot Sites


iandunn:
Pilot sites will need to use many of the same plugins as WordCamps, and those plugins may need some updates to be compatible with post types of events. Some plugins that they _don't_ use may also need to be updated to account for both types.

This doesn't include mu-plugins, those will be done in #906

Depends on #885 

## general checks
- [x] #936
- [ ] Check is any instances of `SITE_ID_CURRENT_SITE` should to be updated to use `WORDCAMP_NETWORK_ID` or `EVENTS_NETWORK_ID`. The latter are more explicit/descriptive if we want the code to only use one of the networks.
- [ ] check all `get_sites()` because they include all networks by default. most queries will probably only want the wordcamp network, but it needs to be decided on a case by case basis because the context matters
- [ ] check all `(get|update|delete)_site_option` calls and see if we want to share that data across networks (based on context). if we do, then they can be updated to use `{action}_network_option( WORDCAMP_NETWORK_ID, ...)` instead. add a comment documenting the reason they're shared
- [ ] Test each plugin to make sure it works on both networks and doesn't have any unintended consequences

## individual plugins
- [x] `camptix` and all related plugins and mu-plugins
- [ ] `wordcamp-payments`
- [ ] `wordcamp-payments-network`
    - [ ] `wp_0_wcbd_sponsor_invoice_index` may need separate tables per network
- [ ] `wc-post-types`

There'll probably be more needed as well. We'll need to test sites to see what we actually run into.

StevenDufresne:
- [x] Update `4-helpers-wcpt.php` to not use hardcoded `wordcamp` cpt.

https://github.com/WordPress/wordcamp.org/blob/18642950aa9b4bcd8647945290e6b86bf2210bb1/public_html/wp-content/mu-plugins/4-helpers-wcpt.php#L54


Edit: Pull Request added #937

StevenDufresne:
I've done a first audit and have come up with this initial list. It may be missing supplemental plugins that need to be added to add functionality that wasn't immediately evident. 

- [x] Camptix
   - Related payment plugins. I wasn't sure which ones were necessary. 
- [x] WordCamp.org Post Types (this seems fine but needs #934)
      - ~~Currently requires: WordCamp Post Type. [See above](https://github.com/WordPress/wordcamp.org/issues/886#issuecomment-1624575560)~~
- [x] WordCamp Budgets (this looks like it works fine)
[...]

Issue #885, opened 2023-06-02
Provision Pilot Event Sites


iandunn:
Pilot sites will need to use the WordCamp infrastructure because they'll use the Budget plugins, automated site provisioning, etc, but they'll also need some customizations since they're not actually WordCamps.

We could use the existing `wordcamp` post type, with a post meta field, taxonomy, etc to distinguish them from traditional camps, or create a new post type entirely. 

Since these are experiments and there are so many unknowns, I'm a little leery of creating a new CPT before we really understand what the needs will be long term. A lot of the codebase is designed to integrate with the `wordcamp` post type, and may need to be updated to work with a new one. That could introduce technical debt if we end up needing something different, etc.

On the other hand, a new post type could help isolate any changes from the `wordcamp` post type, and having a clean break between the two may help make changes to the codebase less fragile. There's also a concern around stats, since pilot events probably shouldn't be counted as WordCamps, etc.

Either way, there are hundreds of things that reference the `wordcamp` post type explicitly or implicitly, via `WCPT_POST_TYPE_ID`, `'post_type' => 'wordcamp'`, `get_wordcamp_post()` etc.

Another idea would be to create a new network inside the Multisite for these events, so that we can have different plugins network-activated, etc. That may remove the need to update some plugins, but definitely not all. 

It may require some experimentation to see which approach will fit best and be practical within the time constraints.

What do you think @StevenDufresne @ryelle @dd32 ?

Related #881 

----
 
_Update: Actionable items:_

These are interdependent with #884 

- [x] Create an additional network. 
    - [x] The README and docker conf should be updated so it can be reproduced across envs. 
- [x] Create a main `events.wordpress.org` site that'd serve as the "central" for the pilot events CPT and other management tools
    - [x] Set `BLOG_ID_CURRENT_SITE` programmatically

iandunn:
@pkevan it'd be good to hear your thoughts on this too

pkevan:
i'll circle back to this tomorrow when i'm traveling and have some time to look in more detail.

pkevan:
> Another idea would be to create a new network inside the Multisite for these events, so that we can have different plugins network-activated, etc. That may remove the need to update some plugins, but definitely not all.

Given we are planning to run this via a events.wordpress.org domain it seems like this would be a good option, that way we are separating it out but could potentially re-use the existing code. 

Any alteration to this can hopefully be done via a network level code, but should it become too burdensome we could separate out too.

[...]

Issue #884, opened 2023-06-02
Support Pilot Event URLs


iandunn:
They'll need to use the WordCamp infrastructure, but a different domain name, like `events.wordpress.org`. This'll probably require something like:

Interdependent with #885 

- [x] [A systems request](https://make.wordpress.org/systems/2023/06/15/events-wordpress-org-served-by-wordcamp-org/) for:
  *  a DNS `A` record that points `events.wordpress.org` to the WordCamp server
  * Updating nginx config to serve from that domain
  * Updating nginx to use the `*.wordpress.org` SSL
- [x] `wp-config` change to set `DOMAIN_CURRENT_SITE`
- [x] Possibly some `sunrise.php` tweaks to support the additional domain, but we didn't need any for buddycamp.org, so it might be fine without
- [x] [Updated nginx `rewrite` rules for regular site paths](https://make.wordpress.org/systems/2023/06/15/events-wordpress-org-served-by-wordcamp-org/#comment-2115)
- [x] Testing the custom tools on the site and on Central to reveal any other problems. Related #886

iandunn:
This is done now that #907 is merged and the sysreq is resolved.

PR #907, opened 2023-06-16
Setup Pilot Events network and URL support

iandunn:
See #884 
See #885 

This handles the interdependent parts of those issues, which covers everything needed to create sites like `events.wordpress.test/barcelona/2023/diversity-day/`, and have requests routed appropriately.

### Testing

⚠️ Do this in a local environment, since more testing is needed before we test this on a w.org sandbox

- [ ] Update your `wp-config.php` and `nginx.conf` to be similar to the Docker files in #912. _( I haven't tested Docker yet, though, so more work may be needed in #910 to make it work )_
- [ ] Manually create sites for `events.wordpress.test` and `events.wordpress.test/vancouver/2023/diversity-day/`. You'll probably have to use `-` instead of `/` in the path when creating them, but then you can edit them to change to `/`.
- [ ] Add those hostnames to your `hosts` file
- [ ] Rerun `mkcert` to update your SSL certs with the new domain
- [ ] Manually change the `site_id` for those new sites in the `wp_blogs` table to be `2`
- [ ] Update `EVENTS_ROOT_BLOG_ID` values to match the blog you created above
- [ ] Don't activate any plugins. We'll have to do #886 and #906 before they'll work
- [ ] Visit the Vancouver site and make sure all the Core functionality works
- [ ] Visit Central and a regular site on the WordCamp network and make sure everything still works

### DIFF
--- a/public_html/wp-content/mu-plugins/latest-site-hints.php
+++ b/public_html/wp-content/mu-plugins/latest-site-hints.php
@@ -2,10 +2,16 @@
 
 namespace WordCamp\Latest_Site_Hints;
 use function WordCamp\Sunrise\get_top_level_domain;
-
 use const WordCamp\Sunrise\{ PATTERN_YEAR_DOT_CITY_DOMAIN_PATH, PATTERN_CITY_SLASH_YEAR_DOMAIN_PATH };
 
 defined( 'WPINC' ) || die();
+
+if ( EVENTS_NETWORK_ID === SITE_ID_CURRENT_SITE ) {
+	// @todo Remove this once https://github.com/WordPress/wordcamp.org/issues/906 is fixed.
+	// If it's needed on the Events network, the constants above will need to be moved to `sunrise.php`, or
+	// defined in `sunrise-events.php` with a pattern designed for Events sites.
+	return;
+}
 
 // Hook in before `WordPressdotorg\SEO\Canonical::rel_canonical_link()`, so that callback can be removed.
 add_action( 'wp_head', __NAMESPACE__ . '\canonical_link_past_home_pages_to_current_year', 9 );
--- a/public_html/wp-content/sunrise.php
+++ b/public_html/wp-content/sunrise.php
@@ -11,6 +11,10 @@ load_network_sunrise();
  */
 function load_network_sunrise() {
 	switch ( SITE_ID_CURRENT_SITE ) {
+		case EVENTS_NETWORK_ID:
+			require __DIR__ . '/sunrise-events.php';
+			break;
+
 		case WORDCAMP_NETWORK_ID:
 		default:
 			require __DIR__ . '/sunrise-wordcamp.php';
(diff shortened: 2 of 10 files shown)


public_html/wp-content/sunrise.php
```diff
@@ -3,154 +3,27 @@
...
-			$redirect .= ltrim( $_SERVER['REQUEST_URI'], '/' );
-		}
+function load_network_sunrise() {
+	$tld = get_top_level_domain();
 
-		if ( $redirect ) {
-			$status_code = 302;
-		}
-	}
+	switch ( DOMAIN_CURRENT_SITE ) {
```
dd32: I'd have been tempted to just use `SITE_ID_CURRENT_SITE` here, the network ID, referring back to a constant..

```suggestion
	switch ( SITE_ID_CURRENT_SITE ) {
		case NETWORK_ID_EVENTS:
			require __DIR__ . '/sunrise-events.php';
			break;
		default:
		case NETWORK_ID_WORDCAMP:
			require __DIR__ . '/sunrise-wordcamp.php';
			break;
	}
```

public_html/wp-content/mu-plugins/latest-site-hints.php
```diff
@@ -2,11 +2,17 @@
 
 namespace WordCamp\Latest_Site_Hints;
 use function WordCamp\Sunrise\get_top_level_domain;
-
 use const WordCamp\Sunrise\{ PATTERN_YEAR_DOT_CITY_DOMAIN_PATH, PATTERN_CITY_SLASH_YEAR_DOMAIN_PATH };
 
 defined( 'WPINC' ) || die();
 
+if ( 2 === SITE_ID_CURRENT_SITE ) {
```
dd32: Probably better to introduce a constant for the Network IDs, to avoid multiple different locations of IDs and switching methods (here it's network ID, elsewhere it's domain)
```suggestion
if ( NETWORK_ID_EVENTS === SITE_ID_CURRENT_SITE ) {
```

dd32 (approved):
Other than some nitpicks around what field is used for switching between the two, this seems good to me (Untested) and I don't really understand the hesitation of just committing and iterating from there.

pkevan:
> just committing and iterating from there.

That would be my preference here too, as I mention in slack, I can't easily see from the code what is needed or not, and what needs fixing.

public_html/wp-content/mu-plugins/latest-site-hints.php
iandunn: That makes sense 👍🏻

public_html/wp-content/sunrise.php
iandunn: 👍🏻

iandunn:
This is deployed to production now, and the root site is working.

Event sites like https://events.wordpress.org/narnia/2023/test-site/ need [some nginx updates](https://make.wordpress.org/systems/2023/06/15/events-wordpress-org-served-by-wordcamp-org/#comment-2115) before they'll work.

Issue #900, opened 2023-06-16
Create a new CPT for pilot events


iandunn:
In #885 we talked about creating a new CPT for pilot events, similar to the `wordcamp` post type (`plugins/wcpt`). Eventually, we'll have an application form which automatically creates these posts, and then a button on the post itself that will provision a new site automatically when the event is approved.

For now, though, we just need the CPT itself.

It should be active on the root `events.wordpress.org` site only, because that's akin to `central.wordcamp.org`.

StevenDufresne:
After looking at #911, I'm going to make a case for not creating a new CPT. I may be missing a critical point, but reading what is documented doesn't seem to provide a strong argument, statistical measurement aside.

Here is what i'm concerned with, based strictly on looking at #911.


### As we turn on plugins (or do it all pre-launch), we'll need to update all `wordcamp` or `WCPT_POST_TYPE_ID`(unless we redefine it in the new CPT).   references 

This isn't a huge deal, but it does open up the chance of missing references and things breaking silently in unexpected ways. A risk added on top of the risks posed by having a new network.

### If we remove functionality from the new CPT, we may break plugin behavior.

It's very likely that we turn on more existing plugins than anticipated. Business logic is spread throughout plugins that consume  the`wordcamp` CPT. Any change or omissions in the new CPT could break things downstream and since we'll probably need to start with a lot of replicated code in the new CPT, it feels like a can of worms to me.

 ### Creating a new CPT doesn't do much in _modernizing_ WordCamp and may introduce more technical debt.

I think adding a new CPT will lead to more updates that make plugins work for both CPTs and not the thinking we should be doing; "How can we pair down the `wordcamp` CPT so it's not so monstrous and make some behavior optional". I think if we focused on abstracting and refactoring the existing CPT, we'll end up in a better place for the whole code base. I also think it's easier to scope/understand and similar in quantity.

### From a user/community member perspective, WordCamp = Pilot Event

Although we are technically pairing down WordCamps and adding some _new_ functionality for this pilot event. These are still considered "NextGen" WordCamps and not something entirely different. The labels and what the users sees should likely be the same as with current WordCamps. I fear we'll walk down a road where this post type gets it's own branding and we have to step all that back post experiment.


Have I missed any counterpoints? Is there anything I'm not considering?

PR #911, opened 2023-06-20
Create a new CPT for pilot events

outdoor2kode:
See #900

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-pilot-event/pilot-event-loader.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-pilot-event/pilot-event-loader.php
@@ -1,0 +1,477 @@
+<?php
+
+// define( 'WCPT_POST_TYPE_ID',   'wordcamp'           );
+// define( 'WCPT_YEAR_ID',        'wordcamp_year'      );
+// define( 'WCPT_PILOT_EVENT_SLUG',           'pilot-events'          );
+// define( 'WCPT_DEFAULT_STATUS', 'wcpt-needs-vetting' );
+// define( 'WCPT_FINAL_STATUS',   'wcpt-closed'        );
+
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/wcpt/wcpt-pilot-event/pilot-event-loader.php
```diff
@@ -0,0 +1,477 @@
+<?php
+
+// define( 'WCPT_POST_TYPE_ID',   'wordcamp'           );
```
StevenDufresne, line 3: Let's make sure we set this so we can use it.

public_html/wp-content/plugins/wcpt/wcpt-pilot-event/pilot-event-loader.php
```diff
@@ -0,0 +1,477 @@
...
+// define( 'WCPT_YEAR_ID',        'wordcamp_year'      );
+// define( 'WCPT_PILOT_EVENT_SLUG',           'pilot-events'          );
```
StevenDufresne, line 5: Or this one.

iandunn: IMO it'd be better to use `WCPT_PILOT_EVENT_SLUG` than `WCPT_POST_TYPE_ID`, to avoid confusion between the two post types

StevenDufresne: Noting, we'll need to go through all the plugins and replace the direct references to `WCPT_POST_TYPE_ID` and call [this function](https://github.com/WordPress/wordcamp.org/pull/937/files#diff-790ef1c27b93f13a8bb852756e3352013fb78410c737c8f361515f0ae5663dcaR22).

PR #915, opened 2023-06-21
Let's Encrypt: Only request certificates for WordCamp sites

iandunn:
Sites on the new Events network will use the `wordpress.org` wildcard SSL.

See #906

`prepare()` is used because in a future commit `1` will be replaced with `EVENTS_NETWORK_ID`. See #907.

This needs to be deployed before #907, because creating the `events.wordpress.org` site triggered a Nagios alert from the Dehydrated client.

xref Slack `G0B1CF7E1/p1687372885952129`


iandunn:
The test failures look like a problem with the test suite rather than an actual bug, so I'm gonna go ahead and deploy this to fix production.

I'll circle back later to fix the tests.

PR #924, opened 2023-06-27
Fix local docker environment

timiwahalahti:
Commit https://github.com/WordPress/wordcamp.org/commit/3f48d5169848a3a24b0c1f0b61dac02ef74b2c94 broke the local Docker environment, since `mu-plugins-private` directory is not included in this public repository.

Moving the required private file inside environment check fixes things, but might have other consequences depending on what plugins are kept private.

### DIFF
--- a/public_html/wp-content/mu-plugins/load-other-mu-plugins.php
+++ b/public_html/wp-content/mu-plugins/load-other-mu-plugins.php
@@ -10,8 +10,11 @@ wcorg_include_network_only_plugins();
  * Load mu-plugins that should run on all networks.
  */
 function wcorg_include_common_plugins() {
-	if ( 'local' !== WORDCAMP_ENVIRONMENT ) {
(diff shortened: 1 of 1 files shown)


public_html/wp-content/mu-plugins/load-other-mu-plugins.php
```diff
@@ -10,9 +10,8 @@
  * Load mu-plugins that should run on all networks.
  */
 function wcorg_include_common_plugins() {
-	require_once dirname( __DIR__ ) . '/mu-plugins-private/wporg-mu-plugins.php';
-
 	if ( 'local' !== WORDCAMP_ENVIRONMENT ) {
+		require_once dirname( __DIR__ ) . '/mu-plugins-private/wporg-mu-plugins.php';
```
iandunn, line 14: Moving it here could have some minor side effects. Nothing huge, but it'd be ideal to keep it above, and add an `if ( file_exists(  ...` condition.

PR #925, opened 2023-06-27
Prevent Camptix DB log overflow

### LINKED ISSUE #649: DB Log list in CampTix Edit Attendee Info CSS shade area extends past right margin
Imported from https://meta.trac.wordpress.org/ticket/1360 
Created by [@faddah](https://profiles.wordpress.org/faddah/):

set-up meta wordcamp.org, our https://portland.wordcamp.org/2015/ site, running WordPress 4.3.2-alpha (that's what it says in the dashboard), using the current verion of the WordCamp Base Redux theme.

steps to reproduce -- 

1)  in a browser, log-in as an admin to your WordCamp.org site using the WordCamp Base Redux theme.

2)  in the Dashboard, once people have started purchasing tickets through the CampTix plug-ins for your WordCamp, go to Dashboard > Tickets > Attendees.

3)  choose a random attendee who has purchased a ticket.

4)  click on them to Edit their Attendee Info.

5)  in the Attendee Info screen for that Attendee, scroll down to the **CampTix DB Log** a the bottom of the screen. 

6)  click on one of the link **data** entries in the **CampTix DB Log** to expand it.

expected behavior --

the db log text should expand and the grey area background goes to the right edge of the screen -- doesn't look the greatest, but necessary for how the text of the db log can expand out to the right if it is lengthy, but all other grey barred entry rows, every other row, in the log below it should only extend as far as the **CampTix DB Log** div right margin.

issue --

instead of every other grey row (odd or even child or sibling, however you are doing it in the CSS) below the expanded row in **CampTix DB Log** only extending as far right as that div's margin within the window, every other grey row extends right all the way to browser window edge. a screen shot is attached showing this behavior. happened in a couple different browsers for me -- google chrome on a chromebook, chrome version 45.0.2454.98 (64-bit), Platform 7262.57.0 (Official Build) stable-channel peppy, Firmware Google_Peppy.4389.93.0, and Apple Safari for iOS on an iPad Air, iOS 9.1.

comments --

this should be a fairly easy CSS fix, or jQuery, if you're using jQuery to select and color the background of the ODD or EVEN child or sibling of the elements.

timiwahalahti:
The Camptix BD log was not styled properly, since the contents of the `<pre>` spread some of the rows in the table way too wide. This PR fixes the styling, making the table look nicer.

Fixes #649 

Props faddah

### DIFF
--- a/public_html/wp-content/plugins/camptix/admin.css
+++ b/public_html/wp-content/plugins/camptix/admin.css
@@ -216,6 +216,10 @@
 	padding: 0;
 }
 
+#tix_db_log .tix-table {
+	table-layout: fixed;
+}
+
 .tix-revenue-summary .tix-sold,
 .tix-revenue-summary .tix-remaining,
 .tix-revenue-summary .tix-sub-total,


iandunn (approved):
Thanks!

PR #934, opened 2023-07-07
Fix shared mu-plugins to work for new network

StevenDufresne:
In working on #886, the following plugins were considered dependencies and have been audited. Seem to work as expected.

Related to #906


iandunn (approved):
LGTM 👍🏻

PR #935, opened 2023-07-10
Add first time participation measurement

### LINKED ISSUE #888: Measure first-time participation
First-time participation will be a key metric for pilot sites.

We could use the old `first-time.php` script for some of these, but volunteers/attendees aren't included in that. We may want to move the others towards a self-reported model for increased accuracy, but that also places an extra burden on the user when filling out forms.

Speakers
- [x] Add First Time being a speakers question to CPT
- [x] Retrieve `_wcb_speaker_first_time` and render the value properly on WordCamp Reports » WordCamp Counts

Sponsors
- [x] Add First Time being a sponsors question to CPT
- [x] Retrieve `_wcb_sponsor_first_time` and render the value properly on WordCamp Reports » WordCamp Counts

Organizers
- [x] Add First Time being an organizer question to CPT
- [x] Retrieve `_wcb_organizer_first_time` and render the value properly on WordCamp Reports » WordCamp Counts

Volunteers. Related #887 
- [x] Retrieve `_wcb_volunteer_first_time` and render the value properly on WordCamp Reports » WordCamp Counts

Attendees. Related #875 
- [x] Add First Time being an organizer question to CPT
- [x] Retrieve `_wcb_attendees_first_time` and render the value properly on WordCamp Reports » WordCamp Counts

Script
- [x] Update first-time.php. 
  - Done in https://wordcamp.trac.wordpress.org/changeset/4494 and https://wordcamp.trac.wordpress.org/changeset/4496

outdoor2kode:
See #888

This PR introduces a measurement for first-time participation of organizers, sponsors, speakers, and volunteers into the WordCamp Counts table. It not only presents the total number of first-time participants but breaks down these figures for each individual event within a specified time period.

### Screenshots

Screen Shot 2023-07-12 at 4 50 24 PM

1. Sandboxed.
2. WP Admin Dashboard => (Top-left side) Dashboard => Reports => [WordCamp Counts](https://central.wordcamp.org/wp-admin/index.php?page=wordcamp-reports&report=wordcamp-counts)
3. Select the desired date range and then click on `Show results`.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-reports/views/html/wordcamp-counts.php
+++ b/public_html/wp-content/plugins/wordcamp-reports/views/html/wordcamp-counts.php
@@ -14,7 +14,8 @@ use DateTime;
 /** @var DateTime $end_date */
 /** @var string $statuses */
 
-$gender_legend = '<span class="description small"><span class="total">Total</span> / F / M / ?</span>';
+$gender_legend     = '<span class="description small">Gender: F / M / ?</span>';
+$first_time_legend = '<span class="description small">Y / N / ?</span>';
 
 ?>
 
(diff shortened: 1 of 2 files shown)


public_html/wp-content/plugins/wordcamp-reports/views/html/wordcamp-counts.php
```diff
@@ -135,9 +212,14 @@
 				</td>
 
 				<td class="number">
-					<span class="total"><?php echo number_format_i18n( $event['totals']['organizer'] ); ?></span>
+					<span class="total">Total: <?php echo number_format_i18n( $event['totals']['organizer'] ); ?></span>
+					<br>
+					FT: <?php echo number_format_i18n( $event['first_times']['organizer']['yes'] ); ?>
```
iandunn, line 225: Is there enough room to write out `First Time Organizers` etc, rather than abbreviating?

public_html/wp-content/plugins/wordcamp-reports/views/html/wordcamp-counts.php
```diff
@@ -135,9 +212,14 @@
 				</td>
 
 				<td class="number">
-					<span class="total"><?php echo number_format_i18n( $event['totals']['organizer'] ); ?></span>
+					<span class="total">Total: <?php echo number_format_i18n( $event['totals']['organizer'] ); ?></span>
```
iandunn, line 223: Does this need its own `<td>`, so that it doesn't get displayed in the same column as gender?

iandunn (commented):
This is looking good 👍🏻

public_html/wp-content/plugins/wordcamp-reports/views/html/wordcamp-counts.php
outdoor2kode: I changed the layout a bit, is this something you had in mind?
| Before | After |
| ------ | ------|
| image | image |

public_html/wp-content/plugins/wordcamp-reports/views/html/wordcamp-counts.php
outdoor2kode: Yeah, it would be something like:
image
I like abbreviations more as there's already full text in the header and the reason I added an extra abbreviated title in every cell is for a long list like below (when scrolling to the bottom, it would be more intuitive if there's a helper text there)
image

outdoor2kode: Or we can make the header sticky, it doesn't seem not easy to implement it.

https://github.com/WordPress/wordcamp.org/assets/18050944/65577608-e6e0-4b14-b743-8cf4c4658040

What do you think?

iandunn: 🤔 Yeah, a sticky header would be nice, but that's probably something that should be done upstream in Core rather than here.

I guess the underlying problem here is that it's so much information to try and squeeze into this format. That's not specific to this PR, though, so it's probably not worth addressing here. I think we can go ahead with what you have, and then revisit the issue later to make the chart easier to digest. That may be something a designer could help with.

outdoor2kode: no problem 👍 
I'll open another issue and ask for feedback from the design and tardis team.

outdoor2kode: Follow-up issue: https://github.com/WordPress/wordcamp.org/issues/993

devinmaeztri:
Thank you @renintw and @iandunn! I've tested, and it looks good to me. At first, I was confused with FT and G and the x/x/x. But I got it now. Just to confirm, is the `?` for folks who opt out, and who don't fall into either option? 

Screenshot 2023-08-03 at 12 46 23

outdoor2kode:
Hey @devinmaeztri 🙂 

In the case of  Gender, `?` would be folks whose gender is `unknown`. Currently, gender stats here aren't the ones filled in by users. It's inferred from users' names through `genderize.io`. And since the gender is identified via the system, the sum of `F+M+?` would always equal the **Total**. (The system will always categorize an individual's gender into one of these three options)

In terms of `FT`, `?` is `unsure`, indicating folks who are unsure about whether they are the first time being a role. This value is filled in by users themselves when applying for a certain role. 
For example, when a person applies for being the speaker, there's a question in the application form asking if they are first time being a speaker. If they select I don't know, then it would be a `?` in the report here (or `Unsure` in the Speaker post edit page ⬇️ ).
image
And since the said question is optional, some users may choose to opt out. This opt-out decision would not be categorized into any of the options (Y/N/?), so the sum of `Y+N+?` here doesn't always equal the **Total**.

Issue #936, opened 2023-07-11
Check `switch_to_blog( BLOG_ID_CURRENT_BLOG )` calls. 


StevenDufresne:
# Summary
 Check `switch_to_blog( BLOG_ID_CURRENT_BLOG )` calls. Some should continue using using `BLOG_ID_CURRENT_BLOG` (which will be the `events.wordpress.test` root blog on the Events network), and some will need to be updated to use `WORDCAMP_ROOT_BLOG_ID`, because they should still switch to `central.wordcamp.test`, even when called by sites on the Events network. https://core.trac.wordpress.org/ticket/25293 and https://core.trac.wordpress.org/ticket/40513 might be relevant.

## References

Checked list items do not require any updates.

- [x] wordcamp.org/public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
```
  677,2: 	switch_to_blog( BLOG_ID_CURRENT_SITE ); // central.wordpress.org
  773,4: 			switch_to_blog( BLOG_ID_CURRENT_SITE );
  800,3: 		switch_to_blog( BLOG_ID_CURRENT_SITE );
  848,2: 	switch_to_blog( $site_id );
```

- [x] wordcamp.org/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
```
  51,2: 	switch_to_blog( BLOG_ID_CURRENT_SITE );
  81,2: 	switch_to_blog( BLOG_ID_CURRENT_SITE );
  114,2: 	switch_to_blog( $site_id );
  254,2: 	switch_to_blog( BLOG_ID_CURRENT_SITE ); // central.wordcamp.org
  327,2: 	switch_to_blog( BLOG_ID_CURRENT_SITE );
  361,2: 	switch_to_blog( BLOG_ID_CURRENT_SITE );
  382,2: 	switch_to_blog( BLOG_ID_CURRENT_SITE );
```

This `switch_to_blog` calls are fine, but the hardcoded `post_type => 'wordcamp'` will need to be addressed. 

- [x] wordcamp.org/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```  
1269,4: 			switch_to_blog( BLOG_ID_CURRENT_SITE ); // central.wordcamp.org .
```
I think we'll keep the sponsors separate for now. So this should be ok.

- [x] wordcamp.org/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php
```  
[...]

PR #937, opened 2023-07-11
Update get_wordcamp calls to have the correct CPT.

StevenDufresne:
See #911, #936

With the events network up and running, there are various plugins that use the helper functions in `4-helpers-wcpt.php`. This PR removes the hardcoded `wordcamp` references in  so we can continue to use that code, but replace it with our soon-to-be minted pilot event post type in #911.

To be honest, I think this is necessary, but if it's not going in the right direction, let me know.

Requires #911 to work functionally but can be merged before.

## Alternate

We could alternatively write a new helper but that would require us update all the consuming plugins and that doesn't seem worth it right now.

### DIFF
--- a/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
+++ b/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
@@ -20,9 +20,9 @@ function get_wordcamp_post_type() {
 	}
 
 	return WCPT_POST_TYPE_ID;
- }
+}
 
- /**
+/**
  * Retrieves the post type string of the wordcamp depending on the current network.
  *
  * @param string $post_type The post type to check.
--- a/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
+++ b/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
@@ -278,7 +278,7 @@ function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(),
 	switch_to_blog( BLOG_ID_CURRENT_SITE );
 
 	if ( empty( $query_options ) ) {
-		$wordcamps = $wpdb->get_results( "
+		$query = $wpdb->prepare( "
 			SELECT $wpdb->posts.ID, $wpdb->posts.post_title, $wpdb->postmeta.meta_value AS start_date
 			FROM $wpdb->posts
 				LEFT JOIN $wpdb->postmeta ON $wpdb->postmeta.post_id = $wpdb->posts.ID
@@ -288,6 +288,9 @@ function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(),
 				$wpdb->postmeta.meta_key = 'Start Date (YYYY-mm-dd)'
 			ORDER BY `$wpdb->posts`.`ID` DESC
 		 ", get_wordcamp_post_type() );
+
+		 $wordcamps = $wpdb->get_results( $query );
+
 	} else {
 		// Default to standard query when query_options is sent.
 		$wordcamps = get_wordcamps( $query_options );
--- a/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
+++ b/public_html/wp-content/mu-plugins/4-helpers-wcpt.php
@@ -277,19 +277,17 @@ function get_wordcamp_dropdown( $name = 'wordcamp_id', $query_options = array(),
 
 	switch_to_blog( BLOG_ID_CURRENT_SITE );
 
-	$post_type = get_wordcamp_post_type();
-
 	if ( empty( $query_options ) ) {
 		$wordcamps = $wpdb->get_results( "
 			SELECT $wpdb->posts.ID, $wpdb->posts.post_title, $wpdb->postmeta.meta_value AS start_date
 			FROM $wpdb->posts
 				LEFT JOIN $wpdb->postmeta ON $wpdb->postmeta.post_id = $wpdb->posts.ID
 			WHERE
-				$wpdb->posts.post_type = '$post_type' AND
+				$wpdb->posts.post_type = '%s' AND
 				( $wpdb->posts.post_status <> 'trash' AND $wpdb->posts.post_status <> 'auto-draft' AND $wpdb->posts.post_status <> 'spam' ) AND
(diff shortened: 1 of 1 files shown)


public_html/wp-content/mu-plugins/4-helpers-wcpt.php
```diff
@@ -8,6 +8,30 @@
  * Helper functions related to the `wordcamp` post type.
```
iandunn: Updating this to briefly explain how there are two networks and post types might help folks understand the context when they're new to this file.

It could also just reference a new part of the README or something though, since we may want a comment like that in many other places.

I don't feel strongly, though.

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
```diff
@@ -8,6 +8,30 @@
...
  */
 
+/**
+ * Retrieves the post type string of the wordcamp depending on the current network.
+ *
+ * @return string
+ */
+function get_wordcamp_post_type() {
```
iandunn, line 22: I think it might be confusing to refer to Pilot Events as WordCamps, since (at least in my mind) they're a separate type of event. What do you think about a more generic name?

_( the same applies to the docblock comments for both of these functions )_

iandunn (approved):
I left a few minor suggestions, but this LGTM 👍🏻

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
```diff
@@ -51,7 +75,7 @@ function get_wordcamp_post( $site_id = null ) {
 	switch_to_blog( BLOG_ID_CURRENT_SITE );
```
iandunn, line 81: Updating the comment for this line would be nice too, since it can also refer to the Pilot Events root site.

_( That's assuming that the new CPT in #911 will be activated on `events.wordpress.org` rather than `central.wordcamp.org` )_

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
StevenDufresne: Yeah, I agree. It isn't clear how these events will evolve. 

My intuition is telling me that the WordCamp branding is so strong that we'll likely not rebrand these to be something else outwardly. Instead we'll modify what we consider a WordCamp. I think what we are building now will be considered the default WordCamp and what is now `wordcamp` will be something like regional/flagship/featured/official or whatever the language. I think in the future, and maybe its not feasible, but the current `wordcamp` cpt slug could/should change.

Using that logic, I've left everything as is and didn't try to generalize the naming conventions. I could easily be convinced otherwise.

StevenDufresne: Alright, I pulled out direct references to `central` and `wordcamp` replacing it with WordCamp to denote that WordCamps could have different CPTs in [ec1a82d](https://github.com/WordPress/wordcamp.org/pull/937/commits/ec1a82d0a61236894f5db286d50abc58a2282bb5).

I also added some context to the main doc block. Mind taking another look? Does it make sense?

public_html/wp-content/mu-plugins/4-helpers-wcpt.php
StevenDufresne: See https://github.com/WordPress/wordcamp.org/pull/937/files#r1260469532

StevenDufresne:
I've made the case against a new CPT, and if that is agreed upon, this would not be needed. 

Setting to draft for now:
https://github.com/WordPress/wordcamp.org/issues/900#issuecomment-1633494392

timiwahalahti:
It looks like this PR broke quite a few things.

New `get_wordcamp_post_type` returns no value (on the docker environment at least) when viewing the dashboard or the site for a single WordCamp.

This always causes at least `get_wordcamps` and `get_wordcamp_post` to return null or empty in [`4-helpers-wcpt.php`](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/mu-plugins/4-helpers-wcpt.php).

I confirmed this by the Coming Soon page, which should show the dates for a WordCamp if those are added on the Central or show "Cancelled" in case the event is cancelled.

Other places I think this might affect:
* some aspects of Camptix, like virtual-only checks, allergy, special accommodations requests
* what data WordCamp dashboard widget shows for organisers
* attendee visa letter generation
* profile activity notifier
* various small tweaks

The `get_wordcamp_post_type` function tries to use `WCPT_POST_TYPE_ID` constant which is defined on [`wcpt` plugin](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php#L3). That plugin is not active on WordCamp sites for obvious reasons.

To fix this, either:
1. Define `WCPT_POST_TYPE_ID` somewhere global, like `wp-config.php`
2. On `get_wordcamp_post_type` function use hard-coded values instead of empty as the default
3. since `4-helpers-wcpt.php` is loaded across the network, hard-code values in `get_wordcamp_post_type` function and change `WCPT_POST_TYPE_ID` references elsewhere to use that function

The last option might cause also unexpected issues, so I'd personally choose option 1 or 2. The first option sounds like the most bulletproof.

timiwahalahti:
Checked against some sites on production:
* [Finland 2023](https://finland.wordcamp.org/2023/wp-admin) attendee visa generation
* [London 2023](https://london.wordcamp.org/2023/) Coming Soon page

Both seem to correctly get the info (dates) from the WordCamp CPT. This makes me wonder whether this PR hasn't yet been deployed to production, `WCPT_POST_TYPE_ID` is defined already somewhere else on the production, but change has not been reflected to the repo, or there's something I don't understand 🤔

Nevertheless, things are broken on the docker env after this PR.

StevenDufresne:
> Define WCPT_POST_TYPE_ID somewhere global, like wp-config.php

I don't mind this option, but there could be paths that are "broken but working" because its not defined. Small risk but a risk nonetheless.

> On get_wordcamp_post_type function use hard-coded values instead of empty as the default

I think this would be best given the circumstances. I'll add back in `wordcamp` as the default as opposed to an empty string.  It'll align it better with the existing functionality.

PR #938, opened 2023-07-11
Update `get_sites()` calls in `mu-plugins` for Pilot Events

iandunn:
WIP

See #906 

check all `get_sites()` because they include all networks by default. most queries will probably only want the wordcamp network, but it needs to be decided on a case by case basis because the context matters

1.
2.
3.


StevenDufresne:
What's left here?

iandunn:
I think we can close it. Updating the `set-skip-feature-flag` command would be nice to do one day, but not high priority.

PR #939, opened 2023-07-12
Add First Time being a organizer/speaker/sponsor question to CPT

### LINKED ISSUE #888: Measure first-time participation

outdoor2kode:
See #888

This PR adds First Time being an organizer/speaker/sponsor question to CPT and prepares CPT stub for form automation in order to allow the WordCamp Counts to retrieve the first_time value.

### Screenshots
**Sponsor**
image
**Speaker**
image
**Organizer**
image

1. Go to individual sites.
2. Go to Sponsors/Organizers/Speakers and click on anyone listed there.
3. You'll see the meta box questions as the above screenshots show.

### Questions
1. I'm curious if the timing for a volunteer/speaker/sponsor to fill in the first-time volunteer/speaker/sponsor can only be through the initial form that's created on a new website? If a user deletes this initial form, is there another way to retrieve it later?

2. Is the timing to fill in the first-time organizer only through the organizer's block editor page side panel? Is there a need to prepare an organizer stub for form automation?

### DIFF
--- a/public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
+++ b/public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
@@ -391,7 +391,7 @@ class MES_Sponsor {
 
 		$text_fields = array(
 			'company_name', 'website', 'first_name', 'last_name', 'email_address', 'phone_number', 'twitter_handle',
-			'street_address1', 'street_address2', 'city', 'state', 'zip_code', 'country', 'first_time',
+			'street_address1', 'street_address2', 'city', 'state', 'zip_code', 'country',
 		);
 
 		foreach ( $text_fields as $field ) {
--- a/public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
+++ b/public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
@@ -391,7 +391,7 @@ class MES_Sponsor {
 
 		$text_fields = array(
 			'company_name', 'website', 'first_name', 'last_name', 'email_address', 'phone_number', 'twitter_handle',
-			'street_address1', 'street_address2', 'city', 'state', 'zip_code', 'country',
+			'street_address1', 'street_address2', 'city', 'state', 'zip_code', 'country', 'first_time',
 		);
 
 		foreach ( $text_fields as $field ) {
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -44,6 +44,20 @@ function register_sponsor_post_meta() {
 			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
 		)
 	);
+	register_post_meta(
+		'wcb_sponsor',
+		'_wcpt_sponsor_first_time',
+		array(
+			'type'          => 'string',
+			'show_in_rest'  => array(
+				'schema' => array(
+					'context' => array( 'edit' ),
+				),
+			),
+			'single'        => true,
+			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
+		)
+	);
 }
 
 /**
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -106,6 +106,20 @@ function register_speaker_post_meta() {
 			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
 		)
 	);
+	register_post_meta(
+		'wcb_speaker',
+		'_wcb_speaker_first_time',
+		array(
+			'type'          => 'string',
+			'show_in_rest'  => array(
+				'schema' => array(
+					'context' => array( 'edit' ),
+				),
+			),
+			'single'        => true,
+			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
+		)
+	);
 }
 
 /**
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -46,7 +46,7 @@ function register_sponsor_post_meta() {
 	);
 	register_post_meta(
 		'wcb_sponsor',
-		'_wcpt_sponsor_first_time',
+		'_wcb_sponsor_first_time',
 		array(
 			'type'              => 'string',
 			'show_in_rest'      => array(
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -48,14 +48,15 @@ function register_sponsor_post_meta() {
 		'wcb_sponsor',
 		'_wcpt_sponsor_first_time',
 		array(
-			'type'          => 'string',
-			'show_in_rest'  => array(
+			'type'              => 'string',
+			'show_in_rest'      => array(
 				'schema' => array(
 					'context' => array( 'edit' ),
 				),
 			),
-			'single'        => true,
-			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
+			'single'            => true,
+			'auth_callback'     => __NAMESPACE__ . '\meta_auth_callback',
+			'sanitize_callback' => __NAMESPACE__ . '\sanitize_first_time_value',
 		)
 	);
 }
@@ -124,14 +125,15 @@ function register_speaker_post_meta() {
 		'wcb_speaker',
 		'_wcb_speaker_first_time',
 		array(
-			'type'          => 'string',
-			'show_in_rest'  => array(
+			'type'              => 'string',
+			'show_in_rest'      => array(
 				'schema' => array(
 					'context' => array( 'edit' ),
 				),
 			),
-			'single'        => true,
-			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
+			'single'            => true,
+			'auth_callback'     => __NAMESPACE__ . '\meta_auth_callback',
+			'sanitize_callback' => __NAMESPACE__ . '\sanitize_first_time_value',
 		)
 	);
 }
@@ -272,14 +274,15 @@ function register_organizer_post_meta() {
 		'wcb_organizer',
 		'_wcb_organizer_first_time',
 		array(
-			'type'          => 'string',
-			'show_in_rest'  => array(
+			'type'              => 'string',
+			'show_in_rest'      => array(
 				'schema' => array(
 					'context' => array( 'edit' ),
 				),
 			),
-			'single'        => true,
-			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
+			'single'            => true,
+			'auth_callback'     => __NAMESPACE__ . '\meta_auth_callback',
+			'sanitize_callback' => __NAMESPACE__ . '\sanitize_first_time_value',
 		)
 	);
 }
@@ -811,4 +814,21 @@ function add_larger_avatar_sizes( $sizes ) {
 	$sizes[] = 512;
 
 	return $sizes;
+}
+
+/**
+ * Sanitize the 'first_time' meta value before storing it.
+ *
+ * @param string $value The unsanitized meta value.
+ *
+ * @return string The sanitized meta value.
+ */
+function sanitize_first_time_value( $value ) {
+	$allowed_values = array( 'yes', 'no', 'unsure' );
+
+	if ( in_array( $value, $allowed_values, true ) ) {
+		return $value;
+	} else {
+		return 'unsure';
+	}
 }
--- a/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
+++ b/public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
@@ -240,6 +240,20 @@ function register_organizer_post_meta() {
 			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
 		)
 	);
+	register_post_meta(
+		'wcb_organizer',
+		'_wcb_organizer_first_time',
+		array(
+			'type'          => 'string',
+			'show_in_rest'  => array(
+				'schema' => array(
+					'context' => array( 'edit' ),
+				),
+			),
+			'single'        => true,
+			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
+		)
+	);
 }
 
 /**
--- a/public_html/wp-content/plugins/wc-post-types/views/sponsors/metabox-sponsor-info.php
+++ b/public_html/wp-content/plugins/wc-post-types/views/sponsors/metabox-sponsor-info.php
@@ -285,37 +285,37 @@
 	</li>
 
 	<li>
-		<label for="_wcpt_sponsor_first_time_yes">
+		<label for="_wcb_sponsor_first_time_yes">
 			<?php esc_html_e( 'Yes', 'wordcamporg' ); ?>
 		</label>
 		<input
 			type="radio"
-			id="_wcpt_sponsor_first_time_yes"
-			name="_wcpt_sponsor_first_time"
+			id="_wcb_sponsor_first_time_yes"
+			name="_wcb_sponsor_first_time"
 			value="yes"
 			<?php checked( $first_time, 'yes' ); ?>
 		/>
 	</li>
 	<li>
-		<label for="_wcpt_sponsor_first_time_no">
+		<label for="_wcb_sponsor_first_time_no">
 			<?php esc_html_e( 'No', 'wordcamporg' ); ?>
 		</label>
 		<input
 			type="radio"
-			id="_wcpt_sponsor_first_time_no"
-			name="_wcpt_sponsor_first_time"
+			id="_wcb_sponsor_first_time_no"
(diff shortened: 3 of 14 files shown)


public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
```diff
@@ -391,7 +391,7 @@ protected function save_post_meta( $post_id, $new_values ) {
 
 		$text_fields = array(
 			'company_name', 'website', 'first_name', 'last_name', 'email_address', 'phone_number', 'twitter_handle',
-			'street_address1', 'street_address2', 'city', 'state', 'zip_code', 'country',
+			'street_address1', 'street_address2', 'city', 'state', 'zip_code', 'country', 'first_time',
```
iandunn: I don't think we'll need to update the Multi-Event Sponsors posts on Central, since they're just a template for the posts that gets created on individual sites.

In fact, when we create the sponsor posts for Multi Event Sponsors on new sites, we should probably just automatically make the value `no`. That would be done in `get_stub_me_sponsors_meta()`

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -44,6 +44,20 @@ function register_sponsor_post_meta() {
...
+		'_wcpt_sponsor_first_time',
+		array(
+			'type'          => 'string',
+			'show_in_rest'  => array(
+				'schema' => array(
+					'context' => array( 'edit' ),
+				),
+			),
+			'single'        => true,
+			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
```
iandunn: It'd probably be helpful to add a `validation_callback` to this (and the new speaker/organizer fields), to make sure it can only be updated with one of the known values.

public_html/wp-content/plugins/wc-post-types/views/sponsors/metabox-sponsor-info.php
```diff
@@ -278,6 +279,49 @@ class="regular-text"
...
+	<li>
+		<label for="_wcpt_sponsor_first_time_yes">
+			<?php esc_html_e( 'Yes', 'wordcamporg' ); ?>
+		</label>
+		<input
+			type="radio"
+			id="_wcpt_sponsor_first_time_yes"
+			name="_wcpt_sponsor_first_time"
+			value="yes"
+			<?php echo 'yes' === $first_time  ? 'checked' : ''; ?>
```
iandunn: The `checked()` function will make this a bit simpler

public_html/wp-content/plugins/wc-post-types/views/sponsors/metabox-sponsor-info.php
```diff
@@ -278,6 +279,49 @@ class="regular-text"
...
+			type="radio"
+			id="_wcpt_sponsor_first_time_no"
+			name="_wcpt_sponsor_first_time"
+			value="no"
+			<?php echo 'no' === $first_time ? 'checked' : ''; ?>
+		/>
+	</li>
+	<li>
+		<label for="_wcpt_sponsor_first_time_unsure">
+			<?php esc_html_e( 'Unsure', 'wordcamporg' ); ?>
```
iandunn: `I don't know` might flow a little better as the label, and match #892. `unsure` is a good value for the key, though 👍🏻 

That's not a big deal, though

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
```diff
@@ -42,7 +42,7 @@ public function print_front_end_styles() {
 		?>
 
 		<style>
-			<?php require_once( __DIR__ . '/front-end.css' ); ?>
+			<?php require_once __DIR__ . '/front-end.css'; ?>
```
iandunn, line 45: The changes in this file (and the related ones in other files) are a good thing to add, but it'd be easier to review if it were a separate PR. They're directly related to the scope of #888, and this PR is already touching a lot of different files.

public_html/wp-content/plugins/wcpt/stubs/post/call-for-speakers.php
```diff
@@ -9,9 +9,9 @@
 <!-- wp:jetpack/contact-form {"subject":"WordCamp Speaker Request","hasFormSettingsSet":"yes"} -->
 <!-- wp:jetpack/field-name {"label":"Name","required":true} /-->
 
-<!-- wp:jetpack/field-email {"label":"Email Address","required":true} /-->
+<!-- wp:jetpack/field-email {"label":"Email","required":true,"requiredText":"(required)","id":"speaker-email"} /-->
 
-<!-- wp:jetpack/field-text {"label":"WordPress.org Username","required":true} /-->
+<!-- wp:jetpack/field-text {"label":"WordPress.org Username","requiredText":"(required)","id":"speaker-username"} /-->
```
iandunn: It looks like this may have removed the `required` attribute; was that intentional?

public_html/wp-content/plugins/wcpt/stubs/post/call-for-sponsors.php
```diff
@@ -9,13 +18,15 @@
...
-<!-- wp:jetpack/field-email {"label":"Email","required":true} /-->
+<!-- wp:jetpack/field-email {"label":"Email","required":true,"requiredText":"(required)","id":"sponsor-email"} /-->
 
 <!-- wp:jetpack/field-telephone {"label":"Phone Number"} /-->
 
 <!-- wp:jetpack/field-select {"label":"Sponsorship Level","options":["Bronze","Silver","Gold"]} /-->
 
 <!-- wp:jetpack/field-textarea {"label":"Why Would you Like to Sponsor WordCamp?","required":true} /-->
 
+<!-- wp:jetpack/field-radio {"label":"Is this your first time being a sponsor at a WordPress event?","requiredText":"(required)","options":["Yes","No","Unsure"],"id":"first-time-sponsor"} /-->
```
iandunn: This doesn't have `required: true`, does Jetpack assume it's required if `requiredText` is present?

iandunn (commented):
I left a few suggestions, but I think this is pretty good overall and heading in the right direction 👍🏻

public_html/wp-content/plugins/wc-post-types/views/sponsors/metabox-sponsor-info.php
outdoor2kode: Didn't know this, Thanks!
See ccc5851038fe2897010eaaa16362d0b08cdaa71c

public_html/wp-content/plugins/wc-post-types/views/sponsors/metabox-sponsor-info.php
outdoor2kode: 7d1d42efec067cafe7d73e5a18f3605158b6c2dd

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
outdoor2kode: Ah, I've been particularly mindful of this. Because I've turned on fixOnSave, and noticed there weren't many auto-fixes this time so I thought to include them as well. But would you suggest that even if there are just two or three of these linter fixes, they should be placed in a separate PR to avoid confusion?

public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
outdoor2kode: Thanks for the context 👍 I've removed it in e7e6481e20e8a08d015200a73f9073a863cb8c53.
And set the value `no` in `get_stub_me_sponsors_meta()` in 05f2f750661a6ca1fa8c7b6ad4ba2ae41b8401c8.

But for ^, I'm a bit not sure how we usually test it. To my understanding, the sponsors in the MES post type on Central seem always to be added to individual sites, and I guess they would be added automatically when an individual site is set up. So is building a new site a common way to test the commit here? (If it's not the common way, what would it be? And at the same time, I'm curious, is the process of automatically creating an individual WordCamp site something I can test in a sandbox?

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -44,6 +44,20 @@ function register_sponsor_post_meta() {
 			'auth_callback' => __NAMESPACE__ . '\meta_auth_callback',
 		)
 	);
+	register_post_meta(
+		'wcb_sponsor',
+		'_wcpt_sponsor_first_time',
```
outdoor2kode: @iandunn I'm wondering what's `wcb`? 
For prefix of `_first_time` for other CPT, I'm all using `wcb`, but for sponsor, I'm using `wcpt` and is due to this [line](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wc-post-types/wc-post-types.php#L1392). If `wcb` is considered a necessary naming here, I guess it needs customizing for `first_time`

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
outdoor2kode: I've added in 8b48ac1f98b2ce67b0cc16e541514627eabeab9a via `sanitize_callback`. Not sure if it's a common one to do so, but I couldn't find any about `validation_callback`, or you were suggesting that I integrate the `validation_callback` into `auth_callback` here?

public_html/wp-content/plugins/wcpt/stubs/post/call-for-speakers.php
outdoor2kode: And this one I was following how you handle WordPress.org Username for volunteers [here](https://github.com/WordPress/wordcamp.org/pull/898/commits/fb0539fa72dd0edf42e4073d9f9e63a2c4319a1f#diff-1b24c2ff03c2faefbbd79cdb6835dd3785886ccd00f0d1dbd39865ce8bc5084eR18), so does it mean that we need speaker's .org username but optional for volunteer?

public_html/wp-content/plugins/wcpt/stubs/post/call-for-sponsors.php
outdoor2kode: I also referred to how you handled this field for volunteers [here](https://github.com/WordPress/wordcamp.org/pull/898/commits/fb0539fa72dd0edf42e4073d9f9e63a2c4319a1f#diff-1b24c2ff03c2faefbbd79cdb6835dd3785886ccd00f0d1dbd39865ce8bc5084eR24), do we want to make this radio question required or optional for all roles, or only optional for volunteers?

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
iandunn: A few lines of linter changes are fine 👍🏻 

In this case I was thinking that changes related to the forms-to-drafts plugin, like the stubs and `wcfd-key` were disconnected enough that they'd make more sense as a separate PR. After taking a second look, though, I think you were right to include it here 👍🏻

public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
iandunn: Yeah, you'd need to create a new site to test it. It's not good to do that on your w.org sandbox, since that changes the production database. I'd do it in a local Docker env instead.

In mine I have a WordCamp post called "new site", and I've moved it through the various post statuses so that it has the "needs site" status. Then you can fill in the URL field and check the `create site` box, and update the post. That'll automatically copy the MES posts to the site.

It you wanna repeat it, you can `wp site delete {site ID}` and then check the box again.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
iandunn: I think it's a legacy from when the `wc-post-type` code was in the old `wordcamp-base` theme. It's a bad practice to put code inside a theme if it's not related to the appearance of a site, because then post types will be unregistered etc if you switch themes. Eventually the code was extracted into a plugin, but the original `wcb` prefix was kept for convenience.

🤔 I'm not sure why the `wc-post-types` plugin uses `wcpt` to prefix sponsor posts, that seems wrong to me. I'm guessing it was just an oversight. I think you were right to keep things consistent 👍🏻 

Kudos for noticing details like that and wanting to understand the background 👍🏻

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
iandunn: Ah, you're right, `validate_callback` is only available on `register_rest_route()`, so `sanitize_callback()` is correct to use here 👍🏻

public_html/wp-content/plugins/wcpt/stubs/post/call-for-sponsors.php
iandunn: Ah, it looks like Jetpack saves `requiredText` even if `required: true` is not present and the default `requiredText` is being used. That felt unexpected to me, but I can see why they did it that way.

I think it's fine to leave it optional 👍🏻

public_html/wp-content/plugins/wcpt/stubs/post/call-for-speakers.php
iandunn: 🤔 That's a good question.

[Removing the requirement for speakers](https://make.wordpress.org/community/2017/03/02/requiring-wordcamp-speakers-to-have-a-wordpress-org-account/) has been discussed before, but there wasn't a consensus, so I think we should keep it required for now.

I'm not sure if it should be required for volunteers. Many of the same arguments apply, but they're also less likely to have one, and less visible than speakers, so it's not clear to me. That might need a discussion and decision from the Community Team before we act on it, so we can probably just ignore it for this PR.

outdoor2kode: Updated in f6c7b46e7de4109d8d379a9f75ccdda327315176

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
outdoor2kode: Updated in 3f85cc7517eac6897539c4d6b24676ad3a34ecb2

public_html/wp-content/plugins/multi-event-sponsors/classes/mes-sponsor.php
outdoor2kode: > In mine I have a WordCamp post called "new site", and I've moved it through the various post statuses so that it has the "needs site" status. Then you can fill in the URL field and check the create site box, and update the post. That'll automatically copy the MES posts to the site.
It you wanna repeat it, you can wp site delete {site ID} and then check the box again.

Much appreciated for sharing this 🙏 I've successfully done testing locally and can see that all copied MES posts have their first_time value set to `no`.

outdoor2kode:
@iandunn Much appreciated to you answering all my questions and I've learned a lot. I've made some commit updates based on your last round of feedback and went ahead to merge the PR 👍 
But you might have missed the two questions that I added to the [PR description](https://github.com/WordPress/wordcamp.org/pull/939#issue-1800023228) later. They won't affect this PR, but they are extended questions that I thought of from this issue. It'd be nice if you could shed some light on them.

iandunn:
Great job 🎉 

> I'm curious if the timing for a volunteer/speaker/sponsor to fill in the first-time volunteer/speaker/sponsor can only be through the initial form that's created on a new website? If a user deletes this initial form, is there another way to retrieve it later?

There isn't an automated way to recreate the form, they'd need to manually create the form with the same questions and add the `wcfd-key` custom field. That's why we have the warning that modifying the form can break the automation.

The post types will still have the questions and data, though, so manually created posts could still collect the data.

> Is the timing to fill in the first-time organizer only through the organizer's block editor page side panel? Is there a need to prepare an organizer stub for form automation?

It's the same as the other post types, except that we don't have the automation set up yet, and we don't have a Call for Organizers post. That's probably less needed than the others, since organizers are often chosen before the site is created, but there could be some camps that would benefit from it. It's never been a high priority though.

PR #945, opened 2023-07-16
Add user switching plugin and upgrade directory to .gitignore

timiwahalahti:
Added a user-switching plugin to be ignored on the repo. The plugin makes it way easier to test different cases that involve working with roles and/or caps. It's highly unlikely that it would be run in production because of security and privacy, so I deemed that it could go into .gitignore without causing problems.

Also added `wp-content/upgrade` to be ignored. The folder is used to store files when WordPress Core performs any upgrades to itself, plugins or themes. These files nor the root folder itself do not add any value to the repo, since it's used only temporarily during upgrade processes.


pkevan:
No strong opinions about the addition of the `user-switching` plugin. I've never used it or had use for it previously.

Happy to wait for others to chime in here.

ryelle:
I like the `user-switching` plugin, though I don't use it often. I just leave the folder untracked when I do, but gitignoring it would be nice  👍🏻

PR #948, opened 2023-07-16
Ensure that special accommodations and allergy emails go only to unique addresses

### LINKED ISSUE #877: CampTix: Duplicate emails for allergy & accommodations notification
The automated emails when someone buys a ticket and flags that they have an allergy or need accommodations are sent to both the "Email Address (Lead organizer)" and "E-mail Address (Event details)" — in some cases these could be the same email, which triggers two of the same email to be sent.

It should check that the emails are unique before looping over to send the email.

See also #872.

timiwahalahti:
Prevent duplicate emails when the lead organiser and city email addresses are the same.

Fixes #877, See #947


pkevan (approved):
Looks good 👍

PR #949, opened 2023-07-16
Prevent WordCamps disabling Coming Soon before they are in schedule

### LINKED ISSUE #791: Prevent disabling Coming Soon mode before WordCamp is s
Sometimes WordCamp organisers are eager to publish their WordCamp and website even before Central has approved their budget and added them to the schedule. Each organiser is told not to announce anything before the event is officially scheduled - on rare occasions this is forgotten or ignored.

To avoid WordCamp going public too soon, we should prevent disabling Coming Soon mode before WordCamp is on the schedule.

timiwahalahti:
Sometimes WordCamp organisers are eager to publish their WordCamp and website even before Central has approved their budget and added them to the schedule. Each organiser is told not to announce anything before the event is officially scheduled - on rare occasions, this is forgotten or ignored.

To avoid WordCamp going public too soon, this change prevents disabling Coming Soon mode before WordCamp is on the schedule.

Fixes #791 

### Screenshots

image

Coming Soon settings when a user tries to save (publish) off mode while WordCamp is not in `wcpt-scheduled` status.

1. Edit any WordCamp on Central, and change the status to something else than "WordCamp Scheduled"
2. Open the dashboard of that WordCamp and open Coming Soon settings
3. If the Coming Soon is off, turn it on and publish
4. Try to turn the setting back to off, you should see the error message
5. Change the status back to "WordCamp Scheduled" on Central
6. Publishing with Coming Soon off should work again

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wccsp-customizer.php
+++ b/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wccsp-customizer.php
@@ -211,16 +211,6 @@ class WCCSP_Customizer {
 			return $validity;
 		}
 
-		/**
-		 * Short circuit when network admin is changing the value.
-		 *
-		 * On typical WordCamp site, at least some organisers do have administrator role so that
-		 * capability can not be used for this check.
-		 */
-		if ( in_array( 'manage_network', wp_get_current_user()->roles ) ) {
-			return $maybe_empty;
-		}
-
 		// Short circuit when deputy is changing the value.
 		if ( current_user_can( 'wordcamp_wrangle_wordcamps' ) ) {
 			return $validity;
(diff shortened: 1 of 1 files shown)


pkevan (commented):
Unsure if this is a valid concern, but is there a chance that this blocking workflow might want to be overridden i.e. by deputies or admins?

timiwahalahti:
> Unsure if this is a valid concern, but is there a chance that this blocking workflow might want to be overridden i.e. by deputies or admins?

Good point! Will refresh the PR soon :)

public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wccsp-customizer.php
```diff
@@ -178,4 +179,58 @@ public function enqueue_customizer_scripts() {
...
+			return $validity;
+		}
+
+		/**
+		 * Short circuit when network admin is changing the value.
+		 *
+		 * On typical WordCamp site, at least some organisers do have administrator role so that
+		 * capability can not be used for this check.
+		 */
+		if ( in_array( 'manage_network', wp_get_current_user()->roles ) ) {
```
pkevan: not sure this is necessary, as a super admin would pass the below condition for deputies too - perhaps we can just write a comment pertaining to this. Also, the return value looks like it should be `$validity`

pkevan (approved):

Issue #734, opened 2022-02-18
Restrict manual WCPT post creation


iandunn:
@NewYorkerLaura [reported](https://wordpress.slack.com/archives/C08M59V3P/p1645192555712969) that she can create wordcamp/meetup posts on Central, despite having a `contributor` role.

This looks intentional from the code, but we might want to consider if it's still desired.

https://github.com/WordPress/wordcamp.org/blob/bddaa4eaf5aea877d9f897aa564d677004db7fca/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php#L114-L128

Do we need manual post creation now that they're automatically created from applications? I'm guessing no, but we need to dig into it more. 

It might be good to still let super admins manually create them, just in case, but maybe not since that would probably result in missing metadata.

timiwahalahti:
Duplicate of #728 

> Do we need manual post creation now that they're automatically created from applications? I'm guessing no, but we need to dig into it more.

Sometimes deputies do need this. For example WCEU, WCUS and for some situations with Meetups. But other than deputies shouldn't be able to create posts. At least few have created post and thought that is the "application" process.

iandunn:
Ah, I missed that issue, thanks!

PR #950, opened 2023-07-16
Prevent users from creating new WordCamp and Meetup posts on central dashboard

### LINKED ISSUE #728: Do not allow creating WordCamps in central for other than deputies
It looks like everyone that has caps to access the WordCamps on the Central site, eg. previous organisers, can use the core "Add new" functionality to add new WordCamp to our tracker tool. Only deputies should be able to do this, as organisers should always submit an application from which the post is automatically generated.

Related: https://secure.helpscout.net/conversation/1768446604/127687/

timiwahalahti:
Prevent users from creating new WordCamp and Meetup posts on dashboard. In most of the cases, all posts should be created thru the public application forms in order to get all needed information and to initiate the vetting process correctly.

Expectation is made for users with administrator and deputy roles, as they need to create events manually from time to time.

Used wp_insert_post_empty_content hook is run fo creation and updates, which is why post ID needs to be checked. The hook short circuits creation of new post when truthy value is returned.

Fixes #728, See #734 

Props @NewYorkerLaura

### Screenshots

image

Message shown when user without admin or deputy role tries to create new event.

1. Log in as user with contributor role
2. Send a WordCamp application via public form https://central.wordcamp.test/wordcamp-organizer-application/
3. Log out and log in as user with admin or deputy role
4. Navigate to https://central.wordcamp.test/wp-admin/edit.php?post_type=wordcamp and you should see the application just sent
5. Navigate to https://central.wordcamp.test/wp-admin/post-new.php?post_type=wordcamp and you should be able to create a new WordCamp
6. Add the user with contributor role to the site
7. Log out and log in as user with contributor role
8. Navigate to central dashboard and try to update the WordCamp you send application for
9. On the central dashboard try to create new WordCamp or Meetup, you should see the error
10. On the central dashboard try to create new post, that should be possible

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
@@ -5,6 +5,8 @@
  * @package WordCamp Post Type
  */
 
+use function WordCamp\Trusted_Deputy_Capabilities\is_deputy as is_deputy;
+
 /**
  * Class Event_Admin
  *
@@ -65,6 +67,7 @@ abstract class Event_Admin {
 
 		add_action( 'send_decline_notification_action',  'Event_Admin::send_decline_notification', 10, 3 );
 
+		add_action( 'wp_insert_post_empty_content', array( $this, 'maybe_prevent_creation_of_new_post' ), 999, 2 );
 	}
 
 	/**
@@ -692,6 +695,59 @@ abstract class Event_Admin {
 			}
 		}
 
+	}
+
+	/**
+	 * Prevent users from creating new WordCamp and Meetup posts on dashboard. In most of the cases, all posts should be created
+	 * thru the public application forms in order to get all needed information and to initiate the vetting process correctly.
+	 *
+	 * Expectatio is made for users with administartor and deputy roles, as they need to create events manually from time to time,
+	 *
+	 * Used wp_insert_post_empty_content hook is run fo creation and updates, which is why post ID needs to be checked. The hook
+	 * short circuits creation of new post when truthy value is retuned.
+	 *
+	 * @param  boolean $maybe_empty Whether the post should be considered "empty".
+	 * @param  array   $postarr      Array of post data.
+	 *
+	 * @return boolean              Whether the post should be considered "empty".
+	 */
+	public function maybe_prevent_creation_of_new_post( $maybe_empty, $postarr ) {
+		$post_type = $postarr['post_type'];
+
+		// Apply only for WordCamp and Meetup post types.
+		if ( $this->get_event_type() !== $post_type ) {
+			return $maybe_empty;
+		}
+
+		// The action hooked into is used also when updating posts, which all users should be able to do based on their caps.
+		if ( ! empty( $postarr['ID'] ) ) {
+			return $maybe_empty;
+		}
+
+		// Doing the checks only on dashboard ensures that other use cases like application forms do still work.
+		if ( ! is_admin() ) {
+			return $maybe_empty;
+		}
+
+		// Allow admins to create new posts in event CPT's.
+		if ( in_array('administrator',  wp_get_current_user()->roles) ) {
+			return $maybe_empty;
+		}
+
+		// Allow deputies to create new posts in event CPT's.
+		if ( is_deputy( get_current_user_id() ) ) {
+			return $maybe_empty;
+		}
+
+		$error = new WP_Error(
+			'not_allowed_to_create_new_wcpt',
+			esc_html( wp_sprintf( 'Only administrators and deputies can create new %s\'s. You should probably be using the public application forms on central.wordcamp.org.', $post_type ) )
+		);
+
+		// Display the error.
+		wp_die( $error );
+
+		return $error;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jamieblomerus (approved):
**LGTM!** This resolves an important issue.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -65,6 +67,7 @@ public function __construct() {
 
 		add_action( 'send_decline_notification_action',  'Event_Admin::send_decline_notification', 10, 3 );
 
+		add_action( 'wp_insert_post_empty_content', array( $this, 'maybe_prevent_creation_of_new_post' ), 999, 2 );
```
pkevan: Looks like this is a filter, so should be run via `add_filter`

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -694,6 +697,59 @@ public function print_admin_notices() {
...
+	 *
+	 * Expectation is made for users with administrator and deputy roles, as they need to create events manually from time to time.
+	 *
+	 * Used wp_insert_post_empty_content hook is run fo creation and updates, which is why post ID needs to be checked. The hook
+	 * short circuits creation of new post when truthy value is returned.
+	 *
+	 * @param  boolean $maybe_empty Whether the post should be considered "empty".
+	 * @param  array   $postarr      Array of post data.
+	 *
+	 * @return boolean              Whether the post should be considered "empty".
```
pkevan: Can also return WP_Error

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -694,6 +697,59 @@ public function print_admin_notices() {
...
+			return $maybe_empty;
+		}
+
+		// The action hooked into is used also when updating posts, which all users should be able to do based on their caps.
+		if ( ! empty( $postarr['ID'] ) ) {
+			return $maybe_empty;
+		}
+
+		// Doing the checks only on dashboard ensures that other use cases like application forms do still work.
+		if ( ! is_admin() ) {
```
pkevan, line 726: This will also return TRUE for ajax requests, do we want to also consider this?

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -694,6 +697,59 @@ public function print_admin_notices() {
...
+			return $maybe_empty;
+		}
+
+		// Allow admins to create new posts in event CPT's.
```
pkevan: For roles checking we should use `current_user_can` against a capability, which can probably also wrap the check for deputy or higher too.

pkevan: the parenthesis below appear to be missing spacing to be consistent with the rest of the code.

pkevan (changes requested):
Thanks for the PR - I've added some inline comments to the PR, mostly it's all minor but good to have in there for consistency.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
timiwahalahti: Can't think of places where new event posts would be created via ajax request. And if someone manages to do that, it's not that a big deal - deputies do catch these manual entries.

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
timiwahalahti: Wanted to separate the admin and deputy checks to make it easier to see for who we are allowing the creation.

There is `wordcamp_wrangle_wordcamps` capability which seems to affect WordCamps and Meetups and should be assigned to deputies and mentors. Though not sure if it's given for admins, so we probably still need that check.

In any case, replacing the `is_deputy` with capability check seems necessary. Looking at the origins of that function again, it seems to check against `$trusted_deputies` array, which sounds a lot like a list that does not contain every deputy.

timiwahalahti: Looks like admins do have the `wordcamp_wrangle_wordcamps` capability, at least in the Docker env.

Thanks for noticing this and suggesting a different approach!

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -694,6 +695,54 @@ public function print_admin_notices() {
 
 	}
 
+	/**
+	 * Prevent users from creating new WordCamp and Meetup posts on dashboard. In most of the cases, all posts should be created
+	 * thru the public application forms in order to get all needed information and to initiate the vetting process correctly.
```
pkevan:
```suggestion
	 * through the public application forms in order to get all needed information and to initiate the vetting process correctly.
```

public_html/wp-content/plugins/wcpt/wcpt-event/class-event-admin.php
```diff
@@ -694,6 +695,54 @@ public function print_admin_notices() {
...
+	 *
+	 * Expectation is made for users with administrator and deputy roles, as they need to create events manually from time to time.
+	 *
+	 * Used wp_insert_post_empty_content hook is run fo creation and updates, which is why post ID needs to be checked. The hook
+	 * short circuits creation of new post when truthy value is returned.
+	 *
+	 * @param  boolean $maybe_empty Whether the post should be considered "empty".
+	 * @param  array   $postarr     Array of post data.
+	 *
+	 * @return mixed                Booleab whether the post should be considered "empty" or WP_Error in case user is not allowed to create post.
```
pkevan:
```suggestion
	 * @return mixed                Boolean whether the post should be considered "empty" or WP_Error in case user is not allowed to create post.
```

pkevan (approved):

jamieblomerus (approved):

jamieblomerus (approved):

PR #956, opened 2023-07-19
Allow switch to wcpt-rejected status from wcpt-needs-vetting

### LINKED ISSUE #865: Setting a WordCamp application status "Declined" after "Needs Vetting"
**What we need**

The ability to set a new WordCamp application status "Declined" after being vetting on WordCamp Tracker.

**Issue**

Currently, a new application can't be set to "Declined" right away, but we sometimes need to do that. The option is available in the list of WordCamp Status, but it is not selectable.

Decline

timiwahalahti:
As per the request, a WordCamp application status can be set directly to "Declined" after "Needs vetting" status. This simplifies the process for deputies.

Fixes #865 

Props @devinmaeztri 

### Screenshots

Before screenshot on the issue. After below.

image

1. Navigate to WordCamp tracker and select WordCamp that needs vetting
2. Open the status selection


pkevan (approved):
👍

devinmaeztri:
I've tested and it works as expected! Thank you so much, everyone!

PR #958, opened 2023-07-19
Add Tracker URL field to WordCamp details report

### LINKED ISSUE #823: Add field "Tracker URL" to report available fields
When we download reports from the [central report page](https://central.wordcamp.org/wp-admin/index.php?page=wordcamp-reports), we can choose the fields that we want to display in the spreadsheet columns (screenshot attached).

Now, we can check the WordCamp URL but we need to be able to also download the link to the tracker (the WordCamp post on central).

Examples:
URL (already available field): https://torino.wordcamp.org/2023/
Tracker (not yet available field):  https://central.wordcamp.org/wp-admin/post.php?post=3155301&action=edit

Can you please add a new "Available field" to the Meetup and WordCamp central report page for the URL to the tracker?

Central report available fields

Please let me know if you need further information on my end. 

Thank you

timiwahalahti:
This is an attempt to add new "Tracker URL" field to WordCamp details report. Couldn't test, since csv export utility is in `mu-plugins-private` for some reason.

```
Warning: require_once(/usr/src/public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync/utilities/class-export-csv.php): failed to open stream: No such file or directory in /usr/src/public_html/wp-content/mu-plugins/2-autoloader.php on line 27
```

Either that utility needs to be moved to a public repository, or this PR can act as a base for someone who has access to private plugins. I guess the latter one will be easier and faster.

Fixes #823 

Props @peiraisotta

1. Navigate to [WordCamp Details](https://central.wordcamp.test/wp-admin/index.php?page=wordcamp-reports&report=wordcamp-details) report
2. Do csv export, the field is selected by default

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-base-details.php
+++ b/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-base-details.php
@@ -171,10 +171,11 @@ abstract class Base_Details extends Base {
 		$meta_keys   = $this->get_meta_keys();
 
 		$row = [
-			'ID'      => $event->ID,
-			'Name'    => $event->post_title,
-			'Created' => get_the_date( 'Y-m-d', $event->ID ),
-			'Status'  => $event->post_status,
+			'ID'          => $event->ID,
+			'Tracker URL' => get_edit_post_link( $event->ID ),
+			'Name'        => $event->post_title,
+			'Created'     => get_the_date( 'Y-m-d', $event->ID ),
+			'Status'      => $event->post_status,
 		];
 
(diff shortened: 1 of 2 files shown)


pkevan:
> Either that utility needs to be moved to a public repository, or this PR can act as a base for someone who has access to private plugins. I guess the latter one will be easier and faster.

Turns out that the file in question is actually here: https://github.com/WordPress/wporg-mu-plugins/blob/trunk/mu-plugins/utilities/class-export-csv.php, but is built in a rather convoluted way (of which I don't know the history of). Is that helpful @timiwahalahti?

PR #962, opened 2023-07-20
Move password reset hooks to more logical place

timiwahalahti:
While looking at `mu-plugins/wcorg-misc.php`, I noticed a few hooks that alter the password reset functionality to guide users into using wordpress.org reset logic. Since there's also `mu-plugins/wcorg-login-message.php` which does the most part of changing login messages to indicate use their wordpress.org credentials, moving the reset changes to that file seemed appropriate.


pkevan (approved):
👍

PR #970, opened 2023-07-21
Do not send organiser reminders with "send X days after the camp ends" trigger if WordCamp didn't happen

### LINKED ISSUE #496: Organizer Reminders: Messages with a "send X days after the camp ends" trigger shouldn't send if the camp didn't happen
The routing that sends messages with a time trigger (before the camp, after the camp, etc.) queries for all camps that have either a public status (scheduled, closed) or one of the pre-planning statuses. In the case of the "X days after the camp ends" trigger, it then sends to the camp based on whether the current date is beyond X days from the camp's Start Date. However, this means that camps that had a start date, but were then postponed (and thus were set back to "pre-planning" status) might still get the email, even though the event never "ended" because it never happened.

One way to address this is to make sure when postponing a camp and moving it back to pre-planning that the start date is also removed. However, there are lots of cases where a camp isn't being postponed, but it is stuck somewhere in pre-planning and the originally hoped for date passes by, thus still triggering the reminder email.

A way to mitigate this in the code would be by adding an extra check within the "after the camp ends" trigger to ensure that the camp has a public status. This shouldn't break any existing reminders because the wording of the trigger "After the camp ends" makes the clear assumption that the event actually happened, and the only valid statuses for events that actually happened are the public ones, scheduled and closed.

timiwahalahti:
We sent organisers reminders with "send X days after the camp ends" trigger even if the WordCamp didn't happen. Especially during the pandemic, many WordCamps were put back to pre-planning but had dates in the tracker. This can also happen from time to time for various reasons.

The `wcor_send_after` emails were sent out due to the `$pending_wordcamps` query grabbing those emails, and in `timed_email_is_ready_to_send` function only checking the end date even the function is called from `send_timed_emails` for various different WordCamp statuses.

This PR adds a check against the status of WordCamp in `timed_email_is_ready_to_send` function for `wcor_send_after` trigger. Only public WordCamp statuses (in schedule, closed) do get the email.

Also, some minor code refactoring was done to make conditions easier to read and tests were updated for this scenario.

Fixes #496 

Props @coreymckrill

1. Schedule some WordCamps with pre-planning statuses and past date(s)
2. Run `wcor_send_timed_emails` cron job

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-organizer-reminders/tests/test-wcor-mailer.php
+++ b/public_html/wp-content/plugins/wordcamp-organizer-reminders/tests/test-wcor-mailer.php
@@ -168,6 +168,13 @@ class Test_WCOR_Mailer extends WP_UnitTestCase {
 			update_post_meta( self::$wordcamp_dayton_post_id, '_timestamp_added_to_planning_schedule', $compare_date );
 		}
 
+		if ( 'wcor_send_after' === $send_when ) {
+			wp_update_post( array(
+				'ID'          => self::$wordcamp_dayton_post_id,
+				'post_status' => 'wcpt-scheduled',
+			) );
+		}
+
 		$wordcamp = get_post( self::$wordcamp_dayton_post_id );
 
 		$this->assertSame( '', $wordcamp->wcor_sent_email_ids );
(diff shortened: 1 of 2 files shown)


outdoor2kode (approved):
Tested and Looks good. Thanks for some refactorings here!

PR #971, opened 2023-07-21
Do not require address for invoice and prevent sending email without attachment

### LINKED ISSUE #255: Invoices: Filling in the Address with a space causes an email without an attachment
When using the Invoices plugin, if you fill in the Address field with a space (Indicating "I don't care about the address") it fails to generate an invoice.

The CPT is created as a draft and an email is sent with `As requested during your purchase, please find attached an invoice for your tickets to "%s".` but has no attachment.

What I expected:
1. Checkout to fail requesting I fill in the required field (or JS trimming the fields client side before validation)
or 
2. Invoice to be created without an Address (I don't see it as a required field, but I can see that it would be in some countries)
or
3. Email not to be sent (Forcing me to contact the Organisers for one)

I would err on the side of 1 if Camptix does client-side validation, or 2 otherwise.
3 should be implemented regardless, as it's pointless sending a invoice email if no invoice can be attached.

timiwahalahti:
It was reported that an invoice email was sent without the invoice attached if the address field was left empty by bypassing the required attribute with a space character.

In this PR, we prevent the email if there's no attachment and make the address field optional.

Fixes #255 

Props dd32

### Screenshots

image

1. Activate Camptix Invoices plugin and enable it on Camptix settings
2. Go to tickets page, buy a ticket with invoice
3. Address field shouldn't be required
4. Invoice should be in email
5. Go buy another ticket with invoice, but before that modify `ctx_get_invoice` to bail early with false value
6. No invoice email received

### DIFF
--- a/public_html/wp-content/plugins/camptix-invoices/admin/js/camptix-invoices.js
+++ b/public_html/wp-content/plugins/camptix-invoices/admin/js/camptix-invoices.js
@@ -3,7 +3,7 @@
 jQuery( document ).ready( function( $ ) {
 	function toggleInvoiceDetailsForm( showForm ) {
 		const $camptixInvoiceDetailsForm = $( '.camptix-invoice-details' );
-		const $camptixInvoiceDetailsFormFields = $camptixInvoiceDetailsForm.find( 'input,textarea,select' );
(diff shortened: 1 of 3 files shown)


timiwahalahti:
Closing in favour of #965 which combines all the Camptix Invoices changes needed to make it network activated.

PR #979, opened 2023-07-25
Remove subrole restriction to show cross link menu item

### LINKED ISSUE #961: Fix `add_wcpt_cross_link` function
Function `add_wcpt_cross_link` in `mu-plugins/wcorg-misc.php` should be placing a 'WordCamp Post' link to the admin bar menu on camp sites. At the moment, that functionality does not seem to work. Let's fix that, since the idea is excellent!

https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/mu-plugins/wcorg-misc.php

iandunn:
Fixes #961 

The capability is needed on all sites for `add_wcpt_cross_link()` to work. 

Note: Being a deputy is required in order to pass the `is_user_member_of_blog()` check in `wp_admin_bar_site_menu()`.


iandunn:
This is working in manual tests, but it looks like the unit tests are failing for an unrelated reason. I'll look into that tomorrow.

PR #987, opened 2023-07-28
Add `wporg-mu-plugins/.../utilities` to docker environment

### LINKED ISSUE #769: Add `wporg-mu-plugins/.../utilities` to local environments
https://github.com/WordPress/wordcamp.org/commit/cb01557207a0dfef12530ca36655f271e04ce2ef and https://github.com/WordPress/wporg-mu-plugins/commit/6b0d108fdf0196ba2a0dfff60a8bcb0ab2987ee5 moved several files from this repo to `wporg-mu-plugins`, because they're also used on w.org.

production wordcamp.org has access to those files, and committers' local environments do too, but other local environments don't yet. production has it because they're part of `mu-plugins-private`, which can't be open-sourced. to get the files in most local envs, we could use composer to install `wporg-mu-plugins`.

this kinda works, but it installs the entire git repo, so the files end up in a subdirectory, which doesn't match the expected path.

```diff
diff --git a/composer.json b/composer.json
index 3a61311f..8778fccc 100644
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,20 @@
                        "dealerdirect/phpcodesniffer-composer-installer": true
                }
        },
+       "extra": {
+               "installer-paths" : {
+                       "public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync": ["wporg/wporg-mu-plugins"]
+               }
+       },
+       "repositories": [
+               {
+                       "type": "vcs",
+                       "url": "git@github.com:WordPress/wporg-mu-plugins.git"
+               }
+       ],
        "require": {
+               "composer/installers": "~1.0",
+               "wporg/wporg-mu-plugins": "dev-trunk",
                "adhocore/jwt": "^1.0"
        },
        "require-dev": {
```

if https://github.com/WordPress/wporg-mu-plugins/issues/141 is done, that would probably make this easier as well.

another option would be to have https://github.com/WordPress/wordcamp.org/blob/production/.docker/bin/extra-setup.sh do something like `svn export https://github.com/wordpress/wporg-mu-plugins/mu-plugins`. That would need to be re-run when breaking changes are made, though.

another option would be to install the repo to `vendor`, and have a constant that defines that path. production would point to `mu-plugins-private` but local envs would point to `vendor`. `wporg-mu-plugins` has a `type`, so we'd still need some composer config like above to overwrite the install path.

timiwahalahti:
Utilities were moved from the repo into `wporg-mu-plugins` repository, which broke for example CSV exports and meetup.com API calls. 

This PR adds that repository and uses Composer `post-update-cmd` [scripts](https://getcomposer.org/doc/articles/scripts.md) to move the utilities from `wporg-mu-plugins` into the path where `2-autoloader.php` expects it to be.

We cannot use `git@github.com` format as the repository URL, since that fails on Docker. We would need to add github.com into `~/.ssh/known_hosts` so it's just easier to use HTTPS instead of SSH. See https://github.com/docker-library/golang/issues/148 for more details.

I'm a bit puzzled about why the repository ends up in root `wp-content` (outside `public_html`) instead of `vendor`, but that seems to be the case, and we clean up a bit later so didn't bother to care too much 🤷‍♂️

In `post-update-cmd` let's first create the needed parent directories. The `-p` flag on the [`mkdir`](https://linux.die.net/man/1/mkdir) command does that. Then move only `utilities` directory, since we don't need anything else from the repo. Lastly do some cleanup and remove the `wp-content` directory where the repository was installed.

Fixes #769

Props @iandunn 

1. Open Docker command shell
2. `cd ../` because shell opens in `public_html` instead of our root directory
3. `rm -r composer.lock` for good measure
4. `php composer.phar install`
5. Check the logs in case of errors
6. Verify that the utilities directory is where it should be, `ls public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync`
7. Open [central.wordcamp.test](central.wordcamp.test) and navigate to [reports](https://central.wordcamp.test/wp-admin/index.php?page=wordcamp-reports)
8. Run an CSV export on some report, for example the [WordCamp Details](https://central.wordcamp.test/wp-admin/index.php?page=wordcamp-reports&report=wordcamp-details)
9. Check that CSV file is generated (should auto download) and looks good

### DIFF
--- a/.docker/config/composer.json
+++ b/.docker/config/composer.json
@@ -26,6 +26,10 @@
 			"url": "https://wpackagist.org/"
 		},
 		{
+			"type": "vcs",
+			"url": "https://github.com/WordPress/wporg-mu-plugins.git"
+		},
+		{
 			"type": "package",
 			"package": [
 				{
@@ -119,6 +123,7 @@
 		"wpackagist-plugin/wp-cldr": "*",
 		"wpackagist-plugin/wp-super-cache": "*",
 		"wordpress-meta/wporg-profiles-wp-activity-notifier": "1.1",
+		"wporg/wporg-mu-plugins": "dev-build",
 		"wpackagist-theme/p2": "*",
 		"wpackagist-theme/twentyten": "*",
 		"wpackagist-theme/twentyeleven": "*",
@@ -133,5 +138,12 @@
 		"wpackagist-theme/twentytwentyone": "*",
 		"wpackagist-theme/twentytwentytwo": "*",
 		"wpackagist-theme/twentytwentythree": "*"
+	},
+	"scripts": {
+    "post-update-cmd": [
+    	"mkdir -p public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync",
+    	"mv wp-content/mu-plugins/wporg-mu-plugins/mu-plugins/utilities public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync",
+    	"rm -rf wp-content/mu-plugins/wporg-mu-plugins"
+    ]
 	}
 }
--- a/.docker/config/composer.json
+++ b/.docker/config/composer.json
@@ -17,6 +17,7 @@
 	"extra": {
 		"installer-paths": {
 			"public_html/wp-content/plugins/{$name}/": ["type:wordpress-plugin"],
+			"public_html/wp-content/mu-plugins/{$name}/": ["type:wordpress-muplugin"],
 			"public_html/wp-content/themes/{$name}/": ["type:wordpress-theme"]
 		}
 	},
@@ -142,8 +143,8 @@
 	"scripts": {
     "post-update-cmd": [
     	"mkdir -p public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync",
-    	"mv wp-content/mu-plugins/wporg-mu-plugins/mu-plugins/utilities public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync",
-    	"rm -rf wp-content/mu-plugins/wporg-mu-plugins"
+    	"mv public_html/wp-content/mu-plugins/wporg-mu-plugins/mu-plugins/utilities public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync",
+    	"rm -rf public_html/wp-content/mu-plugins/wporg-mu-plugins"
     ]
 	}
 }
--- a/.docker/config/composer.json
+++ b/.docker/config/composer.json
@@ -141,10 +141,10 @@
 		"wpackagist-theme/twentytwentythree": "*"
 	},
 	"scripts": {
-    "post-update-cmd": [
-    	"mkdir -p public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync",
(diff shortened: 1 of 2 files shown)


timiwahalahti:
> I'm a bit puzzled about why the repository ends up in root wp-content (outside public_html) instead of vendor

Ahh, of course! 🤦‍♀️ It's because we have `composer-installer` package, which detects the `wordpress-muplugin` type and there's no configuration in `installer-paths` for that type. That makes the `composer-installer` build the "correct" path for MU plugins starting from the root.

To avoid future problems, let's configure the path for  `wordpress-muplugin` on the same go.

iandunn (approved):
This worked for me 👍🏻 

I'm guessing there's probably a way to have it install to the desired path to begin with, rather than using `post-update-cmd`, but it's probably not worth digging into since this works.

timiwahalahti:
Tried to look for native ways to cherry-pick folders from the repo and set path for it. Looks like composer nor `composer-installer` supports it. There's [Composer-Dropin-Installer](https://github.com/Koodimonni/Composer-Dropin-Installer) for example which does something similar, but only for file at the time and that's not sufficient in this case. Also thought that it's cleaner to have `post-update-cmd` rather than an additional repository which does the trickery.

iandunn:
🤔 I was thinking it might work to [set a custom `installer-path` for the specific package](https://getcomposer.org/doc/faqs/how-do-i-install-a-package-to-a-custom-path-for-my-framework.md), rather than _type_ of package:

```json
"installer-paths": {
    "public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync": ["wporg/wporg-mu-plugins"],
}
```

For that to work it may require overriding the `package` definition for the project, but maybe not.

```json
"repositories": [
	{
		"type": "package",
		"package": [
			{
				"name": "wporg/wporg-mu-plugins",
				"type": "vcs",
				"version": "dev-build", /* or maybe just `build` ? */
				"source": {
					"type": "vcs",
					"url": "https://github.com/WordPress/wporg-mu-plugins.git"
					"reference": "dev-build"  /* same as above ? */
				}
			}
		]
	}
]
```

I haven't tested any of that, though, and I wouldn't be surprised at all if it doesn't work. I've spent more hours banging my head against a wall w/ Composer than I care to admit 😆 I think it's totally fine to leave things the way they are.

fumikito:
@iandunn 
Hi, I got the error below while building local env with the installation [doc](https://github.com/fumikito/wordcamp.org/blob/production/.docker/readme.md#initial-setup):

```
docker compose exec wordcamp.test wp site list --field=url                                                                    
WARN[0000] /Users/me/Documents/GitHub/wordcamp.test/docker-compose.yaml: `version` is obsolete 
Warning: require_once(/usr/src/public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync/loader.php): failed to open stream: No such file or directory in /usr/src/public_html/wp-content/mu-plugins/load-other-mu-plugins.php on line 21
Fatal error: require_once(): Failed opening required '/usr/src/public_html/wp-content/mu-plugins-private/wporg-mu-plugins/pub-sync/loader.php' (include_path='.:/usr/local/lib/php') in /usr/src/public_html/wp-content/mu-plugins/load-other-mu-plugins.php on line 21
Error: There has been a critical error on this website.Learn more about troubleshooting WordPress. There has been a critical error on this website.
```

Seems like `pub-sync` is missing in `public_html/wp-content/mu-plugins-private/tm/wporg-mu-plugins`.

PR #990, opened 2023-07-28
Allow bypassing max ticket limit when using coupon code

### LINKED ISSUE #740: CampTix: Allow bypassing the 10 ticket per order limit with a special link?
Sometimes companies bulk purchase WordCamp tickets for employees, and the 10 ticket per order limit means more clicking and multiple invoices to process.

Why the limit in the first place?

The reasoning is not known for sure, but it's probably related to preventing people from buying/reserving a big block of tickets and then scalping them or just not showing up, denying other people access to the event.

One solution might be to allow a WordCamp site to generate a special link that goes to the ticket order form, but without the limits (or at least with much larger limits).

timiwahalahti:
Sometimes, there is a need for a sponsor or a company to buy more than the maximum allowed tickets per purchase. Currently, the maximum number of tickets per purchase is 10.

These needs typically go hand in hand with coupon codes. So this PR adds a bulk buy yes/no field to coupon options, message about changed max quantity when coupon is applied and an filter to override max quantity also after coupon with bypass has been applied. When bulk buy is active, triple the normal max quantity.

Fixes #740

Props @jaz-on, @coreymckrill

### Screenshots

image

image

1. Open WordCamp site dashboard
2. Create ticket
3. Create coupon code, you should see the new setting set value to yes
4. Open tickets page, max quantity should be 10
5. Apply coupon code, max quantity should be 30 and message shown
6. Add following lines to `mu-plugins/sandbox-functionality.php`
```php
add_filter( 'camptix_max_tickets_per_order', function() {
  return 1;
} );
```
7. Open tickets page without coupon, max quantity should be 1
8. Apply coupon, max quantity should be 3
9. Proceed and buy tickets

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -5205,7 +5205,12 @@ class CampTix_Plugin {
 				$coupon->tix_discount_price = (float) get_post_meta( $coupon->ID, 'tix_discount_price', true );
 				$coupon->tix_discount_percent = (int) get_post_meta( $coupon->ID, 'tix_discount_percent', true );
 				$coupon->tix_applies_to = (array) get_post_meta( $coupon->ID, 'tix_applies_to' );
+				$coupon->tix_bypass_max_tickets_per_order = true;
 				$this->coupon = $coupon;
+
+				if ( $coupon->tix_bypass_max_tickets_per_order ) {
+					$max_tickets_per_order = $max_tickets_per_order * 3;
+				}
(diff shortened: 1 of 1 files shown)


jaz-on:
Many thanks @timiwahalahti this will be very helpful for future WCEU editions specifically as we're offering coupons for more than the current limit of tickets allowed in one single order for main sponsors.

outdoor2kode (approved):
Checked the logic and did testing. It looks good.
Screenshot 2024-04-30 at 03 52 22

PR #998, opened 2023-08-06
Enhance next gen event budget tool

### LINKED ISSUE #976: WordPress NextGen Event Budget Enhancements
With the introduction of NextGen WP Events, and the work that is being done to update the organizer dashboard and tools, there are several budget tool updates that could make it more functional for organizers and Central.

 - [x] Under Event Data > Days (please create a drop down for the following)

`Days`

`Hours`

- [x] Instead of Event Data > Tracks (Lets use format options as drop downs instead)

- WP Expertise Level (beginner, intermediate, advanced)
- Focused Activity (workshops, networking, contributing)
- Job Status (students, job seekers, business owners, freelancers)
- Identity Based (women, BIPOC, LGBTQ+, age, accessibility needs)
- Content Topic Focused (hosting, block development, designers)
- Other

- [x] Under Event Data > Speakers, please update `Speakers` to `Facilitators`
- [x] Under Expenses > Can we remove `Speaker Event`, `After Party`, and `Camera Shipping` as expense categories

outdoor2kode:
See #976

This PR replaced, removed, and updated certain fields as requested in the ticket.
The modifications for `Days` and `Hours` will be in a separate PR, as they will involve more extensive changes to the logic than expected and this PR already has enough changes.

### Screenshots

| Central Network | Event Network |
| -----------------|----------------|
| image | image image |
| image | image |
| image | image |

  

1. Create a new (a) og WordCamp site (b) next-gen event site through `WordCamps` >> `Add new`
2. In the dashboard of both sites, go to Budgets.
3. Check if the content is the same as the screenshots above demonstrate as well as requested in the ticket.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/bootstrap.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/bootstrap.php
@@ -23,6 +23,15 @@ if ( is_admin() ) {
 
 	$GLOBALS['wordcamp_budgets']    = new WordCamp_Budgets();
 	$GLOBALS['wcp_payment_request'] = new WCP_Payment_Request();
+
+	/**
+	 * Determines if the current site is a "next generation" WordCamp.
+	 *
+	 * @return bool Returns true if the current site is a "next generation" WordCamp, otherwise false.
+	 */
+	function is_next_gen_wordcamp() {
+		return SITE_ID_CURRENT_SITE === EVENTS_NETWORK_ID;
+	}
 }
 
 if ( is_admin() || wp_doing_cron() || wp_doing_ajax() ) {
--- a/public_html/wp-content/plugins/wordcamp-payments/bootstrap.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/bootstrap.php
@@ -23,15 +23,6 @@ if ( is_admin() ) {
 
 	$GLOBALS['wordcamp_budgets']    = new WordCamp_Budgets();
 	$GLOBALS['wcp_payment_request'] = new WCP_Payment_Request();
-
-	/**
-	 * Determines if the current site is a "next generation" WordCamp.
-	 *
-	 * @return bool Returns true if the current site is a "next generation" WordCamp, otherwise false.
-	 */
-	function is_next_gen_wordcamp() {
-		return SITE_ID_CURRENT_SITE === EVENTS_NETWORK_ID;
-	}
 }
 
 if ( is_admin() || wp_doing_cron() || wp_doing_ajax() ) {
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
@@ -298,12 +298,12 @@ class WordCamp_Budget_Tool {
 		);
 
 		$extra_budget_for_next_gen = array(
-			array( 'type' => 'meta', 'name' => 'wp-expertise-level', 'value' => 0 ),
-			array( 'type' => 'meta', 'name' => 'focused-activity', 'value' => 0 ),
-			array( 'type' => 'meta', 'name' => 'job-status', 'value' => 0 ),
-			array( 'type' => 'meta', 'name' => 'identity-based', 'value' => 0 ),
-			array( 'type' => 'meta', 'name' => 'content-topic-focused', 'value' => 0 ),
-			array( 'type' => 'meta', 'name' => 'other', 'value' => 0 ),
+			array( 'type' => 'meta', 'name' => 'wp-expertise-level', 'value' => '' ),
+			array( 'type' => 'meta', 'name' => 'focused-activity', 'value' => '' ),
+			array( 'type' => 'meta', 'name' => 'job-status', 'value' => '' ),
+			array( 'type' => 'meta', 'name' => 'identity-based', 'value' => '' ),
+			array( 'type' => 'meta', 'name' => 'content-topic-focused', 'value' => '' ),
+			array( 'type' => 'meta', 'name' => 'other', 'value' => '-' ),
 		);
 
 		if ( is_next_gen_wordcamp() ) {
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
@@ -267,7 +267,7 @@ class WordCamp_Budget_Tool {
 	 */
 	private static function _get_default_budget() {
 		// phpcs:disable WordPress.Arrays.ArrayDeclarationSpacing.AssociativeArrayFound
-		return array(
+		$default_budget = array(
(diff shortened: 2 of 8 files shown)


public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
```diff
@@ -191,6 +191,14 @@ public static function enqueue_scripts() {
 				3,
 				true
 			);
+
+			wp_localize_script(
+				'wcb-budget-tool',
+				'networkStatus',
+				array(
+					'isNextGenWordCamp' => is_next_gen_wordcamp(),
```
tellyworth: How certain are we that there will only ever be two types of events (nextgen and OG)?

If it's likely there could be a third type in future then perhaps something other than a binary is/not distinction is worthwhile.

public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
```diff
@@ -288,6 +319,25 @@ private static function _get_default_budget() {
...
+			array( 'type' => 'meta', 'name' => 'job-status', 'value' => '' ),
+			array( 'type' => 'meta', 'name' => 'identity-based', 'value' => '' ),
+			array( 'type' => 'meta', 'name' => 'content-topic-focused', 'value' => '' ),
+			array( 'type' => 'meta', 'name' => 'other', 'value' => '-' ),
+		);
+
+		if ( is_next_gen_wordcamp() ) {
+			// $extra_budget_for_next_gen should be positioned after the 'days' element for proper ordering.
+			$insert_position = array_search( 'days', array_column( $default_budget, 'name' ), true ) + 1;
+			array_splice( $default_budget, $insert_position, 0, $extra_budget_for_next_gen );
```
tellyworth, line 307: `array_splice()` with an offset of zero is a clever way of extending an array. I typically use a construction like `$default = array_merge( $default, $extra)`. I'd be curious to check out the performance differences; certainly array_splice could be more memory efficient.

public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
```diff
@@ -254,12 +262,35 @@ private static function _get_budget() {
 		return $budget;
 	}
 
+	/**
+	 * Helper function to filter out the budget items.
+	 */
```
tellyworth, line 267: I'm a bit puzzled by the need for this. Why use dynamic filtering rather than just leave out the unneeded items from the declaration in `_get_default_budget()`? I'm sure there's a reason but it's not obvious.

tellyworth (approved):
No blockers here, and it looks fine when I tested per the test steps.

public_html/wp-content/plugins/wordcamp-payments/bootstrap.php
```diff
@@ -23,6 +23,15 @@
 
 	$GLOBALS['wordcamp_budgets']    = new WordCamp_Budgets();
 	$GLOBALS['wcp_payment_request'] = new WCP_Payment_Request();
+
+	/**
+	 * Determines if the current site is a "next generation" WordCamp.
+	 *
+	 * @return bool Returns true if the current site is a "next generation" WordCamp, otherwise false.
+	 */
+	function is_next_gen_wordcamp() {
```
StevenDufresne: As noted somewhere else in a different comment. Let's move this somewhere else. 

I think:https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/mu-plugins/3-helpers-misc.php is a good candidate.

public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
StevenDufresne: `array_merge` is definitely more readable.

public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
outdoor2kode: Good question. 🤔  I'm not entirely certain, but adapting the function to accept a different type of event also sounds legit and good now.

public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
outdoor2kode: Actually, the offset is `$insert_position` here, not zero. 
In this case, `$extra_budget_for_next_gen` needs to be inserted immediately after `days`. If using `$default = array_merge($default, $extra);`, $extra would be appended to the end of the $default array.
I may have missed something or misunderstood your point, it'd be nice if you can correct me if that's the case, thanks.

public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
outdoor2kode: Because for og wordcamps, the items are still needed. If leaving out those items directly from the declaration, on og wordcamp sites, they won't show up.

---

Also, since the `_get_default_budget` is getting bigger, it seems to me that it's not easy to read and keep up with. And I'm guessing we'll want to pull out more items from the event network down the line, so I abstract this logic into a separate function. This would facilitate the addition or removal of items as needed. If we were to directly integrate this logic into `_get_default_budget`, I feel it would make the function a bit lengthy, especially as more changes are added down the road."


If we don't add an extra filtering function, we'd just be incorporating one of the following into `_get_default_budget`. It looks okay for now, but I'm thinking it might get a bit lengthy if we add more conditions down the line. What do you think?
```
$default_budget = array_filter($default_budget, function($item) {
    return !($item['name'] === 'tracks' || (isset($item['category']) && $item['category'] === 'speaker-event'));
});
```

```
$default_budget = array_filter($default_budget, function($item) {
    if (isset($item['name']) && $item['name'] === 'tracks') {
        return false;
    }
    
    if (isset($item['category']) && $item['category'] === 'speaker-event') {
        return false;
    }

    return true;
});
```

outdoor2kode: ~Or how do you feel about splitting the default budget retrieval for the next-gen network, like what's done in this [commit](https://github.com/WordPress/wordcamp.org/pull/998/commits/7c5dcc6a7687799f953318b15d0ebc68de380bbd)?~

the function was removed as the original requirements have changed.

public_html/wp-content/plugins/wordcamp-payments/bootstrap.php
outdoor2kode: Updated in d870b2f050b8639fb256e27197a0f2d09506207f

public_html/wp-content/plugins/wordcamp-payments/includes/budget-tool.php
outdoor2kode: Updated in 294127b124dc48fcdf593169ca0da166922206d5

PR #1017, opened 2023-08-24
Add docker override file name to .gitignore

miminari:
We need to override the docker file to build a local environment with docker on m1 Mac, so I add that filename to gitignore.
And I also add .DS_Store files are created by the Mac OS X operating system.

Props @ryelle 

1. Checkout new branch.
2. Add a docker-compose.override.yaml and a .DS_Store file.
3. Git add, git status, you can check to ignore these files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .docker/database
 .docker/wordcamp*.pem
 .docker/test_suite
+docker-compose.override.yaml
 
 
 #
@@ -47,7 +48,7 @@ public_html/wp-content/plugins/wc-post-types/package-lock.json
 #
 public_html/wp-content/plugins/query-monitor/
 public_html/wp-content/plugins/user-switching/
-docker-compose.override.yaml
+
 
 #
 # Plugins
--- a/.gitignore
+++ b/.gitignore
@@ -109,9 +109,3 @@ public_html/wp-content/themes/twentytwentyone
 public_html/wp-content/themes/twentytwentytwo
 public_html/wp-content/themes/twentytwentythree
 public_html/wp-content/themes/wporg-parent-2021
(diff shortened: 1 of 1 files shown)


ryelle:
@miminari Thanks again for the PR! I've pushed two commits to this branch. One to move the docker override file to the "Docker" section. The other change removes the `.DS_Store` from the gitignore. In the comments to this PR https://github.com/WordPress/wordcamp.org/pull/139 it was decided that system files like that should be ignored globally, not in the project configuration.

miminari:
@ryelle Thank you for commit and merge!

2ndkauboy:
Haha, great that this was added. In a different issue I experienced today at WCUS Contributor Day, I also had to add such a file :)

@miminari what overwrite is necessary for an M1? I was about to add an `example.docker-compose.override.yml` and might add your customization into it as well.

miminari:
@2ndkauboy Hi, I just added same codes written on this docs.

https://github.com/WordPress/wordcamp.org/blob/production/.docker/readme.md

'''
Using an Apple ARM64 (Silicon) chip? You may see failed to solve: rpc error: code = Unknown desc =... or failed to solve: mysql:5.7-debian: no match for platform in manifest after running either of commands above. Try adding platform: linux/amd64 to both wordcamp.test and wordcamp.db in docker-compose.yaml. This will instruct Docker to create an image based on the linux/amd64 architecture instead of linux/arm64, i.e.
wordcamp.test:
  build:
    context: .docker
    dockerfile: Dockerfile.php-fpm
  platform: linux/amd64

wordcamp.db:
  build:
    context: .docker
    dockerfile: Dockerfile.mysql
  platform: linux/amd64
'''

PR #1018, opened 2023-08-24
Composer: Update `composer.lock` fixing parent theme install issue

### LINKED ISSUE #1016: Required package "wporg/wporg-parent-2021" is not present in the lock file
When I'm following the README to install the composer dependencies, I get the mentioned error.

**To reproduce**

Follow the steps 1-4 of the README, then run the `composer install` in step 5:

```bash
$ composer install
Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
- Required package "wporg/wporg-parent-2021" is not present in the lock file.
This usually happens when composer files are incorrectly merged or the composer.json file is manually edited.
Read more about correctly resolving merge conflicts https://getcomposer.org/doc/articles/resolving-merge-conflicts.md
and prefer using the "require" command over editing the composer.json file directly https://getcomposer.org/doc/03-cli.md#require-r
```

**Expected behavior**

Composer should install all dependencies.

 - OS: Manjaro Linux 23.0.0
 - Composer: 2.5.8

2ndkauboy:
I've updated the `composer.lock` by removing it and running `composer update`. It seems the issue was, that in 29c821b55b5f077267c59b7ed61c25b5dcc2801b the package was manually moved from `require-dev` to `require`, but in the `composer.lock` it was not moved from `packages-dev` to `packages` as well.

Fixes #1016

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "a5015aea9e9c14085f9efc84446b689b",
+    "content-hash": "29291fe29de76306f51d13fde11ef2cd",
     "packages": [
         {
             "name": "adhocore/jwt",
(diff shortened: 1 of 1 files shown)


iandunn:
Ah, you're right, thanks! I updated it in f7049a5c5dbd35828591cb9f52ce331afa68429d before noticing this PR.

PR #1021, opened 2023-08-28
Only show pay by check/ACH when the US is selected

### LINKED ISSUE #977: WordPress Event/WordCamp Vendor Payment dashboard & Reimbursement dashboard Enhancements
A few additional updates were shared that I think make sense and would not be super complicated additions:

- [x] For payment types, either only showing `pay by check` or `pay by ACH` when supported country (USA) is selected, or just add a note to indicate that those payment types are only available in the USA.

outdoor2kode:
Closes #977

This PR hides the `Direct Deposit` and `Check` payment methods when the selected country is NOT the US.

### Screencasts

https://github.com/WordPress/wordcamp.org/assets/18050944/a445c785-9efb-4c24-a266-abcc8cbcd53b

1. Go to Budget Tool > Vendor Payment and Reimbursement Request > Add New.
2. Under `Payment Details / Payment Information`, select your `Country for payment receipt` as either `United States` or `non-US countries`.
3. You should see it behave as requested or as video above.
4. Save changes, and you should see the data is correctly saved.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
+++ b/public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
@@ -48,18 +48,16 @@ jQuery( document ).ready( function( $ ) {
 				$notes.find('textarea').attr('required', state);
 			}).trigger('change');
 
-			var initialValue = $('#bank_country_iso3166').val();
-			if (initialValue !== 'US') {
-				$('#payment_method_direct_deposit_container, #payment_method_check_container').hide();
-			}
 			$('#bank_country_iso3166').on('select2:select', function(e) {
-				var selectedValue = $(this).val();
+				const selectedValue = $(this).val();
 			
 				if (selectedValue === 'US') {
 					$('#payment_method_direct_deposit_container, #payment_method_check_container').show();
 				} else {
 					$('#payment_method_direct_deposit_container, #payment_method_check_container').hide();
 				}
+
+				$('#row-payment-method').show();
 			});
 		},
 
--- a/public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
+++ b/public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
@@ -46,7 +46,7 @@ jQuery( document ).ready( function( $ ) {
 				$notes.find('textarea').attr('required', state);
 			}).trigger('change');
 
-			$('#payment_receipt_country_iso3166').on('select2:select', function(e) {
+			$('#payment_receipt_country_iso3166').on('change', function(e) {
 				const selectedValue = $(this).val();
 			
 				if (selectedValue === 'US') {
@@ -56,7 +56,7 @@ jQuery( document ).ready( function( $ ) {
 				}
 
 				$('#row-payment-method').show();
-			}).trigger('select2:select');
+			}).trigger('change');
 		},
 
 		/**
--- a/public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
+++ b/public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 jQuery( document ).ready( function( $ ) {
 	'use strict';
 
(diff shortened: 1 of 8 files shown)


public_html/wp-content/plugins/wordcamp-payments/views/payment-request/input-radio.php
```diff
@@ -1,4 +1,4 @@
-<tr id="row-<?php echo esc_attr( str_replace( '_', '-', $name ) ); ?>">
+<tr id="row-<?php echo esc_attr( str_replace( '_', '-', $name ) ); ?>" class="<?php echo true === $is_visible ? 'active' : 'hidden'; ?>">
```
iandunn, line 1: I think it'd be good to use something like `wcb__is-active`, to avoid conflicts with other CSS. 

That was BEMi-sh, but if there's already CSS in the plugin that uses a different convention, then it'd be good to be consistent with that.

public_html/wp-content/plugins/wordcamp-payments/javascript/wordcamp-budgets.js
```diff
@@ -149,32 +150,6 @@ jQuery( document ).ready( function( $ ) {
...
 		},
 
-		/**
-		 * Set the default payment method based on the currency
-		 *
-		 * Don't override any existing payment method choices.
-		 *
-		 * @param {object} event
-		 */
-		setDefaultPaymentMethod : function ( event ) {
```
iandunn, line 159: What was the reason for removing this?

public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
```diff
@@ -47,6 +45,18 @@ jQuery( document ).ready( function( $ ) {
 				$notes.toggle(state);
 				$notes.find('textarea').attr('required', state);
 			}).trigger('change');
+
+			$('#payment_receipt_country_iso3166').on('select2:select', function(e) {
```
iandunn: Is it possible to use a standard event like `change`? That would avoid the need to update this code if we ever disable `select2` or switch to a different dropdown. IIRC select2 will still fire the native event.

public_html/wp-content/plugins/wordcamp-payments/views/payment-request/input-radio.php
outdoor2kode: Yeah, I've noticed that `active` and `hidden` are commonly used in several plugins. 

For example: https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/wordcamp-payments/views/payment-request/metabox-payment.php#L31

public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
renintw: Ah, yeah, you're right. `change` is supposed to be used here.

public_html/wp-content/plugins/wordcamp-payments/javascript/wordcamp-budgets.js
outdoor2kode: Sorry I should've added more details for this change. A couple of reasons to remove this.

1. It kind of didn't work as expected, from what I've observed, it would always default to `Wire` and doesn't change when switching to a different currency. I guess it's due to when the callback is triggered for the first time, 'USD' doesn't equal `newCurrency`, so the `Wire` is selected. And later when a user selects an option, triggering callback again, the function would return as `undefined !== selectedPaymentMethod` would always be true.
2. Currently, once the `Country for payment receipt` is selected, the system lists all payment methods based on whether the chosen country is the US. After this, users pick a payment method, and only then does the relevant form for that method become visible. 
So even though the `setDefaultPaymentMethod` might work initially, it ends up auto-selecting either `Wire` or `Direct Deposit` for the user based on the currency they choose, but the desired outcome is to let users choose between `Deposit` or `Check` only when the payment receipt country is set to the US, rather than basing it on the currency.

I've added the above details to the commit message.

public_html/wp-content/plugins/wordcamp-payments/javascript/payment-requests.js
renintw: Ah, yeah, you're right. `change` is supposed to be used here. Thank you!

outdoor2kode: Ah, yeah, you're right. `change` is supposed to be used here. Thanks!

Issue #1019, opened 2023-08-28
Debug `'get_font_stylesheet_url'` fatal


iandunn:
There are sporadic `get_font_stylesheet_url()` fatals in the error log:

`Call to undefined function WordPressdotorg\MU_Plugins\Global_Fonts\get_font_stylesheet_url() in wp-content/themes/wporg-parent-2021/functions.php:46`

I can't reproduce it, but it's been happening regularly since I added the parent theme in #1013 . It mostly appears to happen with pentesters, but I imagine it'll happen with others once the Events root site launches.

Because it's a fatal, we're not seeing the backtrace etc, but adding some extra logging may help.

iandunn:
In addition to #1020, I added some temporary logging to the parent theme in `r4537-wordcamp`

pkevan:
I managed to reproduce the error, but only through a `curl` request.

The backtrace is exactly as you see it through the logging.

```
[07-Sep-2023 08:21:07 UTC] PHP Fatal error:  Uncaught Error: Call to undefined function WordPressdotorg\MU_Plugins\Global_Fonts\get_font_stylesheet_url() in []wp-content/themes/wporg-parent-2021/functions.php:46
Stack trace:
#0 [~]wp-includes/class-wp-hook.php(308): WordPressdotorg\Theme\Parent_2021\theme_support('')
#1 /[~]wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters(NULL, Array)
#2 [~]wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#3 [~]wp-settings.php(601): do_action('after_setup_the...')
#4 [~]wp-config.php(129): require_once('/home/wordcamp/...')
#5 [~]wp-load.php(55): require_once('/home/wordcamp/...')
#6 [~]wp-blog-header.php(13): require_once('/home/wordcamp/...')
#7 [~]index.php(17): require('[~]...')
#8 {main}
  thrown in [~]wp-content/themes/wporg-parent-2021/functions.php on line 46

```

pkevan:
Seems like it only happens when the subdomain is capitalised e.g. `Events.wordcamp` rather than `events.wordcamp`.

pkevan:
I was able to work around it (I think!) by adding the capitalised subdomain to the wp-config file, but unsure if that's the right thing to do here.

iandunn:
Ah, yeah, that's a great find, thanks!

[...]

PR #1033, opened 2023-09-08
Docker: Convert hostname to lowercase to fix fatal error

iandunn:
See #1019 

Otherwise pentesters using `curl` with a hostname like `Events.wordpress.org` (not the uppercase) will clutter the logs with fatal errors because the Global Fonts mu-plugin isn't loaded. Regular browsers automatically convert the domain to lowercase before sending the request.

The production config will need to be changed in SVN.


iandunn:
Production updated in `r4557-wordcamp`

PR #1038, opened 2023-09-13
Add organizer debrief survey

### LINKED ISSUE #930: New event debrief survey for organizers
Hello team! As a part of the New event infrastructure, we'll also need a new debrief form/survey for organziers to complete once the event is concluded. It would be similar to the WordCamp debrief, and organizers should receive an automated reminder about it.

Below you can find the text of the survey. If needed, [you can find here the Google form](https://forms.gle/ERwS5bjH81yTMGkB9) that we'll be using until the debrief on central is created. Questions with the symbol * are mandatory questions.

Please let me know if you need any more info on my part, thanks a lot!

·····································
[Title] Organizer Survey (event debrief)

If you recently organized a WordPress event, please take the survey below to share your feedback about your experience.

Start Survey

What event did you organize? *

General Information

How many tickets were sold? *

How many people actually attended? *

How many people attended your event as Attendees? *

How many people attended your event as Organizers? *

How many people attended your event as Volunteers? *

How many people attended your event as Organizers? *

How many people attended your event as Speakers/Trainers/Facilitators? *

Did you livestream your event? *
Nope!
Yep!

If you live streamed your event, what service did you use, and how many viewers did you have?

Have the videos and slides been uploaded for wordpress.tv yet? *
Yes, they have all been uploaded
Some, but not all, we are still working on it
We have not started uploading videos and slides yet

If they haven't been uploaded yet, is there something specific holding up the process?

How was the Code of Conduct publicized at your event? (check all that apply)*
It was on the website, linked from main navigation
It was on the website, linked from other pages
It was mentioned in pre-event emails
It was mentioned in opening remarks
It was printed on individual collateral (like name badges, programs, etc)
It was printed on signs
Other:

Were there any reports of Code of Conduct violations at your event? *

Budget

How did you wind up in terms of budget? *
There is a surplus
We broke even
We lost money

If there was a budget surplus/deficit, what was the amount?

Have all of your vendors been paid? *
Yes, all of them
No, we still have more payments that we need to request

If there are outstanding payments, what's the holdup?

Which expenses were unanticipated, or cost more than you expected? *

Have payments been received from all your sponsors? *
Yes, all of them
No, but we have invoiced them all
No, we still have to invoice some late-addition sponsors

If you are still waiting for any sponsor payments, which sponsors are they?

Did you receive in-kind donations or contributions? If yes, what are they?*

Opinions

What do you think went really well at your event? *

What could have gone better at your event? * (This could include logistical things like printing name badges late, bigger things like overall schedule planning, or anything else that you've learned from this year.)

Would you do the event again?*
Yes
No

If not, why?

What were the goals that you wanted to reach by organizing this event?*

Have you reached those goals?*
Yes
No

If not, what blocked you from reaching them - What can be improved for the next time?

After completing the event, do you consider your event to be (check all that apply): *
Doable (less human and financial resources)
Replicable (by other communities in different parts of the world)
Scalable (can be changed in size or scale)
Desirable (seen as useful, important or necessary by the community)

Did you have enough support from WordCamp Central?*
Yes
No

If there were things that made planning this event harder or easier, let us know!

Have you written a recap post of the event yet? If so, please provide the URL so we can reblog it on the WordCamp Central website. *

Is there anything else we should know about how your event went?

Finish Survey

[Thank you message]: Thank you so much for sharing your feedback and providing an excellent experience to your community, we are very grateful for your contribution to the WordPress open source project!

outdoor2kode:
Closes #930

This PR implements what has been concluded in the ticket. The logic refers to #997.

**Flow**

- Manually published Organizer Survey (event debrief) in Central.

- When the plugin is activated:
  - Create a `tix_email` in draft status.
    -  The email will contain a token generated by hashing the access token key and WordCamp ID, to prevent the survey page from being accessed and abused by individuals other than the lead organizers.
  - Schedule a `daily` cron job.

- When the cron job runs:
  - Check if we already sent an email for this camp.
  - Check to make sure the camp ended exactly 2 days ago.
  - Add the lead organizer as an attendee and associate it with `tix_email`.
  - Add `tix_email` to the queue (set to `pending`)
    - _email gets picked up by camptix cron email sending queue `tix_scheduled_every_ten_minutes`_

## Screenshots

| View | Screenshot |
| --- | --- |
| Email queue | image |
| Manually published Organizer Survey (event debrief) page in central | image |
| Draft Survey Email | image |
| Survey Email Sent | image |
| Survey | image |

## How to test

1. Create a Next-Gen Event WordCamp site and fill in `Organizer Name` and `Email Address` inside the Organizing Team.
   - Make the Start Date in the past.
   - Make the End Date the current date.
2. Visit Cenral
   - Go to Pages
     - Manually publish `Organizer Survey (event debrief)`
     - View the content, which should be the same as the description in ticket.
2. Visit Next-Gen
   - Go to Ticket -> E-Mail
     - Expect to see an email `Organizer Survey (event debrief)` in `Draft` mode.
     - The email content is pending an update from the Tardis Team. 
   - Run the following wp cli script to make sure the cron jobs are scheduled
     - `wp cron event list --url=https://events.wordpress.test/{city}/{year}/{type}` 
     - expect to see the following cron job: `wc_organizer_debrief_survey_email`
   - Go back to the Next-Gen Wordcamp Post type and update the End Date to exact 2 days ago.
   - Run this script to set WordCamp to closed `wp post update {post_id} --post_status=wcpt-closed --url=https://central.wordcamp.test/`
   - Run the following cli command:
     - run `wp cron event run wc_organizer_debrief_survey_email --url=https://events.wordpress.test/{city}/{year}/{type}` 
       - Should succeed
       - Status of `Organizer Survey (event debrief)` in Ticket -> E-Mail should be `pending`.
       - An attendee should be added in Ticket -> Attendees with `tix_first_name` and `tix_email` the same as the one you input when creating the event.
     - run `wp cron event run tix_scheduled_every_ten_minutes -url=https://events.wordpress.test/{city}/{year}/{type}`
       - `Organizer Survey (event debrief)` in Ticket -> E-Mail should be published.
       - In, http://localhost:1080/, there should be an email with Subject `Organizer Survey (event debrief)`. 
         - Visit the survey link on the front end and submit.
 4. Make sure none of this exists on the traditional WordCamp.


public_html/wp-content/plugins/wordcamp-organizer-survey/includes/admin-page.php
```diff
@@ -8,9 +8,8 @@
 defined( 'WPINC' ) || die();
```
renintw, line 8: This commit is reverted, can be skipped.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/admin-page.php
```diff
@@ -8,9 +8,8 @@
...
 
 use function WordCamp\OrganizerSurvey\{get_feature_id};
-// use function WordCamp\OrganizerSurvey\Email\{get_email_id};
+use function WordCamp\OrganizerSurvey\Reminder\{get_reminder_id};
```
outdoor2kode: This commit is reverted, can be skipped.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
```diff
@@ -0,0 +1,200 @@
...
+	}
+}
+
+/**
+ * Add temporary attendee since email can only be sent to items in attendee tracker.
+ */
+function add_temp_attendee() {
+	if ( ! get_wordcamp_attendees_id() ) {
+		$attendees_id = wp_insert_post( array(
+			'post_title' => 'Organizer debrief survey',
```
outdoor2kode, line 62: image
image


The title is `(no title)` even though the `post_title` is specified and updated to DB. And after moving it to the Trash, the Title appears. I'm not entirely sure why this happens, but it doesn't affect the overall process.

public_html/wp-content/plugins/wordcamp-organizer-survey/wordcamp-organizer-survey.php
```diff
@@ -0,0 +1,151 @@
...
+ * @return void
+ */
+function activate_on_current_site() {
+	// Debrief Survey.
+	add_debrief_survey_page();
+	add_debrief_survey_email();
+
+	// Flushing the rewrite rules is buggy in the context of `switch_to_blog`.
+	// The rules will automatically get recreated on the next request to the site.
+	delete_option( 'rewrite_rules' );
```
iandunn: Since this function is hooked to `init`, this will run on every page load. It's very expensive, so it's best to only do it during the activation hook.

Is it necessary to call it, though? I don't see any rewrite rules being added.

If it is needed, then `flush_rewrite_rules()` could be used directly, since this isn't being called during a `switch_to_blog()` process.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/admin-page.php
```diff
@@ -0,0 +1,145 @@
...
+	$wordcamps = get_posts( array(
+		'post_type' => WCPT_POST_TYPE_ID,
+		'post_status' => 'wcpt-closed',
+		'posts_per_page' => -1,
+	) );
+
+	$feedback_details = array();
+
+	foreach ( $wordcamps as $camp ) {
+		// It's a bit counter intuitive, but the site ID is the blog ID.
```
iandunn: This is a legacy inconsistency that stems from Core, where it was originally called a "blog", but then later called a "site". These days it's always called a site, but some of the old things like database schemas still say "blog" due to back-compat

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
```diff
@@ -0,0 +1,341 @@
...
+ * Create the Survey page, save ID into an option.
+ */
+function add_page() {
+	$page_id = get_option( SURVEY_PAGE_ID );
+
+	if ( $page_id ) {
+		return;
+	}
+
+	$page_id = wp_insert_post( array(
```
iandunn: 🤔 I wonder if it's better to create a single page on Central, rather than a new page on each camp site? There'll only be 1 respondent per camp, so it seems it might be overkill to do it on every site.

It could remove the need for `admin-page.php`, code to create/publish/delete/protect the page, etc.

I don't feel strongly, though. What do you think?

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
```diff
@@ -0,0 +1,200 @@
...
+ * Add cron jobs to the schedule.
+ */
+function schedule_jobs() {
+	if ( ! wp_next_scheduled( 'wc_organizer_debrief_survey_email' ) ) {
+		wp_schedule_event( time(), 'daily', 'wc_organizer_debrief_survey_email' );
+	}
+}
+
+/**
+ * Add temporary attendee since email can only be sent to items in attendee tracker.
```
iandunn, line 57: This feels like a bit of a hack. I wonder if there's a way to make the CampTix email more flexible, so that it doesn't require an attendee?

Or maybe this plugin shouldn't be so tightly coupled with CampTix?

I can see how it's a practical solution, though, so it may not be worth changing. What do you think?

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
```diff
@@ -0,0 +1,200 @@
...
+	$wordcamp_post = get_wordcamp_post( $blog_id );
+
+	if ( ! $wordcamp_post ) {
+		log( 'Couldn\'t retrieve wordcamp for blog id:', $email_id, $blog_id );
+		return false;
+	}
+
+	$end_date = $wordcamp_post->meta['End Date (YYYY-mm-dd)'][0];
+
+	if ( ! isset( $end_date ) ) {
```
iandunn: IIRC, some sites that are only 1-day events will only have the start date. If the end date is empty and the start date is filled in, then the start date can be used as the end date.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
```diff
@@ -0,0 +1,341 @@
...
+ */
+function get_page_content() {
+	return <<<EOT
+	<!-- wp:paragraph -->
+	<p>If you recently organized a WordPress event, please take the survey below to share your feedback about your experience.</p>
+	<!-- /wp:paragraph -->
+	
+	<!-- wp:jetpack/contact-form {"subject":"New feedback received from your website","style":{"spacing":{"padding":{"top":"16px","right":"16px","bottom":"16px","left":"16px"}}}} -->
+	<div class="wp-block-jetpack-contact-form" style="padding-top:16px;padding-right:16px;padding-bottom:16px;padding-left:16px"><!-- wp:group {"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group"><!-- wp:jetpack/field-text {"label":"What event did you organize?","required":true,"requiredText":"(required)"} /--></div>
```
iandunn: We shouldn't need to ask this, since we already know it based on the current site, who the email is sent to, etc. We can remove the question, but have some code that automatically adds it to the response before it's saved in the database.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
```diff
@@ -0,0 +1,341 @@
...
+	<!-- /wp:group -->
+	
+	<!-- wp:group {"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group"><!-- wp:heading -->
+	<h2 class="wp-block-heading"><strong><strong><strong>Part I. <strong>General Information</strong></strong></strong></strong></h2>
+	<!-- /wp:heading -->
+	
+	<!-- wp:group {"layout":{"type":"constrained"}} -->
+	<div class="wp-block-group"><!-- wp:jetpack/field-text {"label":"How many tickets were sold?","required":true,"requiredText":"(required)"} /--></div>
```
iandunn: This can also be pulled from existing data. There may be other questions that can also be removed for that reason.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
```diff
@@ -0,0 +1,200 @@
...
+		return;
+	}
+
+	// Associate attendee to tix_email as a recipient.
+	$result = add_post_meta( $email_id, 'tix_email_recipient_id', $recipient_id );
+
+	if ( ! $result ) {
+		log( 'Failed to add recipients:', $email_id, $recipient_id );
+	} else {
+		update_post_meta( $email_id, 'tix_email_recipients_backup', (array) $recipient_id );
```
iandunn, line 98: Did you notice this happening, or is it more to protect against a theoretical situation?

iandunn (approved):
I left a few suggestions but this looks good overall

public_html/wp-content/plugins/wordcamp-organizer-survey/wordcamp-organizer-survey.php
outdoor2kode: Ah, yeah. You're right. This should be removed. Previously, when I tried to add the Organizer Reminder, I used `switch_to_blog`, so it's been there ever since.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
outdoor2kode: > I wonder if it's better to create a single page on Central, rather than a new page on each camp site? There'll only be 1 respondent per camp, so it seems it might be overkill to do it on every site.

If we only create a page on the central site, then we can't have the mechanism to set this page as a draft to prevent spam. 

We currently have an anti-spam mechanism where, two weeks after the survey page link sent to organizer, this published page is automatically deleted to prevent organizers from accessing it. 

If we were to create a page on the central site, the timing of changing the page status from draft to publish (add the page to camptix email) triggered by one site would overlap with the anti-spam mechanism's change from publish to draft triggered by another site. 

That said, given that only one person, the lead organizer, is filling out the survey for each event, the chances of spam, while still possible, are considerably reduced.

Perhaps in this scenario, we don't need to worry about preventing spam?

> It could remove the need for admin-page.php, code to create/publish/delete/protect the page, etc.

I understand we can remove the need for code to create/publish/delete/protect the page if moving the page to central, but I'm a bit not sure how this can also affect the need for admin-page.php.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
outdoor2kode: > Or maybe this plugin shouldn't be so tightly coupled with CampTix?

When I was working on this PR previously, I had intended to use the Organizer Reminder to address it. However, my initial approach was in the totally wrong direction, leading me to spend an excessive amount of time on it. That's why I eventually resorted to this hacky but practical method.

Today I spent another day trying to see if the Organizer Reminder could potentially resolve this issue. I believe it's feasible, but the main hurdle is the spam concern mentioned above. If we can bypass the need to set the page to draft for spam prevention, then based on how Organizer Reminder operates, the solution would be: setting up an Automated Job with the email content prepared, including the link to the published survey page. It would then scan the WordCamp posts on a daily basis to identify any WordCamp events that meet the sending-reminder criteria, fetch their data, include it in the context of the automated job, and send the email.

What do you think of this approach?

outdoor2kode: > I can see how it's a practical solution, though, so it may not be worth changing.

If we proceed with merging the current PR, as I've made some preparations here to generalize the plugin, we can open another PR to integrate the Attendee Survey, integrating it into a single Survey Plugin.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
outdoor2kode: Good point, thanks!

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
outdoor2kode: Yes, I noticed there's a warning
```
Warning: count(): Parameter must be an array or an object that implements Countable in /usr/src/public_html/wp-content/plugins/camptix/camptix.php on line 808
```
image


---

After investigating the issue, I discovered that this line was returning an empty string for recipients_backup
https://github.com/WordPress/wordcamp.org/blob/ea44ddd8008b565b3b32da99be3149aacad12d38/public_html/wp-content/plugins/camptix/camptix.php#L807
This would make it counted as `1`. I've pushed ea44ddd8008b565b3b32da99be3149aacad12d38 and 0510637a93f98ad6950eab47cb975c737b4d0426 to fix it.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
outdoor2kode: It seems there's no proper filter or approach to add the predefined data before inserting the response to Feedback for [Jetpack contact form](https://github.com/codeforamerica/cityhow/blob/master/wp-content/plugins/jetpack/modules/contact-form/grunion-contact-form.php#L1038). What's the common way to do so?

Also tried `grunion_pre_message_sent` and `wp_insert_post_data`, but I feel they aren't the best solution here.

Maybe using JavaScript to set the default values? I feel like there's a more elegant way to do so.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
iandunn: > then we can't have the mechanism to set this page as a draft to prevent spam.

🤔 That's a good point. We could manually create the page on Central, and leave it published all the time. We then might be able to require the organizer to log in to submit an entry, and we could automatically populate the "which wordcamp..." field based on that. We could show them an error if their camp was over more than 2 weeks ago, or maybe we wouldn't have to worry about that because we know only the organizer can fill it in.

iandunn: >  how this can also affect the need for admin-page.php.

I was thinking that it wouldn't be needed, since admins could use a workflow like this:

1. Visit https://central.wordcamp.org/wp-admin/edit.php?post_type=feedback
2. change `All sources` to the specific form
3. click `Filter` or `Export`

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
iandunn: That approach with Organizer Reminders sounds good 👍🏻 I also think it's reasonable to stick with this approach if you want. Organizer Reminders may be more ideal, but it also may take more time, so it's a trade off. If we'll want to refactor this later, though, then that also takes time 🤷🏻‍♂️ 

I don't have a strong opinion either way, but I trust your judgement 👍🏻

StevenDufresne: Let's prioritize merging this over architecture changes. This feature hasn't proved its worth yet.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
iandunn: 🤔 Rather than requiring them to log in, we could also just put a token in the URL that we send them, and use that to verify that they're an organizer. That might avoid some folks skipping it because of the extra work of logging in.

It may also avoid some edge cases where there's extra hassles. For example, we store the email address of the lead organizer, but not their w.org username. So if they log in but their w.org account email doesn't match the email we sent the request to, then they might have to update the email to match.

For the token, we might be able to just create a secret key in the config, and then create a hash of their email address and the key; something like `hash_hmac( 'sha1', $email, ORGANIZER_SURVEY_ACCESS_TOKEN_KEY )`. That way we wouldn't have to store any additional metadata in the database. If having the value in the db is needed for other parts of the code, though, then that's not a problem either. In that case we could probably just use `wp_generate_password()` to create the token.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
outdoor2kode: Thanks for providing extra insights on the path moving forward @iandunn and @StevenDufresne, I'll proceed with the current approach.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
outdoor2kode: Thanks for the suggestion 👍 👍 👍 
I've removed the admin page and moved the survey page to central.

And after talking a bit with Steve earlier today, we figure it seems we can get more things on the central site if including the `wordcamp_id` in the URL - capable of answering [as many questions as possible](https://github.com/WordPress/wordcamp.org/pull/1038#discussion_r1329418934) with the event site data retrieved from it. 
So it would be `hash_hmac( 'sha1', $wordcamp_id, ORGANIZER_SURVEY_ACCESS_TOKEN_KEY )`.

The changes starts from this commit [7354b3a](https://github.com/WordPress/wordcamp.org/pull/1038/commits/7354b3a7d7e13dc43a3a2b7f65cf34810e05ef21)

.docker/wp-config.php
```diff
@@ -118,6 +118,8 @@
 define( 'EMAIL_DEVELOPER_NOTIFICATIONS', 'developers@example.test' );
 define( 'EMAIL_CENTRAL_SUPPORT',         'support@wordcamp.test' );
 
+define( 'ORGANIZER_SURVEY_ACCESS_TOKEN_KEY', 'wcorg_debrief_survey_email' );
```
iandunn:
```suggestion
define( 'ORGANIZER_SURVEY_ACCESS_TOKEN_KEY', 'put your unique phrase here' );
```

This might make it more obvious that the value is essentially a password, so it should be a long randomly generated  value, just like the key/salt values above.

public_html/wp-content/mu-plugins/wcorg-network-plugin-control.php
```diff
@@ -94,6 +94,7 @@ function _get_network_plugin_state_list( $state ) {
 	} elseif ( EVENTS_NETWORK_ID === $network_id ) {
 		$network_plugin_state['deactivated'][] = 'tagregator/bootstrap.php';
 		$network_plugin_state['activated'][]   = 'camptix-attendee-survey/camptix-attendee-survey.php';
+		// $network_plugin_state['activated'][]   = 'wordcamp-organizer-survey/wordcamp-organizer-survey.php';
```
iandunn: Should this be commented out?

public_html/wp-content/plugins/wordcamp-organizer-survey/wordcamp-organizer-survey.php
```diff
@@ -0,0 +1,178 @@
...
+function validate_token_on_debrief_survey() {
+	global $wordcamp_post_data;
+
+	if ( is_page( 'organizer-survey-event-debrief' ) ) {
+		$token       = $_GET['t'] ?? '';
+		$wordcamp_id = $_GET['wid'] ?? '';
+
+		$expected_token = hash_hmac( 'sha1', base64_decode( $wordcamp_id ), ORGANIZER_SURVEY_ACCESS_TOKEN_KEY );
+
+		if ( $token !== $expected_token ) {
```
iandunn:
```suggestion
		if ( $token !== $expected_token ) {
		if ( ! hash_equals( $expected_token, $token ) {
```

Using `hash_equals()` avoids [timing attacks](https://security.stackexchange.com/a/83670/8467). It's not a big deal in this scenario, but it's a good habit to get into, and might get flagged by a linter in the future if we don't.

public_html/wp-content/plugins/wordcamp-organizer-survey/wordcamp-organizer-survey.php
```diff
@@ -0,0 +1,178 @@
...
+	global $wordcamp_post_data;
+
+	if ( is_page( 'organizer-survey-event-debrief' ) ) {
+		$dom = new \DOMDocument();
+		$dom->loadHTML($content, LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD);
+
+		$labels = $dom->getElementsByTagName('label');
+		foreach ( $labels as $label ) {
+			// phpcs:disable WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+			if ( 'What event did you organize?' === $label->nodeValue ) {
```
iandunn, line 169: 🤔 I suspect there may an easier way to do this. Do we need to show the question to the user if we're just going to fill it in anyway? 

Another problem is that, If it's in the page, it could be edited by an admin who wouldn't realize that doing that would break this.

If we don't need to show the question, we could insert the value in `$_POST` some time before `Grunion_Contact_Form_Plugin::process_form_submission()` runs.

If you do want to show it, it might be possible to add an `id/class` value to the post content in the database, and then use that in this function. That'd still break if someone edited it, but they'd be less likely to.

public_html/wp-content/plugins/wordcamp-organizer-survey/wordcamp-organizer-survey.php
```diff
@@ -0,0 +1,178 @@
...
+				$input_id         = $label->getAttribute('for');
+				$event_name_field = $dom->getElementById($input_id);
+				$event_name_field->setAttribute('value', $wordcamp_post_data->post_title);
```
iandunn, line 172: I think it'd be best to save the ID rather than the title, since the title can change. It also won't have the year, unless passed through `get_wordcamp_name()`. That function might not work well for NextGen camps though, since some of them are monthly, etc.

If we don't remove the question from the user-facing form (see above), then it would be good to show the user the title, but we could still save the ID. Maybe just by modifying `$_POST` before the value gets saved. Or it could be in a separate field if that's easier.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/cron.php
```diff
@@ -0,0 +1,183 @@
...
+	$email_status = queue_survey_email( $email_id );
+
+	if ( is_wp_error( $email_status ) ) {
+		log( 'Failed updating email status', $email_id, $email_status );
+	} else {
+		log( 'Email status change to `pending`.', $email_id );
+	}
+
+	// Remove the cron job that queues everything.
+	wp_clear_scheduled_hook( 'wc_organizer_debrief_survey_email' );
```
iandunn, line 183: It looks like this will just get added again on the next page load by `Cron\schedule_jobs()`, is that right? if so, is there any advantage to removing it here?

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
outdoor2kode: Known data was added in ad40047c6cf40e48ce4c82170b023d0bc2666690

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
outdoor2kode: In this PR, I've pre-populated the value for the question "What event did you organize?". The data sources for the other questions are different and might need to look for proper functions to retrieve the existing data. I'll handle that in a separate PR.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
```diff
@@ -0,0 +1,148 @@
...
+	return $meta['Location'][0];
+}
+
+
+/**
+ * Return a token to put into the survey URL.
+ *
+ * @return string
+ */
+function get_encryption_token() {
```
iandunn:
```suggestion
function get_verification_token() {
```

Technically this is [hashing, rather than encryption](https://www.thesslstore.com/blog/difference-encryption-hashing-salting/).

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
```diff
@@ -0,0 +1,148 @@
...
+	$wordcamp_id = get_current_blog_id();
+	return hash_hmac( 'sha1', $wordcamp_id, ORGANIZER_SURVEY_ACCESS_TOKEN_KEY );
```
iandunn:
```suggestion
	return hash_hmac( 'sha256', $wordcamp_id, ORGANIZER_SURVEY_ACCESS_TOKEN_KEY );
```

I should have double checked if `sha1` was still considered secure before suggesting it; turns out it's not. `sha256` should be fine though.

Similar to `hash_equals()`, it's not really critical for this application, but a good habit to get into.

The other `sha1` instance will need to be updated to match.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Returns content for the reminder email.
+ *
+ * @return string
+ */
+function get_email_content() {
+	$wordcamp_name   = get_wordcamp_name();
+	$survey_page_url = 'https://central.wordcamp.test/organizer-survey-event-debrief/?t=' . get_encryption_token()
+					 . '&wid=' . base64_encode( get_current_blog_id() );
```
iandunn: What's the advantage of using `base64_encode()` here?

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/**
+ * Returns content for the reminder email.
+ *
+ * @return string
+ */
+function get_email_content() {
+	$wordcamp_name   = get_wordcamp_name();
+	$survey_page_url = 'https://central.wordcamp.test/organizer-survey-event-debrief/?t=' . get_encryption_token()
```
iandunn:
```suggestion
	$tld = get_top_level_domain();
	$survey_page_url = "https://central.wordcamp.$tld/organizer-survey-event-debrief/?t=" . get_encryption_token()
```

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
```diff
@@ -0,0 +1,148 @@
...
+
+	$email .= "<strong>Event Budget</strong>\r\n";
+	$email .= 'Please update your working budget on your event dashboard. If you ran your money outside of WordPress Community Support, PBC, please also balance your budget spreadsheet and prepare your Transparency Report.';
+	$email .= "If there are any issues, please reach out to us at support@wordcamp.org to schedule a budget close-out meeting. Please complete these steps within the next two weeks.\r\n\r\n";
+
+	$email .= "<strong>Event Recording (if any)</strong>\r\n";
+	$email .= sprintf( "If you haven't yet done so, please review the submission guidelines before beginning to edit your videos (or before your videographers starts editing): %s\r\n\r\n", esc_url( 'http://blog.wordpress.tv/submission-guidelines/' ) );
+	$email .= sprintf( "To submit your video for publication to WordPress.tv, just upload them at this page: %s\r\n\r\n", esc_url( 'http://wordpress.tv/submit-video/' ) );
+	$email .= 'Our intrepid team of video moderators will review the videos and schedule them for publication. Our intrepid team of video moderators will review the videos and schedule them for publication. ';
```
iandunn:
```suggestion
	$email .= 'Our intrepid team of video moderators will review the videos and schedule them for publication.';
```

iandunn (approved):
I left some questions/suggestions, but nothing that blocks merging 👍🏻

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/page.php
iandunn: Yeah, I wish there were more filters there 😦 

It seems like it might work to hook into `template_redirect` at priority `5`, since `process_form_submission()` runs at priority `10`. Then you might be able to modify `$_POST` directly.

public_html/wp-content/mu-plugins/wcorg-network-plugin-control.php
outdoor2kode: Ah, thanks for catching it!

public_html/wp-content/plugins/wordcamp-organizer-survey/wordcamp-organizer-survey.php
outdoor2kode: TIL 👍 👍 👍

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
outdoor2kode: Ah, yes, you're right. Messed up on the basics.. Thanks!

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
outdoor2kode: Actually I did consider the fact that `sha1` can be compromised initially. I just thought later on that maybe there was a reason you suggested `sha1`, like perhaps it doesn't need such a high level of security for this particular use case here, and `sha1` performs better in terms of performance 😆

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
outdoor2kode: Because originally, the email was placed here. And with considerations for when users open their emails or click on the link, adding extra protection to ensure others can't easily see the user's email from the screen. I think we can remove it here. Even if the `wordcamp_id` is seen, it's not easy to misuse it.

public_html/wp-content/plugins/wordcamp-organizer-survey/includes/debrief-survey/email.php
outdoor2kode: Ah wasn't aware there's already a function like this. Was planning to handle that before shipping to production. 👍

PR #1055, opened 2023-09-21
Automatically disable Attendance UI

### LINKED ISSUE #620: Automatically disable Attendance UI after event ends
Imported from https://meta.trac.wordpress.org/ticket/4374 
Created by @iandunn:

CampTix's Attendance UI tool asks organizers to remember to disable it after the event is over, but that could be done automatically, to ensure it gets done.

Sometimes camps will enter data after the event, though, so the job should probably run 2 weeks after the camp is over.

The dates of the camp can be gotten via `get_wordcamp_post()->meta`.

timiwahalahti:
Camptix Attendance UI must be turned on to reveal attendee info with a link without user authentication. Currently, we trust organisers to disable the UI when they no longer need it. That can lead to the UI being open and the potential of attendee info leakage.

To mitigate the potential issues, this PR automatically disabled the UI after two weeks from generating the code.

It is done simply by saving the time when the secret was generated and comparing that when the link or Camptix settings are visited. If the link is older than two weeks, the UI will be disabled and the secret removed.

This could have been done with a cron job or event, but that just felt too complicated.

Fixes #620 

### Screenshots

CleanShot 2023-09-21 at 15 08 19@2x

CleanShot 2023-09-21 at 15 07 18@2x

1. Open a WordCamp website dashboard
2. Navigate to tickets -> setup ->  attendance UI `wp-admin/edit.php?post_type=tix_ticket&page=camptix_options&tix_section=attendance-ui`
3. Enable and generate a new code
4. You should be able to see the UI with the link
5. Temporarily change the expiry by modifying `$secret_expiry` value (set to 1 second for example)
6. Try to revisit the link, UI should not be visible
7. Check that settings have reseted

### DIFF
--- a/public_html/wp-content/plugins/camptix-attendance/addons/attendance.php
+++ b/public_html/wp-content/plugins/camptix-attendance/addons/attendance.php
@@ -356,9 +356,9 @@ class CampTix_Attendance extends CampTix_Addon {
 		<label for="camptix-attendance-generate"><?php esc_html_e( 'Generate a new secret link (old links will expire)', 'wordcamporg' ); ?></label>
 		<p class="description">
 			<?php if ( empty( $this->secret_generated ) ) {
-				echo esc_html( sprintf( _( 'Link will expire automatically after two weeks from generating it.', 'wordcamporg' ), $this->secret_expiry ) );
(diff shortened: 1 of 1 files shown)


iandunn (approved):
LGTM 👍🏻

PR #1056, opened 2023-09-21
Include sharable link in "favorite sessions" email

### LINKED ISSUE #186: Include sharable link in "favorite sessions" email
The "favorite sessions" email includes the sessions the user chose, but it'd be helpful to also include the link, so that they can easily apply those sessions to their phone or other device, and so that they can easily get a much better view of their personalized schedule on the  website.

Since it's not the primary purpose of the email, though, it should probably be displayed below the raw schedule itself.

timiwahalahti:
The "favorite sessions" email includes the sessions the user chose, it'd be helpful to include the link, so that they can easily apply those sessions to their phone or other device, and so that they can easily get a much better view of their personalized schedule on the website.

This PR adds that link to that email.

For ease, let's get the schedule page by slug. This way we do not need to pass the page ID just for this, but we neither blindly trust the input sent to REST API as a URL that belongs to the WordCamp site.

Fixes #186 

Props @iandunn 

### Screenshots

CleanShot 2023-09-21 at 17 09 11@2x

1. Go to WordCamp website schedule page
2. Favorite some sessions
3. Send the email
4. Open the link in email, favorite sessions should be indicated

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -69,12 +69,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/WordPress/wporg-parent-2021.git",
-                "reference": "216da2209e16c70fd137ea7123a9cd5d48b89c0c"
+                "reference": "d4e52af3e755c15787293e1fd1e25054af225eec"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/WordPress/wporg-parent-2021/zipball/216da2209e16c70fd137ea7123a9cd5d48b89c0c",
-                "reference": "216da2209e16c70fd137ea7123a9cd5d48b89c0c",
+                "url": "https://api.github.com/repos/WordPress/wporg-parent-2021/zipball/d4e52af3e755c15787293e1fd1e25054af225eec",
+                "reference": "d4e52af3e755c15787293e1fd1e25054af225eec",
                 "shasum": ""
             },
             "type": "wordpress-theme",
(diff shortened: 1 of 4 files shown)


timiwahalahti:
Lint errors (phpcs) are coming from rule that I don't get. 

`'validate_callback' => function( $value, $request, $param ) {` is way better than
`'validate_callback' => function ( $value, $request, $param ) {` 🤔

pkevan:
> Lint errors (phpcs) are coming from rule that I don't get.
> 
> `'validate_callback' => function( $value, $request, $param ) {` is way better than `'validate_callback' => function ( $value, $request, $param ) {` 🤔

For some reason PHPCS is treating the `function` keyword as a function declaration instead of being an anonymous function as it is in this case.

I can see two ways of this being used in the `wc-post-types/inc/rest-api.php` file, including one with a space, perhaps to get over this error.

outdoor2kode (approved):
Codes look good and works as expected.

Screenshot 2024-05-02 at 17 34 23

outdoor2kode:
Note: I timeboxed some time to address the issue where PHPCS was mistakenly identifying anonymous functions as declaration functions. Found that: 1. The `--config-show` reveals that PHPCS doesn't seem to properly read the phpcs.xml.dist file. 2. It is possible to write custom logic to avoid certain patterns being misidentified. Tried, but not working as expected. Given that it exceeded the time I had allocated for it and many other files use `function ()`, I decided not to look into it further at the moment. Noted this for future reference.

PR #1058, opened 2023-09-21
Notify Community Team when new post is pending in Central 

### LINKED ISSUE #718: Notify Community Team when new post is pending in Central
Many WordCamp organisers do have a contributor role on the WordCamp Central website. That means they can write a post, but not publish it. When they are to publish that post, it goes to pending review status.

The team has witnessed a few times that the author doesn't know/understand that no one in Central is looking at pending posts automatically. That results in posts not being published (non at the moment) or delayed.

To avoid these situations, we could send notification to Community Team (support@) when the post status is changed to pending.

timiwahalahti:
Many WordCamp organisers do have a contributor role on the WordCamp Central website. That means they can write a post, but not publish it. When they are to publish that post, it goes to pending review status.

The team has witnessed a few times that the author doesn't know/understand that no one in Central is looking at pending posts automatically. That results in posts not being published (non at the moment) or delayed.

To avoid these situations, send notification to Community Team when the post status is changed to pending.

Fixes #718 

Props @pkevan 

### Screenshots

CleanShot 2023-09-21 at 18 40 41@2x

1. Go to Central and create a new post
2. Check "Pending review" on post sidebar
3. Click "Send as pending"
4. Check your mail (on local, mailcatcher)

### After merge & deploy

- [ ] Activate the plugin on Central

### DIFF
--- a/public_html/wp-content/plugins/notify-central-on-pending-posts.php
+++ b/public_html/wp-content/plugins/notify-central-on-pending-posts.php
@@ -1,0 +1,55 @@
+<?php
+/**
+ * Plugin Name: Notify WordCamp Central on pending posts
+ * Plugin URI: http://wordcamp.org
+ * Description: Send email notification to WordCamp Central when post status becomes pending.
+ * Version: 1.0
+ *
+ * Heavily inspired from Pending Submission Notifications plugin by Razvan Horeanga.
+ */
+
+namespace Notify_Central_Pending_Posts;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	die( 'Invalid request.' );
+}
+
+add_action( 'transition_post_status', __NAMESPACE__ . '\send_notification_email', 10, 3 );
+
+/**
+ * Send the notification email.
+ *
+ * @param string  $new_status New post status.
+ * @param string  $old_status Old post status.
+ * @param WP_Post $post       Post object.
+ */
+function send_notification_email( $new_status, $old_status, $post ) {
+	if ( 'pending' === $new_status && user_can( $post->post_author, 'edit_posts' ) ) {
+		// Prevent many emails from the same post.
+		$sent = get_post_meta( $post->ID, '_ncpp_sent', true );
+		if ( ! empty( $sent ) ) {
+			return;
+		}
+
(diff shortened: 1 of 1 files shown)


pkevan (commented):
Thinking about this - I think it would be better to have this as an mu-plugin, restricting it's use on central via a site id check (or similar), rather than it being ambiguous as a plugin and potentially activated on other WordCamp (or Event) sites.

I'll leave some other feedback inline on the PR.

public_html/wp-content/plugins/notify-central-on-pending-posts.php
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Send the notification email.
+ *
+ * @param string  $new_status New post status.
+ * @param string  $old_status Old post status.
+ * @param WP_Post $post       Post object.
+ */
+function send_notification_email( $new_status, $old_status, $post ) {
+	if ( 'pending' === $new_status && user_can( $post->post_author, 'edit_posts' ) ) {
```
pkevan, line 27: not sure of the purpose of the `edit_posts` check here - would that not be covered by edit privileges anyway?

public_html/wp-content/plugins/notify-central-on-pending-posts.php
```diff
@@ -0,0 +1,55 @@
...
+		$message .= __( 'Title' ) . ": {$post->post_title}\r\n";
+		$message .= __( 'Author' ) . ": {$username->user_login}\r\n";
+		$message .= "\r\n\r\n";
+		$message .= __( 'Edit' ) . ": {$edit_link}\r\n";
+		$message .= __( 'Preview' ) . ": {$preview_link}";
+
+		wp_mail( 'support@wordcamp.org', $subject, $message );
+
+		// Save a pointer that notification has been sent.
+		update_post_meta( $post->ID, '_ncpp_sent', wp_date( 'Y-m-d H:i:s' ) );
```
pkevan, line 53: I think we need a method for clearing this on successful save, in case we switch between statuses such as:

draft => pending
pending => draft
draft => pending
pending => publish

PR #1064, opened 2023-09-21
Show empty tickets with "sold out" message

### LINKED ISSUE #573: CampTix: Poor UX when one ticket type is sold out
Imported from https://meta.trac.wordpress.org/ticket/3493 
Created by @hlashbrooke:

CampTix allows WordCamp organisers to create multiple ticket types. The most common usage of this is to create the following two tikets:

1. A general admission ticket (no more than the local equivlent of $20/day).
2. A "micro-sponsor" ticket, which is generally around 3-5x more expensive than the regular ticket.

This is entirely correct, but the issue comes in when the general admission tickets are sold out and there are only micro-sponsor tickets remaining. This happened recently in the case of WordCamp Greenville and it creates a confusing UX for potential attendees. See these URLs to see the issue:

* Ticket sales page: https://2018.greenville.wordcamp.org/tickets/ (screenshot attached below because this page will be outdated soon)
* Tickets CPT in the dashboard: https://2018.greenville.wordcamp.org/wp-admin/edit.php?post_type=tix_ticket (only visible to organisers)

The issue here is that a potential attendee comes to the ticket sales page and only sees the micro-sponor ticket available, then thinks that the tickets for the WordCamp are $100 each (this did actually happen, so it's not a made-up scenario). If we had some kind of indicator that the other ticket type was sold out, then that could help to mitigate this issue.

I created a quick mockup for a potential solution here (attached below) that does the following if a ticket type is sold out:
1. Does _not_ remove the ticket type from the table.
2. Replaces the quantity select box with a 'Sold out' string.
3. Reduces the opacity of the row to `0.3`.

There may be a better solution for this as this was a very quick mockup, so open to ideas here of course. Essentially I think we need to improve the UX so that it's clear what tickets are sold out to prevent confusion.

timiwahalahti:
Currently sold-out tickets are hidden from the tickets page.

Showing the empty, sold-out, tickets helps to avoid confusion if the camp has several types of tickets (e.g., General Admission, Micro-sponsorship, etc) and the General Admission ticket sells out. If the General Admission ticket was hidden, some users may mistakenly think that the Micro-sponsorship ticket is the "normal" ticket, even though it costs several hundred dollars. Since we value keeping regular tickets accessible by as many people as possible, we don't want anyone getting the impression that WordCamps are expensive to attend.

Fixes #573 

Props @hlashbrooke

### Screenshots

CleanShot 2023-09-22 at 00 39 10@2x

1. Go to a WordCamp site
2. Modify a ticket type to have 0 remaining
3. Ensure you have another ticket type that has tickets left
4. Visit the tickets page

### DIFF
--- a/public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
+++ b/public_html/wp-content/mu-plugins/camptix-tweaks/camptix-tweaks.php
@@ -12,7 +12,6 @@ add_action( 'camptix_admin_notices',                         __NAMESPACE__ . '\s
 add_filter( 'camptix_dashboard_paypal_credentials',          __NAMESPACE__ . '\paypal_credentials'                  );
 add_filter( 'camptix_paypal_predefined_accounts',            __NAMESPACE__ . '\paypal_credentials'                  );
 add_filter( 'camptix_stripe_predefined_accounts',            __NAMESPACE__ . '\stripe_credentials'                  );
-add_action( 'init',                                          __NAMESPACE__ . '\hide_empty_tickets'                  );
 add_action( 'wp_print_styles',                               __NAMESPACE__ . '\print_login_message_styles'          );
 add_filter( 'camptix_require_login_please_login_message',    __NAMESPACE__ . '\override_please_login_message'       );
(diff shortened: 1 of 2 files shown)


iandunn (approved):
Looks good 👍🏻

PR #1069, opened 2023-09-26
Make sure survey is sent AFTER event has completed.

StevenDufresne:
The attendee survey has sent out 2 days before the end date for rome because it was only looking for a date difference and not making sure that happened after the event. This PR fixes that logic.

I've decided to send on the exact day. I did so because this feature is still not well tested and I don't want it to go back in time and send any previous events. I would prefer to restrict it to the exact day for now. I could be persuaded otherwise.

I also added a check to make sure the WordCamp is closed as well.

### DIFF
--- a/public_html/wp-content/plugins/camptix-attendee-survey/includes/cron.php
+++ b/public_html/wp-content/plugins/camptix-attendee-survey/includes/cron.php
@@ -131,7 +131,7 @@ function is_time_to_send_email( $email_id ) {
 
 	$send_date = $end_date + ( DAYS_AFTER_TO_SEND * DAY_IN_SECONDS );
 
-	return date( 'Y-m-d', $send_date ) === current_time( 'Y-m-d' );
+	return gmdate( 'Y-m-d', $send_date ) === current_time( 'Y-m-d', true );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


iandunn (approved):
This LGTM 👍🏻 

I think it might be a little safer to use `>=` rather than `===`, just in case there's a situation where cron is broken and doesn't run for a day or two. That's an edge case though.

public_html/wp-content/plugins/camptix-attendee-survey/includes/cron.php
```diff
@@ -118,6 +118,10 @@ function is_time_to_send_email( $email_id ) {
 		return false;
 	}
 
+	if ( ! 'wcpt-closed' === $wordcamp_post->post_status ) {
```
outdoor2kode: I think here it should be `!==` instead. I've pushed a commit to make this change. If it's not right, please just revert it.

outdoor2kode (approved):
I tested it for both two days before and two days after, and everything looks good. 
And I've replaced `date()` with `gmdate()` to fix the linting error.

Additionally, I rebased on production to check if the following issue was resolved, and after confirming it was, I force-pushed. 

Screen Shot 2023-09-27 at 12 29 25 AM

I've also applied the same logic to the debrief survey cron job. I also tested it for the two days before and after scenarios, as well as cases where the WordCamp status is set to 'closed'. All seems fine.

public_html/wp-content/plugins/camptix-attendee-survey/includes/cron.php
StevenDufresne: eerg, CoPilot. Thanks Ren.

PR #1100, opened 2023-10-19
Mu-plugins: Add margin to fixed navigation on sites with the latest event banner

### LINKED ISSUE #1098: Navigation hidden under "latest site" banner
On some old WordCamp site, the (main) navigation is hidden under the new "latest site/year" banner introduced with #725.

**To reproduce**

Here are two examples of probably many WordCamp sites with this issue:
- https://london.wordcamp.org/2013/
- https://europe.wordcamp.org/2016/

**Expected behavior**

When the banner is shown on a page, the rest of the page should move down, including any fixed navigation.

**Screenshots**

Screenshots of the two example pages with the banner set to `opacity: 0.75` to visualize the navigation behind the banner:

image
image

**WordCamp**

See above (and probably more).

2ndkauboy:
The solution to the linked issue is not really easy, since different WordCamp themes use very different markup and CSS to make the navigation fixed. There is also no way to change the reference point for a fixed positioned element.

This PR will introduce some JavaScript that searches for any `<nav>` item with `position: fixed` and adds a CSS class, that is than used in the existing CSS selector pushing down the navigation as well.

Fixes #1098

### Screenshots

This is how the pages look like with this PR applied (compared to #1098):

image

image

### DIFF
--- a/public_html/wp-content/mu-plugins/latest-site-hints.php
+++ b/public_html/wp-content/mu-plugins/latest-site-hints.php
@@ -71,7 +71,8 @@ function canonical_link_past_home_pages_to_current_year() {
  */
 function add_notification_styles() { ?>
   <style type="text/css">
-		html:not(#specificity-hack) {
+		html:not(#specificity-hack),
+		.wordcamp-latest-site-notify-fixed-position-fix {
 			/* 44 = 10px x2 for padding, 24px for line height. */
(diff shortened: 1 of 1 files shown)


pkevan:
It might be better to assume that some sites have a fixed top nav, and just place this banner somewhere below that - I do wonder about the effectiveness of such an element though, and perhaps we'd be best with something more obvious?

PR #1113, opened 2023-11-27
WCPT: Get user info regardless of role on site

iandunn:
Previously the query only looked at users with a role on the current site, which didn't always include everyone.

See https://core.trac.wordpress.org/ticket/38851
See https://wordpress.slack.com/archives/C08M59V3P/p1700830335858499


iandunn:
Merged manually in [ac3fa93](https://github.com/WordPress/wordcamp.org/commit/ac3fa93c3f8af124936f60cde9f912efe1028886) since GitHub's UI wasn't working

PR #1120, opened 2023-12-01
Update nginx.conf to server .woff2 fonts

StevenDufresne:
We use .woff2 fonts on wordpress.org. This project will inherit some of those styles and fonts because of the events.wordpress.org crossover.
 
https://github.com/WordPress/wporg-mu-plugins/blob/b0fff4be072f71ed7d478d8b96638805aa0ca041/mu-plugins/global-fonts/index.php#L214


iandunn:
Is this working in your tests? If so it LGTM 👍🏻

PR #1124, opened 2023-12-04
Events: Use a showcase like approach for the events list/archive

StevenDufresne:
This is a draft PR exploring the use of components from https://github.com/wordpress/wporg-showcase-2022 to build out the "All Events" page. This is a Work in Progress.

## Why I think this is the best approach
- We don't have to get the google map abstraction perfect since we don't need to do it.
  - The google map block is already more than a google map block. 
- By using a very similar approach to showcase, we'll get almost all styles out of the box
  - We're talking implementation in hours, not days (it took me less than an hour to put this together)
- Consistency  

## What's not perfect here yet
- This currently filters on the full list of retrieved events. 
   - May be a bad idea if there are too many events
- The event retrieval functions are in the google map block plugin.
- By rendering from the server, we don't get the best user-centric datetime for the user.
   - I think the [time block](https://github.com/WordPress/wporg-mu-plugins/tree/trunk/mu-plugins/blocks/time), should help. 

### Current
Note: the filters are not being applied to the results set yet.

events wordpress test__map_type%5B%5D=wordcamp format_type%5B%5D=wordcamp s=asdfasf


public_html/wp-content/themes/wporg-events-2023/package.json
```diff
@@ -12,6 +12,7 @@
 		"postcss-preset-env": "^9.3.0"
 	},
 	"scripts": {
+		"build:block": "wp-scripts build",
```
iandunn:
```suggestion
		"build": "wp-scripts build && postcss postcss/style.pcss postcss/editor.pcss --dir . --ext css",
```

_(then remove the existing `build` command below)_

This way we don't have to remember 2 different commands.

iandunn (approved):
This WFM 👍🏻 

> Note: the filters are not being applied to the results set yet.

This was the part that didn't feel great to me, since it's shoehorning arbitrary data from the `wporg_events` table into a `WP_Query` flow. The opposite approach in https://github.com/WordPress/wporg-mu-plugins/pull/523 has it's own set of problems, though, so I think this PR is at least as good, and probably better.

public_html/wp-content/themes/wporg-events-2023/src/event-list/index.php
```diff
@@ -0,0 +1,119 @@
...
+	return $filtered_events;
+}
+
+/**
+ * Retrieves event list based on the filters.
+ *
+ * @return array
+ */
+function get_event_list(): array {
+	return get_all_upcoming_events();
```
iandunn: If you call `Google_Map\get_events( 'all-upcoming', 0, 0 )` you'll get the same results, but they'll be cached.

public_html/wp-content/themes/wporg-events-2023/templates/page-upcoming.html
```diff
@@ -0,0 +1,15 @@
+<!-- wp:template-part {"slug":"header"} /-->
```
iandunn, line 1: I assumed the URL would be `upcoming-events`, but this WFM too. The menu will need to be updated, though:

https://github.com/WordPress/wordcamp.org/blob/70eaf22ab3f2194a9686e15328a104fe9d21f61c/public_html/wp-content/themes/wporg-events-2023/functions.php#L38-L41

public_html/wp-content/themes/wporg-events-2023/package.json
StevenDufresne: Yeah, that makes more sense. I rammed that in for the prototype.

iandunn:
I merged in the Past Events template in 9cc0b23f6e34a3fd85eb12d6796c2c737e376e3a

StevenDufresne:
Opened up a PR for adjusting the time block:
https://github.com/WordPress/wporg-mu-plugins/pull/534

public_html/wp-content/themes/wporg-events-2023/patterns/front-cover.php
```diff
@@ -46,7 +46,7 @@
 <!-- wp:column {"width":"33.33%","fontSize":"medium","fontFamily":"eb-garamond"} -->
 <div class="wp-block-column has-eb-garamond-font-family has-medium-font-size" style="flex-basis:33.33%"><!-- wp:group {"style":{"border":{"radius":"2px","width":"1px"},"spacing":{"padding":{"left":"var:preset|spacing|30","top":"var:preset|spacing|20","bottom":"var:preset|spacing|20"}}},"borderColor":"light-grey-1","layout":{"type":"constrained"}} -->
 <div class="wp-block-group has-border-color has-light-grey-1-border-color" style="border-width:1px;border-radius:2px;padding-top:var(--wp--preset--spacing--20);padding-bottom:var(--wp--preset--spacing--20);padding-left:var(--wp--preset--spacing--30)"><!-- wp:paragraph {"style":{"layout":{"selfStretch":"fit","flexSize":null},"elements":{"link":{"color":{"text":"var:preset|color|charcoal-0"}}}},"textColor":"charcoal-0","fontSize":"heading-6"} -->
-<p class="has-charcoal-0-color has-text-color has-link-color has-heading-6-font-size">540,537 members</p>
+<p class="has-charcoal-0-color has-text-color has-link-color has-heading-6-font-size">540,537 attendees</p>
```
iandunn, line 49: I don't think it'd be accurate to change this. That number is the # of meetup.com group members (plus WordCamp attendees, but that doesn't affect it significantly), and the number of attendees is likely to be significantly different.

If we really want to say attendees, we could try to estimate the # based on RSVPs, but I don't think it'd be accurate enough to publish. IMO members is the most accurate/meaningful stat we can use here.

public_html/wp-content/themes/wporg-events-2023/src/event-list/index.php
```diff
@@ -47,7 +47,7 @@ function render( $attributes, $content, $block ) {
 	array_walk( $facets, 'sanitize_text_field' );
 	$facets = array_filter( $facets );
 
-	$events = get_events( $attributes['events'], 0, 0, $facets );
+	$events = get_all_upcoming_events();
```
iandunn: @StevenDufresne Did you mean to overwrite this? I'm going to assume not and change it back, since passing `$facets` is needed for filtering, and `get_events()` adds caching, but LMK if there's a reason to use `get_all_upcoming_events()` directly.

public_html/wp-content/themes/wporg-events-2023/package.json
iandunn: done in 855fa570

public_html/wp-content/themes/wporg-events-2023/src/event-list/index.php
iandunn: 0889f6f3

public_html/wp-content/themes/wporg-events-2023/src/event-list/index.php
```diff
@@ -0,0 +1,119 @@
...
+		'format' => get_query_var( 'format_type' ) ?? '',
+		'month'  => get_query_var( 'month' ) ?? '',
+	);
+	array_walk( $facets, 'sanitize_text_field' );
+	$facets = array_filter( $facets );
+
+	$events = get_events( $attributes['events'], 0, 0, $facets );
+
+	// Get all the filters that are currently applied.
+	$filtered_events = array_slice( filter_events( $events ), 0, 10);
```
iandunn, line 53: This is cutting the list off at 10, but we only want that for the homepage. Maybe the block should accept a `limit` attribute that defaults to `50` or something?

public_html/wp-content/themes/wporg-events-2023/templates/search.html
```diff
@@ -0,0 +1,22 @@
...
+<div class="wp-block-group alignfull has-white-background-color has-background" style="border-top-color:var(--wp--preset--color--light-grey-1);border-top-width:1px;padding-top:18px;padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:18px;padding-left:var(--wp--preset--spacing--edge-space)"><!-- wp:wporg/site-breadcrumbs {"fontSize":"small"} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:group {"tagName":"main","style":{"spacing":{"blockGap":"0px"}},"className":"entry-content","layout":{"type":"constrained"}} -->
+<main class="wp-block-group entry-content">
+	<!-- wp:group {"align":"wide","style":{"spacing":{"padding":{"left":"var:preset|spacing|edge-space","right":"var:preset|spacing|edge-space","bottom":"var:preset|spacing|50"}}},"layout":{"type":"default"}} -->
+	<div class="wp-block-group alignwide"
+		style="padding-right:var(--wp--preset--spacing--edge-space);padding-bottom:var(--wp--preset--spacing--50);padding-left:var(--wp--preset--spacing--edge-space)">
+
+		<!-- wp:pattern {"slug":"wporg-events-2023/event-list-filters-archive"} /-->
```
iandunn: This doesn't exist in the repo, do you have it locally? Or should this be `event-list-filters`?

public_html/wp-content/themes/wporg-events-2023/patterns/events-list-filters.php
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+?>
+
+<!-- wp:group {"align":"wide","style":{"spacing":{"margin":{"top":"var:preset|spacing|30","bottom":"var:preset|spacing|30"}}},"layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
+<div class="wp-block-group alignwide" style="margin-top:var(--wp--preset--spacing--30);margin-bottom:var(--wp--preset--spacing--30)"><!-- wp:group {"layout":{"type":"flex","flexWrap":"wrap"}} -->
+<div class="wp-block-group"><!-- wp:search {"showLabel":false,"placeholder":"Search events...","width":100,"widthUnit":"%","buttonText":"Search","buttonPosition":"button-inside","buttonUseIcon":true,"className":"is-style-secondary-search-control"} /--></div>
+<!-- /wp:group -->
+
+<!-- wp:template-part {"slug":"event-filters"} /-->
```
iandunn: I don't see this either, do you have it locally?

iandunn: I added it in 303e77f3 so I can work on it today, but free free to change that if you had something else in mind

public_html/wp-content/themes/wporg-events-2023/templates/search.html
iandunn: Same here as https://github.com/WordPress/wordcamp.org/pull/1124#discussion_r1415849299

public_html/wp-content/themes/wporg-events-2023/templates/page-upcoming.html
iandunn: updated to `upcoming` in 5b413d4ec2ca2d7168d758907d0ab74ea6fcccb2, 0de639a3

public_html/wp-content/themes/wporg-events-2023/src/event-list/index.php
StevenDufresne: Eeeek. No, it was to temporarily restore the table for styling purposes.

public_html/wp-content/themes/wporg-events-2023/templates/search.html
StevenDufresne: Hmmm, I was on late last night. Fuzzy brain. Forgot to add it to git.

public_html/wp-content/themes/wporg-events-2023/patterns/front-cover.php
StevenDufresne: Thanks for that context. I was going off the designs. I'll provide that feedback and switch it back to members today.

public_html/wp-content/themes/wporg-events-2023/src/event-list/index.php
StevenDufresne: Yep, good idea. I was thinking about the grouping as well. Should we pass in a `groupyByMonth` property? Feels too specific.

Alternatively, we can pass a "heading" property to this list block and create a wrapper block where we use the query results to loop and output multiple instances of the event list block. I'll look at that today.

public_html/wp-content/themes/wporg-events-2023/templates/search.html
iandunn: Does it still need to be added? I just worked around it in 303e77f3be3df838a070a6b67ea39348b4c08414, but I'm not sure if that matched what you have locally

public_html/wp-content/themes/wporg-events-2023/patterns/front-cover.php
thetinyl: > IMO members is the most accurate/meaningful stat we can use here.

I definitely see what you mean. I'm just unsure if "members", despite being accurate, is the right term to use. It does have hints of exclusivity and seems directly tied to Meetup.com's terminology, not necessarily WordPress's. When "members" are talked about in WP ways, I've typically seen it explicitly contextualized as "community members"... which is too long (as usual 😅).

StevenDufresne: So summing up the problem, if we use "attendees", we will be underrepresenting the actual number of attendees.

I also find `members` to be an inclusive term. What about just `people`? 

Simple is as simple does.

thetinyl: `people` is accurate and simple, for sure. Also feels too generic/passive. (and what of all the pet contributors? 😎)

I'd suggest one of the following (for now), with my preference being 1:

1. `participants`: technically accurate, the bar for participation can be fluid, i.e., as low as just becoming a member on Meetup.com or as high as attending and contributing at a WordCamp. 
2. `contributors`: perhaps not entirely inclusive/accurate, but directly related to the project and also used elsewhere on the page already.

iandunn: 🤔 that's a good point, we definitely don't want it to feel exclusionary. I like `participants`, thanks for thinking of that!  

I think the problem with `attendees` is more that it _over_represents the number of attendees, since so many people join a meetup group, and even RSVP to an event, but don't actually show up. IMO accuracy is the most important factor, and it'd be better to err on the side of too low than too high. My concern is that inflating stats, even unintentionally, would erode trust.

I'll go ahead and update it to `participants` since we're coming up on the deadline to launch, but I'm happy to continue discussing it if y'all would like to iterate on that, or make any changes.

PR #1126, opened 2023-12-05
Add past events template

iandunn:
This uses the same approach as #1124, and shouldn't be merged until after that is.

See https://www.figma.com/file/jdMk5ssz2Av7KFfEaeK7de/Events?type=design&node-id=1617-13838&mode=dev

_[ todo: add screenshots visual changes done ]_


iandunn:
merging into #1124 to make collaboration easier

PR #1127, opened 2023-12-06
Events: Split event-list by month year

outdoor2kode:
Split the event list by month and year based on the [design](https://www.figma.com/file/jdMk5ssz2Av7KFfEaeK7de/Events?node-id=1617%3A13682&mode=dev).

## Screencast

https://github.com/WordPress/wordcamp.org/assets/18050944/0d9ffeaa-cb52-4724-8fab-2a2a3e51fcad

## Test Steps

Adding the following snippet after `$filtered_events` will get you the same fake data as in the screencast.
```
use stdClass;

for ($i = 0; $i < 10; $i++) {
	$event = new stdClass();

	$event->id = strval(rand(10000000, 99999999));
	$event->type = "event_type_" . ($i + 1);
	$event->title = "Event Title " . ($i + 1);
	$event->url = "https://example.com/event_" . ($i + 1);
	$event->meetup = "Meetup " . ($i + 1);
	$event->location = $i % 2 == 0 ? "online" : "offline";
	$event->latitude = strval(37.75 + $i);
	$event->longitude = strval(-122.42 - $i);
	$event->tz_offset = strval(-28800 + ($i * 3600));

	$timestamp = mktime(0, 0, 0, $i + 1, 1, 2024);
	$event->timestamp = $timestamp;

	$filtered_events[] = $event;
}
```

## Question

I think we are going to just display the events naturally as their dates approach, right? 
Like by the end of December, the list will naturally be divided into two sections - one for December and one for January, as January events begin to surface. 
Or is there a requirement to explicitly show 10 events for December and 10 for January?


StevenDufresne:
I was able to test this with real data:

https://events.wordpress.test/?s=HelpDesk

Looks good.

StevenDufresne:
This looks good, I pushed some changes so we can remove the style sheet that uses `<!-- wp:heading -->` instead. 
That should remove a resource that needs to be loaded since the list doesn't have any styles.

The one thing to note is that this uses the data and groups on the date. If by chance a month doesn't have results, like in a search, it won't show the in-between months that have no events. I think that's fine but people typically think chronologically and showing empty months in between January and April for example, may be necessary down the road.

Issue #1128, opened 2023-12-06
Events: Landing page performance


adamwoodnz:
Current Lighthouse score

![Screenshot 2023-12-07 at 10 22 13 AM](https://github.com/WordPress/wordcamp.org/assets/1017872/aff78b3b-4e5b-407f-b7bc-2bee442b2e96)

Key problems:

![Screenshot 2023-12-07 at 10 22 33 AM](https://github.com/WordPress/wordcamp.org/assets/1017872/d313793e-f0ac-4dcb-81f0-482a4b92eb9e)

adamwoodnz:
Converting images to webp will help

StevenDufresne:
>Converting images to webp will help

I turned on [JetPack site acceleration](https://jetpack.com/support/site-accelerator/?site=events.wordpress.org) which should convert to WebP. I see it working on `/organize-and-event` but not on the homepage. Not sure why yet.

What are the acceptance criteria for this ticket? As in, when can we close it?

pkevan:
I guess some of this was probably down to the page not being live as such, a quick reference in lighthouse shows a page speed of 86 (logged out). 

![Screenshot 2023-12-12 at 10 19 30](https://github.com/WordPress/wordcamp.org/assets/2290623/d981eeb8-e669-41ba-9064-b3678ae801dd)

I'll update the lighthouse cron crawler with the current URL (without `?preview`).

adamwoodnz:
> I guess some of this was probably down to the page not being live as such, a quick reference in lighthouse shows a page speed of 86 (logged out).

Yeah I expected a bit of a bump once all the live infrastructure was in place, that's great.

adamwoodnz:
> What are the acceptance criteria for this ticket? As in, when can we close it?

I think we're done 🙂

StevenDufresne:
I'm still confused as to why the JetPack gallery is not converting the image to WebP. I wonder if they need to be served from the media library? Probably the case because the `/organize-an-event` page serves the feature image as WebP but not the one linked to the file system.

[Reference](https://jetpack.com/support/site-accelerator/)
[...]

PR #1144, opened 2023-12-07
CampTix: Only enqueue CSS when used, for performance

iandunn:
See #1128 

This speeds up most pages a little bit by not enqueuing an unnecessary file. It's also moved to the footer so it doesn't block the initial page render. That wasn't valid HTML when this plugin was originally written, but it is now.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -486,7 +486,7 @@ class CampTix_Plugin {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


pkevan:
This change appears to break the styling of the attendees page: https://github.com/WordPress/wordcamp.org/issues/1210.

pkevan:
I have a potential fix coming, it might need to be added to other places, but it seems like other shortcodes depend on the stylesheet.

iandunn:
Thanks!

PR #1154, opened 2023-12-07
Events: Stop large cumulative layout shift caused my map height

adamwoodnz:
See #1128 

There is a large shift in the page content when the map loads, as the container has no height set beforehand. This has a significant impact on the Lighthouse score. This PR sets the height on the initial container, and the map to fill it, which we can do because there is no search field or other associated UI surrounding the map which needs to also fit in the container.

There should be no visual change, other than the layout shift not happening. The layout should stay set from initial load.


iandunn (approved):
LGTM 👍🏻

PR #1159, opened 2023-12-08
Events: Add the social tag code.

### LINKED ISSUE #1158: Events: Add social media share image & description.
We need to add social media share tags. 

We can use the [showcase code](https://github.com/WordPress/wporg-showcase-2022/blob/625b65aa1876da9c68c5c08e00b6fbdabb8acde2/source/wp-content/themes/wporg-showcase-2022/functions.php#L453C26-L453C26) as inspiration.

@WordPress/meta-design Can we get an image? [Here's showcase's image](https://github.com/WordPress/wporg-showcase-2022/blob/main/source/wp-content/themes/wporg-showcase-2022/images/social-image.png) for sizing.

StevenDufresne:
This PR adds social media tags and descriptions for shares.

**To Do**
- Replace image
- Replace description copy

Fixes #1158


public_html/wp-content/themes/wporg-events-2023/functions.php
```diff
@@ -39,6 +40,41 @@ function enqueue_assets() {
 	);
 }
 
+/**
```
iandunn, line 45: There's some existing code that does similar things:

* https://github.com/WordPress/wordcamp.org/blob/72b6a8e6ea6d3bc6dfb4c070ef2a093c5f744110/public_html/wp-content/mu-plugins/jetpack-tweaks/sharing.php
* https://meta.trac.wordpress.org/browser/sites/trunk/wordpress.org/public_html/wp-content/mu-plugins/pub/site-branding.php?rev=11778

Would those work, maybe with some modifications, rather than duplicating it?

StevenDufresne: Ah, nice. I wasn't aware of that code. Yes, we should re-use thanks.

dd32 (approved):

PR #1163, opened 2023-12-09
Events: Update templates to use WP Query

### LINKED ISSUE #1152: Events: Update theme to use WP_Query instead of custom queries direct to DB.
We are currently using custom DB queries to retrieve data for event.wordpress.org. As a result, It's getting in the way of using some of our existing components #1136, changing our block architecture and making this tricky overall.

We should switch to using `WP_Query`.

iandunn:
Follows #1156 
Fixes #1152 

Now that the events data is available via a query block, we can update the templates/patterns/etc to use that instead of the custom rendering in the event-list block. That block may still be needed to access some of the data, though, since patterns can't have any content that's specific to a post in the loop.


iandunn:
There isn't a tangible need for this anymore, so I'm going to close it. We may want to revisit it when we add pagination for #1178

PR #1166, opened 2023-12-11
Add in the query-total to the search results.

StevenDufresne:
This PR builds on changes added in #1156, to add the `<!-- wp:query-total /-->` component to the filters template. 

Ultimately #1163 will replace this code but since it isn't clear when that will land, we don't need to delay this feature as it already works and is helpful for users.

## Templates

I duplicated the filters template which isn't very DRY but I don't see a better option.

Ideally we could add the `query-filters` into a template part but that had to be reversed in https://github.com/WordPress/wordcamp.org/commit/72b6a8e6ea6d3bc6dfb4c070ef2a093c5f744110 since it breaks the `overflow: scroll` behaviour on mobile for the filters.

Alternatively, we could move the search container to a new template, but that will add another dev and not really simplify things.

### Screenshots

| Page | Screenshot |
|--------|--------|
| `/upcoming-events` | events wordpress test_upcoming-events__month=05 |
| `?s={keyword} | events wordpress test__format_type=in-person s=help |
|`/` (no count) | events wordpress test__preview | 

1. Visit the homepage, ensure there are no results
2. Navigate to `/upcoming-events` and use filters, verify that the count matches the result list
3. Search for something, verify that the results match the count.

### DIFF
--- a/public_html/wp-content/themes/wporg-events-2023/patterns/events-list-filters-with-count.php
+++ b/public_html/wp-content/themes/wporg-events-2023/patterns/events-list-filters-with-count.php
@@ -1,0 +1,29 @@
+<?php
+/**
+ * Title: Events List Filters With Event Count
+ * Slug: wporg-events-2023/event-list-filters-with-count
+ * Inserter: no
+ */
+
+?>
+
+<!-- wp:group {"align":"wide","className":"wporg-events__filters","style":{"spacing":{"margin":{"top":"40px","bottom":"40px"}}},"layout":{"type":"flex","flexWrap":"wrap","justifyContent":"space-between"}} -->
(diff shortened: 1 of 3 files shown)


adamwoodnz (approved):
👍 Works for me

Out of scope for this PR, but I noticed that if I have filters set and enter/change search term the filters are cleared.

StevenDufresne:
> 👍 Works for me
> 
> Out of scope for this PR, but I noticed that if I have filters set and enter/change search term the filters are cleared.

Good catch. That's also the case for wp.org/showcase. We'll follow up.

StevenDufresne:
Logged: https://github.com/WordPress/wordcamp.org/issues/1167

PR #1181, opened 2023-12-13
Events: Fix list titles are not underlined + Improve CSS clearness

outdoor2kode:
This PR fixes 
1. List titles are not underlined. ([Ref](https://github.com/WordPress/wordcamp.org/pull/1172/files#r1424806482))
2. Align the format of `/community-callout.pcss` with others.
3. Try explicitly adding back the '&', and see how it feels? Do y'all prefer this? Just take this chance to align the style. If there's no particular preference, ~I will revert the commit. Because actually, mostly, the '&' wasn't added in the first place.~ **Update**: I'll keep it as this is how wporg-mu-plugins does.

### DIFF
--- a/public_html/wp-content/themes/wporg-events-2023/postcss/blocks/event-list.pcss
+++ b/public_html/wp-content/themes/wporg-events-2023/postcss/blocks/event-list.pcss
@@ -1,7 +1,7 @@
 .wporg-marker-list__container {
 	padding-left: 0;
 
-	.wporg-marker-list-item {
+	& .wporg-marker-list-item {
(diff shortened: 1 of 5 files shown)


iandunn (approved):
LGTM 👍🏻 

I don't care about the `&` from a style perspective, either way is fine with me. The only thing that matters to me is that we include it **_if_** it's necessary for things to transpile correctly. Feel free to remove it if the built styles work without it in browsers that don't natively support nested rules.

outdoor2kode:
I'll keep that commit as that is how `wporg-mu-plugins` does.

PR #1185, opened 2023-12-19
Events: Restore local event time

### LINKED ISSUE #1170: Events: Replace UTC time with local time.
We currently display times to users using UTC. It is more user-friendly to use the user's timezone.

**Screenshots**

Screenshot 2023-12-11 at 1 59 41 PM

iandunn:
Fixes #1170

This changes back to displaying the event date/times in JS, so they can be converted to the visitor's local time instead of being in UTC.

### DIFF
--- a/public_html/wp-content/themes/wporg-events-2023/package.json
+++ b/public_html/wp-content/themes/wporg-events-2023/package.json
@@ -12,12 +12,16 @@
 		"postcss-import": "^15.1.0",
 		"postcss-preset-env": "^9.3.0"
 	},
+	"eslintConfig": {
+		"extends": "../../../../.eslintrc.js"
+	},
+	"prettier": "../../../../.prettierrc.js",
 	"scripts": {
 		"start": "npm run watch",
 		"watch": "wp-scripts start & npm run build:css -- --watch",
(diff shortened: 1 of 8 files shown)


iandunn:
This is mostly working, just needs some clean up. The one big problem left is that the block fails to render in the template editor. I think it's a similar problem to https://github.com/WordPress/wporg-mu-plugins/issues/500

It's a dynamic block, and the PHP `render()` passes some data to JS via `wp_add_inline_script()`. That works well on the front end, but in the editor the JS variable is undefined. `Edit()` uses `<ServerSideRender>`.

I guess the "correct" way to do it would be to fork `view.js` into `edit.js` as a React component, but it doesn't seem like that should be necessary. It'd waste time and introduce tech debt.

@renintw @adamwoodnz @StevenDufresne, do you have any thoughts on a way to reuse `view.js` in the template editor, or is the only way to fork it?

iandunn:
I removed `view.js` from `editorScript` since it wasn't working, and added an early return if the global isn't defined. That avoids the explicit error in the editor, but doesn't fix the underlying problem. We can fix that in a future PR if any of you have ideas on that, though.

Issue #929, opened 2023-06-29
NextGen WP Event Attendee Satisfaction Survey


devinmaeztri:
**Current WordCamp Attendee Survey:** Attendee Survey is not mandatory for organizer to send it to the attendees, and the feedback goes to the organizers, instead of WordCamp Central (we don't have access to the feedback directly).

**Expected Attendee Satisfaction Survey for new WP events:** As part of the learning from experimentation and a default process, we want the Attendee Satisfaction Survey to be mandatory for attendees and they can opt out. We also want it to be attached to the central event dashboard, so WordCamp Central can have access to the feedback of the survey for every event. 

### Attendee Satisfaction Survey

**Start Survey**	

If you recently attended a WordPress event, please take the survey below to help us make them even better in the future.

**Part I. Tell us about you!**

1. How long have you been using WordPress? [Required]

- More than 1 year
- 1 year 
- Less than a year 
- I don't use WordPress yet

2. Which of these best describes you?  [Allow more than one answer]  [Required]

- Personal Blogger 
- Company Blogger 
- Designer 
- Junior Developer 
- Senior Developer 
- Project Manager 
- System Administrator/IT Professional 
- Sales/Marketing/PR 
- Business Owner 
- WordPress Fan 
- Other: (Please specify.)

3. Please provide a link to the WordPress Meetup chapter where you are actively participating.
4. Have you attended WordPress events before? [Required]

- Yes 
- No

5. If yes, please tell us what event, where and when.
6. Which of the following role(s) have you contributed to in the past? [Allow more than one answer]

- WordPress event Organizer
- WordPress event Volunteer
- WordPress event Sponsor
- WordPress event Speaker/Trainer/Facilitator
- Other: (Please specify.)

7. Did this event allow you to experience any of the following roles? [Allow more than one answer] [Required]

- First-time WordPress event Attendee
- First-time WordPress event Organizer
- First-time WordPress event Volunteer
- First-time WordPress event Sponsor
- First-time WordPress event Speaker/Trainer/Facilitator
- Other: (Please specify.)

**Part II. Tell us about the event you attended!**

1. What WordPress event did you attend? * (title, city, year)  [Required]
2. How did you hear about the event? [Required]
3. How satisfied are you with the event? [Required]

- Extremely satisfied
- Satisfied
- Neutral, 
- Dissatisfied
- Extremely unhappy

4. What aspects of the event were valuable to you?  [Allow more than one answer] [Required]

[...]

PR #1186, opened 2023-12-20
Fix Attendee Survey, it's missing some next gen events.

StevenDufresne:
The attendee survey is not pulling all the Next Gen events because sometimes `$blog_id = get_post_meta( $camp->ID, '_site_id', true );` returns an empty string.

This PR updates that line to use `get_wordcamp_site_id` ([ref](https://github.com/WordPress/wordcamp.org/blob/05a520922cc35980d227e26cc415f09913125104/public_html/wp-content/mu-plugins/4-helpers-wcpt.php#L85)) since it handles that case.

Incidentally, this PR updates the WordCamp query to only start looking after this feature was launched in #929.

## Screenshots

| Before | After |
|--------|--------|
| Screenshot 2023-12-20 at 12 18 49 PM | Screenshot 2023-12-20 at 12 18 21 PM |

### DIFF
--- a/public_html/wp-content/plugins/camptix-attendee-survey/includes/admin-page.php
+++ b/public_html/wp-content/plugins/camptix-attendee-survey/includes/admin-page.php
@@ -61,13 +61,18 @@ function get_feedback_details() {
 		'post_type' => WCPT_POST_TYPE_ID,
 		'post_status' => 'wcpt-closed',
 		'posts_per_page' => -1,
+		'date_query'     => array(
+			array(
+				'after'     => '2023-09-01', // The feature doesn't exist before this date.
+				'inclusive' => true,
+			),
+		),
 	) );
 
 	$feedback_details = array();
 
 	foreach ( $wordcamps as $camp ) {
-		// It's a bit counter intuitive, but the site ID is the blog ID.
-		$blog_id = get_post_meta( $camp->ID, '_site_id', true );
+		$blog_id = get_wordcamp_site_id( $camp );
 
 		switch_to_blog( $blog_id );
 


iandunn (approved):
🤔 those are both good changes to make regardless, but I don't think there should be any NextGen events that don't have a `_side_id` field.

That get's assigned [when the site is created](https://github.com/WordPress/wordcamp.org/blob/05a520922cc35980d227e26cc415f09913125104/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-new-site.php#L145-L154), the fallback code in `get_wordcamp_site_id` was just to handle old sites that were created before we started keeping track of the ID on the `wordcamp` post.

There may be a bug somewhere else that also needs to be fixed to address the root cause.

StevenDufresne:
Yeah I'm not sure why but the `get_wordcamp_site_id` also seems to protect against that as if it's somewhat expected:

https://github.com/WordPress/wordcamp.org/blob/05a520922cc35980d227e26cc415f09913125104/public_html/wp-content/mu-plugins/4-helpers-wcpt.php#L90

I followed the code backwards but it isn't clear why it was added.

iandunn:
I remember adding it, and the reasons/timeline was something like this:

* 2006 - 2014: No `wordcamp` posts had a `_site_id` value, it wasn't a thing that existed anywhere in the codebase.
* 2015: We started storing the `_site_id` value to the `wordcamp` post when we created a new site. From this point on, all new sites should have a `_site_id` value in their `wordcamp` post. If they don't, something has gone wrong.
* 2015: In 0e62c91a, I added the code to lookup a site based on the `_site_id`. I included the `! $site_id` fallback because most sites that existed at this point didn't have a `_site_id` meta key on their `wordcamp` post
* 2020: I created `bin/wp-cli/single-use/wcpt/set_site_id_post_meta.php` in `r3580-wordcamp`, and ran it. This retroactively added the `_site_id` to most of the `wordcamp` posts that were missing it. At this point, the fallback code shouldn't be needed, unless a bug is causing sites to not have the `_site_id`

PR #1193, opened 2024-01-12
Add functionality for flagship event landing pages 

iandunn:
See https://www.figma.com/file/zDDX2dyqXSt8ASEeZ8h3LQ/WordCamp-Archives?type=design&node-id=4467-1930&mode=dev

* Modifies sunrise.php to not redirect `{flagship}.wordcamp.org` to `{flagship.wordcamp.org/{latest year}`
* Modifies sunrise.php to show a landing page instead of the redirect
* Adds a new theme to show that landing page. I chose a theme to try and keep things straightforward, rather than having to muck with Core internals too much. We can't just output some HTML from sunrise because we need to wait until WP is fully loaded.

This is the functionality only, the design will need to be implemented in a follow-up PR. This PR's functionality will be disabled on production until that PR merges.


iandunn:
I think the functionality is mostly done. Once the design is finalized I can circle back to implement that.

PR #1211, opened 2024-01-17
Update shortcodes.php

### LINKED ISSUE #1210: Attendees: script and styles not enqueueing
The scripts and styles within camptix no longer render on shortcode output, for the attendees , example: https://netherlands.wordcamp.org/2023/bezoekers2023/

Some initial debugging shows that https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/camptix.php#L5416 isn't being called. It could perhaps be related to the shortcode hardening which shipped in 6.2.1 & 6.2.2.

pkevan:
Adds css style back for shortcode, without it, the attendees list is a single column

Fixes #1210


pkevan:
Screenshot 2024-01-17 at 15 45 28

vs 

Screenshot 2024-01-17 at 15 44 51

Issue #1205, opened 2024-01-16
Add Event schema


iandunn:
> Why don’t we have any [event schema](https://developers.google.com/search/docs/appearance/structured-data/event) on these pages? ([See here](https://validator.schema.org/#url=https%3A%2F%2Fevents.wordpress.org%2F))
> https://make.wordpress.org/community/2024/01/16/share-your-feedback-about-the-new-wordpress-events-page/#comment-31484

It sounds like we should only add it for individual events, like https://events.wordpress.org/rome/2023/training/ and https://asia.wordcamp.org/2024/, rather than each event on pages like https://events.wordpress.org/upcoming-events/:

> Currently, the event experience on Google only supports pages that focus on a single event. We recommend focusing on adding markup to your event posting pages instead of pages that list schedules or multiple events.
> https://developers.google.com/search/docs/appearance/structured-data/event#technical-guidelines

PR #1214, opened 2024-01-19
WCPT: Save venue street name and number

iandunn:
We previously didn't save those, but they'll be needed for #1205. This PR updates them when a WordCamp post is saved, so that new posts will have the data. We'll also need to backfill the data for existing posts with a cli script.


iandunn:
The script was added in `r4714-wordcamp`, and I ran it on my sandbox to backfill the production database.

PR #1221, opened 2024-01-23
Events: Make filter URLs pretty

### LINKED ISSUE #1196: Make clean filter URLs on events.wordpress.org
Right now the overview of WordCamps on events.wordpress.org has the following URL: `https://events.wordpress.org/upcoming-events/?event_type%5B%5D=wordcamp`

I'd prefer that to be: `https://events.wordpress.org/upcoming-events/wordcamp/`

Is that doable?

iandunn:
Fixes #1196 

WIP

When finished, applying filters should result in a URL like `https://events.wordpress.test/upcoming-events/filtered/format/online/` and 
`https://events.wordpress.test/upcoming-events/filtered/format/in-person/type/wordcamp/month/january,february/country/united-states,germany/`. Currently that's only partially working.

When finished, you should also be able to start at `https://events.wordpress.test/upcoming-events/` and apply filters to get to a "pretty" URL like the one above, rather than the key-value pairs shown in #1196. This part will probably require some modification to the `query-filters` block, to append the selected filters to the form action in JS.

### DIFF
--- a/public_html/wp-content/themes/wporg-events-2023/inc/events-query.php
+++ b/public_html/wp-content/themes/wporg-events-2023/inc/events-query.php
@@ -1,17 +1,26 @@
 <?php
 
 namespace WordPressdotorg\Events_2023;
-use WP_Query, WP_Post, WP_Block;
+use WP, WP_Query, WP_Post, WP_Block;
 use WordPressdotorg\MU_Plugins\Google_Map;
 
 defined( 'WPINC' ) || die();
+
+// Match URLs like `{pagename}/filtered/{facets}`, e.g., `/upcoming-events/filtered/type/meetup/format/in-person/month/05/country/US/`.
+// This intentionally doesn't have the starting/ending delimiters and flags, so that it can be used with
+// `add_rewrite_rule()`.
+const FILTERED_URL_PATTERN       = '([\w-]+)/filtered/(.+)';
+const PRETTY_URL_VALUE_DELIMITER = '-';
 
 // Misc.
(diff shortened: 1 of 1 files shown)


iandunn:
I setup a rewrite rule so that URLs like `https://events.wordpress.test/upcoming-events/filtered/format/online/` will be treated the same as `https://events.wordpress.test/upcoming-events/?format_type=online`. That part works.

The other main part of this is to redirect the user to that type of URL when they select filters in the `query-filters` block. That currently isn't working, the filters still get added to the URL as encoded key-value pairs, because they're being created as  `input` fields. 

One solution would be so enqueue some JS from this theme, and update the `<form action="...">` on the fly when the user selects a value from an instance of the block, and also remove the `<input>` fields from the form on submit. That doesn't feel very elegant, though.

Another solution would be to modify the `query-filters` block directly, but I think that'd add tech debt.

This is related to previous work around treating the arbitrary event data work as `WP_Query` results. See  https://github.com/WordPress/wordcamp.org/pull/1124 and https://github.com/WordPress/wordcamp.org/pull/1156. 

@StevenDufresne , @ryelle: i'm curious what you think about this.

iandunn:
Ah, I see now that [Showcase submits to the URL with key-value pairs, then redirects to the "pretty" one](https://github.com/WordPress/wporg-showcase-2022/blob/2f59d8230bdfb75f3b99f247c72b29a4c93abb0a/source/wp-content/themes/wporg-showcase-2022/functions.php#L609-L635). That should work here too 👍🏻

PR #1239, opened 2024-02-02
Latest Site Hint: Don't hint when the year is the same.

dd32:
I've created an additional site for Canada 2024, such that they can have an English and French site. See https://wordpress.slack.com/archives/C08M59V3P/p1706836174954409

However, this has caused the latest site hint to suggest this:
> WordCamp Canada 2024 (FR) is over. Check out the next edition!

This change updates the code to strip the `-identifier` part of `/2024-fr/` from the url for the comparison.

1. Visit https://canada.wordcamp.org/2024-fr/
2. Notice the banner directing to  https://canada.wordcamp.org/2024/
3. Apply PR
4. Repeat 1 & 2, but the banner is no longer visible.
5. Check an older site, for example, https://montreal.wordcamp.org/2019-fr/ still shows the banner.

### DIFF
--- a/phpunit-database-testcase.php
+++ b/phpunit-database-testcase.php
@@ -20,6 +20,7 @@ class Database_TestCase extends WP_UnitTestCase {
 	protected static $slash_year_2016_site_id;
 	protected static $slash_year_2018_dev_site_id;
 	protected static $slash_year_2020_site_id;
+	protected static $slash_year_2020_dev_site_id;
 	protected static $yearless_site_id;
 	protected static $slash_year_with_yearless_site_id;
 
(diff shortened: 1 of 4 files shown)


dd32:
Just noting; I didn't get a chance to add a unit test today, I'll follow up with one next week

dd32:
Github actions are completely broken due to unrelated changes, and I can't run unit tests locally due to incompatibilities in how this codebase is setup.

I can't duplicate the failing unit test outside of the unit tests.

pkevan:
I tried to fix the tests, but it's going to be stuck between a unit test fail and a static analysis fail atm!

pkevan:
Having trouble testing this PR - both locally and on a sandbox - will continue looking later.

pkevan:
Testing the PR works, just can't figure out why the unit tests are failing

pkevan:
Did this another way using a [toggle](https://github.com/WordPress/wordcamp.org/pull/1260), since there appears to be a few edges cases (probably more too).

Issue #1075, opened 2023-10-02
Custom fonts are not loaded correctly in block themes


ryelle:
In #868, I added code to register google fonts loaded in the "Appearance > Fonts" setting into the editor using the Fonts API. Something must have changed from gutenberg 16.6 to 16.7, because they aren't loaded correctly anymore. The Fonts API was experimental to begin with, and it has seen activity in this cycle, so it likely just needs to be updated to use the current API.

**To reproduce**

View [us.wordcamp.org/2023/](https://us.wordcamp.org/2023/).

Or to reproduce yourself:

1. Use a block theme like Twenty Twenty-Three
2. Add some google fonts to the Appearance > Fonts page
3. Create a page and add a block that supports font styles (eg, heading, paragraph)
4. Enable the font family controls in Typography
5. Only the core fonts appear

**Expected behavior**

The fonts that were added in Appearance > Fonts should be in this list, so that they can be used in the editor.

**Screenshots**

| Gutenberg 16.6, working | Gutenberg 16.7, not working |
|---|---|
| ![working](https://github.com/WordPress/wordcamp.org/assets/541093/35dd2de6-aee1-4b54-91e5-51504ebca99d) | ![broken](https://github.com/WordPress/wordcamp.org/assets/541093/66e7c866-8f18-496f-b89a-37869f750750) |
| <img width="293" alt="Screenshot 2023-10-02 at 2 27 14 PM" src="https://github.com/WordPress/wordcamp.org/assets/541093/0c6440e9-f060-4e87-9056-9011f6efade3"> | <img width="282" alt="Screenshot 2023-10-02 at 2 26 23 PM" src="https://github.com/WordPress/wordcamp.org/assets/541093/e3737d60-5f32-4bff-aca2-a275d13db8b9"> |

iandunn:
I'm seeing these locally:

```sh
> wp site list --network_id=2
Deprecated: Function wp_fonts is <strong>deprecated</strong> since version Gutenberg 16.3 with no alternative available. in wp-includes/functions.php on line 5463

Deprecated: Function WP_Fonts::__construct is <strong>deprecated</strong> since version Gutenberg 16.3.0 with no alternative available. in wp-includes/functions.php on line 5463

Deprecated: Function WP_Fonts::register_provider is <strong>deprecated</strong> since version Gutenberg 16.3.0 with no alternative available. in wp--includes/functions.php on line 5463
```

[...]

PR #1245, opened 2024-02-09
Store custom fonts in the upload directory.

dd32:
This is the first step forward for #1075, this change should allow WordCamps to install custom fonts themselves.

With this in place, the installed fonts end up with a URL like `https://example.wordcamp.test/2023/files/fonts/UMBCrPdDqW66y0Y2usFeQCH18mulUxBvI9p7TqbCHJ8BRq0b.woff2`

This doesn't support the `wc-fonts` `@import` functionality, and I can't immediately see a way to make those fonts available in the Editor; which is what #868 was doing.

See #1075

### DIFF
--- a/public_html/wp-content/plugins/wc-fonts/wc-fonts.php
+++ b/public_html/wp-content/plugins/wc-fonts/wc-fonts.php
@@ -22,9 +22,6 @@ class WordCamp_Fonts_Plugin {
 		// Temporary workaround until we can use the core font library on WordCamp.org.
 		// See https://github.com/WordPress/gutenberg/pull/57697.
 		add_filter( 'wp_theme_json_data_theme', array( $this, 'inject_fonts_theme_json' ) );
-
-		// Store fonts in the WordCamp uploads.
-		add_filter( 'font_dir', array( $this, 'font_dir' ) );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


public_html/wp-content/plugins/wc-fonts/wc-fonts.php
```diff
@@ -22,6 +22,9 @@ public function __construct() {
 		// Temporary workaround until we can use the core font library on WordCamp.org.
 		// See https://github.com/WordPress/gutenberg/pull/57697.
 		add_filter( 'wp_theme_json_data_theme', array( $this, 'inject_fonts_theme_json' ) );
+
+		// Store fonts in the WordCamp uploads.
+		add_filter( 'font_dir', array( $this, 'font_dir' ) );
```
ryelle: Maybe this should be in [gutenberg-tweaks.php](https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/mu-plugins/gutenberg-tweaks.php) instead?

ryelle:
> This doesn't support the wc-fonts @import functionality, and I can't immediately see a way to make those fonts available in the Editor; which is what https://github.com/WordPress/wordcamp.org/pull/868 was doing.

At the time, core didn't have a UI for selecting fonts, it was only supported via code - so I hijacked the wc-fonts UI. Now that core does, I think it makes more sense to just use that. Seems to be what a few people have been doing already.

ryelle (commented):
I don't know if this is ready for testing, but I tried on my sandbox (with the uploads workaround from dev-docs), and got this error:

> No font faces were installed. Fetch error: Unable to create directory fonts. Is its parent directory writable by the server? Fetch error: Unable to create directory fonts. Is its parent directory writable by the server?

dd32:
> but I tried on my sandbox (with the uploads workaround from dev-docs), and got this error:

I also ran into that, but I don't think it should be run into in production, as core should be attempting to create the `wp-content/fonts` directory anyway.. I think that error only happens when the temporary folder doesn't exist yet (ie. /tmp/uploads doesn't exist)

I patched around it, and then removed it thinking it probably wasn't needed. (It was an additional `wp_mkdir_p()` call)

public_html/wp-content/plugins/wc-fonts/wc-fonts.php
dd32: I'm tempted to agree. I'm going to move it over there and merge it, see if it works in production.

dd32:
> but I tried on my sandbox (with the uploads workaround from dev-docs), and got this error:

Ahhh I think because the devdocs instructions doesn't actually change the upload directory, just where the file ends up on disk, such that existing uploads continue to be resolved without causing everything to 404.

dd32:
Testing has confirmed that this is working.

https://testing.wordcamp.org/2022/ has installed a font, and used it on the front-end.

Issue #1255, opened 2024-02-26
Stripe checkout sessions may timeout, with valid transaction


dd32:
**Describe the bug**

As a follow up to #1227, it's possible for a ticket purchaser to purchase a ticket on the stripe side, but not have a ticket assigned to them on WordCamp.org.

**Why does this happen?**
The stripe flow is as such:
1. Click Purchase
2. Setup attendee, set status as not-paid/draft
3. Redirect to Stripe
4. Pay on Stripe
5. Redirect back to WordCamp
6. Set attendee status
7. Everyone is happy.

If step 5/6 doesn't happen, 7 doesn't happen.

The previous Stripe flow was something like this:
1. Capture authorization
2. Submit form
3. setup attendee
4. Capture payment & set status
5. Everyone is happy

Steps 2~4 would happen in the same HTTP request, so there was no potential for a transaction being missed.

**Potential fixes**
In order to combat this, we could do one of...
 - When a Stripe checkout session begins, start running a cron task to check current sessions for non-completed tickets, looking for a completed transaction attached
 - Implement the Stripe Webhook system, so that we get a webhook request when the ticket purchase completes
 - When the ticket is to go to timeout, ensure that there exists no stripe transaction for the checkout.

Using Stripe Webhooks is the obvious choice. The paypal integration uses something similar.

**To reproduce**

1. Attempt to purchase a ticket
9. Click on Purchase; redirect to Stripe
10. *prevent requests to $city.WordCamp.org from your browser*
11. Complete the transaction on Stripe, watch the failed network requests after redirect back to us.
12. See ticket still in draft

This is an example of such a ticket: https://vienna.wordcamp.org/2024/wp-admin/post.php?post=3074&action=edit

dd32:
> Implement the Stripe Webhook system, so that we get a webhook request when the ticket purchase completes


The way the Paypal code works, is that the `notify_url` is set, which pings the individual WordCamp site with the webhook/IPN.

Stripe seems to only have global webhooks, which means that the webhook ping wouldn't be site-specific.

That means we'd have to have a generic webhook endpoint that then switched to the appropriate site before processing it.
_(edit: Probably can't switch, so it'd be a self-call like `Stripe -> Generic WordCamp webhook -> self-request to WordCamp-specific webhook API`.)_

> When the ticket is to go to timeout, ensure that there exists no stripe transaction for the checkout.

#1256 is a PR that implements this option; which might be good enough, depending on how often these missed transactions occurs.

[...]

Issue #1227, opened 2024-01-26
Update to latest Stripe Checkout


iandunn:
There's a new version of Checkout, and folks have started reporting problems w/ the old one, like payments failing because they don't support 3DS.

* https://stripe.com/docs/payments/checkout/migration#api-products
* https://stripe.com/docs/payments/accept-a-payment?integration=checkout
* https://stripe.com/docs/payments/checkout/migration#parameter-conversion

The places we need to update to use the new Checkout flow are:

- [ ] sponsorship payment 
- [ ] camptix stripe addon

We could update `WordCamp\Utilities\Stripe_Client` to use the new Checkout flow (create a session ID and redirect to their stripe.com), or we could switch to using [the official client](https://github.com/stripe/stripe-php/) instead.

Either way, it may be best to replace `CampTix_Stripe_API_Client` with the client chosen above, rather than updating and maintaining a 2nd custom client. If we stick with `WordCamp\Utilities\Stripe_Client`, then we'll need to port some of the functionality over, like refunds.

PR #1256, opened 2024-02-26
Stripe: Check if transaction actually completed before setting as timeout.

dd32:
This PR attempts to check if a stripe transaction was actually processed, before an attendee is moved from `draft` to `timeout`.

See #1255 & #1227

_(This is untested so far)_

1. Make a transaction
2. Either trigger #1255 OR change the status of the attendee to `draft`
3. Either wait for the timeout, or trigger `do_action( 'camptix_pre_attendee_timeout', $id )`
4. Check the attendee is now in a paid state, and got the emails.


pkevan (approved):
Code looks good - will test to see it works.

pkevan:
Not forgotten about this one - will test tomorrow!

pkevan:
Tested this earlier, but struggled to manage to create a fake timeout for some reason. I suspect it will be fine to release, and see if it makes things better.

PR #1260, opened 2024-03-01
Add/latest site hints toggle

pkevan:
Allows the latest site hint to be turned off per site.

Props ryelle


pkevan:
Tried to add unit tests, but realised that the sites aren't actually created, so it's not possible to test the feature.

PR #1262, opened 2024-03-11
Camptix: Harden against junk input

dd32:
Pentesters on the Camptix checkout can cause a high number of warnings by providing junk input to many attendee fields.

Such as:
```
E_WARNING trim() expects parameter 1 to be string, array given
https://example.wordcamp.org/2020/tickets/?tix_action=checkout
plugins/camptix/camptix.php:7079

E_WARNING array_key_exists(): The first argument should be either a string or an integer
URL https://example.wordcamp.org/2022/tickets/?tix_action=checkout
File plugins/camptix/camptix.php:7068

E_WARNING strip_tags() expects parameter 1 to be string, array given
URL https://example.wordcamp.org/2020/tickets/?tix_action=checkout
File plugins/camptix/camptix.php:7099
```

This PR applies some sanitization and validation to the attendee data before processing.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -7065,6 +7065,9 @@ class CampTix_Plugin {
 		foreach( (array) $_POST['tix_attendee_info'] as $i => $attendee_info ) {
 			$attendee = new stdClass;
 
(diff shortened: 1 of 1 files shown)


pkevan (approved):
Thanks @dd32 - the changes look good, and think it's worth merging these to help reduce the amount of warnings

PR #1264, opened 2024-03-11
Close ticket sales automatically after the event finishes

### LINKED ISSUE #1263: Tickets set to "Auto availability" should not be available for purchase after an event has ended
When a ticket is set to auto-availability, or at least without an end date, the ticket will be purchasable forever.

Ideally:
 - The ticket should not be able to be purchased after the event has finished.
 - The ticket (even if an end date of 2038-01-01 is set) should not be possible to purchase once the WordCamp has "closed" (The internal WordCamp status for "This event is finalised").

Screenshot 2024-03-11 at 4 13 42 PM

dd32:
When the WordCamp is over according to the WordCamp post, ticket sales should be closed.

Fixes #1263

| Before | After |
| --- | --- |
|  Screenshot 2024-03-11 at 4 28 32 PM | Screenshot 2024-03-11 at 4 30 22 PM |

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -6803,26 +6803,51 @@ class CampTix_Plugin {
 			$post = get_post( $post );
 		}
 
-		if ( ! $post ) return false;
-		if ( $post->post_type != 'tix_ticket' ) return false;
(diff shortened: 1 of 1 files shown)


outdoor2kode (approved):
Tested and it works as described.

PR #1266, opened 2024-03-13
Check against post meta

pkevan:
Change check to use the post meta retrieved above, rather than $session_id, which doesn't exist.

From PHP notice:

```
Undefined variable: session_id
URL
https://europe.wordcamp.org/2024/wp-cron.php?doing_wp_cron=1710337138.6555919647216796875000
File
[..]plugins/camptix/addons/payment-stripe.php:649
```


pkevan:
Fixed in https://github.com/WordPress/wordcamp.org/commit/3bf5c0b836bf42c44388b07fca0a5144424fd479

PR #1271, opened 2024-03-18
fix: Update camptix code to resolve the issue of Stripe Payment Inten…

### LINKED ISSUE #1270: Using `[` in ticket names created with Camptix causes an error with the Stripe API.
**Describe the bug**

When using Stripe payment for ticket purchases in Camptix, if the ticket name contains `[`, an error occurs in the Stripe API response. This is because the Stripe API specification interprets the string inside `[]` as an object key. To resolve this issue, the `clean_metadata` function needs to replace `[]` with another character like `()`.

**To reproduce**

- Create a new ticket and include the `[` string, e.g., `6/2 – Micro Sponsor [ 10USD ]`
- Enable Camptix Stripe Payment (it's okay to use sandbox mode)
- Try to order this ticket from the Ticket page
- You will encounter an error when creating a new Stripe Checkout Session
- 
**Expected behavior**

The ticket name needs to be included in the metadata of the Stripe Payment Intent object when creating Stripe Checkout Sessions API.
e.g.: `$args['payment_intent_data']['metadata']['6/2 [Sponsor ticket]'] = '1';`

**Expected Behavior
The ticket name needs to be included in the metadata of the Stripe Payment Intent object when creating Stripe Checkout Sessions API.
e.g.: `$args['payment_intent_data']['metadata']['6/2 [Sponsor ticket]'] = '1';`

## Actual Behavior**

Due to the Stripe API specification, a text inside the `[]` will be converted to an object key like the following:

```
$args['payment_intent_data']['metadata']['6/2 [Sponsor ticket]'] = '1';
->
{
  "metadata": {
    "6/2": {
      "Sponsor ticket": "1"
    }
  }
}
```
**Suggestion**
In the `clean_metadata` function, we may need to replace the `[]` string with another character like `()` instead.

```php
$key = $this->trim_string( $key, 40, '' );
$key = str_replace( '[', '(', str_replace( ']', ')', $key ) );
```

When we avoid passing the `[]` string as the key of the metadata, the Stripe API request will work correctly.

This idea seems reasonable, and if you approve, I'll proceed with creating a new pull request for this.

**Screenshots**

スクリーンショット 2024-03-18 16 38 23

**WordCamp**

If this is a problem on a specific WordCamp's site, list the site or page URL here.

At the WordCamp Kansai 2024, we suffered from this error:
https://kansai.wordcamp.org/2024/wp-admin/post.php?post=3772&action=edit

スクリーンショット 2024-03-18 16 39 59

 - Device: [e.g. laptop, iPhone6]

**Security Issues**

To report a security issue, please visit the WordPress HackerOne program: https://hackerone.com/wordpress.

hideokamoto:
It's a patch for this issue: #1270 .

When calling the Stripe API, including a string with '[' in the metadata key can cause unexpected behavior.
Therefore, I added a process to replace '[' with '('.

Fixes #1270

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/payment-stripe.php
+++ b/public_html/wp-content/plugins/camptix/addons/payment-stripe.php
@@ -994,6 +994,7 @@ class CampTix_Stripe_API_Client {
 
 			// Trim the key to 40 chars.
 			$key = $this->trim_string( $key, 40, '' );
+			$key = str_replace( [ '[' , ']' ], [ '(', ')' ], $key );
 
 			// Trim the val to 500 chars.
 			$val = $this->trim_string( $val );
--- a/public_html/wp-content/plugins/camptix/addons/payment-stripe.php
+++ b/public_html/wp-content/plugins/camptix/addons/payment-stripe.php
@@ -992,11 +992,11 @@ class CampTix_Stripe_API_Client {
 				return $cleaned;
 			}
 
-			// Trim the key to 40 chars.
-			$key = $this->trim_string( $key, 40, '' );
-
 			// Remove unsupported characters from the key.
 			$key = str_replace( array( '[', ']' ), '', $key );
+			
+			// Trim the key to 40 chars.
+			$key = $this->trim_string( $key, 40, '' );
 
 			// Trim the val to 500 chars.
 			$val = $this->trim_string( $val );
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/camptix/addons/payment-stripe.php
```diff
@@ -994,6 +994,7 @@ protected function clean_metadata( $metadata = array() ) {
 
 			// Trim the key to 40 chars.
 			$key = $this->trim_string( $key, 40, '' );
+			$key = str_replace( array( '[', ']' ), array( '(', ')' ), $key );
```
dd32:
```suggestion

			// Remove unsupported characters from the key.
			$key = str_replace( array( '[', ']' ), '', $key );
```

As done with the payment descriptor, I think we should just strip these characters instead of replacing them.

PR #1273, opened 2024-03-22
Various notice fixes

dd32:
As usual, various code-formatting and notice fixes caused by pentester and search engine requests.

These are:
```
E_WARNING: Invalid argument supplied for foreach() in plugins/camptix/camptix.php:5260

E_NOTICE Undefined index: 'Is this the first time you have volunteered at a WordPress event?' in plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php:409
```

Code formatting was removing short-if's and converting `isset()` to `!empty()` when we expect a value, and removing cases where both were checked needlessly.

1.
2.
3.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -5169,8 +5169,9 @@ class CampTix_Plugin {
 		}
 
 		// Allow [camptix attr="value"] but not [camptix_attendees] etc.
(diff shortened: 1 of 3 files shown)


dd32:
:grumble: :grumble: After [58e0481](https://github.com/WordPress/wordcamp.org/pull/1273/commits/58e048152b2af2f78a805afb7bd67d1dc3c92f1b)
>  109 | ERROR | Using short ternaries is not allowed as they are rarely used correctly

IMHO this rule should be removed..  but I've removed the ternaries instead, since the null operator wasn't doing anything anyway.

PR #1275, opened 2024-03-28
test: add unit test for the camptix stripe addon

hideokamoto:
Add a new PHPUnit test case for the class `Camptix_Payment_Stripe_Addon` on the Camptix plugin.

1.  Run the `phpunit` command
2. If it's not failed, it's okay.

### DIFF
--- a/public_html/wp-content/plugins/camptix/tests/addons/test-payment-stripe.php
+++ b/public_html/wp-content/plugins/camptix/tests/addons/test-payment-stripe.php
@@ -14,7 +14,7 @@ class Test_Camptix_Payment_Stripe_Addon extends \WP_UnitTestCase {
 				'USD', 10, 1000, // 10USD should be 1000
 			),
 			array(
-				'EUR', 10, 1000, // 10USD should be 1000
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/camptix/tests/addons/test-payment-stripe.php
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Provide a test case for the function "CampTix_Payment_Method_Stripe->get_fractional_unit_amount".
+	 **/
+	public function currencyAmountProvider() {
+		return array(
+			array(
+				'USD', 10, 1000, // 10USD should be 1000
+			),
+			array(
+				'EUR', 10, 1000, // 10USD should be 1000
```
pkevan: comment should reference EUR

pkevan (approved):
Left a minor comment, which I can fix up in a followup.

Unsure if this will automatically be added to phpunit (through github actions), but again, minor and can be rectified.

pkevan:
> Unsure if this will automatically be added to phpunit (through github actions)

runs as expected with the additional tests against this PR.

PR #1229, opened 2024-01-29
Try using MariaDB LTS to reduce need for Apple Silicon workarounds.

dd32:
Tested on Apple M1 Max

Fixes #, See #

Props <username>, <username>

### How to test the changes in this Pull Request:

1.
2.
3.

PR #1279, opened 2024-04-02
Use Oracle Linux mysql for arm64 support

dd32:
See https://github.com/WordPress/wordcamp.org/pull/1261 https://github.com/WordPress/wordcamp.org/pull/1229 for background

It seems that didn't actually work as expected; as although it's now using Mysql 8, it's not actually using a arm64 image.

This PR changes it to use the oracle linux version of the mysql docker image, which does support arm64. The debian version doesn't.

Props @renintw for pushing me to look at this again

### DIFF
--- a/.docker/Dockerfile.mysql
+++ b/.docker/Dockerfile.mysql
@@ -1,9 +1,8 @@
 FROM mysql:8.0
 
-RUN apt-get update \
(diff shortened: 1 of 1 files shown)


outdoor2kode (approved):
Thanks @dd32

PR #1295, opened 2024-04-23
Verify the order before checking for error flags

dd32:
While looking at the checkout process as part of #1296 I noticed that `verify_order()` is run after the check for error_flags, which means if `verify_order()` flags any issues with the order, the checkout still proceeds.

Some of the checks within the `verify_order()` method is duplicated within the checkout process already, but this ensures that any issues encountered during validation do properly stop the checkout process.

This might reduce the number of `ticket_excess` issues that some WordCamps can run into during high ticket loads.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -7172,12 +7172,12 @@ class CampTix_Plugin {
 		if ( ! is_email( $receipt_email ) )
 			$this->error_flags['no_receipt_email'] = true;
 
+		$this->verify_order( $this->order );
+		
 		// If there's at least one error, don't proceed with checkout.
 		if ( $this->error_flags ) {
 			return $this->form_attendee_info();
 		}
-
-		$this->verify_order( $this->order );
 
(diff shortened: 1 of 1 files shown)


dd32:
This is less problematic than I originally thought.

`verify_order()` is being called multiple times.

During checkout; verify_order() is called:
https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/camptix.php#L5296-L5299

Errors from that instance of verify_order() are respected.

Then form_checkout() is run
https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/camptix.php#L5386-L5387

which then calls verify_order again:
https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/camptix.php#L7175-L7180

In that 2nd instance of verify_order running, it'll not respect any new errors thrown by verify_order.. but there shouldn't be any at that point, so the errors are being ignored, because there's unlikely to be any errors at all.

As a result, this is fairly safe to merge I think, as nothing should be changing, since the check is already happening.. but this is also now a low-priority thing

PR #1296, opened 2024-04-23
Try caching get_remaining_tickets() in APCu to allow tickets to be 'reserved' during checkout, avoiding race condition

dd32:
**Draft PR** due to inability to properly test this right now, and unsure if it's the proper direction forward or worth it.

Ref: https://wordpress.slack.com/archives/C08M59V3P/p1712066063709249

WordCamp Sylhet ran into an issue where WordCamp tickets were oversold, due to a race condition during the checkout process.

The ticket are counted on every pageload, and in the event of a few hundred people attempting to checkout at the same instant, resulted in multiple tickets inserting tickets after the initial `do we have available tickets` check passed.

This PR attempts to add some caching/locking using APCu to the process.

This appears to work at first glance, but is not entirely complete.
 - failed payments locks a ticket for up-to 10 minutes (possibly a good thing?) (fixable)
 - changing the available tickets number doesn't reflect for the up-to 10 minutes (fixable)
 - Hard to test due to low requests/second sandbox throughput

While writing this PR, I've noticed something that was rather odd.
It appears that **some** of the order verification is checked after the check for errors, resulting in some of the validation being skipped during checkout, although it seems duplicative.. ba27a3a791b030fa7cc2813be53533deaffda295 moved that and I've split it into #1295

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -6838,6 +6838,21 @@ class CampTix_Plugin {
 	 */
 	function get_remaining_tickets( $post_id, $via_reservation = false ) {
 		$remaining = 0;
+		$apcu_key  = 'remaining_tickets:' . get_current_blog_id() . ':' . $post_id . ':' . ( $via_reservation ?: '' );
+		if ( function_exists( 'apcu_enabled' ) && apcu_enabled() ) {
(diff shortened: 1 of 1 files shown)


dd32:
I'm closing this, as I don't think this actually resolved anything. I suspect something like #1297 would be better suited.

PR #1320, opened 2024-05-15
Treat UGX as a decimal based currency.

dd32:
Stripe documents several currencies as Zero-decimal currencies,  these currencies should be charged as “$1 = 1” in the stripe API.

https://docs.stripe.com/currencies#zero-decimal

However, for backwards compatibility, specific zero-decimal currencies must be treated as decimal currencies, this includes UGX and ISK (which is not documented as a zero-decimal in the above documentation)

This means that when we've attempted to charge `1,000 UGX` we've actually charged `10.00 UGX = 10 UGX`.

1. Create a UGX Camptix ticket
2. Charge it, ensure that the ultimate Stripe charge is correct

or
1. Observe Stripe docs, make change :)

### DIFF
--- a/public_html/wp-content/mu-plugins/utilities/class-stripe-client.php
+++ b/public_html/wp-content/mu-plugins/utilities/class-stripe-client.php
@@ -191,7 +191,7 @@ class Stripe_Client {
 		$currency_multipliers = array(
 			// Zero-decimal currencies.
 			1   => array(
-				'BIF', 'CLP', 'DJF', 'GNF', 'JPY', 'KMF', 'KRW', 'MGA', 'PYG', 'RWF', 'UGX', 'VND', 'VUV', 'XAF',
+				'BIF', 'CLP', 'DJF', 'GNF', 'JPY', 'KMF', 'KRW', 'MGA', 'PYG', 'RWF', 'VND', 'VUV', 'XAF',
 				'XOF', 'XPF',
 			),
 			100 => array(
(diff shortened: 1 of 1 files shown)


outdoor2kode:
Observations shared by @dd32:

UGX is a zero-decimal currency in stripe.
https://docs.stripe.com/currencies#zero-decimal
> For zero-decimal currencies, still provide amounts as an integer but without multiplying by 100. For example, to charge ¥500, provide an amount value of 500.

We provided Stripe with a value of 380000 UGX
Stripe then charged 3800.00 UGX.

But this is a bug on our end… because although Stripe documents UGX as a zero-decimal currency, and states you should pass it without multiplying by 100, further down the page it has a special notes section for some currencies….

https://docs.stripe.com/currencies#special-cases
> UGX was a decimal-based currency, but is now effectively a zero-decimal currency. To maintain backwards compatibility, you must pass in amounts with two decimals. For example, to charge 5 UGX, provide an amount value of 500.

Yup.. so we should’ve told them to charge 380000**00**..

> Effective 0:00 UTC on 2023-04-14, ISK becomes a zero-decimal currency. To maintain backwards compatibility, you must pass in amounts with two decimals

ISK too.. except ISK isn’t listed as a Zero-decimal currency…

[https://github.com/WordPress/wordcamp.org/blob/ddc9cad237a4d57e092455d60cb69e0622e[…]ic_html/wp-content/mu-plugins/utilities/class-stripe-client.php](https://github.com/WordPress/wordcamp.org/blob/ddc9cad237a4d57e092455d60cb69e0622e579b2/public_html/wp-content/mu-plugins/utilities/class-stripe-client.php#L193-L196)
This has also affected WooCommerce Stripe previously: https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2674

outdoor2kode (approved):
Thank you Dion for the detailed observation and the quick fix, it's awesome..

PR #1353, opened 2024-08-01
Making the "$wordcamp is over. Check out the next edition" text translateable.

unapersona:
Some strings in the wordcamp.org sites are not translateable. 

This PR make the "$wordcamp is over. Check out the next edition", visible in all past WordCamps, translateable.

### Screenshots
imagen

### DIFF
--- a/public_html/wp-content/mu-plugins/latest-site-hints.php
+++ b/public_html/wp-content/mu-plugins/latest-site-hints.php
@@ -127,9 +127,9 @@ function show_notification_about_latest_site() {
 
 	echo '<div class="wordcamp-latest-site-notify"><p>' .
 		wp_sprintf(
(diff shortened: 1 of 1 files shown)


pkevan:
we probably want to wrap this in `wp_kses` to avoid anything potentially malicious being added to this. It might be why there wasn't any translations added as it isn't straightforward.

dd32:
> we probably want to wrap this in `wp_kses` to avoid anything potentially malicious being added to this. It might be why there wasn't any translations added as it isn't straightforward.

Generally WordPress (&WordPress.org) trusts translations and relies upon our filters / warnings in translate.wordpress.org for this kinda stuff. There's plenty of "unsafe" `__()`'s and `_e()`'s around the WordCamp codebase already.

Issue #1367, opened 2024-09-11
Please stop sending the automated survey emails to WP Events (former next-gen)


peiraisotta:
Following up on [this question](https://github.com/WordPress/wordcamp.org/issues/1345#issuecomment-2304607121) on issue 1345.

Currently once a WP Event (former next gen) ends, attendees receive an automated email inviting them to fill a post event survey. 

As we've created a new survey for all events attendees (WordCamp + WP Event former next-gen) please disable the current emails sent to WP Event attendees as soon as possible. The next former next-gen event is schedule on September 28.

Thank you so much.

peiraisotta:
Sorry, I forgot I already created this issue for that: https://github.com/WordPress/wordcamp.org/issues/1363 you can deprecate this one.

peiraisotta:
Sorry, I'm seeing now that I already created an [issue](https://github.com/WordPress/wordcamp.org/issues/1363) for this. Please feel free to close or deprecate this request.

PR #1368, opened 2024-09-12
Fix/delete attendee survey

StevenDufresne:
This PR deletes the `camptix-attendee-survey` plugin as it's no longer used.

Related: https://github.com/WordPress/wordcamp.org/issues/1367
Original PR: https://github.com/WordPress/wordcamp.org/pull/997


pkevan:
Presume this is ready for review? Just still in draft?

pkevan:
Preferably when deleting, we should remove them in two commits, one for the activation stuff in https://github.com/WordPress/wordcamp.org/pull/1368/files#diff-c8e4e2f5dfad9d832ecf6d9b10ed8ad38c3f97d3b90f6d41a18a35aa4ba3d525, then a second for the unused code.

While not a huge issue on WordCamp, it is better to do it this way to avoid unnecessary fatals in the deploy process.

pkevan (approved):

pkevan (approved):

PR #1384, opened 2024-10-04
Sunrise: Redirect domain to latest year site that is not deleted.

dd32:
https://wordpress.slack.com/archives/C08M59V3P/p1728017732714399

As reported here, accessing the root-domain for kolkata is redirecting to a incorrect URL, this happens as a duplicate 2025 site was accidentally created with a space after it:

```
$ curl -I https://kolkata.wordcamp.org/
HTTP/1.1 302 Found
....
Location: https://kolkata.wordcamp.org/2025/%20/
```

After this change, it correctly references the non-deleted site:

```
$ curl -I https://kolkata.wordcamp.org/
HTTP/1.1 302 Found
...
Location: https://kolkata.wordcamp.org/2025/
```

### DIFF
--- a/public_html/wp-content/sunrise-wordcamp.php
+++ b/public_html/wp-content/sunrise-wordcamp.php
@@ -636,8 +636,12 @@ function get_latest_site( string $domain ) {
 		SELECT `blog_id`, `domain`, `path`
 		FROM $wpdb->blogs
 		WHERE
-			( domain =    %s AND path != '/' ) OR -- Match city/year format.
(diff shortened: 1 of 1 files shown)


dd32:
Notes
 - Some deleted sites (deleted = 1) are also public (public=1)
 - Some not-public sites are not-deleted (for example testing.wordcamp.org)
 - There's also archived, which are a mix of public=0 and deleted=1

That's why I've included both public and deleted here, which seems to cover the correct subset of sites that shouldn't be redirected to.

PR #1394, opened 2024-10-11
WC Post Types: Fix timezone offset when sessions happen after DST

### LINKED ISSUE #1385: Session time input changes time unexpectedly
The session time input is not working correctly, something is causing it to jump around while entering times. This appears to only happen if the session is after daylight saving time ends.

In the video, you can see the number is skipping down by two despite me only hitting the down arrow once. Then I hit up, and it doesn't always change. Moving over to toggle AM/PM, the time also changes. Back to the input, entering 12 automatically updates to 11.

https://github.com/user-attachments/assets/935180e9-2bfa-4434-b106-ac7459d7f2c0

Video recorded on MacOS + firefox.

Thanks to platcatheo for [reporting this in Slack](https://wordpress.slack.com/archives/C08M59V3P/p1728074980106939).

ryelle:
This change updates how the session dates are converted when the site and client timezones are different. Rather than creating a fixed timezone offset on page load, this converts the times into the server timezone for the given date — this way, the offset is correctly applied if a future date is in a different timezone (DST starts or ends).

Fixes #1385.

The `date-fns` libraries added are also used in core/gutenberg, though they are bundled in the built script, not as externals. Since this in the editor, and it's not a huge library (not like moment), I think this is okay. Once an admin is editing sessions, it'll be cached from page to page.

1. Set your site's timezone to a location with DST, for example Los Angeles or New York (different than your current timezone).
2. DST for those locations ends on Nov 3rd, so try setting the time on a session in December.
3. The session date/time UI should have a label with the site's timezone, and the time is in that timezone.
4. It should work, there should be no glitching or skipping numbers.
5. Once saved, if you view the session in the list table, it should be the same - also in the site timezone.

Try with the site's timezone the same as your own & no during DST to ensure no regressions.

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/js/src/components/date-control/index.js
+++ b/public_html/wp-content/plugins/wc-post-types/js/src/components/date-control/index.js
@@ -1,10 +1,17 @@
 /**
+ * External dependencies
+ */
+import { format } from 'date-fns';
+import { TZDate } from '@date-fns/tz';
+
+/**
  * WordPress dependencies
  */
-// eslint-disable-next-line @wordpress/no-unsafe-wp-apis -- Date settings OK.
-import { __experimentalGetSettings, dateI18n } from '@wordpress/date';
+import { getSettings } from '@wordpress/date';
 import { BaseControl, TimePicker } from '@wordpress/components';
 import { sprintf } from '@wordpress/i18n';
+
+const TIMEZONELESS_FORMAT = "yyyy-MM-dd'T'HH:mm:ss";
 
 /**
  * Using the site settings, generate the offset in ISO 8601 format (`+00:00`).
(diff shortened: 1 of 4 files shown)


public_html/wp-content/plugins/wc-post-types/js/src/components/date-control/index.js
```diff
@@ -35,20 +42,44 @@ export default function( { date, label, onChange } ) {
...
+
+	// First get the server timezone from date settings. This could be a string
+	// like "America/New_York" or an offset like "-1".
+	let serverTimezone = settings.timezone.string;
+	if ( ! serverTimezone ) {
+		// If it's a offset, convert it to ISO 8601 format.
+		serverTimezone = getTimezoneOffset( settings.timezone );
+	}
+
+	// Because this date is a timestamp, it's understood to be a UTC value, and
```
outdoor2kode, line 56: 👍

outdoor2kode (approved):
LGTM, thanks for the fix!

Issue #1397, opened 2024-10-17
WordCamp Login Change - UI Improvement Needed Urgently


aashishpahwa:
The recent changes of WordCamp making it mandatory to create an account for buying tickets is hampering ticket sales greatly.There are several first-timers too and people don’t want to take an extra step to get the ticket.  Is there a way to go away with this requirement?
Tickets sales have slumped after this.

The UI is not optimized as well. There's just a paragraph text to tell them to login, not even a login button or CTA.Several potential attendees don't know how to proceed.

Here's how WordCamp Delhi's website looks to the visitors right now - 

![Image](https://github.com/user-attachments/assets/73340fce-bd7c-43d8-a6c5-7b7f9d7366b0)

There isn't a CTA now at all.

I'd suggest keeping the 'Register' CTA as it is and redirecting them to a login or signup page as an interstitial.

Currently, no CTA means people don't know how to proceed. 

![Image](https://github.com/user-attachments/assets/6b54ed0e-9e57-4e0f-9675-adb3c43aac6c)

Even the wordpress.org login page can be improved to cater to these first timers. It should tell them about it being one of the steps to buy the ticket - 

![Image](https://github.com/user-attachments/assets/5b4881e2-1567-453b-a625-c647e88bdf78)

Honestly, the change affected us and might hamper our promise to the sponsors.

markmemayank:
Hi @aashishpahwa great suggestion and this should be addressed at earliest.

GautamMKGarg:
Nice suggesion

kumaranup594:
@aashishpahwa @markmemayank create a small video and add top of that with thumbnail clearly mentioned - Please watch to buy ticket "something like this I mean".

adamwoodnz:
> Even the wordpress.org login page can be improved to cater to these first timers. It should tell them about it being one of the steps to buy the ticket

We customise the login form introduction copy when the request is from wordcamp.org

<img src="https://github.com/user-attachments/assets/36362be9-1f2e-46e2-9f69-afd32d164a0e" width="400">

We could likely do the same for the account creation form introduction copy.

adamwoodnz:
> I'd suggest keeping the 'Register' CTA as it is and redirecting them to a login or signup page as an interstitial.

Someone will be looking into this option, thanks for the suggestion.

aashishpahwa:
Thanks, team, for quickly solving the ticketing UX issue. But it still needs some work.

[...]

PR #1399, opened 2024-10-18
Improve the logged-out ticket registration flow

tellyworth:
This changes two things:

* On the ticket purchase screen, the UI to purchase tickets is restored for logged-out users
* When a logged out user clicks the Register button, they'll be taken through the login/user registration process, and land back on the main ticket selection page after completing that flow

This doesn't solve all issues to do with logged-out ticket purchases, but it's an improvement on the current situation.

The main drawback is that any selected tickets or coupons are not saved when redirecting through the login/signup process; the user will have to re-enter their selections. That could certainly be improved, but as a short-term improvement this might be ok while we do more testing and iteration.

See #1397

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/require-login.php
+++ b/public_html/wp-content/plugins/camptix/addons/require-login.php
@@ -14,9 +14,6 @@ class CampTix_Require_Login extends CampTix_Addon {
 	 */
 	public function __construct() {
 		// Registration Information front-end screen
-		add_filter( 'camptix_register_button_classes',                array( $this, 'hide_register_form_elements' ) );
-		add_filter( 'camptix_coupon_link_classes',                    array( $this, 'hide_register_form_elements' ) );
-		add_filter( 'camptix_quantity_row_classes',                   array( $this, 'hide_register_form_elements' ) );
 		add_action( 'camptix_notices',                                array( $this, 'ticket_form_message' ), 8 );
 		add_filter( 'camptix_ask_questions',                          array( $this, 'hide_additional_attendee_questions_during_checkout' ), 10, 5 );
 		add_filter( 'camptix_form_register_complete_attendee_object', array( $this, 'add_username_to_attendee_object' ), 10, 3 );
@@ -65,7 +62,7 @@ class CampTix_Require_Login extends CampTix_Addon {
 		}
 
 		if ( ! is_user_logged_in() && isset( $_REQUEST['tix_action'] ) ) {
-			wp_safe_redirect( wp_login_url( add_query_arg( $_REQUEST, $camptix->get_tickets_url() ) ) );
+			wp_safe_redirect( wp_login_url( $camptix->get_tickets_url() ) );
 			exit();
 		}
 		
--- a/public_html/wp-content/plugins/camptix/addons/require-login.php
+++ b/public_html/wp-content/plugins/camptix/addons/require-login.php
@@ -13,7 +13,7 @@ class CampTix_Require_Login extends CampTix_Addon {
 	 * Register hook callbacks
 	 */
 	public function __construct() {
-		// Registration Information front-end screen
+		// Registration Information front-end screen.
 		add_action( 'camptix_notices',                                array( $this, 'ticket_form_message' ), 8 );
 		add_filter( 'camptix_ask_questions',                          array( $this, 'hide_additional_attendee_questions_during_checkout' ), 10, 5 );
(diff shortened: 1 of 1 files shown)


pkevan:
> The main drawback is that any selected tickets or coupons are not saved when redirecting through the login/signup process; the user will have to re-enter their selections.

I wonder if we should/could disable the selection, prompting the user that they need to login before selecting?

StevenDufresne:
> I wonder if we should/could disable the selection, prompting the user that they need to login before selecting?

pkevan:
Screenshot 2024-10-18 at 10 09 41

I wonder if this is clearer/more obvious if we can't save any state within the ticket page?

GautamMKGarg:
> The main drawback is that any selected tickets or coupons are not saved when redirecting through the login/signup process; the user will have to re-enter their selections.

Hi Team

Can't we just save it in session or cookie and fetch these information later?

public_html/wp-content/plugins/camptix/addons/require-login.php
```diff
@@ -59,16 +56,37 @@ public function __construct() {
 	public function block_unauthenticated_actions() {
 		/** @var $camptix CampTix_Plugin */
 		global $camptix;
+		// Continue normal request, this is not a tickets page.
+		if ( ! isset( $_REQUEST['tix_action'] ) ) {
+			return;
+		}
+
```
ryelle, line 63: I changed the logic a little here so we don't need to keep checking for this being set.

ryelle:
In https://github.com/WordPress/wordcamp.org/pull/1399/commits/66361773bd504982fe85ea106ff58e5e492009ca, I updated the login's redirect URL to include the selected tickets & optionally a coupon (if set), so that a user can select tickets, hit register, and when they come back from logging in, it will drop them on the checkout form. This also works when creating an account - once confirmed, the form redirects back to the ticket checkout.

https://github.com/user-attachments/assets/1c9dcb72-ee30-436f-b927-cca351cc454c

I also added the "Apply coupon" flow as an exception to the login requirement, so you can apply a coupon first without logging in, then when you select a ticket and hit register, it carries through the checkout.

aashishpahwa:
Creating an account isn't leading to redirection but - "Your account is pending approval". Can we add that on the WordCamp website itself?

image

aashishpahwa:
The UX still needs work for new users - those new to WordPress. It'll be not easy for them to complete 3 steps - then remember the website name - add that to address bar - go to tickets section again - buy tickets.

aashishpahwa:
Here's the process that a new user needs to follow - https://scribehow.com/embed-preview/Buy_A_Ticket_For_WordCamp_Delhi__Co0obiBrSxGvRZdQ4V1DRA?as=video

Not at all UX friendly.

StevenDufresne:
A couple things to point out in your test.

1. You chose a spammy username (a have a spammy looking email address) and as a result your account needed manual approval. Looking at the logs over the last 10 days, those are not common. 
2. Since you are waiting for manual approval, you see the extra profile fields. Users who are **not** stuck in pending (and most account aren't) only see the "Confirm your email" screen which does not include those inputs. You can read the background on that [here](https://wordpress.slack.com/archives/C02QB8GMM/p1729063497321459).
3. Users who confirm their email, upon complete, are redirected to the ticket page. They do not need to log back in.
   - You appear to have not waited for approval on your pending account.
   - You therefore had no session and needed to navigate back manually and log in again (to a different account).

aashishpahwa:
Hello Steven

I understood your point. But I guess I wasn't able to put my point here. 

Here's a process of me using real username and real email - 

https://scribehow.com/shared/Creating_an_Account_for_WordCamp_Delhi_2024__SmvphzvrTeuTWs2undaV_A

**Still, 15 steps before I can click on checkout.** 

Given that the biggest web community in the world is coming up with this UX getting new members is not recommended. 
People are just not doing it. We sold a total of 3 tickets total in the last five days. (We were selling 10-15 per day before this)

I'd suggest creating an account and redirecting back to the ticketing page while the person gets an email to confirm their account in the backend. 

It'll be efficient.

StevenDufresne:
Thanks for the follow-up!

> I'd suggest creating an account and redirecting back to the ticketing page while the person gets an email to confirm their account in the backend.

I can understand how email confirmation can make account creation feel disjointed. However, I personally don't feel confident removing that step seeing that it may cause attendees to lose access to their ticket. Important information is communicated via email. I think best case scenario would be to support social logins.

PR #1403, opened 2024-10-21
Don't show login banner if no tickets exist.

StevenDufresne:
We currently show the "Login" banner regardless of whether the camps are selling tickets. This is particularly problematic for past events.

Looking at WordCamp Asia 2024, you see:

Screenshot 2024-10-21 at 10 22 27 AM

This PR makes it so we only show the banner when tickets are for sale.

## Future considerations

In the case where the WordCamp hasn't been completed and new tickets will be going on sale, we could provide a banner to suggest users create an account ahead of time. I don't think that will be a compelling call to action without a useful feature in return like maybe "Create an account to be notified when tickets are back on sale".

I'll also follow up with another PR that changes that ticket-for-sale message.

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/require-login.php
+++ b/public_html/wp-content/plugins/camptix/addons/require-login.php
@@ -123,7 +123,7 @@ class CampTix_Require_Login extends CampTix_Addon {
 		}
 
 		// Warn users that they will need to login to purchase a ticket
-		if ( ! is_user_logged_in() && $camptix->has_tickets_available() && ! $this->user_is_editing_ticket()  ) {
+		if ( ! is_user_logged_in() && $camptix->has_tickets_available() && ! $this->user_is_editing_ticket() ) {
 			$camptix->notice( apply_filters(
 				'camptix_require_login_please_login_message',
 				sprintf(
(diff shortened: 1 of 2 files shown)


public_html/wp-content/plugins/camptix/addons/require-login.php
```diff
@@ -121,7 +121,7 @@ public function ticket_form_message() {
 		}
 
 		// Warn users that they will need to login to purchase a ticket
-		if ( ! is_user_logged_in() ) {
+		if ( $camptix->has_tickets_available() && ! is_user_logged_in() ) {
```
pkevan: we should probably check the status of the WordCamp as well, as an event can end with tickets still available.

pkevan: oh, I guess it could rely on https://github.com/WordPress/wordcamp.org/pull/1404/

StevenDufresne: Good call.

StevenDufresne: Actually, we don't need to check the WordCamp status. When a camp is closed, tickets are all set to unavailable.

PR #1404, opened 2024-10-21
Change ticket messaging when camp is closed.

StevenDufresne:
Related: #1403

When a WordCamp is closed, we still display messaging that appears like a ticket may be sold later. This PR updates that message to explain the camp is closed.

I also created an `is_wordcamp_closed` function as opposed to checking the post type so we can start to refactor. I updated the only other reference in this file as well.

### Screenshots

Example: https://asia.wordcamp.org/2024/tickets/

| Before | After |
|--------|--------|
| Screenshot 2024-10-21 at 10 47 14 AM | Screenshot 2024-10-21 at 10 41 45 AM |


public_html/wp-content/plugins/camptix/camptix.php
```diff
@@ -5485,10 +5485,14 @@ function form_start() {
 			$this->error( __( "It doesn't look like your form submitted any attendee information. Please try again.", 'wordcamporg' ) );
 		}
 
-		if ( ! $available_tickets ) {
+		if ( ! $available_tickets && ! $this->is_wordcamp_closed() ) {
```
pkevan, line 5488: unsure of the intention here, but guess it might want to be `OR` instead of `AND` since we're checking negatives

public_html/wp-content/plugins/camptix/notice.diff
```diff
@@ -0,0 +1,74 @@
+Index: camptix.php
```
pkevan: guess this was included in the PR by mistake?

StevenDufresne: 🤦‍♂️

public_html/wp-content/plugins/camptix/camptix.php
StevenDufresne: I agree it reads a bit weird. It is likely better as `! available_tickets && $this->is_wordcamp_open()` but that function doesn't exist. It cannot be an `OR` clause based on how the code currently functions, however.

pkevan: the problem is that this notice would then only show if the wordcamp has closed **and** there are no available tickets, which is different to the current function which would show this message when there aren't available tickets regardless of status.

StevenDufresne: > the problem is that this notice would then only show if the wordcamp has closed

The logic is `! $this->is_wordcamp_closed()` or specifically `! 'wcpt-closed'`, meaning anything before closed which I think is fine. I think I'm missing what the problem is. I could create another helper function that targets `wcpt-scheduled` if that makes this more readable.

pkevan: yeah - but if there are available tickets, and the wordcamp is closed, this wouldn't trigger...

StevenDufresne: Correct, if there are available tickets we don't want to show `__( 'Sorry, but there are currently no tickets for sale. Please try again later.', 'wordcamporg' )`. Tickets being available is the default state and is handled at the bottom.

Additionally, if the event is closed or past the event day, [there will be no available tickets](https://github.com/WordPress/wordcamp.org/blob/450a3716e6e8f44638f4547d7aa908f3fc0c77b6/public_html/wp-content/plugins/camptix/camptix.php#L6878).

Issue #1393, opened 2024-10-11
New wordpress.org login prevents users to edit tickets bought without accounts.


MarcZijderveld:
### Describe the bug

New required wordpress.org login prevents users who have bought tickets for a wordcamp to edit their tickets without creating a WordPress account.

### To reproduce

1. Go to 'Ticket edit link in email'
2. Prompted with login screen

### Expected behavior

Editing tickets for previously bought tickets (without WordPress account) should not require login.

### WordCamp

https://netherlands.wordcamp.org/2024/

### System (please complete the following information)

 - Device: Framework Laptop
 - OS: Ubuntu 24.04
 - Browser: Google Chrome
 - Version: 129.0.6668.70

pkevan:
related slack thread: https://wordpress.slack.com/archives/C08M59V3P/p1728654678770999?thread_ts=1728647015.720739&cid=C08M59V3P

ryelle:
There are a few other bugs in this flow:

Post-login, the access link parameters are stripped, so you just land on the regular tickets page.

Even once logged in, I don't think users can edit their tickets with the emailed access link. For example, my WCUS access link lands me on this page. The "status" has taken over where the edit link goes.

![Image](https://github.com/user-attachments/assets/09f3eba6-145a-47dc-8115-10ef7c3fd34b)

This seems to be because I bought it before the requirement, so it's not attached to my user account, and `show_buyer_attendee_status_instead_of_edit_link` takes over to display this message.

PR #1405, opened 2024-10-21
Allow users to edit tickets without needing to log in.

StevenDufresne:
Related: https://github.com/WordPress/wordcamp.org/issues/1393

WordPress.org accounts were [added](https://github.com/WordPress/wordcamp.org/commit/14478e7575c6ca4e25f880fb4012f4812d122f70) as a requirement for purchasing tickets.

Since this feature was added after users purchased tickets _the old way,_ there can be a mismatch between the attendee associated with the ticket and the logged-in account. This PR removes the need to log in to edit the ticket. This returns it to previous functionality.

Props [MarcZijderveld](https://github.com/MarcZijderveld)

This should be removed in the future when there are no more tickets sold without wp.org accounts.

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/require-login.php
+++ b/public_html/wp-content/plugins/camptix/addons/require-login.php
@@ -72,7 +72,7 @@ class CampTix_Require_Login extends CampTix_Addon {
 		}
 
 		// Temporary: We don't want to block users from editing tickets.
-		// See: https://github.com/WordPress/wordcamp.org/issues/1393
+		// See: https://github.com/WordPress/wordcamp.org/issues/1393.
 		$user_is_editing_ticket = in_array( $_REQUEST['tix_action'], array( 'access_tickets', 'edit_attendee' ) );
 
 		if ( ! is_user_logged_in() && ! $user_is_editing_ticket ) {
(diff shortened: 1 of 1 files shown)


public_html/wp-content/plugins/camptix/addons/require-login.php
```diff
@@ -71,7 +71,11 @@ public function block_unauthenticated_actions() {
 			return;
 		}
 
-		if ( ! is_user_logged_in() ) {
+		// Temporary: We don't want to block users from editing tickets.
+		// See: https://github.com/WordPress/wordcamp.org/issues/1393.
+		$user_is_editing_ticket = in_array( $_REQUEST['tix_action'], array( 'access_tickets', 'edit_attendee' ) );
```
pkevan, line 76: Ideally we should check that the ticket doesn't have an assigned username already, since the workaround should only cover those tickets that were purchased before this was possible.

dd32: I agree, but I'm not overly worried about that myself.  If someone has the ticket access (or edit) token even if they're logged into the wrong account I don't see that as being a major concern.

dd32 (approved):
LGTM, noting that I don't think we strictly need to validate whether they have an edit/access token or not, since they'll either have one, have access through an account, or just end up in a dead end.

pkevan (approved):
👍

PR #1421, opened 2024-10-30
Force unconfirmed tickets to log in.

### LINKED ISSUE #1420: Editing unconfirmed tickets clears the username.
### Describe the bug
Original Thread: https://wordpress.slack.com/archives/C08M59V3P/p1730140369131709.

In order to allow users who purchased tickets before requiring wp.org accounts to edit their ticket, I removed the forced login in #1405. 

However, that PR mistakenly didn't remove code that sets the user to the logged-in user for unconfirmed tickets.

**How are tickets unconfirmed?**

When a user purchases multiple tickets in one session, we create all the tickets but assign `[[ unconfirmed ]]` to the `tix_username`. Code in [require-login.php](https://github.com/WordPress/wordcamp.org/blob/2bbc491545d8b05c9dc360fa1c8abc6278686f04/public_html/wp-content/plugins/camptix/addons/require-login.php) looks for these unconfirmed tickets and provides a [confirm registration](https://github.com/WordPress/wordcamp.org/blob/2bbc491545d8b05c9dc360fa1c8abc6278686f04/public_html/wp-content/plugins/camptix/addons/require-login.php#L151) interface that, when submitted, assigns the logged in user to the ticket being edited. 

Since we no longer require login, this code is in a weird spot. Currently, when unconfirmed tickets are updated, the `tix_username` is emptied. This has [downstream impact](https://github.com/WordPress/wordcamp.org/blob/2bbc491545d8b05c9dc360fa1c8abc6278686f04/public_html/wp-content/plugins/camptix/addons/require-login.php#L797) as other components rely on this meta value.

We have a few options:
1. Remove the filter that assigns the ticket on form save.
   - This feels risky because it would probably lead to more filters needing to be commented out.   
3. Force users to login if the ticket is unconfirmed.
   - This is how the code expects things to be, we should probably go with this.  

### Expected behavior

The `tix_username` should not be empty.

StevenDufresne:
Fixes https://github.com/WordPress/wordcamp.org/issues/1420

Props mkrndmane, ryelle.

This PR forces login for editing tickets that are currently unconfirmed.

## Changes
- Adds `user_must_confirm_ticket` that checks if the ticket is unconfirmed. [Reference](https://github.com/WordPress/wordcamp.org/pull/1421/files#diff-f9f30f21b09915ba2e39f241af9901508aa3d8e8d533fd234296914c79a8efb1R859)
- Passes along all `tix_` parameters found in `$_REQUEST` to the final destination. [Reference](https://github.com/WordPress/wordcamp.org/pull/1421/files#diff-f9f30f21b09915ba2e39f241af9901508aa3d8e8d533fd234296914c79a8efb1R84)
  - I don't understand is why the [original code checked](https://github.com/WordPress/wordcamp.org/blob/2bbc491545d8b05c9dc360fa1c8abc6278686f04/public_html/wp-content/plugins/camptix/addons/require-login.php#L79) for `isset( $_REQUEST['tix_tickets_selected'] )` before passing along the parameters. I would assume we could pass them along in all cases.

## How to test
1. Purchase multiple tickets. ([Stripe test cards](https://docs.stripe.com/testing#testing-interactively), [Stripe test keys](https://docs.stripe.com/keys#obtain-api-keys))
    - Set a ticket to `I don't know who will use this ticket yet`
    - Set a ticket First Name/Last Name/Email.
2. While viewing an `Attendees` post through the admin interface list:
    - Confirm the `tix_username` is set to `[[ unconfirmed ]]`
    - Copy the edit token link address.
3. Visit that link in an incognito session.
4. Confirm you are asked to login.
   - Note if you login with the same account you purchased your ticket with, [this filter will stop you](https://github.com/WordPress/wordcamp.org/blob/2bbc491545d8b05c9dc360fa1c8abc6278686f04/public_html/wp-content/plugins/camptix/addons/require-login.php#L38C80-L38C104). You can comment out this one and the one below it to test if you don't have a second account to use
 5. Enter new details into the fields.
 6. Click the "Confirm Registration" button.
 7. View the attendee through the admin interface and confirm that their `tix_username` is set to the user that was used to log in.
 8. Copy the edit token link again.
 9. Log out.
 10. Paste that link in your browser.
 11. Update information and click "Save Attendee information"
 12. Notice that you are able to update the ticket information without logging in once the user is no longer `[[ unconfirmed ]]`.


public_html/wp-content/plugins/camptix/addons/require-login.php
```diff
@@ -71,16 +71,22 @@ public function block_unauthenticated_actions() {
...
 			$args = array();
-			// If this was a registration, pass through the selected tickets and coupon.
-			if ( 'attendee_info' === $_REQUEST['tix_action'] && isset( $_REQUEST['tix_tickets_selected'] ) ) {
-				$args['tix_action'] = $_REQUEST['tix_action'];
-				$args['tix_tickets_selected'] = $_REQUEST['tix_tickets_selected'];
-				if ( isset( $_REQUEST['tix_coupon'] ) ) {
-					$args['tix_coupon'] = $_REQUEST['tix_coupon'];
+			if ( in_array( $_REQUEST['tix_action'], array( 'attendee_info', 'edit_attendee' ) ) ) {
+				// Pass along all `tix_` information.
+				foreach ( $_REQUEST as $key => $value ) {
```
pkevan: we'll want to be explicit in passing the values from $_REQUEST (i.e. checking they are what we expect) and checking their value before adding to `add_query_arg`

StevenDufresne: Sure. I can do that. How about the changes in [5346a21](https://github.com/WordPress/wordcamp.org/pull/1421/commits/5346a2122cc0d528f8f3717b784e93bd412ece66)?

pkevan: Nice!

pkevan:
Unsure if this has been tested, but could this logic be extended to support users transferring their tickets to another username? Or is that supported elsewhere? I suppose on transfer of a ticket, it would just switch back to `[[ unconfirmed ]]`

ryelle:
> I don't understand is why the [original code checked](https://github.com/WordPress/wordcamp.org/blob/2bbc491545d8b05c9dc360fa1c8abc6278686f04/public_html/wp-content/plugins/camptix/addons/require-login.php#L79) for isset( $_REQUEST['tix_tickets_selected'] ) before passing along the parameters. I would assume we could pass them along in all cases.

Pretty sure that was just because I used `$_REQUEST['tix_tickets_selected']` right below and wanted to make sure it wouldn't error.

I've tested out the edit ticket flow and a logged out user can correctly edit a ticket bought before the require-login, and not edit after 👍🏻 

Unfortunately, in testing the regular ticket purchase flow I'm seeing this error:

Screenshot 2024-10-31 at 3 20 18 PM

I also realized the current redirect probably doesn't pass through the reservation info, but I was interrupted by ^ before I could properly test. A reservation is set up on a ticket (in the edit-ticket screen), and ends up being a link like `somecamp.wordcamp.org/2024/tickets/?tix_reservation_id=speakers&tix_reservation_token=k3vlopd1G3vWewZA`.

ryelle:
Oh, the error is because `tix_tickets_selected` should be an array of [ticket-id] => quantity, not an int.

StevenDufresne:
I think I've covered all the relevant parameters. Can I get another review?

public_html/wp-content/plugins/camptix/addons/require-login.php
```diff
@@ -71,26 +71,78 @@ public function block_unauthenticated_actions() {
...
 			wp_safe_redirect( add_query_arg( 'wcname', get_bloginfo( 'name' ), wp_login_url( $tickets_url ) ) );
 			exit();
 		}
 	}
 
+	/**
+	 * Get sanitized ticket parameters from request array.
+	 *
+	 * @param array $request_data Array of request data to sanitize.
+	 * @return array Sanitized parameters.
```
pkevan, line 95: i guess technically we could return an empty array too, if no parameters match the allowed list.

public_html/wp-content/plugins/camptix/addons/require-login.php
```diff
@@ -71,26 +71,78 @@ public function block_unauthenticated_actions() {
...
-			// If this was a registration, pass through the selected tickets and coupon.
-			if ( 'attendee_info' === $_REQUEST['tix_action'] && isset( $_REQUEST['tix_tickets_selected'] ) ) {
-				$args['tix_action'] = $_REQUEST['tix_action'];
-				$args['tix_tickets_selected'] = $_REQUEST['tix_tickets_selected'];
-				if ( isset( $_REQUEST['tix_coupon'] ) ) {
-					$args['tix_coupon'] = $_REQUEST['tix_coupon'];
-				}
+		if ( ! is_user_logged_in() ) {
+
+			// Temporary: We don't want to block users from editing tickets unless they are unconfirmed.
```
pkevan, line 76: how long will this be temporary 😉

pkevan (approved):
untested, but code looks ok.

public_html/wp-content/plugins/camptix/addons/require-login.php
StevenDufresne: I have a script that will tell me when there are no more tickets purchased without accounts. Should be about February 2025.

public_html/wp-content/plugins/camptix/addons/require-login.php
StevenDufresne: Yep, $args will return empty if nothing matches. Do you think we should note that here?

MakarandMane:
I will test it & get back here, found any issue.

public_html/wp-content/plugins/camptix/addons/require-login.php
pkevan: yes, `@return array Empty array or sanitized parameters.` or similar would work.

MakarandMane:
@StevenDufresne There are some more issue I have report about this. 

---

**Bug Report: Username Cleared When Editing Confirmed Ticket Without Login**  

**Steps to Reproduce:**  
1. In WordPress admin, navigate to **wp-admin → Tickets → Attendees** and filter **Published** tickets.  
2. Select a **confirmed** ticket that has a username attached.  
3. Copy the unique edit URL for that ticket.  
4. Open a new incognito window or log out of the website.  
5. Paste the edit URL in the browser and attempt to modify the ticket.  
6. Save the changes.  

**Expected Behavior:**  
- The username associated with the confirmed ticket should **not** be cleared.  

**Actual Behavior:**  
- The username field gets **cleared** after editing the ticket while not being logged in.  

**Additional Context:**  
- A previous fix restricted unconfirmed tickets from being edited without logging in.  
- However, confirmed tickets are now affected when edited via the unique URL without authentication.  

**Priority:** Medium (Affects data integrity for confirmed tickets)  

**Suggested Fix:**  
- Ensure that the username is preserved when editing confirmed tickets, even if the user is not logged in.  
- Validate the ticket edit process to maintain username data integrity.

PR #1431, opened 2024-11-25
Fix localised WordCamp logs

### LINKED ISSUE #1300: Don't translate wordcamp statuses in logs
**Describe the bug**

On https://central.wordcamp.org/reports/wordcamp-status-report/ a PHP Notice is being triggered:
```
E_NOTICE Undefined index:
File wp-content/plugins/wordcamp-reports/views/html/wordcamp-status.php:41
```

This is caused by `$active_camp` containing this data:
```
array (size=4)
  'name' => string 'WordCamp ......' (length=28)
  'logs' => 
    array (size=1)
      1 => 
        array (size=3)
          'timestamp' => int ..........
          'user_id' => int .........
          'message' => string 'Needs Vetting &rarr; 已拒絕' (length=30)
  'latest_log' => 
    array (size=3)
      'timestamp' => int ...........
      'user_id' => int ...........
      'message' => string 'Needs Vetting &rarr; 已拒絕' (length=30)
  'latest_status' => string '' (length=0)
```

`'latest_status'` is empty, as the new status is extracted from the log text, expecting that it's only in English.

https://github.com/WordPress/wordcamp.org/blob/d151fcd45ed673b42e569c5f6c142b7b1f5ce5ae/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-base-status.php#L77-L121

When logging, the post status should use the english non-translated text, or, it should log the IDs in the metadata for it.

This is more than just a PHP Notice; this means that others have to translate the logs to see what changed.

**To reproduce**

1. Use a locale that has translated the post_status labels
2. Change status
3. Note log including translated post status label.

**Expected behavior**

English logs.

**WordCamp**

https://central.wordcamp.org/wp-admin/post.php?post=5701004&action=edit

dd32:
The current WordCamp status code filters `locale` which is the incorrect way to switch locales, causing logs to contain localised post_status labels, causing the logs to contain a localised status which is unmatchable and causes PHP Notices and broken UIs.

Fixes #1300 probably

There's numerous other uses of `set_locale_to_en_us` in the codebase, which likely do not work. The locale can't be changed at runtime simply by changing the value of the filter, the filter only changes which locale is initially loaded.


public_html/wp-content/plugins/wcpt/wcpt-event/class-event-loader.php
```diff
@@ -17,6 +17,7 @@ public function __construct() {
 		add_action( 'plugins_loaded', array( $this, 'includes' ) );
 		add_action( 'init', array( $this, 'register_post_types' ) );
 		add_action( 'init', array( $this, 'register_post_statuses' ) );
+		add_action( 'change_locale', array( $this, 'register_post_statuses' ) ); // Re-register them when the locale changes, to have localised statuses.
```
dd32:
```suggestion
		add_action( 'init', array( $this, 'register_post_types' ) );
		add_action( 'init', array( $this, 'register_post_statuses' ) );
		// Re-register them when the locale changes, to have localised statuses.
		add_action( 'change_locale', array( $this, 'register_post_types' ) );
		add_action( 'change_locale', array( $this, 'register_post_statuses' ) );
```

The post types can probably be reloaded too, to ensure they're properly translated, but I don't think they're used anywhere in logs so are likely less problematic.

PR #1438, opened 2024-12-10
Add/badges

### LINKED ISSUE #1428: Allow badges to be created by event organisers
### Components / Tools

WP.org profile badges

### Description

Allow event organiser (administrators on a site) to create a list of people who should get badges for various WordCamp activities.

There currently exists some automation, but it's largely unreliable and requires various fields to be set in order to work.

The proposal is to allow a list of wp.org usernames to be gives badges through an admin interface, with a possible extension to include the various custom post types e.g. volunteers where no automation currently exists.

Related #550.

pkevan:
Adds an admin interface into Camptix to allow administrators to assign badges to users.

Fixes #1428

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -67,6 +67,7 @@ class CampTix_Plugin {
 
 		require( dirname( __FILE__ ) . '/inc/class-camptix-addon.php' );
 		require( dirname( __FILE__ ) . '/inc/class-camptix-payment-method.php' );
+		require( dirname( __FILE__ ) . '/inc/class-camptix-badges.php' );
 
 		if ( defined( 'WP_CLI' ) && WP_CLI ) {
 			require_once( dirname( __FILE__ ) . '/inc/class-wp-cli-commands.php' );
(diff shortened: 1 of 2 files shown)


pkevan:
@renintw would you mind taking a quick look at this when you get a chance?

dd32:
Due to events having `Badges` perhaps this should be labeled `Profile Badges` to properly explain what they are?

pkevan:
> Due to events having `Badges` perhaps this should be labeled `Profile Badges` to properly explain what they are?

good point - i'll add some specific language around on the tool page and adjust the menu item accordingly so there isn't confusion.

ePascalC:
@pkevan Why would this be a submenu under Tickets? If this is to only add badges for attendees, then it would fit there. But if this module will also be used for Volunteers, like I see in the code, then it seems the wrong place in the menu.

Also please use https://developer.wordpress.org/reference/functions/validate_username/ just to ensure correct wordpress.org usernames (unless Profiles\badge_api is already doing this)

pkevan:
> Why would this be a submenu under Tickets? If this is to only add badges for attendees, then it would fit there. But if this module will also be used for Volunteers, like I see in the code, then it seems the wrong place in the menu.

There isn't really a suitable place for this to go, hence it's location. 

Initially I was going to add multiple badges to the interface which was why it ended up related to Camptix's menus.

PR #1456, opened 2025-02-12
Avoid a PHP Warning when accessing the submitted-badges confirmation page upon refresh

dd32:
```
E_WARNING: Undefined array key "_wpnonce"
wp-content/plugins/camptix/inc/class-camptix-badges.php:14

URL: https://asia.wordcamp.org/2025/wp-admin/edit.php?post_type=tix_ticket&page=camptix_badges&badge-submit=1
```

Reloading / Accessing the URL  above causes PHP Warnings to be thrown, and `Invalid Request` to be shown in the UI.. when seemingly it probably shouldn't have attempted to perform any actions.

@pkevan this is untested, but I don't see the need to use an extra query-var here, and the existence of the POST data is probably enough, is there something I'm missing?

1.
2.
3.

### DIFF
--- a/public_html/wp-content/plugins/camptix/inc/class-camptix-badges.php
+++ b/public_html/wp-content/plugins/camptix/inc/class-camptix-badges.php
@@ -11,7 +11,7 @@ use WordPressdotorg\Profiles;
  */
 function process_badges() {
 
-	if ( ! current_user_can( 'manage_options' ) || ! wp_verify_nonce( $_POST['_wpnonce'], 'badge-submission' ) ) {
+	if ( ! current_user_can( 'manage_options' ) || empty( $_POST['_wpnonce'] ) || ! wp_verify_nonce( $_POST['_wpnonce'], 'badge-submission' ) ) {
 		return __( 'Invalid request', 'wordcamporg' );
 	}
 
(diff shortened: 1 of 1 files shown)


pkevan:
> @pkevan this is untested, but I don't see the need to use an extra query-var here, and the existence of the POST data is probably enough, is there something I'm missing?

Quite possibly just me following a pattern from another file in the Camptix plugin (like here: https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/plugins/camptix/camptix.php#L2190), but can't remember if it did anything. 

I'll test and report back - thanks!

public_html/wp-content/plugins/camptix/inc/class-camptix-badges.php
```diff
@@ -48,7 +48,8 @@ function process_badges() {
  * Outputs manage badge admin screen, and allows processing of submit.
  */
 function menu_badges() {
-	if ( isset( $_GET['badge-submit'] ) && ( 1 == $_GET['badge-submit'] ) ) {
+	// phpcs:ignore WordPress.Security.NonceVerification.Missing
+	if ( ! empty( $_POST ) ) {
```
pkevan, line 52: I wonder if we should check something more specific here, like `$_POST['action']`?

pkevan (approved):

public_html/wp-content/plugins/camptix/inc/class-camptix-badges.php
dd32: I considered it, but then figured who cares :) Either the handler will pass or it won't.. 

Appreciate you looking and merging @pkevan!

PR #1457, opened 2025-02-12
Unit Tests: Run under PHP 8.1, same as production

dd32:
PR is for testing the github actions.

WordCamp.org has switched to PHP 8.1, so the unit tests should be running under the same environment.

There's likely more steps needed here.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-remote-css/app/synchronize-remote-css.php
+++ b/public_html/wp-content/plugins/wordcamp-remote-css/app/synchronize-remote-css.php
@@ -77,7 +77,9 @@ function fetch_unsafe_remote_css( $remote_css_url ) {
  */
 function sanitize_unsafe_css( $unsafe_css ) {
 	if ( ! class_exists( 'Jetpack_Custom_CSS_Enhancements' ) ) {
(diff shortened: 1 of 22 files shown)


public_html/wp-content/plugins/wordcamp-remote-css/app/synchronize-remote-css.php
```diff
@@ -77,7 +77,9 @@ function fetch_unsafe_remote_css( $remote_css_url ) {
  */
 function sanitize_unsafe_css( $unsafe_css ) {
 	if ( ! class_exists( 'Jetpack_Custom_CSS_Enhancements' ) ) {
-		if ( version_compare( JETPACK__VERSION, '11.6', '<' ) ) {
+		if ( version_compare( JETPACK__VERSION, '13.7.1', '>' ) ) {
+			return ''; // Custom CSS not included in Jetpack anymore.
```
dd32, line 81: This is obviously a big TODO; to replace the Jetpack CSS sanitization dependency with our own bundled here.
This is needed, and was one of the reasons for this PR in the first places.

This change should be reverted from the PR really.

PR #1470, opened 2025-05-01
Notify: Allow segmenting the Notification functionality by ticket confirmed/unconfirmed status.

dd32:
The intention of this is to allow WordCamps an easy way to bulk notify attendees whose tickets have not yet been confirmed (linked to a WordPress.org user, and properly filled out)

### Screenshots

Screenshot 2025-05-01 at 4 10 55 pm

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/require-login.php
+++ b/public_html/wp-content/plugins/camptix/addons/require-login.php
@@ -46,7 +46,7 @@ class CampTix_Require_Login extends CampTix_Addon {
 		add_filter( 'camptix_attendees_shortcode_query_args',         array( $this, 'hide_unconfirmed_attendees' ) );
 		add_filter( 'camptix_private_attendees_parameters',           array( $this, 'prevent_unknown_attendees_viewing_private_content' ) );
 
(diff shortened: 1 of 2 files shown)


dd32:
This has had ZERO testing.

public_html/wp-content/plugins/camptix/addons/require-login.php
```diff
@@ -936,6 +940,79 @@ public function hide_unconfirmed_attendees( $query_args ) {
...
+						array(
+							'key' => 'tix_username',
+							'compare' => 'NOT EXISTS',
+						),
+						// The email is the standard unknown attendee.
+						array(
+							'key' => 'tix_email',
+							'value' => $this->get_unknown_attendee_info()['email'],
+							'compare' => '=',
+						),
```
ryelle, line 1000: If the email is the unknown value, the message sent by notify won't go anywhere, so do we need this part of the query?

ryelle:
I've rebased and done some testing (and fixed a variable name), seems to filter attendees correctly now. It would be nice to merge this so that I could use the filters to add #1511.

PR #1053, opened 2023-09-21
Send copy of call for speakers email to the speaker itself

timiwahalahti:
When the speaker submits a talk to WordCamp, send a "receipt" of that to the speaker. It will help people keep track of what talks they’ve submitted to WordCamps, and when.

Instead of sending out separate emails and handling logic for that, I decided to just modify the email that goes out in any case. This helps to keep the code a tad cleaner, but has also the added benefit of organisers being able to reply to that email and continue the discussion with the speaker on a thread that has all of the submitted information. In addition, the same code can be extended for sponsor and volunteer applications easily.

Fixes #376

Props @melchoyce

### Screenshots

CleanShot 2023-09-21 at 11 36 13@2x

### How to test the changes in this Pull Request:

1. Head to WordCamp site and open a Call for Speakers post
2. Fill the form
3. Check the email (on Docker, use Mailcatcher for that)

PR #1487, opened 2025-05-30
Forms to drafts: Update email sent to speaker after submission

ryelle:
In https://github.com/WordPress/wordcamp.org/pull/1053, an email was added, sent to speakers when they submit a talk on the Call for Speakers form. This PR just updates the language of the email a little, to include the WordCamp name and fix a grammar mistake.

Props Kristin Smith on the WCUS team for flagging the current email.

### Screenshots

No screenshot, but here's what the new email will read:

> Hello,
>
> Thank you for your interest in speaking at [WordCamp City]! We have received your submission, and a copy is included below for your records.
>
> Please do not reply to this email, as it is automatically generated. If you have any questions, send an email to [city@wordcamp.org].
>
> [form submission]

1. Install an email catcher plugin or use mailcatcher (visit http://localhost:1080/ to view on Docker env)
2. Submit a talk using the Call for Speakers form
3. You should see an email go to the speaker's email address
4. It should look like above, with the bracketed content replaced with the current WordCamp details

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
+++ b/public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
@@ -450,9 +450,19 @@ class WordCamp_Forms_To_Drafts {
 		// Get the key from submitted data.
 		$form_key = $this->get_form_key_by_id( absint( $_POST['contact-form-id'] ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
 
-		// Add speaker as a copy on the email.
+		// Update headers on call for speakers responses.
 		if ( 'call-for-speakers' === $form_key ) {
+			// Add speaker as a copy on the email.
 			$headers .= "Cc: {$reply_to_addr}\r\n";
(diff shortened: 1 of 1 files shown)


dd32:
> Please do not reply to this email, as it is automatically generated.

Could we fix that perhaps?

ryelle:
> > Please do not reply to this email, as it is automatically generated.
>
> Could we fix that perhaps?

I've updated the email headers to set the Reply-To to the WordCamp's email, so when speakers get their copy and reply, it should send to the WordCamp's real email. But technically it's the same email that's sent to the WordCamp as record of submission. So now the email sent to whatever@wordcamp.org also has reply-to whatever@wordcamp.org. I suspect this is fine, I don't think any WordCamps are managing speakers by email responses, they likely use the feedback export. I'm not sure if this is a change that would require an announcement somewhere, though.

dd32:
>  I suspect this is fine

I suspect this is fine too..  I guess ideally the email organisers got had reply-to speaker, and what speakers got would be reply-to wordcamp, but :shrug: organisers should be more likely to realise the email is going to the wrong place..

Issue #1490, opened 2025-06-03
Campus Connect: Create Secondary sites


dd32:
Campus Connect sites have a Primary URL, `events.wordpress.org/campusconnect/2025/location` but also need a student space for each university.

These student sites will be created as `campus.wordpress.org/university-campus`. There may be multiple groups per Campus Connect event.

This is also useful for WordCamps which have a primary site, and a secondary site, for example WordCamp Canada with their English and French sites: https://canada.wordcamp.org/2025/ https://canada.wordcamp.org/2025-fr/

jonoalderson:
_screams in subdomain_

jonoalderson:
Putting these on a new/different subdomain means that they're starting from scratch, duplicating efforts, and competing with themselves / each other / wordcamps. 

Unless there's a _really_ strong reason why they should be on a different hostname - and unless they have the marketing/branding/resourcing/strategy/budget to actively compete in their own right - it'd be preferable to use an existing hostname.

dd32:
@jonoalderson Apologies for the scare :) These sites won't actually be public, they're intended on being "private groups", and will be `noindex`'d entirely. It's just a namespace to get it away from all the other public content, and allow it to be clearly different.

jonoalderson:
Also, whilst I'm dreaming, having https://canada.wordcamp.org/2025/ and https://canada.wordcamp.org/2025/fr/ instead of https://canada.wordcamp.org/2025/ https://canada.wordcamp.org/2025-fr/ would be _so_ much better, more scalable, etc.

jonoalderson:
> [@jonoalderson](https://github.com/jonoalderson) Apologies for the scare :) These sites won't actually be public, they're intended on being "private groups", and will be `noindex`'d entirely. It's just a namespace to get it away from all the other public content, and allow it to be clearly different.

Ah, phew, thanks! Panic mode cancelled!

dd32:
> having https://canada.wordcamp.org/2025/ and https://canada.wordcamp.org/2025/fr/ 

I'm not sure why we have it setup as `2025-fr` for any reason other than they used to be `2025-fr.city.wordcamp.org`. Decision from before my time, which AFAIK is currently engrained into the nginx configurations :)

Ideally, we won't be creating second public sites for WordCamps for too much longer though, as hopefully there'll be multilingual support within Gutenberg / WordPress at some point, which would negate the need for them entirely.

Only one or two WordCamps each year request the second site.

PR #1492, opened 2025-06-03
Organizer Survey / Speaker Feedback: Limit actions to the current network

dd32:
While working on #1490 I may have caused the Organizer Debrief / Speaker Feedback routines to run, which was not expected nor intended. I'm not sure if it actually processed anything, but regardless, these two plugins shouldn't attempt to operate on sites outside of their network when they're activated/deactivated network wide.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
@@ -282,8 +282,7 @@ function get_site_ids_without_skip_flag() {
 
 	$blog_ids = $wpdb->get_col(
 		$wpdb->prepare(
-			"
-			SELECT b.blog_id
+			"SELECT b.blog_id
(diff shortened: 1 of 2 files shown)


public_html/wp-content/plugins/wordcamp-speaker-feedback/wordcamp-speaker-feedback.php
```diff
@@ -280,13 +280,17 @@ function get_assets_url() {
...
-		SELECT b.blog_id
-		FROM $wpdb->blogs AS b
-		LEFT OUTER JOIN $wpdb->blogmeta AS m
-		ON b.blog_id = m.blog_id AND m.meta_key = 'wordcamp_skip_feature' AND m.meta_value = 'speaker_feedback'
-		WHERE m.meta_value IS NULL
-	" );
+	$blog_ids = $wpdb->get_col(
+		$wpdb->prepare(
+			"
+			SELECT b.blog_id
```
dd32:
```suggestion
			"SELECT b.blog_id
```

PR #1498, opened 2025-06-19
Post Types: Fix term conflict on "teams" for Organizers, Volunteers

ryelle:
Currently, the "team" taxonomy is broken on Organizers. For example, we have organizers published on the "Programming" team, but https://us.wordcamp.org/2025/team/programming/ does not work. The same is true if you [try to view in wp-admin](https://us.wordcamp.org/2025/wp-admin/edit.php?team=programming&post_type=wcb_organizer). The issue is that both `wcb_organizer_team` and `wcb_volunteer_team` use the same `query_var=team`, so it's trying to find a volunteer term for "programming" and an organizer term for "programming" and a post with both… which is nothing.

To fix this, I first thought to combine the teams into one taxonomy, but Volunteers & Organizers are split up differently (volunteers might be room monitors, reg desk, etc, while organizers are on programming, web, sponsors…). This also has the benefit of not removing any existing content, because the taxonomy names stay the same. Now Volunteer teams will be `volunteer_team=whatever` and Organizers can stay with `team=whatever`.

Props ryelle

### Screenshots

I also added the team to the CPT list table, so it's easy to see who's on what team.

Screenshot 2025-06-19 at 12 00 16 PM

1. Set up some organizers, volunteers, and teams for each
2. You should see the team on the CPT list table for the Organizer or Volunteer
3. You can filter by team in wp-admin
4. The URLs for team archives should work

### DIFF
--- a/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
+++ b/public_html/wp-content/plugins/wc-post-types/wc-post-types.php
@@ -1763,44 +1763,20 @@ class WordCamp_Post_Types_Plugin {
 			'wcb_organizer_team',
 			'wcb_organizer',
 			array(
-				'labels'       => $labels,
-				'rewrite'      => array( 'slug' => 'team' ),
-				'query_var'    => 'team',
-				'hierarchical' => true,
-				'public'       => true,
-				'show_ui'      => true,
-				'show_in_rest' => true,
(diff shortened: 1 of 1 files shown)


ryelle:
@dd32 Would you be able to review/merge this, since I still don't have sandbox access back?

PR #1499, opened 2025-06-19
Config: Add the missing "CAMPUS_NETWORK_ID" to prevent PHP warning on docker

ryelle:
On the production branch, it currently throws a warning:

> Warning: Use of undefined constant CAMPUS_NETWORK_ID - assumed 'CAMPUS_NETWORK_ID' (this will throw an Error in a future version of PHP) in /usr/src/public_html/wp-content/sunrise.php on line 91

So I've added the const to `wp-config.php` near the other network consts. 

1. Spin up the docker env
2. Visit a test site
3. It works, there is no PHP warning

### DIFF
--- a/.docker/wp-config.php
+++ b/.docker/wp-config.php
@@ -42,7 +42,7 @@ const WORDCAMP_NETWORK_ID   = 1;
 const WORDCAMP_ROOT_BLOG_ID = 5;
 const EVENTS_NETWORK_ID     = 2;
 const EVENTS_ROOT_BLOG_ID   = 47;
-const CAMPUS_NETWORK_ID     = 9999;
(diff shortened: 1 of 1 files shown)


.docker/wp-config.php
```diff
@@ -42,6 +42,7 @@
 const WORDCAMP_ROOT_BLOG_ID = 5;
 const EVENTS_NETWORK_ID     = 2;
 const EVENTS_ROOT_BLOG_ID   = 47;
+const CAMPUS_NETWORK_ID     = 9999;
```
dd32:
```suggestion
const CAMPUS_NETWORK_ID     = 3;
```

PR #1515, opened 2025-09-02
Attempt to automatically connect sites to Jetpack

dd32:
There are some edge-cases with Jetpack that appears to possibly be due to the delay in connecting Jetpack sites.

As the original reason for Jetpack not being auto-connected is no longer true - It was that `city.year.wordcamp.org` SSL was not available at site creation time, we should be able to automatically connect now.
 - *.wordcamp.org SSL covers CITY.wordcamp.org/YEAR
 - events.wordpress.org/*/*/*/ sites always have SSL
 - capus.wordpress.org/*/ sites always have SSL


public_html/wp-content/mu-plugins/wordcamp/jetpack-new-sites.php
```diff
@@ -28,26 +26,79 @@ function auto_connect_new_sites( $new_value, $old_value ) {
...
+		$jetpack_network = \Jetpack_Network::init();
+		// Wrap it in a callable check, as this is reaching deeper into Jetpack than reasonable.
+		if ( is_callable( array( $jetpack_network, 'do_subsiteregister' ) ) ) {
+			$jetpack_connection_result = $jetpack_network->do_subsiteregister( $site_id );
+
+			// Log this for debugging later.
+			if ( is_wp_error( $jetpack_connection_result ) ) {
+				trigger_error( 'Jetpack subsiteregister failed for site ' . $site_id . ': ' . $jetpack_connection_result->get_error_message(), E_USER_WARNING );
+			}
+		}
```
dd32:
```suggestion
		$jetpack_network           = \Jetpack_Network::init();
		$jetpack_connection_result = new \WP_Error( 'not_callable', 'Jetpack_Network::do_subsiteregister() not callable.' );
		// Wrap it in a callable check, as this is reaching deeper into Jetpack than reasonable.
		if ( is_callable( array( $jetpack_network, 'do_subsiteregister' ) ) ) {
			$jetpack_connection_result = $jetpack_network->do_subsiteregister( $site_id );
		}

		// Log this for debugging later.
		if ( is_wp_error( $jetpack_connection_result ) ) {
			trigger_error( 'Jetpack subsiteregister failed for ' . site_url(). ': ' . $jetpack_connection_result->get_error_message(), E_USER_WARNING );
		}
```

PR #1517, opened 2025-09-04
Camptix: Add "Purchase Week" to the ticket summary options

ryelle:
This adds a new summary option for the ticket purchase data, grouping by "purchase week". Showing purchases by date can be too detailed and hard to see trends when ticket sales are open for more than a few weeks, so showing by week can give data without requiring the organizer to download and graph the dates (which I've definitely done).

The only downside is the column title is just the week number, not very human friendly. I used this to check the weeks: https://www.epochconverter.com/weeks/2025

1. Go to the summarize tab in Tickets > Tools
2. Select "Purchase week" to summarize by
3. Show summary
4. You should see the tickets purchased by week numbers

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -2399,7 +2399,7 @@ class CampTix_Plugin {
 			'purchase_time' => __( 'Purchase time', 'wordcamporg' ),
 			'purchase_datetime' => __( 'Purchase date and time', 'wordcamporg' ),
 			'purchase_dayofweek' => __( 'Purchase day of week', 'wordcamporg' ),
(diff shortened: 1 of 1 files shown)


dd32:
I wonder if it would be better to include a Date range or Year in the value? Something like `Purchase Week: 2025 Week 50 (Dec 8-14)` or `Purchase Week: 2026 Week 1`?

_(Yeah I realise that maybe it's harder to implement)_

dd32 (approved):

dd32 (approved):

ryelle:
Updated the label to "Week 2025-29 (starting Jul 14 2025)", so that when downloaded, it can still be sorted by the label but it also has meaning to a human reading it.

Screenshot 2025-10-31 at 12 20 04 PM

PR #1537, opened 2025-10-22
Try skipping man-db updates, as we're not using man and actions keep stalling on it.

dd32:
Fixes #, See #

Props <username>, <username>

1.
2.
3.


dd32:
This reduces the `Install SVN` step from 80~120seconds to ~20 seconds. Merging.

Issue #1427, opened 2024-11-13
Incorrect Payment Details Displayed in PDF Invoice for WordCamp Ahmedabad 2025 Booking


tirth-doshi-techno:
### Describe the bug

During the ticket booking process for WordCamp Ahmedabad 2025, when ABC completes the payment through Google Pay, the confirmation email is incorrectly sent to the secondary user listed on the booking. Additionally, the attached PDF invoice displays the secondary user’s name as the payer, despite ABC being the one who completed the payment. This misrepresentation causes confusion about the payer’s identity in the official record.

### To reproduce

1. Start the booking process for WordCamp Ahmedabad 2025.
2. Enter Payer's name as the primary attendee.
3. Add a secondary attendee under the same booking.
4. Complete the payment via Google Pay, with same name as the Primary name.
5. Review the confirmation email sent after the transaction.

### Expected behavior

The confirmation email and PDF invoice should be sent to Payer's name as the primary attendee and payer. The PDF should correctly display Payer's name as the payer, irrespective of additional attendees in the booking.

### Screenshots

> Confirmation Message on WordCamp website
![Image](https://github.com/user-attachments/assets/89eacda1-12a0-4d9a-a829-afc752690dbb)

> Payment Details Displaying Secondary User’s Name
![Image](https://github.com/user-attachments/assets/105127f3-d3dd-413d-a247-4c353c75e552)

dd32:
This is caused by the Instamojo code looping over the attendees and using the last attendee as the one on record:

https://github.com/WordPress/wordcamp.org/blob/86f6983d84c7fa59aee16ae165f2d063e61a047c/public_html/wp-content/plugins/campt-indian-payment-gateway/inc/instamojo/class-camptix-payment-method-instamojo.php#L235-L240

https://github.com/WordPress/wordcamp.org/blob/86f6983d84c7fa59aee16ae165f2d063e61a047c/public_html/wp-content/plugins/campt-indian-payment-gateway/inc/instamojo/class-camptix-payment-method-instamojo.php#L276-L281

PR #1546, opened 2025-10-29
Camptix_Payment::get_order() should return the first user as the attendee_id

dd32:
At present CampTix_Payment_Method::get_order() returns the LAST attendee added as part of the transaction as the attendee_id.

This is usually incorrect, as the first user is often the person purchasing the ticket, and in the case of requiring user logins, will always be the purchaser.

Partially related to #1427


dd32:
This would break some implementations which store metadata against only the attendee_id, #1551 might be a better route.

Issue #1559, opened 2025-11-10
Event Trackers: Add webhooks on tracker creation / status change


supernovia:
### Description
Add support for webhooks triggered on Event Tracker creation and on status change (e.g., when a tracker moves to Scheduled).

This would allow integrations with tools like Zapier, enabling automated actions such as posting Slack notifications, [creating GitHub issues](https://github.com/WordPress/Community-Team/issues/765), or triggering other workflow automations. It would also make it easier to connect the Event Tracker with other reporting or community tools.

### Stakeholders
@Piyopiyo-Kitsune and the Community Team

### How does this new feature contribute to the goals
We're mainly hoping for more options for automating and notifications with event changes, so there's more visibility and support for these events, especially in planning.

### Acceptance Criteria
We're open to shaping on this one.

### Priority
Low: for now we can manually ping and update folks, and for now we may be able to use Help Scout and Zapier to achieve an on-new-event ping in Slack.

dd32:
AFAIK there's a few ways that this would be used, but I'm not 100% sold on any of them yet.

> New event notification

Using a webhook seems limiting here, primarily that it'd be limited to either being setup by developers or WordCamp admins, which limits the ability for other community members to get the same data feed.

For example, we already have Slack notifications of new events:
https://wordpress.slack.com/archives/C02RQC7RN/p1762342709270799

<img width="665" height="220" alt="Image" src="https://github.com/user-attachments/assets/2b79ed27-ac0d-43a2-8b85-46d8e206f6f6" />

I'm wondering is there's a better way to create notifications for others, For example, being able to subscribe to an RSS feed of the events in https://events.wordpress.org/upcoming-events/filtered/country/IN/type/other-wordcamp/ 

A Feed could be subscribed to in other Slack instance channels too if needed.

> Other automations

I guess it would depend on what these were as well, as without a use-case (and known data) it's not super easy to design a webhook data payload.

I think a better solution (for now) would be to create an issue for each automation that needs integrating, the trigger, and payload needed, and we can set those up at that time.

> creating GitHub issues

I'm **really** questioning this one as worthwhile, looking at https://github.com/WordPress/Community-Team/issues/765 it doesn't seem to provide any benefit to community members, other than making work more visible (Which could be done much easier other ways). I don't really understand the point of https://github.com/WordPress/Community-Team at all :)

supernovia:
This one is nice to have, but pretty low-priority:

[...]

PR #1561, opened 2025-11-13
Add a Feed of the Upcoming Events page

dd32:
Spurred by #1559 and a desire to have Events notifications in other places, it got me thinking about using RSS feeds.

This PR implements RSS feeds for urls such as:
- https://events.wordpress.org/upcoming-events/
- https://events.wordpress.org/upcoming-events/filtered/country/AU/
The feed url is suffixing `/feed/` to those URLs.

Some shortcomings were encountered:
 - The current event query code doesn't return the description, and always queries for 500 elements. https://github.com/WordPress/wporg-mu-plugins/pull/713 (This is a precursor to this feed PR)
 - RSS doesn't seem to handle future-dated content very well.. if at all.. as this is explicitly not allowed by the spec really
 - As the date isn't deemed valid by some clients, the order of the items in the feed is somewhat not predictable

Not implemented:
 - Page headers and links to the RSS feed to make it visible.
 - Feed Title is `<title>Posts filtered by: Australia &#8211; WordPress Events</title>` which could do with changing to `Upcoming Events filtered by: `

### Screenshots

Example of a RSS reader, as you can see, the event order is `December 2025, Jan 2026, Dec 2025, Nov 2025` despite the feed being ordered as `Nov, Dec, Dec, Jan`.

Screenshot 2025-11-13 at 4 59 37 pm

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -123,6 +123,8 @@
 		<exclude name="WordPress.Arrays.ArrayDeclarationSpacing.ArrayItemNoNewLine" />
 		<exclude name="PEAR.Functions.FunctionCallSignature.ContentAfterOpenBracket" />
 		<exclude name="PEAR.Functions.FunctionCallSignature.CloseBracketLine" />
+		<exclude name="Squiz.Functions.MultiLineFunctionDeclaration.SpaceAfterFunction" />
+		<exclude name="PEAR.Functions.FunctionCallSignature.MultipleArguments" />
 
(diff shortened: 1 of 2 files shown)


dd32:
> RSS doesn't seem to handle future-dated content very well.. if at all.. as this is explicitly not allowed by the spec really

Atom feeds might help here, we can set `pubDate` to when we're first aware of an event (which annoyingly we don't store), and `Updated` to the date of when the event occurs.. it seems that some clients handle future dates in the updated field.

PR #1567, opened 2025-12-08
Revenue Report: Use a direct comparison on the timestamp field.

dd32:
This avoids MySQL needing to run `UNIX_TIMESTAMP()` on every value in the column when querying the table.

Untested as of yet.


dd32:
Included in #1579

PR #1569, opened 2025-12-17
Add 'Actual Attendees' and 'Series Event' to event post types

### LINKED ISSUE #1558: Event Trackers: Add Extra fields
### Description
Add two new fields to the Event Tracker for improved reporting and program metrics:

- `Actual Attendees` – numeric field, editable by organizers only after the event has taken place.
- `Series Event` – a tracker checkbox for event supporters, used to indicate when an event is part of a multi-venue or multi-session series (e.g., Campus Connect Ajmer workshops held across several campuses).

### Stakeholders
@Piyopiyo-Kitsune and the Community Team

### How does this new feature contribute to the goals
These fields will help measure event reach, growth, and engagement trends while reducing the need for manual reporting after events.
- Having attendee counts directly in the tracker ensures consistent, reliable data and makes it easier to identify successful event models.
- Being able to quickly identify “series” events can help highlight repeatable formats and measure collective impact across campuses or sessions.

### Acceptance Criteria
Actual Attendees: numeric; organizer can edit in their tracker after an event, available in report.
Series Event: event supporters can edit in tracker as a checkbox; available in report

(and if you want to be extra nice, have them auto-checked for the WPCC report 🙏 )

dd32:
Fixes #1558

Some magic sugar is added, in that the number is automatically calculated at event close based on either a) tickets sold, or b) if the attendee checkin app has used, how many have checked in.

I'm not 100% that setting this automatically is wanted, as this may cause organisers to not fill it in.. however, I make an assumption that it's more likely that it's going to be WordCamp Admins who are setting this field, so perhaps filling this in with the best estimate would be useful for reporting purposes? I might split that into a new PR.

### DIFF
--- a/public_html/wp-content/plugins/camptix-attendance/addons/attendance.php
+++ b/public_html/wp-content/plugins/camptix-attendance/addons/attendance.php
@@ -44,6 +44,8 @@ class CampTix_Attendance extends CampTix_Addon {
 			return;
 		}
 
+		add_action( 'tix_scheduled_daily', array( $this, 'cron_stats_update_attended_count' ) );
+
 		add_filter( 'wp_ajax_camptix-attendance', array( $this, 'ajax_callback' ) );
 		add_filter( 'wp_ajax_nopriv_camptix-attendance', array( $this, 'ajax_callback' ) );
 
@@ -94,6 +96,7 @@ class CampTix_Attendance extends CampTix_Addon {
 	 * Sets are removes the attended flag for a given camptix_id.
 	 */
 	public function _ajax_sync_model() {
+		global $camptix;
 		if ( empty( $_REQUEST['camptix_id'] ) )
 			return;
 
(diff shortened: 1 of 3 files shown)


supernovia:
> Some magic sugar is added, in that the number is automatically calculated at event close based on either a) tickets sold, or b) if the attendee checkin app has used, how many have checked in.

Oh wow! That sounds super nice, actually!  

> I might split that into a new PR

Would it make sense to do one for the fields, and one for the sugar?  Personally I'm A-OK with going ahead with both!

dd32:
I'm good with going ahead with it as is here, with the auto calculation. 

I thought I'd YOLO merged this before going afk, oops. If I have time I'll sneak on to do this, but otherwise after Jan 5th.

supernovia:
Thanks! Now that we're all back, if you have a minute to try it we can test to see how things go!

PR #1572, opened 2026-01-13
Update Meetup tracker sync

dd32:
This updates the Meetup listings to:
 - Include Lat / Lon coords for groups
 - ~Include the Member count on the overview~
 - Automatically remove a group from the chapter-post_status when it's no longer in existence
 - ~Properly exclude the WordPress organizer from the listings~

Split some items into #1573

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-admin.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-admin.php
@@ -371,8 +371,8 @@ if ( ! class_exists( 'Meetup_Admin' ) ) :
 			$event_hosts = array();
 			if ( isset( $group_leads ) && is_array( $group_leads ) ) {
 				foreach ( $group_leads as $event_host ) {
-					if ( WCPT_WORDPRESS_MEETUP_ID === $event_host['id'] ) {
-						// Skip WordPress admin user.
+					// Skip WordPress admin user.
+					if ( WCPT_WORDPRESS_MEETUP_ID === (int) $event_host['id'] ) {
 						continue;
 					}
 					$event_hosts[] = array(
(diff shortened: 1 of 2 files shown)


public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-admin.php
```diff
@@ -352,8 +358,37 @@ public static function update_meetup_data( $post_id ) {
...
+
+				wp_update_post( array(
+					'ID'          => $post_id,
+					'post_status' => 'wcpt-mtp-canceled',
+				) );
+
+				$self_loop_prevent = false;
+
+				return;
+			}
```
dd32, line 388: I'm not actually sure I want to automatically remove groups from the post_status, but given the number of meetups who were listed as being active but no longer existed for whatever reason..

public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-admin.php
```diff
@@ -381,8 +416,30 @@ public static function update_meetup_data( $post_id ) {
...
+						),
+					)
+				);
+			}
+
 			update_post_meta( $post_id, 'Meetup Co-organizer names', $event_hosts );
 			update_post_meta( $post_id, 'Meetup Location (From meetup.com)', $group_details['localized_location'] );
+			update_post_meta( $post_id, 'Meetup Coordinates', "{$group_details['lat']},{$group_details['lon']}" );
+			update_post_meta( $post_id, '_lat', $group_details['lat'] );
+			update_post_meta( $post_id, '_lon', $group_details['lon'] );
```
dd32, line 442: I was adding this such that I could source the location of all WP Meetups via this listing.

I'm not sure that this is viable however, as there isn't a strict 1:1 mapping of groups to entries here, as the entries are more applications-for-meetups tracking, and there can be multiple entries for a singular active meetup.

Successfulsebunya:
I personally suggest that this remove be made manually or set a trigger to be able to do it properly. This will help on accidental deletions and further errors that could arise from the same

PR #1579, opened 2026-01-19
Optimize Camptix log for report queries

### LINKED ISSUE #1578: Ticket Revenue report times out
### Describe the bug

The Ticket revenue report times out with `No Data` for larger time periods:
https://central.wordcamp.org/wp-admin/index.php?page=wordcamp-reports&report=ticket-revenue

https://github.com/WordPress/wordcamp.org/blob/9a06191d66e146e9949905071786fc4876d19b73/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-ticket-revenue.php#L184-L187

This happens as the database query times out.

Brief investifation:
- `UNIX_TIMESTAMP( timestamp ) BETWEEN 1738281600 AND 1768867199` means it can't utilise any indexes on the timestamp column (There are none)
- The `section` column could be indexed to split the table that the queries need to scan
- The log table appears to contain a lot of irrelevant data, caused by bugs elsewhere that were never investigated ("Something is wrong" is the 5th most logged thing) or are logged way too much ("reviewed attendee timeouts" is 1/3 of the table, and appears to be logged for every historical event daily)

dd32:
See #1578

This PR:
 - Adjusts the table to have an index on the timestamp column
 - Uses a query syntax that can make use of that index

This resolves #1578 for small data sets, but still runs into limits on larger time frames.

The main issue is that the query is filtering by `message LIKE *X* OR message LIKE *Y*` which results in it needing to perform a full table scan of the matched rows.
A further index on the `message` field to group by the first X characters may help here, The first 10 char should be fairly non-unique set of data points. However, the query isn't anchored to the prefix so would be unable to use it at present.

### DIFF
--- a/public_html/wp-content/plugins/camptix-network-tools/camptix-network-tools.php
+++ b/public_html/wp-content/plugins/camptix-network-tools/camptix-network-tools.php
@@ -12,7 +12,7 @@
 
 class CampTix_Network_Tools {
 	private $options;
-	private $db_version = 20260119;
+	private $db_version = 20260121;
 	const PLUGIN_URL    = 'http://wordpress.org/plugins/camptix-network-tools';
 
 	function __construct() {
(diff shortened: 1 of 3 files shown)


dd32:
> needing to perform a full table scan of the matched rows.

As such, reducing the size of the table would also assist. As noted in the above ticket, there's a lot of irrelevant logs.. Roughly 40% of the logs are currently "Marked X attendees as timed out"

PR #1587, opened 2026-01-21
Call for Speakers: Mark non-logged-in submissions as spam.

### LINKED ISSUE #1586: Call for Speakers: Login required not working
### Describe the bug

Various jetpack forms on WordCamp are protected by a login requirement.
This reduces spam, as most pentesters and spambots aren't logged in (and when they are, we can combat it better)

It appears that this is no longer applying, the visuals show login as required, but bypassing the notice allows the form to work as normal.

Reported via slack: https://wordpress.slack.com/archives/C08M59V3P/p1768898321619449

### To reproduce

1. Go to Speaker Submission form
2. Edit HTML to bypass overlay
3. Submit form
4. See it in WordCamp backend.

### Expected behavior

An error to prevent the form submission, denoting that it requires login to process.

### Screenshots / Screencasts

Image

dd32:
Fixes #1586 

Props <username>, <username>

1.
2.
3.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
+++ b/public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
@@ -78,8 +78,7 @@ class WordCamp_Forms_To_Drafts {
 	 * @return string
 	 */
 	public function force_login_to_use_form( $content ) {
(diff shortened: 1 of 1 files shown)


dd32:
Jetpack doesn't seem to have any good logic to abort a contact form submission, the best I could find was marking it as spam.

Marking as spam seems like a good option here though.

PR #1589, opened 2026-01-28
Update list of fields included in WPCC reports

dd32:
This adds extra enabled-by-default fields to the WPCC report.

Props supernovia


dd32:
Merged this as part of fcc0d1ce3f228f6772608d8202f1cf4791ebee6d

PR #1590, opened 2026-01-28
Add a Meetup Sponsors report.

dd32:
This is a private report that lists meetup group sponsors.


public_html/wp-content/plugins/wordcamp-reports/classes/report/class-meetup-sponsors.php
```diff
@@ -0,0 +1,334 @@
...
+
+		$safe_html = [
+			'p' => [],
+			'img' => [
+				'src'    => true,
+				'alt'    => true,
+				'style'  => true,
+			],
+			'a' => [
+				'href'   => true,
```
dd32, line 221:
```suggestion
				'href'   => true,
				'rel'    => true,
				'target' => true,
```

PR #1615, opened 2026-02-13
Add EUR SEPA Transfer payment method

### LINKED ISSUE #1609: Payments: Add SEPA transfers
Related: #1608

### Description

We now have a EUR bank account based in Europe.

This brings two new possibilities for us:
 - Payments in EUR can be made directly to that account, Instructions were added in #1591
 - EUR Payments can be made from that account to other EUR bank accounts via SEPA, at lower cost and without currency conversion.

To make use of this we need to:
 - Allow Vendor Payments & Reimbursements to select SEPA as the payment method
 - Allow Accounting to export the approved EUR payments in the bulk-import format for the bank.
   - Exclude these matching EUR payments from the `Wire` export

I propose we rename the existing `Wire` Method to `Wire (SWIFT / IBAN)` to suggest that it uses the SWIFT Network.
I propose we add a new payment type listed of `EUR SEPA Transfer` _(I would call it a 'Direct Deposit', as US and AU do, but it looks like EU just calls it a 'Bank transfer' or 'SEPA transfer')_

This payment method is only available when:
 - The currency is EUR
 - The country selection is within the SEPA network 

The SEPA method would request:
 - Account Name
 - BIC
 - IBAN 
While it appears that BIC is generally optional, it appears as if it may be required by our bank, which is using 2013-era documentation.. Clarification will be saught.

The export format that we need to generate is: `SEPA Credit Transfer – ISO 20022 XML` not the custom CSV used for US accounts.
Noting:
> We only support pain 001.002.003 and pain 001.003.03 German format within the bounds of the EPC guidelines referenced above.
Which seems to be greatly outdated and replaced a few times over :) 

### Stakeholders

Accounting

dd32:
## Summary
- Adds a new **EUR SEPA Transfer** payment method for Vendor Payments and Reimbursement Requests
- Conditionally shown only when currency is EUR and country is in the SEPA zone (36 countries)
- Collects Account Name, BIC, and IBAN (all encrypted at rest)
- Adds SEPA Credit Transfer ISO 20022 XML export (pain.001.003.03 format) for the network dashboard
- Debtor bank details are configurable via `wcb_sepa_debtor_bic` and `wcb_sepa_debtor_iban` filters
- SEPA payments are automatically excluded from Wire exports (different payment method value)

### Note on Wire rename
The issue proposes renaming "Wire" to "Wire (SWIFT / IBAN)". This was intentionally deferred because the payment method value is used as both the stored value and the display label throughout the codebase. Renaming it would require a data migration for existing Wire payments. This could be addressed in a follow-up with a proper migration strategy.

## Test plan
- [ ] Create a Vendor Payment, select EUR currency and a SEPA country (e.g., DE) — verify "EUR SEPA Transfer" radio option appears
- [ ] Select a non-EUR currency — verify SEPA option is hidden
- [ ] Select EUR with a non-SEPA country — verify SEPA option is hidden
- [ ] Select EUR SEPA Transfer, fill in Account Name, BIC, IBAN — verify fields save and are encrypted
- [ ] Repeat the above for Reimbursement Requests
- [ ] Export SEPA payments from the network dashboard — verify valid pain.001.003.03 XML output
- [ ] Export Wire payments — verify SEPA entries are not included

Closes #1609

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
@@ -280,7 +280,7 @@ class WordCamp_Budgets {
 		$methods = array(
 			'Direct Deposit' => __( 'Direct Deposit', 'wordcamporg' ),
 			'Check'          => __( 'Check', 'wordcamporg' ),
-			'sepa_transfer'  => __( 'EUR SEPA Transfer', 'wordcamporg' ),
+			'sepa_transfer'  => __( 'Euro SEPA Transfer', 'wordcamporg' ),
 			'Wire'           => __( 'Wire', 'wordcamporg' ),
 		);
 
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
@@ -267,20 +267,52 @@ class WordCamp_Budgets {
 	}
 
 	/**
-	 * Get a list of valid payment methods
+	 * Get payment methods as slug => label pairs.
 	 *
-	 * @param $post_type
+	 * Existing payment method slugs retain their original string values
+	 * for backward compatibility with stored post meta.
 	 *
-	 * @return array
+	 * @param string $post_type
+	 *
+	 * @return array Associative array of slug => translated label.
 	 */
-	public static function get_valid_payment_methods( $post_type ) {
-		$methods = array( 'Direct Deposit', 'Check', 'Wire', 'EUR SEPA Transfer' );
+	public static function get_payment_methods( $post_type ) {
+		$methods = array(
+			'Direct Deposit' => __( 'Direct Deposit', 'wordcamporg' ),
+			'Check'          => __( 'Check', 'wordcamporg' ),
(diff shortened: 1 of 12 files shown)


public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
```diff
@@ -274,7 +274,7 @@ public static function validate_amount( $amount ) {
 	 * @return array
 	 */
 	public static function get_valid_payment_methods( $post_type ) {
-		$methods = array( 'Direct Deposit', 'Check', 'Wire' );
+		$methods = array( 'Direct Deposit', 'Check', 'Wire', 'EUR SEPA Transfer' );
```
dd32: @Claude Let's abstract this out to a new Payment Methods method somewhere, which returns a 'slug' and 'textual label' such that we can have these labels translated.

For backwards compat, the existing methods slugs should retain as their existing values.

The slug for the new EUR SEPA transfer should be 'sepa_transfer'.

dd32: Done in 5cb94b5e. Added `get_payment_methods()` returning slug => translatable label pairs, with existing methods retaining their current strings as slugs and the new SEPA transfer using `'sepa_transfer'`. Also added `get_payment_method_label()` helper for the network list tables.

Issue #1588, opened 2026-01-28
Event site's old URL should redirect to the new URL when renamed


Piyopiyo-Kitsune:
### Describe the bug

https://events.wordpress.org/campusconnect/2025/cebu/ was renamed to https://events.wordpress.org/campusconnect/2026/cebu/, but https://events.wordpress.org/campusconnect/2025/cebu/ does not redirect to the new address. 

https://events.wordpress.org/campusconnect/2025/cebu/ should redirect to the latest year event if no site exists at 2025.

### Expected behavior

The old site redirects to the new site when the URL is renamed.

dd32:
#1613 handles when a site is renamed due to a typo, ie. `https://events.wordpress.org/TEST/2025/gatherpress/` => `https://events.wordpress.org/testing/2025/gatherpress/`

#1623 handles accessing a non-existent site, where location+type has happened later, ie. `https://events.wordpress.org/testing/2000/gatherpress/` but something like `https://events.wordpress.org/TEST/2000/gatherpress/` will still 404.

PR #1623, opened 2026-02-23
Redirect old event URLs to the latest year for the same city/type

dd32:
## Summary
- Adds `get_latest_event_url()` to `sunrise-events.php` that redirects old event URLs (e.g., `/rome/2023/training/`) to the latest year with the same city/type combination (e.g., `/rome/2025/training/`)
- Wired into `set_network_and_site()` **after** the renamed-site check, before the NOBLOGREDIRECT fallback
- Added integration tests in `test-sunrise-events.php` using `Database_TestCase`

Part of #1588

## Test plan
- [x] `get_latest_event_url()` returns redirect URL when an older year is requested and a newer year exists
- [x] `get_latest_event_url()` returns `false` when the current/latest year is requested (no redirect loop)
- [x] `get_latest_event_url()` returns `false` for unknown cities
- [x] `get_latest_event_url()` returns `false` for non-matching paths


jonoalderson:
Yessss!

PR #1626, opened 2026-02-26
Stripe: Clarify that EUR also supports non-Credit Card payments.

dd32:
We also support a few net banking options via Stripe in certain countries.

See #1625

I'm not clear on what the terminology should be for EUR here, `banking` feels too generic.

AI keeps wanting to say SEPA, but it's not SEPA :) (I wish we supported SEPA for ticket purchases..)

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/payment-stripe.php
+++ b/public_html/wp-content/plugins/camptix/addons/payment-stripe.php
@@ -57,10 +57,16 @@ class CampTix_Payment_Method_Stripe extends CampTix_Payment_Method {
 		add_action( 'template_redirect', array( $this, 'template_redirect' ) );
 		add_action( 'camptix_pre_attendee_timeout', array( $this, 'pre_attendee_timeout' ) );
 
-		// Use specific name for INR as we support UPI via Stripe.
-		if ( 'INR' === $this->camptix_options['currency'] ?? '' ) {
-			$this->name = 'Credit Card or UPI (Stripe)';
-			$this->description = 'Credit card and UPI processing, powered by Stripe.';
+		// Use specific name for EUR/INR as we support some local payment methods via Stripe.
+		switch ( $this->camptix_options['currency'] ?? '' ) {
+			case 'INR':
+				$this->name = 'Credit Card or UPI (Stripe)';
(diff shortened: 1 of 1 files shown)


dd32:
I don't think we run any GeoIP on requests to WordCamp, I was going to suggest having it be country-specific... 

- iDEAL - Netherlands
- Bancontact - Belgium
- EPS - Austria

Showing `banking` to anyone outside of those countries is pretty pointless.

Maybe it's fine to leave it as-is, as Credit Card is the primary method for the rest of EUR.

gedex:
Merged the `production` branch in to fix the conflict, then followed the point above about GeoIP: the Stripe name and description now name iDEAL, Bancontact or EPS only when the event venue is in NL, BE or AT, and every other EUR event keeps "Credit Card (Stripe)". Tested locally for NL, BE, AT, DE and no country; details in the description.

<kbd>
image
</kbd>

public_html/wp-content/plugins/camptix/addons/payment-stripe.php
```diff
@@ -62,10 +62,28 @@ public function camptix_init() {
...
+			case 'EUR':
+				// iDEAL, Bancontact and EPS each only work in one country, so only name the
+				// one that works where the event is held. Other EUR events keep the default.
+				$local_methods = array(
+					'NL' => 'iDEAL',
+					'BE' => 'Bancontact',
+					'AT' => 'EPS',
+				);
+				$wordcamp      = function_exists( 'get_wordcamp_post' ) ? get_wordcamp_post() : false;
+				$country       = $wordcamp ? strtoupper( $wordcamp->meta['_venue_country_code'][0] ?? '' ) : '';
```
mcliwanow: I didn't find a way to get user's country. I think venue's country code might be a good proxy to show the country specific payment methods 👍 

I'd even consider showing all of them (iDEAL, EPS, Bancontact) for all EUR payments. If you're not from a country supporting those, it likely doesn't hurt to see them. If you're purchasing from a country that is supporting one of them, but it's different from venue's country, then it might be helpful to see a familiar payment method.

I don't have super strong feelings about it. Just wanted to mention an alternative option.

@gedex, which one do you prefer?

PR #1631, opened 2026-03-03
Improve SEPA transfer form fields and IBAN handling

### LINKED ISSUE #1609: Payments: Add SEPA transfers

dd32:
## Summary
- Renames "Account Name" to "Account Holder Name" for clarity
- Adds placeholder text to all SEPA fields (Account Holder Name, BIC, IBAN) to guide users on expected format
- Enforces IBAN minimum length of 15 characters (Norway has the shortest IBAN) via both HTML `minlength` and JS regex validation
- Strips whitespace from IBAN and BIC values in `generate_sepa_xml()` as a safety net, ensuring IBANs entered with spaces are properly structured in the XML export

Fixes #1609

## Test plan
- [ ] Select Euro SEPA Transfer as payment method and verify the field is now labeled "Account Holder Name"
- [ ] Verify placeholder text appears in all three SEPA fields
- [ ] Enter an IBAN shorter than 15 characters and verify validation rejects it
- [ ] Enter an IBAN with spaces (e.g. `GB29 NWBK 6016 1331 9268 19`) and verify the SEPA XML export contains the IBAN without spaces
- [ ] Verify the same form improvements appear on reimbursement requests

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/payment-request.php
@@ -451,7 +451,7 @@ class WCP_Payment_Request {
 	 * @param bool    $readonly
 	 * @param bool    $required
 	 */
-	protected function render_text_input( $post, $label, $name, $description = '', $variant = 'text', $row_classes = array(), $readonly = false, $required = true ) {
+	protected function render_text_input( $post, $label, $name, $description = '', $variant = 'text', $row_classes = array(), $readonly = false, $required = true, $placeholder = '' ) {
 		$value = $this->get_field_value( $name, $post );
 		array_walk( $row_classes, 'sanitize_html_class' );
 		$row_classes = implode( ' ', $row_classes );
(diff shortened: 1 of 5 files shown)


dd32:
Screenshot 2026-03-03 at 1 39 00 pm

PR #1649, opened 2026-03-12
Match thumbnail aspect ratio to singular view banner

### LINKED ISSUE #1617: Central Schedule: The thumbnail aspect ratio should match the singular view
### Describe the bug

WordCamp Central has a schedule listing upcoming WordCamps / Events, it also has a singular view for each WordCamp.

The singular view is not directly linked to AFAIK.

The requirements specify a banner image for the singular page, but when an image of that size is uploaded, the thumbnail on the schedule listing doesn't match it's aspect ratio.

We should make them match.

Singular:
https://central.wordcamp.org/wordcamps/wordpress-campus-connect-stitnu-al-farabi/

Image

Listing:
https://central.wordcamp.org/schedule/
Image

Note the image sides are cut off.

This is partially because of the super wide banners, which don't make sense for thumbnails.

Historically, WordCamp banners had a lot of nothingness on the sides, to account for varying screen resolutions, where the sides would be progressively hidden on smaller screens.

As an example of what would happen if we pushed these thumnails to the same aspect ratio.. it'd require a 230px wide, by 37px tall image, which looks out of place for the 80% of events that never upload these banner assets
_(ignore the stretched images for other events)_
Image

### To reproduce

1. Upload a banner with 1852px wide × 320px tall as per requirements
2. View the above result.

### Expected behavior

The thumbnails and banner should look good.

### Solutions

This site needs a new coat of paint, seriouisly.

For affected events, I'm going to suggest creating images that target the thumbnail aspect ratio, rather than the banner aspect ratio..

dd32:
## Summary

- Changes the `wccentral-thumbnail-large` image size from 926x160 (5.79:1 ratio) to 926x418 (2.22:1 ratio) to match the `wccentral-thumbnail-small` thumbnail size (82x37, 2.22:1 ratio).
- This ensures the banner image on the singular WordCamp view uses the same aspect ratio as the schedule listing thumbnail, so uploaded images are not cropped differently between the two views.
- Existing images will need to be regenerated (via a thumbnails regeneration) for the new size to take effect.

## Context

The schedule listing at `central.wordcamp.org/schedule/` uses `wccentral-thumbnail-small` (82x37) while the singular view uses `wccentral-thumbnail-large` (926x160). These had very different aspect ratios (2.22:1 vs 5.79:1), causing banner images to appear cropped on the sides when shown as thumbnails.

Fixes #1617

## Test plan

- [ ] Verify the image size registration in `functions.php` uses the updated dimensions
- [ ] Upload a test featured image and confirm the generated `wccentral-thumbnail-large` crop matches the aspect ratio of `wccentral-thumbnail-small`
- [ ] Check the singular WordCamp view displays the banner image correctly at the new ratio
- [ ] Confirm the schedule listing thumbnails still display correctly

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/public_html/wp-content/themes/wordcamp-central-2012/functions.php
+++ b/public_html/wp-content/themes/wordcamp-central-2012/functions.php
@@ -54,7 +54,7 @@ class WordCamp_Central_Theme {
 
 		// Add some new image sizes, also site shot is 205x148, minimap is 130x70
 		add_image_size( 'wccentral-thumbnail-small', 82, 37, true );
-		add_image_size( 'wccentral-thumbnail-large', 926, 160, true );
+		add_image_size( 'wccentral-thumbnail-large', 926, 418, true );
 		add_image_size( 'wccentral-thumbnail-past', 130, 60, true );
 		add_image_size( 'wccentral-thumbnail-hero', 493, 315, true );
 


dd32:
Closing - submitted to wrong repo. Correct PR is on dd32/wordcamp.org.

PR #1651, opened 2026-03-12
Fix media library pagination for non-admin users

### LINKED ISSUE #1316: Non admin users can't view more than the first page in media library grid
### Describe the bug

Non-admin users are restricted from having the `load more` button active, so they cannot see more than the first page.

See: https://wordpress.slack.com/archives/C08M59V3P/p1715444166309679

### To reproduce

1. Go to a media library using an Editor level account e.g. https://europe.wordcamp.org/2024/wp-admin/upload.php, and look at attachment in `grid view`
2. Wait for screen to load with attachments.
3. Scroll down to bottom of page, and notice that the `load more` button is hidden (by CSS) and not active.

### Expected behavior

Users should be able to scroll through results with the button

Media library with Firefox

dd32:
## Summary

- Non-admin users (editors) could not load more than the first page of media items in the media library grid view (fixes #1316)
- Root cause: the `the_posts` filter in `wordcamp-payments/includes/privacy.php` removed payment attachments from query results *after* the query ran, returning fewer items than `posts_per_page`, which made the media library JavaScript conclude there were no more items to load
- Fix: switch from `the_posts` to `posts_clauses` to exclude other users'  payment attachments at the SQL level, so `found_posts` and actual results stay consistent

## Test plan

- [ ] Log in as an editor (non-admin) user on a WordCamp site that has payment attachments
- [ ] Go to Media > Library and switch to grid view
- [ ] Verify that scrolling down loads more media items (infinite scroll / load more works)
- [ ] Verify that the "Showing X of Y media items" count is accurate
- [ ] Log in as an admin and verify payment attachments are still fully visible
- [ ] Verify that non-admin users still cannot see other users' payment file attachments

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
@@ -9,7 +9,7 @@ use WCP_Payment_Request;
 defined( 'WPINC' ) || die();
 
 
-add_filter( 'the_posts',                          __NAMESPACE__ . '\hide_others_payment_files', 10, 2 );
+add_filter( 'posts_clauses',                      __NAMESPACE__ . '\exclude_others_payment_files_from_query', 10, 2 );
 add_filter( 'wp_unique_filename',                 __NAMESPACE__ . '\obscure_payment_file_names', 10, 2 );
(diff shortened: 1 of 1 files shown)


dd32:
Closing - submitted to wrong repo. Correct PR is on dd32/wordcamp.org.

PR #1677, opened 2026-03-24
Show error when attaching a file already used by another request

dd32:
## Summary

- When a user selects an existing media file that is already attached to a different payment/reimbursement request, the file was silently ignored. This caused confusion as the attachment appeared to not save.
- Displays an inline error notice explaining the file belongs to another request and prompting the user to upload a fresh copy instead.
- Resolves a long-standing TODO from 2016.

## Context

- Original Slack discussion (2016): https://wordpress.slack.com/archives/meta-wordcamp/p1459185670000179
- Recent report: https://wordpress.slack.com/archives/C08M59V3P/p1774274020560509

## Test plan

- [x] Create a payment or reimbursement request and attach a file
- [x] Create a second request
- [x] On the second request, open the media modal and select the file already attached to the first request
- [x] Verify an inline error notice appears below the files list explaining the file is already attached elsewhere
- [x] Verify uploading a fresh file to the second request works normally


dd32:
The attachment shouldn't show as attached if it's not attached:

Screenshot 2026-03-24 at 2 22 55 pm

PR #1715, opened 2026-05-08
CampTix: Fire camptix_payment_result on every payment_result call

### LINKED ISSUE #1713: [Bug] Invoice generation broken since May 5, tickets purchased with invoice request not generating or sending invoices
### Summary

Since **Tuesday, May 5**, invoices are no longer being generated or sent for ticket purchases where the attendee requested an invoice. The billing data is being saved correctly, but the generation and delivery steps appear to be silently failing.

This has been confirmed across at least two events:
- WordCamp Portugal
- WordCamp Europe

---

### Steps to reproduce

1. Purchase a ticket on a WordCamp event site.
2. Check the "I need an invoice" option and fill in the billing details.
3. Complete the purchase.
4. **Expected:** Invoice is generated and sent to the attendee.
5. **Actual:** Billing data is saved, but no invoice is generated or sent.

---

### Example ticket

The following order has billing details filled in but no invoice was generated:

> https://portugal.wordcamp.org/2026/wp-admin/post.php?post=2659&action=edit
> (also IDs 2660, 2689, 2690, 2692, 2693, 2727, 2728, 2735, 2745, 2772, 2775, 2776, 2780, 2819, 2821, 2822, 2823, 2838)

> List of invoices generated and sent: https://portugal.wordcamp.org/2026/wp-admin/edit.php?post_type=tix_invoice
---

### Impact

- Affects all ticket purchases with invoice requests from **May 5 onwards**.
- Dozens of pending invoices across at minimum WordCamp Portugal and WordCamp Europe.
- Attendees have not received invoices they are legally/fiscally entitled to.

---

### Priority

**High / Urgent** — This affects fiscal compliance for attendees. Manual resolution of dozens of invoices is not scalable.

---

### Questions / Requests

1. @dd32 Was there a deployment or code change on May 5 that touched the invoice generation logic or the ticketing/payment flow?

2. Is there a way to **retroactively trigger invoice generation** for all affected orders? A manual fix order-by-order would be… entertaining.

3. If a bulk fix isn't available, could one be added as part of the fix? 

Huge thanks for looking into it!

**Props to Alvaro Gois do Santos** for finding the bug and informing about it!

wputasic:
Fire `camptix_payment_result` on every `payment_result()` call rather than only on status transitions, so addon listeners (notably `CampTix_Addon_Invoices::maybe_create_invoice`) get a chance to run when the centralized Stripe webhook (#1705) and the interactive return path race for the same payment token. Add an idempotency guard in `maybe_create_invoice` so the now-possibly-double-firing action doesn't mint duplicate invoices.

Fixes #1713

### Background

Since #1705, two callers can invoke `CampTix_Plugin::payment_result()` for the same payment token:

1. The buyer's interactive return on the event site (`payment_return` template_redirect, `$interactive = true`).
2. The centralized Stripe webhook running on the central site after `switch_to_blog( $site_id )` (`$interactive = false`).

`payment_result()` previously fired `do_action( 'camptix_payment_result', … )` only when `$status_changed === true`. Whichever caller runs first transitions the attendee `draft → publish`; the second sees `$status_changed === false` and returns without firing the action. `CampTix_Addon_Invoices::maybe_create_invoice` is the only listener on this hook, so no `tix_invoice` post was created.

#1708 (option-loading refactor) made the webhook reliably win the race — the attendee is now in `publish` before the buyer's browser redirects back. This is the source of the regression reported in #1713.

1. On a multisite dev environment with CampTix and `camptix-invoices` active on an event site, configure Stripe to point to the centralized webhook at `/wp-json/camptix/v1/stripe-webhook` on the central site.
2. Purchase a ticket on the event site, check "I need an invoice", complete payment via Stripe.
3. Confirm exactly one `tix_invoice` post is created and the invoice email is sent.
4. Repeat with the buyer closing the tab before redirect (webhook-only path) — invoice still created.
5. Repeat with the Stripe webhook disabled (interactive-only path) — invoice still created.
6. Replay the same Stripe webhook event a second time — only one `tix_invoice` exists (idempotency guard in `maybe_create_invoice`).

### DIFF
--- a/public_html/wp-content/plugins/camptix-invoices/includes/class-camptix-addon-invoices.php
+++ b/public_html/wp-content/plugins/camptix-invoices/includes/class-camptix-addon-invoices.php
@@ -222,35 +222,30 @@ class CampTix_Addon_Invoices extends \CampTix_Addon {
 			return;
 		}
 
-		// Bail if we've already created an invoice for this transaction.
-		// payment_result() can fire camptix_payment_result more than once for the
-		// same payment_token (centralized Stripe webhook + interactive return).
-		$txn_id = get_post_meta( $attendees[0]->ID, 'tix_transaction_id', true );
-		if ( $txn_id ) {
-			$existing = get_posts(
-				array(
-					'post_type'      => 'tix_invoice',
-					'post_status'    => 'any',
-					'posts_per_page' => 1,
-					'fields'         => 'ids',
-					'meta_query'     => array( // @codingStandardsIgnoreLine
-						array(
-							'key'   => 'transaction_id',
-							'value' => $txn_id,
-						),
-					),
-				)
-			);
-			if ( $existing ) {
-				return;
-			}
+		// Race-safe idempotency guard: payment_result() can fire camptix_payment_result
+		// more than once for the same payment_token (centralized Stripe webhook + interactive
+		// return run within ~1-2 seconds of each other). add_post_meta() with $unique=true
+		// is rejected by WordPress if the key already exists for this post, so only the first
+		// caller successfully claims invoice creation; concurrent callers bail. Keyed on the
+		// attendee rather than on tix_transaction_id so that no-transaction purchases
+		// (100% coupon, Stripe payment_status=no_payment_required) are also covered.
+		if ( get_post_meta( $attendees[0]->ID, '_invoice_id', true ) ) {
+			return;
+		}
+		if ( ! add_post_meta( $attendees[0]->ID, '_invoice_id', 0, true ) ) {
+			return;
 		}
 
 		$order      = get_post_meta( $attendees[0]->ID, 'tix_order', true );
 		$invoice_id = self::create_invoice( $attendees[0], $order, $metas );
-		if ( ! is_wp_error( $invoice_id ) && ! empty( $invoice_id ) ) {
-			self::send_invoice( $invoice_id );
+		if ( is_wp_error( $invoice_id ) || empty( $invoice_id ) ) {
+			// Release the lock so a later call can retry.
+			delete_post_meta( $attendees[0]->ID, '_invoice_id' );
+			return;
 		}
+
+		update_post_meta( $attendees[0]->ID, '_invoice_id', $invoice_id );
+		self::send_invoice( $invoice_id );
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


public_html/wp-content/plugins/camptix/camptix.php
```diff
@@ -6691,6 +6691,13 @@ function payment_result( $payment_token, $result, $data = array(), $interactive
 		$from_status = $attendees_status;
 		$to_status = $attendees[0]->post_status;
 
+		// Fire the result action on every call, not just on status transitions.
+		// Webhooks and interactive returns can both invoke payment_result() for
+		// the same payment_token; whichever runs second sees $status_changed === false.
+		// Listeners (e.g. CampTix_Addon_Invoices::maybe_create_invoice) must be
+		// idempotent — see the matching guard there.
+		do_action( 'camptix_payment_result', $payment_token, $result, $data );
```
dd32: I think we should add the `$status_changed` param here too, such that other listeners can be informed whether we think it's changed.

```suggestion
		do_action( 'camptix_payment_result', $payment_token, $result, $data, $status_changed );
```

public_html/wp-content/plugins/camptix-invoices/includes/class-camptix-addon-invoices.php
```diff
@@ -218,13 +218,39 @@ public static function maybe_create_invoice( $payment_token, $result ) {
...
-			$invoice_id = self::create_invoice( $attendees[0], $order, $metas );
-			if ( ! is_wp_error( $invoice_id ) && ! empty( $invoice_id ) ) {
-				self::send_invoice( $invoice_id );
-			}//end if
-		}//end if
+		if ( ! $metas ) {
+			return;
+		}
+
+		// Bail if we've already created an invoice for this transaction.
```
dd32: Due to the length of time it takes for the invoice creation process to run, I think we should do something like
```php
if (
  false !== get_post_meta( $post, '_invoice_id' ) ||
  ! add_post_meta( $post, '_invoice_id', 0, true )
) {
  return;
}
```

such that only one process attempts to generate the invoice.

The get_posts() check will only cause it to return after creation, but due to how close the webhook and interactive returns run (1-2 seconds) it seems likely that this'll be a real racey situation.

PR #1735, opened 2026-05-26
Latest Site Hints: Resolve the next edition from schedule blogmeta

dd32:
## Problem

The "Check out the next edition" banner on past WordCamp sites, and the canonical bare-domain redirect (`city.wordcamp.org` → newest year), both picked the **highest-numbered year path** for a city. When a site for a future edition is created in advance but not yet scheduled, they pointed at that empty placeholder.

Reported case: `europe.wordcamp.org/2025/` linked to `/2027/` (status `needs-schedule`, no dates) instead of the real next edition `/2026/`. Same class of bug for other cities (e.g. `sofia.wordcamp.org`).

The canonical redirect previously worked around this with a **hardcoded per-flagship date guard** (#1736 shared that guard between the redirect and the banner). It exists because `get_canonical_year_url()` runs during the `sunrise.php` bootstrap — *before* the `wordcamp` post type is registered — so it couldn't read each camp's schedule. This PR removes the hardcode by making the schedule queryable that early.

## Approach

Denormalize each WordCamp's schedule onto its **site's blogmeta** (same mechanism as `old_home_url`). A new `WordCamp\Schedule_Meta` mu-plugin writes, whenever a `wordcamp` post is saved:

| Key | Value |
|---|---|
| `_wc_event_start` | Start Date (Unix ts). Written **only** for live, non-cancelled scheduled editions. |
| `_wc_event_end` | End Date, falling back to Start Date (Unix ts). Always written alongside the start. |
| `_wc_event_status` | The `wordcamp` post status. Observability/flexibility only; not load-bearing. |

The writer is **status-aware**: a scheduled-then-cancelled camp keeps its dates on the post, but its date keys are deleted, so downstream "has a start date" reliably means "live edition." It's idempotent (full re-derivation on every `save_post`), so quick/bulk edits, the auto-close cron, and programmatic updates all stay in sync.

Both readers now resolve the right edition from real dates:

- **`get_canonical_year_url()`** → `get_current_edition_site() ?: get_latest_site()`. Prefers the current edition (upcoming or finished within the last month) over a newer unscheduled placeholder, else the newest site. **`get_flagship_canonical_url()` and the hardcoded dates are removed.**
- **`get_latest_home_url()`** (banner) → same decision via a `_wc_event_end` JOIN, dropping the `switch_to_blog()` + `get_posts()` cross-blog lookup.

`get_current_edition_site()` keeps the SQL simple — one `_wc_event_end` JOIN, `event_end.meta_value + %d >= %d` with no CAST (MySQL coerces the digit-string timestamp), readable `blog`/`event_end` aliases.

`wp wordcamp-schedule-meta backfill` seeds the meta for existing camps.

## Deploy ordering

1. Ship the writer, 2. run the backfill, 3. then the readers go live. Until backfilled, a site with no meta simply looks like a placeholder and the readers fall back to "newest site" — i.e. **prior behavior**, never a fatal.

## Tests

Run against the WordCamp test DB harness (PHP 8.3, multisite), all green:

- `test-wordcamp-schedule-meta.php` (4) — writer writes dates for live camps, falls back end→start, **removes** dates on cancellation, skips camps with no linked site.
- `test-latest-site-hints.php` (14) — existing cases plus new placeholder-skip / fall-through.
- `test-sunrise.php` (71) — existing canonical cases (unchanged behavior with no schedule meta) plus the new placeholder-skip case.

With no schedule meta, both readers fall back to the newest site, so the existing data-provider tests are unaffected. `phpcs` clean on changed lines.

> Supersedes the hardcoded flagship guard from #1736 (merged into `production` and merged in here).

### DIFF
--- a/public_html/wp-content/mu-plugins/latest-site-hints.php
+++ b/public_html/wp-content/mu-plugins/latest-site-hints.php
@@ -1,7 +1,6 @@
 <?php
 
 namespace WordCamp\Latest_Site_Hints;
-use function WordCamp\Sunrise\get_flagship_canonical_url;
 use const WordCamp\Sunrise\{ PATTERN_YEAR_DOT_CITY_DOMAIN_PATH, PATTERN_CITY_SLASH_YEAR_DOMAIN_PATH, PATTERN_CITY_YEAR_TYPE_PATH };
 
 defined( 'WPINC' ) || die();
@@ -154,10 +153,33 @@ function show_notification_about_latest_site() {
  * @return bool|string
  */
 function get_latest_home_url( $current_domain, $current_path ) {
-	global $wpdb;
+	/*
+	 * `maybe_add_latest_site_hints()`, `canonical_link_past_home_pages_to_current_year()`, and
+	 * `show_notification_about_latest_site()` each call this during the same request, so memoize the result
+	 * to avoid repeating the database work.
+	 */
+	static $cache = array();
+	$cache_key = $current_domain . '|' . $current_path;
 
-	$wordcamp = get_wordcamp_post();
-	$end_date = absint( $wordcamp->meta['End Date (YYYY-mm-dd)'][0] ?? 0 );
+	if ( ! array_key_exists( $cache_key, $cache ) ) {
+		$cache[ $cache_key ] = determine_latest_home_url( $current_domain, $current_path );
+	}
+
+	return $cache[ $cache_key ];
+}
+
+/**
+ * Resolve the home URL of the most recent relevant event in a city.
+ *
+ * See `get_latest_home_url()`, which memoizes this.
+ *
+ * @param string $current_domain
+ * @param string $current_path
+ *
+ * @return bool|string
+ */
+function determine_latest_home_url( $current_domain, $current_path ) {
+	global $wpdb;
 
 	/**
 	 * In rare cases, the site for next year's camp will be created before this year's camp is over. When that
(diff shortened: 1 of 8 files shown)


dd32:
Rebuilt this branch on the latest `production` (now includes #1736 and #1737) and removed an accidentally-committed `composer.lock`. This direction supersedes the hardcoded flagship guard, so the now-dead `get_flagship_canonical_url()` (moved to `sunrise.php` in #1737) is removed here.

Addressed the Copilot review:

- **Writer was too loose** (`$is_live`): now gated on an allowlist of the public statuses (`wcpt-scheduled`/`wcpt-closed`) rather than "has a date and not cancelled", so pre-planning camps with tentative dates and cancelled/rejected camps are excluded. Added a test asserting a `wcpt-needs-schedule` camp with a Start Date does **not** get `_wc_event_*` written.
- **Repeated DB work**: `get_latest_home_url()` now memoizes per request (it's called 3× — banner, canonical link, notification).
- **Blog-status filters**: the banner candidate queries now filter `( public AND NOT deleted )`, matching `get_latest_site()`, so deleted/non-public placeholders aren't linked.
- **Scheduled-vs-placeholder coverage**: added placeholder-skip and fall-through tests for both the banner and the sunrise canonical path.

The `SUBSTR(domain,1,4) REGEXP` comment Copilot flagged is pre-existing and unrelated to this change, so I left it alone.

Full suite passes locally (PHP 8.3, multisite): writer 5, banner 14, sunrise 71. `phpcs` clean on changed lines.

PR #1741, opened 2026-06-02
BD: Addition of SurjoPay payment gateway

nazmulch11:
Props to @tareq1988

### How to test the changes in this pull request:

1. Updated the SSLCommerz integration to the latest version following the v4 documentation.
2. Added support for SurjoPay, another popular payment gateway in Bangladesh.
3. Fixed several common issues and updated the codebase to the latest standards.
4. Changed the default Docker port from 80 to 8888, as port 80 is often occupied by other applications on users' machines.
5. Updated the README documentation to provide clearer setup instructions and ensure future maintainability.


dd32:
## Codex-powered security review

Reviewed current PR head `e7c79bed` against `production` base `cb09ad45`. I would not merge this yet. Some earlier bot findings appear fixed in the current head, but there are still security and operational risks.

> **Claude:** I verified each point against the current head. Per-point notes are interleaved below (prefixed **Claude — my review of this point**), and committable code suggestions are posted as inline review comments on this PR.

### Blockers

1. **SurjoPay production API endpoint should not be manually configurable by site admins.**  
   `public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-surjopay.php:63-72` builds auth/checkout/verification URLs from saved `api_endpoint`, and `:580-584` only applies `esc_url_raw()`. `Base_Gateway::api()` then POSTs gateway credentials and attendee PII to that URL. If CampTix payment settings are editable by site/event admins, this allows arbitrary outbound requests and credential/PII exfiltration.

   The shurjoPay REST API docs list fixed base URLs:
   - Production: `https://engine.shurjopayment.com/api`
   - Sandbox: `https://sandbox.shurjopayment.com/api`

   Source: https://web2.shurjopay.com.bd/frontend-developers/direct-integration-with-api

   Recommendation: remove the free-form `api_endpoint` setting and hard-code the known base URL from `sandbox` mode, then append `/get_token`, `/secret-pay`, and `/verification`. The admin settings should only need username, password, prefix/merchant prefix, and sandbox mode. If shurjoPay has exceptional per-merchant endpoints, handle that through centrally managed config or a strict allowlist, not arbitrary admin input.

   > **Claude — my review of this point:** Confirmed — real, keep it a blocker. The strongest risk is SSRF (an arbitrary authenticated outbound POST from the server, able to reach internal hosts) plus attendee PII flowing to an attacker-controlled host; the credential-exfil angle is weaker since the admin sets those credentials themselves. SSLCommerz already hard-codes both its sandbox and production base URLs, so SurjoPay should match that. Inline suggestion posted.


2. **Redirect allowlisting is bypassable.**  
   `class-gateway-surjopay.php:198-207` accepts any host that `str_ends_with( $host, 'shurjopayment.com' )`, so `evilshurjopayment.com` would pass. SSLCommerz also still redirects to `GatewayPageURL` via `wp_redirect()` without host validation at `class-gateway-sslcommerz.php:176-188`.

   Recommendation: require `https`, exact host or dot-suffix checks such as `host === 'shurjopayment.com' || str_ends_with( host, '.shurjopayment.com' )`, and equivalent SSLCommerz host allowlisting.

   > **Claude — my review of this point:** Confirmed — `evilshurjopayment.com` satisfies `str_ends_with( $host, 'shurjopayment.com' )`. Lower severity on its own (the URL comes from the authenticated gateway response) but directly exploitable once combined with point 1. Inline suggestion (require https + dot-suffix host) posted.


3. **SurjoPay success verification is not bound tightly enough to the local order.**  
   `verify_payment()` checks `sp_code === 1000` and only compares `order_id` if the response includes it (`class-gateway-surjopay.php:414-456`). It does not require `order_id`, amount, currency, store ID, or payable status to match the local CampTix order before marking attendees paid at `:347-350` and `:480-484`.

   Recommendation: require order ID equality and verify amount/currency/store against `get_order( $payment_token )` before `PAYMENT_STATUS_COMPLETED`.

   > **Claude — my review of this point:** Partly — I'd narrow it. The order_id↔token binding you're asking for already exists: `payment_return()` runs `payment_token_matches_order_id()` (a `hash_equals()` against the stored `tix_surjopay_order_id`) *before* calling `verify_payment()`. The genuinely missing control is **amount/currency** verification before `PAYMENT_STATUS_COMPLETED` — note SSLCommerz already does `(float) $order['total'] === (float) $response->amount`. So scope this down to the amount/currency check. Inline suggestion posted.


4. **SurjoPay cancellation can be triggered without a matching gateway order ID.**  
   `payment_cancel()` only checks `payment_token`; if `order_id` is omitted, it proceeds to `PAYMENT_STATUS_CANCELLED` (`class-gateway-surjopay.php:375-404`). A leaked token could cancel an unpaid draft order.

   Recommendation: require the saved `tix_surjopay_order_id` to match for cancellation, or verify the order status with shurjoPay before changing CampTix status.

   > **Claude — my review of this point:** Confirmed, low severity (it only cancels a draft that would time out anyway, and needs a valid `payment_token`). Cheap fix — require the `order_id` to match. Inline suggestion posted.


5. **Gateway responses are logged/stored raw in several paths.**  
   `Base_Gateway::api()` logs non-2xx response bodies (`class-gateway-base.php:177-182`). SurjoPay logs raw auth/checkout/verification responses and stores raw verification objects in `tix_transaction_details` (`class-gateway-surjopay.php:103-105`, `:177-179`, `:436-451`, `:347-350`, `:480-484`). These responses may contain tokens, customer PII, card/payment metadata, or gateway diagnostics.

   Recommendation: add a SurjoPay-specific `prepare_transaction_for_log()` and use it for all logs and `payment_result()` data; never log bearer tokens, credentials, phone/email, or raw gateway bodies.

   > **Claude — my review of this point:** Partly — I'd narrow it. Storing raw transaction **PII** is acceptable here: these are ticket purchases tied to an attendee whose name/email/phone we already store, so the verification object in `tix_transaction_details` isn't a new exposure. The real issue is **secrets**: `Base_Gateway::api()` logs raw non-2xx bodies and SurjoPay never calls the existing `prepare_transaction_for_log()` helper (SSLCommerz does), so bearer tokens / credentials can reach the logs. Recommend redacting tokens/credentials in logs (reuse the base helper); no need to strip attendee PII from stored transaction details. Inline suggestion posted on the auth-log line.


6. **Root `composer.lock` is added without a `composer.json` change.**  
   This PR adds a 5,607-line lockfile while `composer.json` is unchanged. That changes dependency/deploy behavior far outside the payment-gateway scope and should not ride along with this security-sensitive feature.

   Recommendation: remove it from this PR or split it into a dedicated dependency-lock PR with CI/deploy validation and dependency audit output.

   > **Claude — my review of this point:** Agree — out of scope. One more unrelated change to split out: `mu-plugins/theme-templates/bootstrap.php` adds a `did_action( 'wp' )` guard to `inject_offline_template()`, which has nothing to do with BD payments.


### Additional merge risks

- `readme.md:22-26` and `:55-59` still claim a ticket hold / auto-release system, but the current PR removed that code. This can mislead reviewers/operators about oversell protection during offsite payments.
- `docker-compose.yaml:14-16` maps `8888:8888`; if nginx still listens on container port 80, local testing breaks. Use `8888:80` or update nginx.
- `class-phone-field.php:172-185` now enforces BD phone collection for every registration while the addon is active, and `:244-267` exports it. This is a privacy/data-minimization change and should be explicitly justified and documented.
- SurjoPay password settings render through `field_text` (`class-gateway-surjopay.php:545-549`), so credentials are shown as plain text in the admin UI. Prefer a password field / secret-handling pattern.

> **Claude — on the additional risks:** Agree on the readme, docker port, and plaintext-password items. On the phone field I'd rank it a touch higher than "privacy/data-min": `validate_phone_on_register()` enforces **BD-only** formats for *every* attendee on *every* camp where the addon is active, so it's also a functional regression that would block non-BD attendees — not only a data-minimisation question.

### Notes

Current head appears to have fixed earlier findings around hardcoded README credentials, predictable SurjoPay order IDs, SSLCommerz signature failure handling, and removal of the ticket-hold MU-plugin from this PR.

Verification performed: reviewed PR metadata/comments/diff, inspected current PR head locally, ran `git diff --check`, and ran `php -l` on the changed PHP files.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-surjopay.php
```diff
@@ -0,0 +1,588 @@
...
+
+	/**
+	 * Build API endpoint URLs based on sandbox/production mode
+	 */
+	private function setup_api_urls() {
+		$is_sandbox = ! empty( $this->options['sandbox'] );
+
+		$this->api_endpoint = $is_sandbox
+			? 'https://sandbox.shurjopayment.com'
+			: untrailingslashit( $this->options['api_endpoint'] ?? '' );
```
dd32: **Claude suggestion (Blocker 1)** — hard-code the production base URL instead of reading a free-form admin setting (mirrors how SSLCommerz hard-codes both hosts). You can then also drop the `api_endpoint` settings field and its `validate_options()` line.

```suggestion
		$this->api_endpoint = $is_sandbox
			? 'https://sandbox.shurjopayment.com'
			: 'https://engine.shurjopayment.com';
```

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-surjopay.php
```diff
@@ -0,0 +1,588 @@
...
+			array(
+				'order_id' => $order_id,
+				'amount'   => $order['total'],
+			)
+		);
+
+		// Validate the checkout URL host is a known shurjoPay domain before redirecting.
+		$checkout_url = esc_url_raw( $response->checkout_url );
+		$host         = wp_parse_url( $checkout_url, PHP_URL_HOST );
+		if ( ! $host || ! str_ends_with( $host, 'shurjopayment.com' ) ) {
```
dd32: **Claude suggestion (Blocker 2)** — require https and an exact/dot-suffix host match so `evilshurjopayment.com` can't pass.

```suggestion
		$checkout_url = esc_url_raw( $response->checkout_url );
		$scheme       = wp_parse_url( $checkout_url, PHP_URL_SCHEME );
		$host         = (string) wp_parse_url( $checkout_url, PHP_URL_HOST );
		$host_allowed = ( 'shurjopayment.com' === $host || str_ends_with( $host, '.shurjopayment.com' ) );
		if ( 'https' !== $scheme || ! $host_allowed ) {
```

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-surjopay.php
```diff
@@ -0,0 +1,588 @@
...
+		// Authenticate to get a fresh token for verification.
+		if ( ! $this->authenticate() ) {
+			$camptix->error( 'Payment verification authentication failed.' );
+			return;
+		}
+
+		// Verify the payment with shurjoPay.
+		$verified = $this->verify_payment( $order_id );
+
+		if ( $verified ) {
```
dd32: **Claude suggestion (Blocker 3)** — the order-id↔token binding already exists upstream via `payment_token_matches_order_id()`; the missing control is amount/currency. Confirm them against the local order before completing. ⚠️ Verify the shurjoPay verification response field names (`amount` / `currency`) against your account before committing.

```suggestion
		$verified = $this->verify_payment( $order_id );

		// Confirm the gateway-reported amount/currency match the local order before completing.
		$order = $this->get_order( $payment_token );
		if (
			$verified
			&& isset( $verified->amount, $verified->currency )
			&& (float) $order['total'] === (float) $verified->amount
			&& 'BDT' === strtoupper( (string) $verified->currency )
		) {
```

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-surjopay.php
```diff
@@ -0,0 +1,588 @@
...
+		global $camptix;
+
+		$payment_token = sanitize_text_field( trim( $_REQUEST['tix_payment_token'] ?? '' ) );
+		$order_id      = sanitize_text_field( trim( $_REQUEST['order_id'] ?? '' ) );
+
+		if ( ! $payment_token ) {
+			return;
+		}
+
+		if ( $order_id && ! $this->payment_token_matches_order_id( $payment_token, $order_id ) ) {
```
dd32: **Claude suggestion (Blocker 4)** — require a matching `order_id` for cancellation so a leaked token alone can't cancel a draft order.

```suggestion
		if ( ! $payment_token || ! $order_id ) {
			return;
		}

		if ( ! $this->payment_token_matches_order_id( $payment_token, $order_id ) ) {
```

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-surjopay.php
```diff
@@ -0,0 +1,588 @@
...
+			)
+		);
+
+		if ( ! $response ) {
+			$this->log( 'Surjo Pay: Authentication failed' );
+			return false;
+		}
+
+		if ( empty( $response->token ) || empty( $response->store_id ) ) {
+			$this->log( 'Surjo Pay: Invalid auth response', null, (array) $response );
```
dd32: **Claude suggestion (Blocker 5, narrowed)** — PII in stored transaction details is fine here; the concern is secrets. Reuse the existing base redaction helper so a bearer token / credentials can't land in logs.

```suggestion
			$this->log( 'Surjo Pay: Invalid auth response', null, $this->prepare_transaction_for_log( (array) $response ) );
```

dd32 (commented):
**Claude:** Inline code-review suggestions for some of the fixes from the security review above. Each is a committable `suggestion` block. Note on Blocker 5: I narrowed it — storing attendee PII is acceptable (we already store it), so the suggestion only redacts secrets, not PII.

nazmulch11:
Thanks for the review. I addressed the unresolved security comments locally:

- SurjoPay production/sandbox API URLs are now hard-coded from the official docs, and the free-form `api_endpoint 'setting' was removed.
- SurjoPay redirect validation now requires HTTPS and an exact/dot-suffix `shurjopayment.com`host.
- SurjoPay verification now checks the gateway response against the local order amount, `BDT`currency, and `customer_order_id`before completing payment.
- SurjoPay cancellation now requires both a `tix_payment_token` and a matching `order_id`.
- Auth/API logging now uses the redaction helper so credentials/tokens are not logged.
- SSLCommerz `GatewayPageURL` is now HTTPS/host allowlisted and performs validation checks on amount/currency according to the SSLCommerz v4 docs.
- Password fields are rendered as password inputs.
- Phone validation is scoped to paid SSL Commerz/SurjoPay checkouts.
- Removed/reverted the unrelated `composer.lock`, Docker port, and theme-template changes from the effective branch diff.
- Updated the README to remove stale ticket-hold/auto-release/API-endpoint claims.

I verified against:
- SSLCommerz v4 docs: https://developer.sslcommerz.com/doc/v4/
- SurjoPay direct API docs: https://web2.shurjopay.com.bd/frontend-developers/direct-integration-with-api

Local checks pass:
- `php -l` on touched PHP files
- `git diff --check`
- `BASE_REF=production php .github/bin/phpcs-branch.php`

I’ll still do a real sandbox checkout/return test for both gateways before the merge, since live gateway responses are the final source of truth.

dd32:
## Codex-powered automated review

Reviewed current PR head `a9ff512a64ddc63a5aa462b477e82e02948c97f6` against `production`. I would not merge this yet without checking the items below.

These are automated-review findings and should be treated as potential issues, not confirmed production bugs. I welcome being wrong on any of them; if one of these is already handled by behavior I missed, please push back with the relevant context.

### Potential blockers

1. **Checkout phone validation appears to be hooked to a non-existent action/filter.**
   `public_html/wp-content/plugins/camptix-bd-payments/includes/class-phone-field.php` registers `camptix_form_register_custom_error_flags`, but I could not find that hook firing in the current CampTix checkout path. `camptix_form_attendee_info_errors` looks useful for rendering notices, but not for performing validation before attendee drafts are created.

   **Suggested fix:** Use `camptix_checkout_start` for validation, set flags with `$camptix->error_flag( 'tix_phone_missing' )` / `tix_phone_invalid`, and keep `camptix_form_attendee_info_errors` only for rendering those messages.

2. **SSLCommerz return/cancel/fail handling now runs before Require Login without equivalent validation on every path.**
   `class-gateway-sslcommerz.php` moves `template_redirect` to priority `6`, before `CampTix_Require_Login`. That may be necessary for gateway returns, but `payment_cancel()` and `payment_failed()` appear able to call `payment_result()` from a request containing only `tix_payment_token`.

   **Suggested fix:** Only process SSLCommerz `payment_return`, `payment_failed`, and `payment_cancel` after validated SSLCommerz POST/cookie data is present. Otherwise return and let Require Login block normal unauthenticated requests.

3. **SurjoPay raw verification responses appear to be stored/logged in some paths.**
   `class-gateway-surjopay.php` stores `$verified` directly in `transaction_details`, and some verification failure logs pass raw `(array) $payment`. shurjoPay verification responses can include card/customer/bank details.

   **Suggested fix:** Store/log an allowlisted subset, or always pass verification responses through `prepare_transaction_for_log()` before logging and before `payment_result()` transaction details.

4. **`Base_Gateway::get_attendees_by_payment_token()` duplicates CampTix core behavior.**
   CampTix already provides `get_attendees_from_payment_token()`, with request-level caching and the statuses CampTix payment handling supports. The custom helper uses `post_status => any` and skips the core cache.

   **Suggested fix:** Remove the custom query helper, or make it a thin adapter around:

   ```php
   protected function get_attendees_by_payment_token( $payment_token ) {
       global $camptix;

       return $camptix->get_attendees_from_payment_token( $payment_token );
   }
   ```

   `sanitize_payment_value()` also appears unused and can be removed.

### Validation

Local checks performed:

- `php -l` passed for all changed PHP files.
- `git diff --check upstream/production...upstream/pr/1741` passed.
- `BASE_REF=production php .github/bin/phpcs-branch.php` passed on the PR head.

I could not run PHPUnit locally because the WordPress test suite bootstrap is not installed at the configured temp path.

I also checked the SSLCommerz v4 and shurjoPay direct API docs while reviewing the gateway response assumptions. The broad direction looks reasonable, but the items above should be checked/fixed before merge, followed by real sandbox checkout/return tests for both gateways.

References:
- https://developer.sslcommerz.com/doc/v4/index.html
- https://web2.shurjopay.com.bd/frontend-developers/direct-integration-with-api

nazmulch11:
Hi @dd32,

Thanks for the detailed review. I've addressed all four findings:

Phone validation hook (class-phone-field.php) Moved phone validation from camptix_form_register_custom_error_flags (which CampTix doesn't fire during checkout) to the existing camptix_checkout_start action. The method now receives $attendee_info and $order directly from the action and calls $camptix->error_flag() instead of returning a modified array. The $order is also passed into selected_gateway_requires_phone() so it no longer reads from the global scope.

SSLCommerz return/cancel/fail gating (class-gateway-sslcommerz.php) Added a $has_validated_gateway_data flag that is only set to true after the temporary POST cookie passes ipn_hash_verify() in early_template_redirect(). template_redirect() now returns early for payment_return, payment_failed, and payment_cancel unless that flag is set. payment_notify (IPN) is unaffected — it verifies its own hash independently.

SurjoPay transaction detail sanitization (class-gateway-surjopay.php) Added prepare_surjopay_transaction_details() which allowlists 14 safe fields from the verification response and passes them through the base-class redactor before anything is stored or logged. Applied in payment_return(), pre_attendee_timeout(), and switched raw (array) $payment in the two failure-path logs inside verify_payment() to prepare_transaction_for_log().

Attendee lookup (class-gateway-base.php) Replaced the custom get_posts() query in get_attendees_by_payment_token() with $camptix->get_attendees_from_payment_token() and removed the unused sanitize_payment_value() helper.

Verification: php -l passes on all four changed files.

dd32:
## Codex-powered automated follow-up review

Reviewed updated PR head `8e309893499637b122ffa9e88eb74f36f2016666` against `production`. These are automated-review notes and should be treated as potential issues, not confirmed production bugs. I welcome being wrong if gateway behavior or deployment context covers something I missed.

I would not merge this yet without checking the first item.

### Potential blocker

1. **SSLCommerz cancel/fail still need to bind the signed transaction to the URL payment token.**
   `public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php:390-402` now correctly checks that signed `tran_id` matches the URL `tix_payment_token` for `payment_notify()` / `payment_return()`. However, `payment_cancel()` and `payment_failed()` at `:423-461` still call `payment_result()` using the URL token after only the temporary signed POST-cookie gate.

   That gate proves the POST body was signed by SSLCommerz, but it does not prove the signed body belongs to the URL-supplied `tix_payment_token`. A signed cancel/fail payload for one order could potentially be replayed against another order by changing only the URL token, because the signed field appears to be `tran_id`, not `tix_payment_token`.

   Suggested fix: apply the same `hash_equals( (string) $payment_token, (string) ( $_POST['tran_id'] ?? '' ) )` check in both `payment_cancel()` and `payment_failed()` before calling `payment_result()`. On mismatch, log and return without changing attendee state.

### Suggested fixes / follow-ups

2. **SSLCommerz timeout recovery still logs a raw validator response.**
   In `class-gateway-sslcommerz.php:559`, `pre_attendee_timeout()` logs `$response` directly with an attendee ID, which means the default CampTix meta logger can persist that raw gateway response. The next lines already use `prepare_transaction_for_log()` for stored transaction details, so the log call should probably do the same:

   ```php
   $camptix->log(
       'SSLCommerz checkout timed out, but order succeeded.',
       $attendee_id,
       $this->prepare_transaction_for_log( (array) $response )
   );
   ```

3. **`prepare_transaction_for_log()` may overlap with existing WordCamp logger redaction.**
   `Base_Gateway::prepare_transaction_for_log()` in `class-gateway-base.php:45` duplicates part of the redaction policy already present in `WordCamp\Logger\redact_keys()` (`public_html/wp-content/mu-plugins/1-logger.php:73`). This is not necessarily a bug: CampTix logging goes through `camptix_log_raw`, and the default CampTix meta logger does not appear to call the shared WordCamp logger redactor. The gateway-local helper may therefore be needed for transaction data.

   Suggested fix: either explicitly reuse the shared redactor where safe, move/add gateway-sensitive keys to the shared redaction policy, or add a short comment explaining why this gateway helper is intentionally separate for CampTix transaction logs. That would avoid two redaction lists silently drifting apart.

### Resolved from the previous review

- The checkout phone validation path now looks right: validation is on `camptix_checkout_start`, with rendering through `camptix_form_attendee_info_errors`. `camptix_form_edit_attendee_custom_error_flags` is an edit-attendee hook, not the checkout validation hook.
- `get_attendees_by_payment_token()` is now only a thin adapter around CampTix core's `get_attendees_from_payment_token()`. It could be inlined, but I no longer see the earlier custom-query/caching concern.
- SurjoPay stored transaction details are now allowlisted via `prepare_surjopay_transaction_details()`, which addresses the raw-storage concern I raised earlier.

### Validation

Local checks performed on the updated PR head:

- `php -l` passed for the changed PHP files.
- `git diff --check upstream/production...upstream/pr/1741` passed.
- `BASE_REF=production php .github/bin/phpcs-branch.php` passed.

I still could not run PHPUnit locally because the configured WordPress test suite bootstrap is not installed at the temp path.

PR #1745, opened 2026-06-11
CI: cache node_modules and Composer downloads to speed up install steps

dd32:
## Why

The `npm install and build` step in the linter workflow consistently takes ~4 minutes (e.g. [this run](https://github.com/WordPress/wordcamp.org/actions/runs/27326842870/job/80729815954)), of which the webpack build is only ~20 seconds. The `cache: npm` on `setup-node` was already hitting, but it only caches downloaded tarballs (`~/.npm`) — `npm ci` still deletes `node_modules` and re-extracts all ~2,900 packages (plus lifecycle scripts) from scratch on every run. That reify phase is the ~3.5 minutes of silence in the log.

## What

For both `linter.yml` and `unit-tests.yml`:

- **Cache the installed `node_modules` tree** (root + workspaces) keyed on `package-lock.json` and the Node major version, and skip `npm ci` entirely on an exact hit. The existing `cache: npm` is kept as a fallback so cache-miss runs are still download-free. This should take the step from ~4 minutes to roughly 30-45 seconds on the common path.
- **Cache the Composer cache directory.** `composer.lock` is intentionally removed before install so the latest `wporg/wporg-mu-plugins` `dev-build` hash is always picked up — that behaviour is unchanged (and now documented with a comment), but re-resolution no longer needs to re-download every package. Keyed on `composer.json` with a `restore-keys` fallback since the resolved set drifts between runs.
- Pass `--no-audit --no-fund` to `npm ci` on the miss path, matching the existing `COMPOSER_NO_AUDIT` behaviour for Composer.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -46,14 +46,38 @@ jobs:
       env:
           COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+    - name: Get Composer cache directory
+      id: composer-cache
+      run: echo "dir=$(composer config cache-dir)" >> "$GITHUB_OUTPUT"
+
+    - name: Cache Composer downloads
+      uses: actions/cache@v4
+      with:
+        path: ${{ steps.composer-cache.outputs.dir }}
+        key: composer-${{ runner.os }}-${{ hashFiles('composer.json') }}
+        restore-keys: |
+          composer-${{ runner.os }}-
+
     - name: composer install
+      # The lock file is removed so that the latest wporg-mu-plugins dev-build is always installed.
       run: |
         rm composer.lock || true
         composer install
 
+    - name: Cache node_modules
+      id: node-modules-cache
+      uses: actions/cache@v4
+      with:
+        path: |
+          node_modules
+          public_html/wp-content/**/node_modules
+        key: node-modules-${{ runner.os }}-node20-${{ hashFiles('package-lock.json') }}
+
     - name: npm install and build
       run: |
-        npm ci
+        if [ "${{ steps.node-modules-cache.outputs.cache-hit }}" != 'true' ]; then
+          npm ci --no-audit --no-fund
+        fi
         npm run build --workspaces --if-present
 
     - name: Lint JavaScript and Styles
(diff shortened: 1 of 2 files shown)


dd32:
## Measured speedup

To validate the caching, the first CI run on this PR populated the caches (cache miss), and the workflows were then re-run to exercise the cache-hit path. All checks passed on both runs. Numbers below compare a recent `production` run ([27326842870](https://github.com/WordPress/wordcamp.org/actions/runs/27326842870/job/80729815954)) against the cache-hit re-run on this PR.

### Linting workflow (the worst offender)

| Step | Before | After (cache hit) |
| --- | --- | --- |
| `composer install` | 44s | 13s (3s restore + 10s install) |
| `npm install and build` | 4m 01s | 47s (27s `node_modules` restore + 20s build, `npm ci` skipped) |
| **Whole job** | **5m 47s** | **2m 11s** |

### Unit tests workflow

| Job | Before | After (cache hit) |
| --- | --- | --- |
| JavaScript | 4m 28s | 43s |
| PHP Tests (8.3) | 1m 28s | 1m 08s |
| PHP Tests (8.4) | 2m 22s | 1m 19s |
| PHP Tests (8.5) | 1m 47s | 1m 09s |

### Why the old setup was slow

The `cache: npm` on `setup-node` was already hitting, but it only caches downloaded tarballs (`~/.npm`) — `npm ci` still deletes `node_modules` and re-extracts all ~2,900 packages plus lifecycle scripts on every run. That extraction phase was ~3.5 minutes of the 4-minute install step. Caching the installed `node_modules` tree keyed on `package-lock.json` skips all of it when the lock file is unchanged.

### Notes

- The first run after `package-lock.json` changes is still a full `npm ci`, now slightly faster via `--no-audit --no-fund` (~2m 30s observed). Subsequent runs hit the cache.
- `composer.lock` is still removed before install, so the latest `wporg/wporg-mu-plugins` `dev-build` is always picked up — the Composer cache only avoids re-downloading packages (178 MB), not re-resolution.
- GitHub scopes caches per branch with fallback to the base branch, so PR branches will reuse caches saved by `production` runs once this is merged.

PR #1747, opened 2026-06-11
Update npm packages and WordPress Scripts 32

dd32:
## AI disclosure

This is an AI-generated and AI-managed PR. Codex is updating the branch and PR description under dd32's direction, with follow-up commits expected as the WordPress Scripts 32 migration is worked through.

## Summary

- updates Node package pins across the npm workspaces and refreshes the lockfile
- moves the WordPress build toolchain to `@wordpress/scripts` 32.4.0 across the workspaces
- updates direct WordPress package peers required by that toolchain, including `@wordpress/element` 8.0.0 and `@wordpress/eslint-plugin` 25.4.0
- keeps the Jest `uuid` test mock so newer WordPress packages can run under the current Jest setup

## Change in direction

This PR originally used a conservative package update that intentionally stayed on the latest 31.x WordPress Scripts line. Per follow-up direction, the v32 migration is now being handled on this same PR branch as follow-up commits instead of a separate PR.

The first v32 commit is intentionally the raw package graph update. Follow-up commits will address lint/build fallout from the WordPress Scripts 32 and ESLint stack changes.

## Notes

The `uuid@9` install warning is removed, and the old root `eslint@8` warning is removed by the v32 toolchain update. Some remaining install deprecation warnings still come from transitive packages in the current WordPress/Jest/Webpack dependency graph.

## Verification

- `npm install --no-audit --no-fund`
- `git diff --check`

Earlier, before the v32 follow-up commit, the conservative package update was also checked with:

- `npm ci --no-audit --no-fund`
- `npm run build --workspaces --if-present`
- `npm run lint:js --workspaces --if-present`
- `npm run lint:css --workspaces --if-present`
- `npm run test --workspace=wordcamp-blocks`
- `npm audit --json`
- `npm audit --omit=dev --json`

JavaScript lint/build migration checks are still in progress for the v32 follow-up commits.


dd32:
Change in direction update: this PR was originally scoped as a conservative dependency refresh that stayed on the latest WordPress Scripts 31.x line. Per follow-up direction, the WordPress Scripts 32 migration is now being handled on this same PR branch instead of in a separate PR.

I pushed `caf3ae9c` (`Update WordPress Scripts to 32.4.0`) to `dd32:node-deps-audit` as a follow-up commit. 

The v32 commit is intentionally the raw package graph update.
Follow-up commits will address lint/build fallout from the WordPress Scripts 32 and ESLint stack changes.

dd32:
Linting migration options after the `@wordpress/scripts` 32 update:

The current v32 checkpoint intentionally pushed the raw package graph first. Running JS lint under the repo's `.nvmrc` Node 20/npm 10 setup shows two broad classes of fallout:

- toolchain/config fallout around the ESLint 10 / WordPress Scripts 32 stack, including import resolver failures such as `typescript with invalid interface loaded as resolver`
- stricter formatting/rule fallout in existing source, including Prettier formatting churn, `object-shorthand`, `no-var`, `no-undef` for legacy browser globals, and JSDoc return-description checks in places that were previously passing

Options forward:

1. Stay on WordPress Scripts 31.x for now.
   - Pros: smallest diff, keeps the previously green conservative package update, avoids the lint migration blast radius.
   - Cons: does not complete the requested v32 move, leaves more of the deprecated dependency chain in place, and postpones the same migration work.

2. Move to WordPress Scripts 32 but pin/downgrade ESLint toward the older peer graph.
   - Pros: may reduce the immediate resolver/peer friction.
   - Cons: fights the current `@wordpress/scripts@32.4.0` dependency graph, makes this repo diverge from the upstream WordPress Scripts toolchain, and risks hiding compatibility problems that will come back on later updates.

3. Keep WordPress Scripts 32 and broadly relax/suppress the new lint failures.
   - Pros: fastest way to get a green lint run with a smaller source diff.
   - Cons: weakens the signal from the modernized lint stack, can hide real unresolved imports/globals, and leaves the repo with custom exceptions that future updates have to carry.

4. Keep WordPress Scripts 32 / ESLint 10 and migrate the lint setup deliberately.
   - Pros: aligns with the current WordPress Scripts toolchain, keeps the dependency update meaningful, and leaves the repo closer to a modern/default WordPress JS standard.
   - Cons: larger follow-up diff; likely includes a mix of config fixes, autofixable formatting changes, and small targeted source updates for intentional legacy globals.

The option I am taking is option 4: keep the v32 package graph and fix forward. The next commits should first resolve the lint toolchain/config problem so import resolution is trustworthy, then apply autofixable formatting where safe, then make narrow manual changes for intentional browser globals or genuinely changed lint rules. I am not planning to roll this back to v31 or broadly disable the new rules unless a specific rule proves incompatible with how this repo has to run.

dd32:
Final v32 lint migration status:

The WordPress Scripts 32 follow-up work is now pushed to `dd32:node-deps-audit`, with the branch head at `7ec8feee`.

Validation is green:

- `npm run lint:js --workspaces --if-present`
- `npm run lint:css --workspaces --if-present`
- `npm run lint:pkg-json --workspaces --if-present`
- `npm run test --workspace=wordcamp-blocks`
- `npm run build --workspaces --if-present`

Notes from validation:

- JS lint exits successfully. It still reports three existing `react-hooks/exhaustive-deps` warnings in `wordcamp-blocks`, but no JS lint errors.
- Builds exit successfully. The build output still reports existing Sass `@import` deprecation warnings, especially in the schedule block styles and speaker feedback Sass.

What changed after the raw v32 package update:

- Added `eslint-import-resolver-typescript` as a direct root dev dependency so the WordPress ESLint preset can load the intended resolver instead of falling through to the `typescript` package.
- Migrated JS linting to a root `eslint.config.cjs` flat config, and pointed workspace `lint:js` scripts at it explicitly.
- Applied the WordPress Scripts 32 lint autofixes as a separate mechanical commit.
- Fixed the remaining manual lint issues, including the stable `getSettings` date API, translator comment placement, obsolete `require-jsdoc` disables, and a few targeted inline exceptions for intentional legacy patterns.

Recommended future JavaScript/TypeScript cleanup:

- Move existing Sass `@import` usage to `@use` / `@forward` before Dart Sass 3.0 makes those warnings hard failures.
- Decide whether to fix or intentionally suppress the three remaining React hook dependency warnings. They are warnings today, but they are worth auditing because they can indicate stale closure behavior.
- Start migrating package-level `eslintConfig` blocks out of `package.json`; with ESLint 10 / WordPress Scripts 32, the reliable path is flat config.
- Consider introducing TypeScript incrementally in isolated packages/components only after the JS lint baseline stays stable. The current flat config is TypeScript-capable, but the codebase should avoid a broad TS conversion until build and package boundaries are clearer.
- Replace legacy browser globals with explicit imports or localized global declarations where practical, especially in older admin/front-end scripts.

dd32:
Follow-up status after the JavaScript/TypeScript cleanup pass:

- Followup 1: migrated the remaining Sass `@import` usage in the touched workspaces to `@use`. The schedule shared Sass values/mixins now live in a non-emitting helper partial, and the consuming Sass files use an explicit `layout.*` namespace rather than wildcard module imports.
- Followup 2: fixed the three remaining React hook dependency warnings. Workspace JS lint now passes with no warnings.
- Followup 3: package-level `eslintConfig` / `eslintIgnore` cleanup was completed in `fc60dd14`. This removed the legacy package metadata from 9 `package.json` files and deleted 55 lines. It should not change current lint behavior because all JS lint scripts now pass the root flat config explicitly with `--config=../../../../eslint.config.cjs`.
- Followup 4: skipped per direction.
- Followup 5: kept the legacy browser-localized globals windowless in source, but narrowed the flat-config global allowances to the exact files that use them. This avoids broad workspace globals without adding noisy runtime `window.*` changes.

Validation is green:

- `npm run lint:js --workspaces --if-present`
- `npm run lint:css --workspaces --if-present`
- `npm run lint:pkg-json --workspaces --if-present`
- `npm run test --workspace=wordcamp-blocks` (2 suites, 15 tests)
- `npm run build --workspaces --if-present`

Recommended future JavaScript/TypeScript cleanup:

- Keep PHP-localized/browser globals exact-file scoped in the flat config until those scripts are converted to module-style imports/enqueued dependencies.
- Introduce TypeScript incrementally around stable data boundaries or new components, not as a broad conversion in this dependency PR.
- Avoid runtime code changes solely for lint compliance when a narrow lint declaration more accurately describes how WordPress provides the value.

PR #1731, opened 2026-05-21
Campus Connect: admin notice and audit log on Approved For Pre-Planning (issue #1714 point 5)

dorsvenabili:
## Summary

Implements point 5 of #1714: when a Campus Connect application is approved for pre-planning, show a temporary success notice to the wrangler and write a permanent audit log entry.

> **Dependencies:** requires #1728 (point 1, CC dropdown) and #1730 (point 4, email trigger) to be merged first. Point 4 introduces the `wcpt_cc_approved_for_pre_planning` action this PR hooks into.

### `wcpt-wordcamp/wordcamp-admin.php` only (31 lines)

**Constructor:** registers `handle_cc_approved_for_pre_planning()` on the `wcpt_cc_approved_for_pre_planning` action.

**New method `handle_cc_approved_for_pre_planning( WP_Post $post )`:**
1. Writes a `_note` post meta entry — same format used by the manual "Add Private Note" metabox — with:
   - `timestamp`: current Unix time (auto-rendered as a readable date in the log UI)
   - `user_id`: current user
   - `message`: "Application approved for pre-planning. Organizer notification email queued."
2. Adds notice key `5` to `$this->active_admin_notices`, which the existing `add_admin_notices_to_redirect_url` / `print_admin_notices` pipeline picks up and displays after the save redirect.

**`get_admin_notices()`:** adds entry for key `5` — a green (`updated`) banner reading: "This Campus Connect application has been approved for pre-planning. An organizer notification email has been queued and a note has been added to the log."

The notice is shown once (after the redirect that follows saving). The log note is permanent.

> **Note on `wcpt_add_private_note()`:** that helper is introduced in #1729 (point 3). This PR uses `add_post_meta()` directly with the same `_note` format to avoid a cross-PR dependency. Once #1729 is merged the direct call can optionally be replaced.

## Test plan

- [ ] Transition a CC post to `wcpt-approved-pre-pl`. After the save redirect, confirm a green success banner appears at the top of the post edit screen.
- [ ] Confirm the banner text reads "This Campus Connect application has been approved for pre-planning. An organizer notification email has been queued and a note has been added to the log."
- [ ] Confirm a new entry appears in the private notes log with the correct timestamp and the message "Application approved for pre-planning. Organizer notification email queued."
- [ ] Reload the post edit screen. Confirm the green banner is gone (it is a one-time notice, not persistent).
- [ ] Transition a non-CC post to any status. Confirm no notice and no note are added.
[...]

PR #1730, opened 2026-05-21
Campus Connect: email trigger on Needs Orientation

dorsvenabili:
## What

Fires a Campus Connect–specific email trigger when a CC tracker entry is transitioned to **Needs Orientation** (`wcpt-needs-orientati`).

Previously the trigger fired on Approved For Pre-Planning (`wcpt-approved-pre-pl`), which is the wrong point in the CC workflow — orientation emails should go out as soon as the application is accepted and the organiser needs to begin onboarding, not at the pre-planning stage.

## Changes

**`wcpt-wordcamp/wordcamp-admin.php`** — `trigger_schedule_actions()`
- Status guard: `wcpt-approved-pre-pl` → `wcpt-needs-orientati`
- Action name: `wcpt_cc_approved_for_pre_planning` → `wcpt_cc_needs_orientation`

**`wordcamp-organizer-reminders/wcor-mailer.php`**
- Trigger key: `wcor_cc_approved_for_pre_planning` → `wcor_cc_needs_orientation`
- Trigger name: "Campus Connect application approved for pre-planning" → "Campus Connect application needs orientation"
- Action name: `wcpt_cc_approved_for_pre_planning` → `wcpt_cc_needs_orientation`
- Callback: `send_trigger_cc_approved_for_pre_planning()` → `send_trigger_cc_needs_orientation()`

No other logic changed. The `wcpt-needs-orientati` status is registered by PR #1728.

## Test plan

- [ ] Open a CC tracker entry (event_subtype = campusconnect) set to any status before Needs Orientation.
- [ ] Transition it to **Needs Orientation** via the status dropdown.
- [ ] Confirm `wcpt_cc_needs_orientation` fires: add a test hook (`add_action('wcpt_cc_needs_orientation', fn($p) => error_log('fired: '.$p->ID))`) and check the log.
- [ ] Confirm no email fires on transition to Approved For Pre-Planning for a CC entry.
- [ ] Confirm non-CC entries are unaffected (the `campusconnect` subtype guard prevents false positives).
- [ ] In Automated Reminders admin, confirm the "Campus Connect application needs orientation" trigger appears and can have emails attached to it.

Issue #1714, opened 2026-05-08
Automate vetting and approval workflow for Campus Connect events


peiraisotta:
### Summary
Automate the initial vetting of Campus Connect (WPCC) applications with an agent that writes private notes and moves the post to a new "Needs Action" status. Streamline the Campus Connect status list and fire the organizer follow-up email when an application is approved.

### Context
Campus Connect event (WPCC) applications submitted via [events.wordpress.org/campusconnect/organize](https://events.wordpress.org/campusconnect/organize/) land on [central.wordpress.org](http://central.wordpress.org/) as wordcamps entries with post-meta key `event_subtype` equal to `campusconnect` and WordCamp Status "Needs Vetting". Today every step (vetting, notes, status transitions, organizer email, site provisioning) is manual, and some of the steps and statuses aren't actually needed: they're there only because the WPCC workflow currently reuses the UX and process of WordCamp applications. The volume is growing and the criteria are documented well enough for an agent to handle the first pass. We want to simplify the process and automate vetting in order to reduce WPCC application waiting times, currently days to weeks, as much as possible.

### Proposed solution
We'll connect a vetting agent to the wordcamps CPT, scoped to entries where `event_subtype` equals `campusconnect` and status is "Needs Vetting". The agent writes its notes to the "Add Private Note" meta field, then transitions the status to a new "Needs Action" so a human picks it up from there. The agent communicates with the server via an MCP HTTP transport endpoint. The queue runner is at [wpcc_queue_runner_rsm.py](https://github.com/WordPress/community-agents/blob/main/wpcc_queue_runner_rsm.py) in `WordPress/community-agents`.

We'll also tighten the WordCamp Status dropdown **exclusively for Campus Connect entries** to ten statuses: Needs Vetting, Needs Action, Needs Orientation, Needs More Info, On Hold, Approved For Pre-Planning, WordCamp Scheduled, WordCamp Closed, Declined, Cancelled. All other statuses must be **hidden** (fully absent from the rendered `<select>`, not merely disabled) when `event_subtype` equals `campusconnect`. Non-Campus-Connect entries must not be affected: they keep their full existing status list unchanged.

Two statuses are **exclusive to Campus Connect**: "Needs Action" (`wcpt-needs-action`) and "Needs More Info" (`wcpt-needs-more-info`). They must not appear in the status dropdown for any other event subtype (WordCamp, DoAction, Student Club, Other Event).

[...]

PR #1748, opened 2026-06-11
Campus Connect: log silent email-trigger no-ops; fix overclaiming notice text (issue #1714 point 4)

peiraisotta:
## Summary

Follow-up to #1730/#1731. Testing the Needs Orientation transition on a real Campus Connect application showed the private note and admin banner appear, but **no organizer email is sent**.

**Root cause — configuration, not code:** `WCOR_Mailer::send_triggered_emails()` only sends Automated Reminder posts (`organizer-reminder` CPT) whose `wcor_which_trigger` meta matches the fired trigger. No reminder post on Central is assigned to `wcor_cc_needs_orientation` — the trigger key first reached production with #1730's merge, so none can exist yet. The code path itself is verified working: `test_cc_needs_orientation_trigger_message_sent` passes precisely because the test fixture creates and assigns a reminder post — the step production is missing. **The email content must be created in wp-admin** (Automated Reminders → new reminder → trigger "Campus Connect application needs orientation"); that's the actual fix and needs no deploy.

This PR fixes the two things that made the gap invisible and misleading:

### `wordcamp-organizer-reminders/wcor-mailer.php`
`send_triggered_emails()` was the **only** silent failure path in the mailer — `mail()` already logs invalid recipients and `wp_mail()` failures. Now logs via `\WordCamp\Logger\log()` when:
- a trigger fires with **no reminder posts assigned** to it (the case hit here), and
- reminders are assigned but **every one is skipped or fails** (subtype restriction, already-sent dedupe, send failure).

Both events are rare and always worth explaining, so log volume is negligible. No behaviour change when emails send normally.

### `wcpt/wcpt-wordcamp/wordcamp-admin.php`
The admin notice and audit note claimed "the organizer notification email **has been triggered**" — stated as fact even when no email exists to send, which is exactly what misled testing. Reworded both to claim only what the code guarantees (the trigger fired), following the same merge-resolution precedent applied to the point 5 text (#1731 [comment](https://github.com/WordPress/wordcamp.org/pull/1731#issuecomment-4677731428)). The notice now also names the Automated Reminders trigger so a wrangler knows where to look when no email arrives.

## Not in this PR (one-time admin action, no deploy needed)

Create the Automated Reminder on Central: **Automated Reminders → Add New** → title = subject, body = email content → "When to send": on a trigger → "Campus Connect application needs orientation" → recipients: "the organizing team" (resolves to the application's `E-mail Address`/`Email Address` meta) → restrict event subtype to Campus Connect. Note every WCOR email is already hard-CC'd to `support@wordcamp.org` by `mail()`, satisfying that requirement in #1714 automatically.

## Test plan

- [ ] `php -l` clean on both files (verified)
- [ ] Existing `test-wcor-mailer.php` suite passes unchanged (no behavioural change when reminders are assigned; `test_cc_needs_orientation_trigger_message_sent` and `test_repeatable_trigger_sends_multiple_times` cover the send paths)
- [ ] Transition a CC post to Needs Orientation with **no** reminder assigned → log entry "Trigger fired, but no reminder posts are assigned to it" appears; notice/note no longer claim an email was sent
- [ ] Assign a reminder to the trigger, repeat → email sends, CC'd to support@wordcamp.org, no log entry

See #1714 (point 4).

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
@@ -972,7 +972,7 @@ if ( ! class_exists( 'WordCamp_Admin' ) ) :
 				array(
 					'timestamp' => time(),
 					'user_id'   => get_current_user_id(),
-					'message'   => __( 'Application moved to Needs Orientation. Organizer notification email triggered.', 'wordcamporg' ),
+					'message'   => __( 'Application moved to Needs Orientation. Organizer notification email trigger fired.', 'wordcamporg' ),
 				)
 			);
 
@@ -1340,7 +1340,7 @@ if ( ! class_exists( 'WordCamp_Admin' ) ) :
 
 				5 => array(
 					'type'   => 'updated',
-					'notice' => __( 'This Campus Connect application has been moved to Needs Orientation. The organizer notification email has been triggered and a note has been added to the log.', 'wordcamporg' ),
+					'notice' => __( 'This Campus Connect application has been moved to Needs Orientation and a note has been added to the log. The organizer notification email is sent by the Automated Reminder assigned to the "Campus Connect application needs orientation" trigger — if no reminder is assigned, no email goes out.', 'wordcamporg' ),
 				),
 			);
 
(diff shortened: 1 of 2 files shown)


dd32:
I'm not fully sure I understand this PR, it seems to be logging stuff, but AFAIK, these logs aren't actually shown or accessed anywhere?

PR #1749, opened 2026-06-12
SSLCommerz: add Live environment guards and structured logging

nazmulch11:
### Description
Fixes SSLCommerz checkout failures caused by rejecting the live epay-gw.sslcommerz.com redirect host, improves payment diagnostics, and adds a configurable minimum transaction amount.
The changes also preserve reconciliation references, protect callback processing against mismatched orders, and separate stored transaction details from redacted diagnostic logs.
### Key Changes
Checkout and redirect validation
- Accepts these exact HTTPS checkout hosts:
  - Sandbox: sandbox.sslcommerz.com
  - Live: securepay.sslcommerz.com and epay-gw.sslcommerz.com
- Continues rejecting arbitrary subdomains and unrelated hosts.
- Shows buyer-facing notices for checkout failures while keeping gateway diagnostics in redacted logs.
### Minimum transaction amount
- Adds Minimum Transaction Amount under Tickets > Setup > Payment > SSLCommerz.
- Defaults to 0 BDT (disabled) for stores without a saved value; merchants can optionally configure a minimum.
- Applies the local check in both sandbox and live mode before starting a gateway session.
- Clearing the field or saving zero disables the local check. SSLCommerz may still enforce its own merchant-specific minimum.
- Invalid, negative, or non-finite values produce a settings error; omitted settings retain their existing values.
Callback handling
- Rejects invalid IPN signatures before looking up attendees, logs the rejection, and leaves order status unchanged.
- Rejects transaction IDs that do not match the URL’s payment token, preventing a replayed callback from changing another order.
- Logs missing-attendee and mismatched browser callbacks; rejected cancel/fail callbacks also display a buyer notice.
- Valid, matching cancel/fail callbacks continue calling payment_result() with the corresponding status.
- A matching, signature-valid IPN that fails server-side transaction validation calls payment_result(FAILED).
- Logs signature-valid notifications with no eligible attendee, including potentially expired reservations.
Transaction storage and diagnostics
- Preserves bank_tran_id and tran_id in transaction details for reconciliation, including timeout recovery.
- Redacts those references separately when producing diagnostic logs.
- Restores transport error messages and decoded non-2xx JSON diagnostics for SSLCommerz and the shared HTTP client used by SurjoPay.
- Redacts sensitive fields and known echoed credentials, and removes URLs and email addresses from free-text diagnostics.
- Retains only body length for non-JSON error responses.
- Avoids short-name substitutions corrupting unrelated words, such as “Al” within “allowed.”
- Associates logs with attendees where a suitable attendee context is available. Invalid-signature and missing-attendee logs intentionally have no attendee association.
Settings and local development
- Preserves valid store-password bytes rather than passing secrets through sanitize_text_field().
- Rejects invalid UTF-8 or control characters with a settings error, retaining the previous password.
- Avoids double-unslashing values already processed by the WordPress Settings API.
- Allows an HTTPS callback-base override through camptix_sslcommerz_callback_base_url only when the WordPress environment type is local.
### Timeout recovery
- Explicitly requests JSON from the session-query endpoint.
- Uses the documented flat response for sessionkey queries; the element array belongs to queries using tran_id.
- Checks transaction ID, amount, and currency before completing the order non-interactively.
How to Test
Use a development or sandbox environment for simulated callbacks and failure cases.
1. Sandbox checkout
   - Enable sandbox mode and configure sandbox credentials.
   - Purchase tickets with an order total at or above the configured minimum.
   - Verify redirect to sandbox.sslcommerz.com, successful return, and completed attendee/order state.
   - Confirm transaction details retain bank_tran_id and tran_id.
2. Minimum amount and settings
   - Configure a minimum of 10.00 to exercise the local threshold.
   - Verify a total below 10 BDT shows an error without starting a gateway session.
   - Test totals equal to and above the threshold.
   - Clear the setting and save; verify it becomes zero.
   - Verify invalid values show a settings error.
   - Confirm omitted credentials remain unchanged and valid password characters survive saving exactly.
3. Live redirect validation
   - Configure live credentials and disable sandbox mode.
   - Verify checkout accepts HTTPS URLs on securepay.sslcommerz.com or epay-gw.sslcommerz.com.
   - Verify HTTP, unrelated hosts, arbitrary subdomains, and sandbox hosts are rejected in live mode.
4. Invalid IPN signature
   - Submit an IPN with an invalid signature or tampered signed payload.
   - Verify no attendee lookup or order status change occurs.
   - Confirm the log contains:
     SSLCommerz IPN hash verification failed; order status unchanged.
5. Order binding and callback failures
   - Replay a valid signed payload for order A using order B’s URL token.
   - Verify order B remains unchanged and the mismatch is logged.
   - Verify matching cancel/fail callbacks call the corresponding payment-result handler.
   - Verify missing or mismatched POST transaction IDs produce a rejection log and buyer notice.
   - Verify a matching, signature-valid IPN that fails server-side validation calls payment_result(FAILED).
6. Late notifications and timeout recovery
   - Submit a signature-valid notification with no eligible attendee and verify a diagnostic is logged.
   - Simulate a successful flat session-query response for a draft reservation.
   - Verify matching transaction ID, amount, and currency allow non-interactive completion.
   - Confirm bank references are retained in transaction details and redacted from the timeout log.
7. Diagnostic hygiene
   - Simulate transport failures and non-2xx JSON responses for both gateways.
   - Verify useful diagnostic messages remain without exposing fixture credentials, tokens, URLs, or customer data.
   - Confirm non-JSON response bodies are not dumped.
   - Verify attendee associations where available; do not expect them for invalid-signature or missing-attendee logs.
8. SurjoPay regression
   - Verify checkout and redirect behavior with appropriate test credentials.
   - Confirm shared API error diagnostics remain useful and redacted.
Validation
- Regression coverage verifies notification-order precedence and SSLCommerz-specific diagnostic redaction.
- PHPCS passes for both modified gateway files and the new test file.
- git diff --check passes.
- Checkout was confirmed working after adding the observed live redirect host.
- Automated HTTP tests use mocked responses; they do not verify live settlement.
- Full-repository PHPUnit remains blocked locally by missing GatherPress. The standalone CampTix configuration reports an existing missing-testsuite-name warning.
- Credentialed SurjoPay end-to-end testing remains outstanding.


public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -599,7 +658,17 @@ protected function api( $method, $url, $args = [], $headers = [] ) {
...
 
-		$response = wp_remote_request( $full_url, $request_args );
+		$response  = wp_remote_request( $full_url, $request_args );
+		$http_code = is_wp_error( $response ) ? 0 : wp_remote_retrieve_response_code( $response );
+
+		$camptix->log(
+			sprintf( 'SSLCommerz API %s %s → HTTP %d.', $method, $url, $http_code ),
+			null,
+			[
+				'full_url' => $full_url,
```
obenland:
```suggestion
```

This will leak store passwords in plaintext into logs.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -183,14 +217,39 @@ public function payment_checkout( $payment_token ) {
...
+			array_merge(
+				array_diff_key( $args, array_flip( [ 'store_passwd' ] ) ),
+				[
+					'sandbox'  => $this->options['sandbox'],
+					'base_url' => $this->options['sandbox']
+								? 'https://sandbox.sslcommerz.com'
+								: 'https://securepay.sslcommerz.com',
+				]
+			)
+		);
```
obenland, line 232:
```suggestion
		$camptix->log(
			'SSLCommerz payment initiation payload.',
			$attendee->ID,
			array_merge(
				$this->prepare_transaction_for_log( $args ),
				[ 'sandbox'  => $this->options['sandbox'] ]
			)
		);
```

Personal identifiable information and store_id are logged on every checkout. Let's reuse existing tools to avoid that.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -378,6 +429,18 @@ public function payment_notify() {
 		$transaction_id = sanitize_text_field( $_REQUEST['tran_id'] ?? '' );
 		$val_id         = sanitize_text_field( $_REQUEST['val_id'] ?? '' );
 
+		$camptix->log(
+			'SSLCommerz payment_notify received.',
+			null,
+			[
+				'payment_token'  => $payment_token,
+				'transaction_id' => $transaction_id,
+				'val_id'         => $val_id,
```
obenland:
```suggestion
```

`prepare_transaction_for_log()` unsets `val_id`, let's keep with that intention.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -96,13 +96,51 @@ public function payment_checkout( $payment_token ) {
...
+		 * reachable (e.g., a *.test domain), you can hook this filter to substitute
+		 * a tunnel URL (ngrok, Cloudflare Tunnel, etc.) so that SSLCommerz's live
+		 * servers can reach the callback and IPN endpoints.
+		 *
+		 * @param string $base_url The tickets page URL.
+		 */
+		$callback_base_url = apply_filters(
+			'camptix_sslcommerz_callback_base_url',
+			$camptix->get_tickets_url()
+		);
```
obenland: Let's add an environment check here to make sure this can only get filtered on local.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -96,13 +96,51 @@ public function payment_checkout( $payment_token ) {
...
+		 * SSLCommerz Live enforces a minimum transaction amount configured in the
+		 * merchant account. The default minimum is 10 BDT — amounts below this are
+		 * rejected with "Transaction amount is not allowed as per admin configuration!".
+		 *
+		 * Filters the minimum allowed amount so it can be adjusted without code changes
+		 * if the merchant account configuration changes.
+		 *
+		 * @param float $min_amount Minimum amount in BDT.
+		 */
+		$min_amount = (float) apply_filters( 'camptix_sslcommerz_min_amount', 10.0 );
```
obenland: $min_amount belongs in gateway settings. The plugin already has a settings screen; a merchant-configured value is merchant-configurable data. Also, skipping the check in sandbox means developers never encounter it

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -688,7 +755,22 @@ protected function ipn_hash_verify( $store_passwd, $data ) {
...
+
+		if ( ! $result ) {
+			global $camptix;
+			$camptix->log(
+				'SSLCommerz IPN hash mismatch.',
+				null,
+				[
+					'verify_key'  => $data['verify_key'] ?? '',
+					'keys_hashed' => array_keys( $new_data ),
+					'has_sign'    => isset( $data['verify_sign'] ),
```
obenland: $data['verify_sign'] is always true

obenland (changes requested):
Hi @nazmulch11, could you give this another look?

A couple of general observations beyond the individual ones:

* `$post_id` is null on all four new log calls. The PR's stated goal is visibility in "WP Admin → CampTix DB Log", but that metabox is per-post — passing `$attendee->ID` where available makes these actually show up next to the attendee.
- `api()` now writes a row to a network-wide table on every call, success or failure. Would it be enough to log only non-2xx responses?

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -183,14 +217,39 @@ public function payment_checkout( $payment_token ) {
...
+				'status'         => $status,
+				'failedreason'   => $response_data['failedreason'] ?? '',
+				'GatewayPageURL' => $gateway_page_url,
+				'sessionkey_set' => ! empty( $response->sessionkey ),
+			]
+		);
+
 		if ( 'SUCCESS' === $status && ! empty( $gateway_page_url ) ) {
-			if ( ! $this->is_allowed_https_host( $gateway_page_url, array( 'sandbox.sslcommerz.com', 'securepay.sslcommerz.com' ) ) ) {
+			if ( ! $this->is_allowed_https_host( $gateway_page_url, array( 'sslcommerz.com' ) ) ) {
```
obenland: Are we sure we want to allow _any_ subdomain here?

bor0:
@nazmulch11 hi! I took another look at the PR. A few things:

1. @obenland's threads are still unresolved with no reply. Can you please confirm what changed for each?
2. What happens for existing sites with no `min_amount` saved yet?
3. Does the settings save path still preserve existing `merchant_id`/`store_password`?
4. Have you confirmed that the live `GatewayPageURL` host always matches the new exact allowlist?
5. Did you retest SurjoPay too, since it shares the changed base-class code?
6. Have you tested this against live credentials or only Sandbox?
7. Have you checked `wp_camptix_log` after testing for leaked secrets/PII?
8. Have you tested an IPN hash-mismatch case to confirm it's rejected and logged?
9. Can you expand "How to test" to cover more cases? For example, sandbox happy path, live happy path, below-minimum rejection, IPN hash mismatch, and a log-content check?

nazmulch11:
Hi @bor0 and @obenland — thank you for reviewing this, and apologies for not responding to the individual threads earlier.

I addressed the code feedback in `5c8717f` and will reply to each completed thread separately.

1. Existing sites without `min_amount` fall back to `10.0` through `wp_parse_args()`, so no settings resave is required.
2. Settings validation starts with the existing options and only replaces submitted values, preserving the current merchant ID and store password.
3. Redirect validation now uses the exact environment-specific hostname: `sandbox.sslcommerz.com` for sandbox and `securepay.sslcommerz.com` for live mode. Subdomains are not accepted.
4. Sensitive values—including passwords, tokens, session keys, customer names, emails, and phone numbers—are excluded or redacted from logs.
5. Invalid IPN hashes return `PAYMENT_STATUS_FAILED` and are logged against the attendee without storing sensitive signature data.
6. The “How to test” section now covers sandbox checkout, live redirection, below-minimum rejection, IPN hash mismatch, DB-log hygiene, and SurjoPay regression testing.
7. All current CI jobs pass, including linting and PHP tests on PHP 8.3, 8.4, and 8.5.


Could you please take another look when convenient?

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -688,7 +915,20 @@ protected function ipn_hash_verify( $store_passwd, $data ) {
 		$hash_string = rtrim( $hash_string, '&' );
 		$hash_string = md5( $hash_string );
 
-		return hash_equals( $hash_string, $data['verify_sign'] );
+		$result = hash_equals( $hash_string, $data['verify_sign'] );
+
+		if ( ! $result ) {
+			global $camptix;
+			$camptix->log(
```
bor0: This log call sits on a path that unauthenticated requests can reach. `ipn_hash_verify()` is called from `early_template_redirect()`, which is hooked at `template_redirect` priority 5 — intentionally ahead of `CampTix_Require_Login::block_unauthenticated_actions` (see the comment near line 49). No valid payment token is needed to get here, and `$post_id` may well be `null`.

Each call that lands in this branch now costs a `posts_per_page => -1` meta_query in `get_attendee_id_for_log()`, a row in the network-wide `wp_camptix_log` table, and an `error_log()` line via camptix-network-tools. Previously the same request returned `false` with no writes at all, so this is a new cost per request on a route that isn't gated.

Worth noting too that `keys_hashed` is derived from the request's own `verify_key`, so the logged payload is caller-supplied rather than gateway-supplied.

Could we either move the logging behind a check that we're dealing with a known payment token, or rate-limit / drop it for requests that don't resolve to an attendee?

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -96,13 +97,57 @@ public function payment_checkout( $payment_token ) {
...
+		if ( (int) round( (float) $order['total'] * 100 ) < (int) round( $min_amount * 100 ) ) {
+			$camptix->log(
+				'SSLCommerz payment rejected: order total is below the configured minimum transaction amount.',
+				$attendee->ID,
+				[
+					'order_total' => $order['total'],
+					'min_amount'  => $min_amount,
+				]
+			);
+			return CampTix_Plugin::PAYMENT_STATUS_FAILED;
```
bor0, line 120: This returns `PAYMENT_STATUS_FAILED` without calling `$camptix->error()` first. Core handles `PAYMENT_STATUS_FAILED` by re-rendering `form_attendee_info()` (camptix.php ~6427), so with no error flag set the buyer clicks "Pay with SSLCommerz", gets bounced straight back to the attendee-info form with no explanation, and can loop there indefinitely.

The equivalent early-returns in the SurjoPay gateway all set an error first (`class-gateway-surjopay.php:137,143,149,166`). Something like:

```php
$camptix->error( __( 'The total is below the minimum amount this payment method accepts.', 'wordcamporg' ) );
```

Since the minimum-amount guard is the headline feature here, this is the one path where the user-facing message really matters.

bor0 (commented):
Nice hardening pass overall — the exact-host redirect validation and the redaction allowlist both read well. Two things I'd like to see addressed before this goes in; leaving them inline.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Thanks for catching this. I removed logging from ipn_hash_verify() and made it a side-effect-free validation method. I also removed the attendee lookups that were only used for logging in early_template_redirect(), so invalid unauthenticated requests no longer trigger those meta queries or write to the CampTix and PHP logs.  
Hash-verification failures are still logged in payment_notify(), but only after the payment token resolves to a known attendee, and the log remains scoped to that attendee.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Thanks, fixed. The minimum-amount guard now calls $camptix->error() with a translated message before returning PAYMENT_STATUS_FAILED. When CampTix re-renders the attendee-information form, the buyer will see a clear explanation instead of being returned silently and potentially repeating the checkout loop.

nazmulch11:
Thanks @bor0! I've updated the PR:

- Removed the attendee lookup and failure logging from `ipn_hash_verify()` to avoid unauthenticated DB queries and log overhead (`7a778bab`).
- Added `$camptix->error()` notices for the minimum amount check, session failure (with `failedreason`), and redirect validation to avoid silent checkout reload loops (`7a778bab`, `c13e4943`).
- Updated `is_allowed_https_host()` to allow `sslcommerz.com` HTTPS subdomains after finding live transactions redirect via `epay-gw.sslcommerz.com` (`c13e4943`).

Could you please take another look?

bor0 (commented):
Thanks for the follow-ups — both of my earlier notes are addressed on `c13e4943` (apologies, I'd reviewed a stale tip). Two new things came in with those commits, though; both inline.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -183,28 +221,77 @@ public function payment_checkout( $payment_token ) {
...
+		$camptix->log(
+			'SSLCommerz payment initiation response.',
+			$attendee->ID,
+			$response_log
+		);
 
 		if ( 'SUCCESS' === $status && ! empty( $gateway_page_url ) ) {
-			if ( ! $this->is_allowed_https_host( $gateway_page_url, array( 'sandbox.sslcommerz.com', 'securepay.sslcommerz.com' ) ) ) {
-				$camptix->log( 'SSLCommerz unexpected redirect host.', null, array( 'url' => $gateway_page_url ) );
+			if ( ! $this->is_allowed_https_host( $gateway_page_url, array( 'sslcommerz.com' ) ) ) {
```
bor0: This widened the allowlist rather than tightening it. `is_allowed_https_host()` defaults `$allow_subdomains` to `true` (`class-gateway-base.php:131`), so passing just `array( 'sslcommerz.com' )` accepts any `*.sslcommerz.com` host, where the previous revision matched the two exact hosts and passed `false` explicitly.

Two consequences:

- The sandbox/live split is lost on the redirect. A site configured for Live will now follow a `GatewayPageURL` pointing at `sandbox.sslcommerz.com`, which is the case the "Live environment guards" in the title are meant to cover.
- `false` is no longer passed for `$allow_subdomains` anywhere in the plugin, so the parameter this PR adds has no caller.

The PR description still says "exact host matching with subdomains disabled ... `sandbox.sslcommerz.com` for sandbox mode and `securepay.sslcommerz.com` for live mode", and the `phpcs:ignore` justification on the `wp_redirect()` at line 284 still says "Scheme and exact host are validated above" — both now describe behaviour the code doesn't implement. Was the loosening deliberate? If it was to accommodate a host shape I'm not seeing, it'd be worth adding that host to the exact-match list instead and updating the two justifications.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -183,28 +221,77 @@ public function payment_checkout( $payment_token ) {
...
-			wp_redirect( $gateway_page_url );
+			wp_redirect( $gateway_page_url ); // phpcs:ignore WordPress.Security.SafeRedirect.wp_redirect_wp_redirect -- Scheme and exact host are validated above.
 			exit;
 		}
 
-		$camptix->log( 'SSLCommerz session initiation failed.', null, $this->prepare_transaction_for_log( $response_data ) );
+		$camptix->log( 'SSLCommerz session initiation failed.', $attendee->ID, $response_log );
+
+		if ( ! empty( $response_data['failedreason'] ) ) {
+			$camptix->error( esc_html( $response_data['failedreason'] ) );
```
bor0: The scrubbed and unscrubbed copies of `failedreason` look inverted here. The version that goes into the log at line 288 is run through `prepare_gateway_message_for_log()`, which replaces `merchant_id`, `store_password`, the payment token, and the buyer's name/email/phone with `[redacted]` — but the copy shown on the public tickets page three lines later is `$response_data['failedreason']` raw.

So the sanitised copy goes to the private DB log and the unsanitised one to the browser. Anything SSLCommerz echoes back in `failedreason` (submitted fields, merchant identifiers, configuration detail) renders to the visitor. Running the user-facing string through the same helper would line the two up:

```php
$camptix->error( $response_log['failedreason'] ?? __( 'A payment error has occurred...', 'bd-payments-camptix' ) );
```

Minor, while you're here: `do_notices()` already runs errors through `wp_kses()` (camptix.php:7238), so the `esc_html()` double-escapes — a reason containing `&` or `'` displays as `&amp;` / `&#039;`.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
bor0: Resolved on `c13e4943` — the log call was dropped from `ipn_hash_verify()` entirely, along with the two `get_attendee_id_for_log()` calls on the pre-auth path in `early_template_redirect()`. Nothing writes there for unauthenticated requests now. Sorry for the noise; I was looking at `8df19e82`.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
bor0: Resolved on `c13e4943` — `$camptix->error()` now precedes the return, so the buyer gets a message instead of a silent bounce back to the attendee-info form. Sorry for the noise; I was looking at `8df19e82`.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Thanks for catching this. I’ve restored exact HTTPS host matching based on the configured mode: sandbox.sslcommerz.com for sandbox and securepay.sslcommerz.com for live, explicitly passing false for $allow_subdomains. This now matches the PR description and redirect comment.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Agreed, the public error should use the redacted message too. I’ve changed it to use $response_log['failedreason'], with a generic fallback when empty. I also removed the redundant esc_html() call to avoid the double-escaping you flagged.

nazmulch11:
Thanks @bor0 — I’ve addressed both points: restored exact, mode-specific HTTPS redirect hosts with subdomains disabled, and updated public errors to use the redacted message without double-escaping. Could you take another look when you have a chance?

bor0 (commented):
Thanks — both of my earlier notes are confirmed fixed on `65511df3`: the redirect now pins the exact host per mode with `$allow_subdomains` explicitly `false`, and the public error uses the redacted `failedreason` with the generic fallback still intact (`''` is unset at line 258 before the `! empty()` check). Dropping `esc_html()` was the right call too.

I then did a fuller pass over the whole diff rather than just the two threads, and came away with a cluster of findings — mostly from the logging refactor in the earlier commits, not from today's push. Two I'd call blocking; the rest are smaller. All inline.

**Blocking**

1. Four IPN/callback paths that previously called `payment_result( …, PAYMENT_STATUS_FAILED, … )` now just `return` the constant, which `template_redirect()` discards — so the order status never changes.
2. `bank_tran_id` no longer reaches `tix_transaction_details`, which removes the reference needed to reconcile or refund with SSLCommerz.

**Smaller:** the substring-based redaction can corrupt the message it redacts (and that text is now user-facing, which is partly my doing); `api()`'s error logging lost its diagnostics for both gateways; `sanitize_text_field()` on `store_password`; a cleared minimum amount is silently ignored; and the `10.0` default is retroactive on live sites.

One question on `pre_attendee_timeout()` at the end — not something this PR introduced, but it touches the function.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -411,8 +525,12 @@ public function payment_notify() {
 			}
 		}
 
-		$camptix->log( 'SSLCommerz IPN hash verification failed.', null, $payment_data );
-		return $camptix->payment_result( $payment_token, CampTix_Plugin::PAYMENT_STATUS_FAILED, $payment_data );
+		$camptix->log(
+			'SSLCommerz IPN hash verification failed.',
```
bor0: This is the one I'd most like resolved before merge.

`payment_result()` is no longer called on the failure paths — this line and the `TRAN_ID_MISMATCH` branch at 512 both `return CampTix_Plugin::PAYMENT_STATUS_FAILED` instead. But `template_redirect()` discards that return value (line 464), so **nothing acts on it**: the order isn't marked failed, it stays in `draft` until the reservation times out.

On `production` both branches ended in `return $camptix->payment_result( $payment_token, PAYMENT_STATUS_FAILED, $payment_data );`, which did change the status.

Concrete case: the store password gets rotated while orders are in flight. Every genuine IPN then fails hash verification, and instead of a batch of clearly-failed orders you get a batch of orders silently stuck in `draft` with only a log line to explain it.

If the intent is "don't act on input we couldn't verify" — which is a perfectly reasonable position for the hash-failure branch specifically — could you add a comment saying so? And test step #4 in the description ("Verify the order returns `PAYMENT_STATUS_FAILED`") would need rewording, since that return is no longer observable anywhere.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -428,8 +546,14 @@ public function payment_cancel() {
 			return $camptix->error( 'empty token' );
 		}
 
-		$transaction_id      = sanitize_text_field( $_REQUEST['tran_id'] ?? '' );
-		$transaction_details = $this->prepare_transaction_for_log( $_POST );
+		$transaction_id = sanitize_text_field( $_REQUEST['tran_id'] ?? '' );
+		$attendee_id    = $this->get_attendee_id_for_log( $payment_token );
+
+		if ( ! $attendee_id || ! hash_equals( (string) $payment_token, (string) $transaction_id ) ) {
```
bor0, line 573: Same pattern as the IPN branches, but harder to justify here: `payment_cancel()` (this line) and `payment_failed()` (581) bail with a bare `return PAYMENT_STATUS_FAILED` when the attendee lookup misses or `tran_id` doesn't match the token — discarded by the caller, so the order is left in `draft` — and with no log entry either.

These two run *after* `early_template_redirect()` has already validated the payload, so the "unauthenticated input" argument doesn't apply the way it does to the IPN path. A cancel callback whose payload doesn't carry `tran_id` currently leaves the order stuck with nothing recorded to explain why, which is the opposite of what this PR is going for.

At minimum a `$camptix->log()` before each return.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -461,15 +591,83 @@ public function payment_failed() {
...
+
+	/**
+	 * Allowlist SSLCommerz response fields that are safe to persist and log.
+	 *
+	 * @param array $data Raw SSLCommerz transaction data.
+	 *
+	 * @return array
+	 */
+	private function prepare_sslcommerz_transaction_details( $data ) {
+		$allowed_keys = [
```
bor0, line 644: `prepare_sslcommerz_transaction_details()` isn't only a logging filter — its result goes into `$payment_data['transaction_details']`, which `payment_result()` persists to the `tix_transaction_details` post meta. So this allowlist decides what's **stored** on the order, not just what's logged.

`bank_tran_id` isn't on it. That's SSLCommerz's own transaction reference — neither a secret nor PII — and it's the value you need to reconcile or refund a payment with SSLCommerz support. On `production` this path used `prepare_transaction_for_log()`, which kept it. After this change a completed order stores only `status`, `amount`, `currency`, `card_type` and friends, so a dispute six months later has nothing to look up.

Could you add `bank_tran_id` (and probably `tran_id`) to the allowlist? Same loss applies on the timeout-recovery path at line 763, which uses this helper too.

Worth noting the interaction: `prepare_transaction_for_log()` now redacts `bank_tran_id` for *log* output, which is fine and separate — it's only the persisted copy that shouldn't lose it.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -461,15 +591,83 @@ public function payment_failed() {
...
+
+	/**
+	 * Prepare a gateway status message for logging without known sensitive values.
+	 *
+	 * @param mixed $message          Gateway response message.
+	 * @param array $sensitive_values Values that must not appear in the log.
+	 *
+	 * @return string
+	 */
+	private function prepare_gateway_message_for_log( $message, $sensitive_values ) {
```
bor0: The substring approach can corrupt the message it's trying to redact. `get_attendee_customer_info()` builds `'name' => trim( $first_name . ' ' . $last_name )`, so a short name matches unrelated text — an attendee named "Al" turns

> Transaction amount is not allowed as per admin configuration!

into

> Transaction amount is not [redacted]lowed as per admin configuration!

Short merchant IDs have the same problem.

This got sharper with today's change, and that part is on me: routing the redacted copy to the buyer was right for the leak, but it also means `[redacted]` markers and any mangling now render on the tickets page rather than staying in the DB log.

Redacting by field name — the way `prepare_transaction_for_log()` and `prepare_sslcommerz_transaction_details()` both do — avoids this entirely. If the values genuinely have to be matched as substrings, requiring a minimum token length (say 6+ characters) before replacing would cover the realistic cases without touching short ones.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-base.php
```diff
@@ -198,18 +249,23 @@ protected function api( $method, $url, $args = array(), $headers = array() ) {
...
 		$code = wp_remote_retrieve_response_code( $response );
 		$body = wp_remote_retrieve_body( $response );
 
 		if ( $code < 200 || $code >= 300 ) {
-			$decoded_body = json_decode( $body, true );
-			$log_body     = is_array( $decoded_body ) ? $this->prepare_transaction_for_log( $decoded_body ) : array( 'body_length' => strlen( $body ) );
-
-			$this->log( sprintf( 'API request returned HTTP %d', $code ), null, $log_body );
+			$this->log(
+				sprintf( 'API request returned HTTP %d', $code ),
```
bor0, line 339: These two branches lost the information that made them useful.

The non-2xx branch now logs `array( 'body_length' => strlen( $body ) )` and nothing else, dropping the decoded error body. The `WP_Error` branch above (253) swapped `get_error_message()` for `get_error_code()`, which for the WP HTTP API is almost always the constant `http_request_failed` — true of a DNS failure, a TLS error and a timeout alike.

The code being replaced already ran the decoded body through `prepare_transaction_for_log()`, so the redaction goal was met without giving up the diagnostics.

This is the shared base, so SurjoPay inherits it: a broken SurjoPay integration now produces log rows with no actionable content. Given the PR's stated goal, I'd keep the redacted body here and `get_error_message()` above.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -377,14 +483,18 @@ public function payment_notify() {
 		$payment_token  = sanitize_text_field( trim( $_REQUEST['tix_payment_token'] ?? '' ) );
 		$transaction_id = sanitize_text_field( $_REQUEST['tran_id'] ?? '' );
 		$val_id         = sanitize_text_field( $_REQUEST['val_id'] ?? '' );
+		$attendee_id    = $this->get_attendee_id_for_log( $payment_token );
+
+		if ( ! $attendee_id ) {
```
bor0, line 517: Small one, worth a log line before the return.

`get_attendee_id_for_log()` goes through `get_attendees_from_payment_token()`, which queries `draft, pending, publish, cancel, refund, failed` (camptix.php:6636) — `timeout` isn't in that list, and timed-out attendees are set to exactly that status at camptix.php:6192. So a signature-valid IPN arriving after the reservation timeout now returns here with nothing written anywhere.

To be fair about the scope: `payment_result()` uses the same status list (camptix.php:6701), so that IPN wouldn't have been *processed* before this PR either — it would have hit `Could not find attendees by payment token` and died. So the regression is the loss of that one log line, not the loss of the payment. Still the case where a log entry is most valuable, though, and `pre_attendee_timeout()` only covers it when a session key was stored.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -231,17 +341,21 @@ public function validate_options( $input ) {
 		$output = $this->options;
 
 		if ( isset( $input['merchant_id'] ) ) {
-			$output['merchant_id'] = $input['merchant_id'];
+			$output['merchant_id'] = sanitize_text_field( $input['merchant_id'] );
 		}
 
 		if ( isset( $input['store_password'] ) ) {
-			$output['store_password'] = $input['store_password'];
+			$output['store_password'] = sanitize_text_field( $input['store_password'] );
```
bor0: `sanitize_text_field()` on a secret is risky in a way it isn't on the merchant ID above. It trims whitespace, strips `<` sequences and removes invalid UTF-8 and octets — so a store password containing any of those is silently altered on save.

The failure mode is unpleasant to debug: afterwards every API call *and* every IPN hash verification fails, with only the gateway's generic "invalid credential" text to go on, and the settings screen shows a password field that looks correctly filled in.

`wp_unslash()` plus a length/charset check would keep the value byte-exact while still validating it.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -231,17 +341,21 @@ public function validate_options( $input ) {
...
 		}
 
 		if ( isset( $input['sandbox'] ) ) {
 			$output['sandbox'] = (bool) $input['sandbox'];
 		}
 
+		if ( isset( $input['min_amount'] ) && is_numeric( $input['min_amount'] ) ) {
```
bor0: `is_numeric( '' )` is false, so an admin who clears this field — the natural way to express "no minimum" — has the save silently dropped: the previous value stays in `$output` (seeded from `$this->options` at 341) and the field re-renders with it, so there's no indication anything was ignored.

Either treat `''` as `0`, or surface a validation error via `add_settings_error()`.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -39,6 +39,7 @@ public function camptix_init() {
 				'merchant_id'    => '',
 				'store_password' => '',
 				'sandbox'        => true,
+				'min_amount'     => 10.0,
```
bor0, line 42: Flagging for the release notes rather than the code: because the defaults are merged on every `camptix_init()`, this applies retroactively to sites that already have SSLCommerz configured. Any camp currently selling tickets under ৳10 starts failing checkout the moment this deploys, with no admin action and no warning — they'd only find out from the buyer-facing error.

Intentional per the description, and the threshold matches what live merchant accounts enforce anyway. But an admin notice on the settings screen, or at least a line in the deploy notes, seems worth it.

Tiny thing: the description says existing sites fall back "via `wp_parse_args()`" — it's `array_merge()` here. Same effect (saved options win, since they're the second argument), just worth matching the text to the code.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -531,13 +729,14 @@ public function pre_attendee_timeout( $attendee_id ) {
...
+			],
+			[],
+			$attendee_id
 		);
 		if ( ! $response ) {
 			return;
 		}
 
-		// If the transaction wasn't successful, bail out.
 		if ( ! in_array( $response->status, [ 'VALID', 'VALIDATED' ], true ) ) {
```
bor0: Not introduced by this PR, but it touches this function so it seems worth asking: this reads `$response->status` from `merchantTransIDvalidationAPI.php`, and my understanding is that endpoint nests its transaction data under an `element` array rather than returning `status` at the top level — unlike `validationserverAPI.php`, which `verify_transaction()` reads the same way and where the flat shape is correct.

If that's right, this check always fails and the timeout-recovery path never fires, which would also make it the reason the "charged but timed out" case has been hard to observe. I couldn't confirm the response shape against the SSLCommerz docs — do you have a captured response from this endpoint? Happy for this to be a follow-up either way, I just don't want to assume it works.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Thanks — the intent is explicitly to leave the order unchanged when the signature cannot be verified. I added a code comment, a clear rejection log, and regression coverage. The same applies to a signed transaction whose ID doesn’t match the URL token: failing that order would allow a replayed callback to affect another order. A matching, signature-valid IPN that fails server-side validation still calls payment_result(FAILED). The hash-mismatch test expectation should be “rejected and logged, with no order status change.”

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Added a log and buyer notice before both rejection returns. The transaction ID now comes from the validated POST payload. Although the payload is signed, it must still match the URL’s order token before changing that order. Matching callbacks continue to call payment_result(CANCELLED) or payment_result(FAILED). Regression tests cover both matching and rejected callbacks.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Added both bank_tran_id and tran_id to the persisted transaction-details allowlist. This covers notification processing and timeout recovery. Log redaction remains separate, and tests verify reference retention without exposing those values in the timeout diagnostic log.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Fixed the short-value matching so a name such as “Al” no longer changes “allowed.” Short values use letter/number boundaries, and replacements happen in one pass so redaction markers aren’t processed again. Buyer-facing errors now use a generic localized message; gateway diagnostics remain in redacted logs. Added regression coverage for these cases.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-base.php
nazmulch11: Restored transport error messages alongside their error codes, plus decoded JSON diagnostics for non-2xx responses. Sensitive fields and known echoed credentials are redacted before logging. Non-JSON bodies retain only their length. This covers both the shared client used by SurjoPay and SSLCommerz’s separate client, with tests exercising both error paths.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Added a log before returning when a signature-valid callback has no eligible attendee, noting that the reservation may have timed out. Signature verification happens before the attendee lookup. This restores visibility without changing CampTix’s handling of expired reservations. Added regression coverage.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Removed sanitize_text_field() from the password save path. Valid UTF-8 strings without control characters are preserved exactly; invalid input produces a settings error and retains the previous password. WordPress already unslashes the options before this validator runs, so I avoided a second wp_unslash() that would alter backslashes. Tests cover spaces, markup characters, percent sequences, quotes, and backslashes.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Clearing the field now saves 0.0, disabling the local minimum check. An omitted field preserves the existing value. Invalid, negative, or non-finite values produce a settings error instead of being silently ignored. Added regression coverage.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Added settings help explicitly stating that the 10 BDT default applies to existing stores and both modes. It also explains that clearing the field or entering zero disables the local check, while the gateway may still enforce its own minimum. You’re also right that the default merge uses array_merge(), not wp_parse_args().

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: I checked the [official SSLCommerz documentation](https://developer.sslcommerz.com/doc/v4/index.html): querying this endpoint by sessionkey returns a flat transaction response; querying by tran_id returns the element array. This function uses sessionkey, so the flat parser is appropriate. I added explicit JSON format and a regression test for recovery using that response shape. I don’t have a captured live response from this endpoint.

nazmulch11:
Thanks @bor0 — I’ve addressed the review findings:
- Restored transaction references in stored details while keeping logs redacted.
- Added explicit callback rejection logs and notices. Invalid signatures and mismatched order tokens intentionally leave orders unchanged to prevent replay attacks; valid, matching callbacks still call payment_result().
- Restored redacted API diagnostics for both gateways.
- Fixed short-name redaction, preserved password bytes, and made clearing the minimum save zero.
- Documented the default minimum’s effect on existing stores in settings help.
The remaining checkout failure came from rejecting epay-gw.sslcommerz.com. I added that exact HTTPS host to the live allowlist, preserving sandbox separation and rejecting arbitrary subdomains. Checkout now works.
For timeout recovery, the official documentation confirms a flat response for sessionkey queries; the nested element response applies to tran_id queries.
Validation: 13 tests / 43 assertions and PHPCS pass. Full-repository PHPUnit is blocked locally by missing GatherPress, and credentialed SurjoPay end-to-end testing remains outstanding.
Could you please take another look?

bor0:
@nazmulch11 thanks for the updates — I re-reviewed at `667ce5d`. CI is green on 8.3/8.4/8.5, and the core of the change looks right: `tran_id` is set to `$payment_token` at session creation, and `payment_notify()`, `payment_cancel()`, `payment_failed()`, `verify_transaction()` and `pre_attendee_timeout()` all `hash_equals()` the signed `tran_id` against the URL token before touching an order. Exact per-mode host allowlisting is back. Nice work.

Most of what's left is in the logging/redaction layer, plus one deploy-behaviour default.

### Confirmed issues

**1. The redaction pattern list gets poisoned by `0`, mangling the diagnostics this PR is restoring** — `class-gateway-base.php:169`

The "is this key sensitive?" probe is `prepare_transaction_for_log( [ $key => $value ] ) !== [ $key => $value ]`, but `SSLCommerz::prepare_transaction_for_log()` runs `array_filter()` on its result. So *any* falsey value — `0`, `'0'`, `false` — returns `[]`, compares unequal, and is collected as a "secret" to strip from free text. The session-init payload carries `'emi_option' => 0`, so `0` becomes a literal redaction pattern:

```
in:  cURL error 28: Operation timed out after 30000 ms with 0 bytes received
out: cURL error 28: Operation timed out after 30000 ms with [redacted] bytes received
```

Probing with an explicit key-sensitivity helper (or `parent::prepare_transaction_for_log()`) would avoid this.

**2. `usort()` sorts by pattern length, not value length, inverting the intended longest-first order** — `class-gateway-base.php:145`

Values under 6 chars get a 23-character `(?<![\pL\pN])…(?![\pL\pN])` wrapper, so a short value's pattern can outrank a longer value's. Where the short value is a prefix of the longer one, the remainder leaks:

```
values: 12345, 12345-6789
in:     ref 12345-6789 failed
out:    ref [redacted]-6789 failed
```

Sorting on the original value length, before wrapping, fixes it.

**3. The `wp_unslash()` comment is inaccurate** — `class-gateway-sslcommerz.php:358`

The comment says the Settings API has already unslashed the value, but `CampTix_Payment_Method::_camptix_validate_options()` (`camptix/inc/class-camptix-payment-method.php:190`) re-reads `$input = $_POST["camptix_payment_options_{$this->id}"]` directly, so `validate_options()` gets slashed input.

This is pre-existing rather than a regression — but the PR's stated goal is to preserve valid store-password bytes, and it doesn't get there: a password containing `"`, `'` or `\` gains a backslash on each save. Once the stored bytes drift, `md5( store_password )` stops matching and every IPN check fails, which under the new early `return` leaves orders silently unpaid instead of FAILED. The control-character regex doesn't catch backslashes. Please add `wp_unslash()` (same for the new `sanitize_text_field( $input['merchant_id'] )`).

### Worth changing before merge

**4. `min_amount => 10.0` is opt-out rather than opt-in** — `class-gateway-sslcommerz.php:42`

Existing stores have no saved key, so the floor applies with no admin action and keeps applying until someone saves the settings page. Any sub-10-BDT total — a coupon or discounted order, or a cheap tier — returns `PAYMENT_STATUS_FAILED` before a session starts. You flag this in the description as a deployment note; defaulting to `0.0` would ship the same feature without the silent behaviour change for live camps.

**5. Failure reasons are dropped from stored transaction details** — `class-gateway-sslcommerz.php:644`

The allowlist in `prepare_sslcommerz_transaction_details()` omits `error`, `failedreason` and `error_reason`; previously the whole redacted POST was stored. For a declined payment, organizers now see `status: FAILED` in wp-admin and nothing about why. These are gateway-supplied free text rather than customer data — worth allowlisting, run through `prepare_gateway_message_for_log()` if you want them scrubbed.

**6. Expired reservations get a callback-verification error** — `class-gateway-sslcommerz.php:573` and `:605`

`get_attendees_from_payment_token()` excludes `timeout` status (`camptix.php:6636`). After the 24h timeout sweep, a buyer returning to a stale checkout page and clicking Cancel hits `$attendee_id === null` and sees *"We could not verify this payment callback. Please contact the event organizers before trying another payment."* Splitting "no attendee" from "token mismatch" would let the expired case show a plain "your reservation expired" notice instead.

### Minor

**7.** `payment_notify()` still reads `$_REQUEST['tran_id']` for the value written to `tix_transaction_id` (`:513`), while the binding check on `:534` correctly uses `$_POST`. Benign under the default `request_order = "GP"`, but inconsistent with the deliberate `$_POST` tightening you made in the browser callbacks.

**8.** Only the first of the three `preg_replace()` calls is null-guarded (`base:148`, vs `:150` and `:151`); a PCRE failure returns `null`, which then reaches `sanitize_text_field( null )`.

**9.** The new test file lives in `camptix/tests/addons/` but exercises `camptix-bd-payments`. It does get picked up by the `CampTix` testsuite so CI runs it — but a `camptix-bd-payments/tests/` directory with its own `phpunit.xml.dist` entry would match how every other plugin in the repo is wired.

### Checked and sound

Exact-host matching, including the `@evil.example` and `.evil.example` cases; `is_allowed_https_host()`'s `$allow_subdomains = true` default preserving SurjoPay behaviour; the signature-first reordering in `payment_notify()`; the integer-cent minimum comparison and the `'1e999'`/negative validation paths; `$response->status ?? ''` null-safety; recursive redaction in the base class covering nested `transaction_details`; SurjoPay's `$attendees[0]->ID` being guarded by the earlier `empty()` check.

@obenland's one remaining open thread (`class-gateway-sslcommerz.php:235`, PII and `store_id` in the initiation-payload log) is addressed at this commit — the log now carries only amounts and counts. Worth resolving so it isn't blocking.

nazmulch11:
@bor0 Thanks for the thorough review; it addressed all nine points:

The redaction pattern was poisoned by 0 Extracted is_sensitive_key() in the base class and replaced the prepare_transaction_for_log() array-comparison probe in prepare_api_diagnostics() with a direct key check. The 'emi_option' => 0 value can no longer poison the pattern list.

usort() by pattern length—Now deduplicating and sorting by original value length (longest first) before wrapping short values in boundary assertions. The old usort() on wrapped patterns is removed.

wp_unslash() and inaccurate comment—Added wp_unslash() for both merchant_id and store_password in validate_options(). Fixed the comment to explain that CampTix reads directly from $_POST (which is slashed by wp_magic_quotes()). Passwords with ", ', or \ no longer drift on each save.

min_amount default — Changed from 10.0 to 0.0 (opt-in). Updated the settings help text. Existing stores won't silently reject sub-10-BDT orders.

Failure reasons dropped - Added error, failedreason, and error_reason to the $allowed_keys in prepare_sslcommerz_transaction_details(). Organizers can now see why a payment was declined.

Expired reservations - Split the combined! $attendee_id || ! hash_equals(...) guard in payment_cancel() and payment_failed() into two separate checks. Token mismatch → security error. Missing attendee → "your reservation has been cancelled or has expired."

$_REQUEST → $_POST in payment_notify()—Changed tran_id and val_id reads from $_REQUEST to $_POST, consistent with the deliberate $_POST tightening in browser callbacks.

Null-guard preg_replace() — Added ?? '' to the URL and email regex replacements in prepare_gateway_message_for_log() (lines 171–172), preventing sanitize_text_field(null) on PCRE failure.

Test file location—Moved test-payment-sslcommerz.php from camptix/tests/addons/ to camptix-bd-payments/tests/ with a dedicated CampTix BD Payments testsuite entry in phpunit. xml.dist, matching how other gateway plugins are wired.

Also resolved @obenland's remaining open thread - the initiation-payload log at :235 already carries only amounts and counts at this commit.

nazmulch11:
@bor0, I appreciate your detailed review! I've addressed all nine points:

Redaction pattern issue resolved—I extracted is_sensitive_key() into the base class and replaced the array-comparison check in prepare_api_diagnostics() with a direct key check. Now, the 'emi_option' => 0 value won't disrupt the pattern list anymore.

usort() by pattern length — We're now deduplicating and sorting by the original value length (longest first) before wrapping shorter values in boundary assertions. The previous usort() on wrapped patterns has been removed.

wp_unslash() and comment clarification — I added wp_unslash() for both merchant_id and store_password in validate_options(). I also updated the comment to clarify that CampTix reads directly from $_POST (which is slashed by wp_magic_quotes()). This means passwords with ", ' or \ won't drift on each save anymore.

min_amount default — I've changed this from 10.0 to 0.0 (opt-in). The settings help text has been updated, ensuring that existing stores won't silently reject orders below 10 BDT.

Failure reasons included — I've added error, failedreason, and error_reason to the $allowed_keys in prepare_sslcommerz_transaction_details(). Now, organizers can see why a payment was declined.

Expired reservations — I split the combined check in payment_cancel() and payment_failed() into two separate checks. A token mismatch will now trigger a security error, while a missing attendee will show the message: "your reservation has been cancelled or has expired."

$_REQUEST changed to $_POST in payment_notify() — I updated tran_id and val_id reads from $_REQUEST to $_POST, aligning with the tighter $_POST handling in browser callbacks.

Null-guard for preg_replace() — I added ?? '' to the URL and email regex replacements in prepare_gateway_message_for_log() (lines 171–172), preventing sanitize_text_field(null) on PCRE failure.

Test file relocation — I moved test-payment-sslcommerz.php from camptix/tests/addons/ to camptix-bd-payments/tests/, creating a dedicated CampTix BD Payments testsuite entry in phpunit.xml.dist, similar to how other gateway plugins are organized.

Additionally, I resolved @obenland's remaining open thread — the initiation-payload log at :235 now only carries amounts and counts at this commit.

bor0 (commented):
Re-reviewed at `4ecf382`. Everything from my last round is addressed — thanks for working through it.

I re-ran the two redaction cases rather than eyeballing them, and both behave now:

```
ref 12345-6789 failed  ->  ref [redacted] failed      (was: ref [redacted]-6789 failed)
emi_option => 0        ->  not collected as a pattern  (was: collected, ate every standalone 0)
```

Extracting `is_sensitive_key()` is the right shape for the probe fix, and `d9c650ff8` is the correct complement to the `wp_unslash()` change — the test now feeds `wp_slash( $password )`, so it actually exercises the production input shape. Moving the tests to `camptix-bd-payments/tests/` with its own `phpunit.xml.dist` entry is also the right call. Moving `val_id` to `$_POST` alongside `tran_id` goes a step further than I asked for, and is correct: the signature is verified over `$_POST`, so that is the only array the value should be read from.

CI is green on 8.3/8.4/8.5 including the new `CampTix BD Payments` suite.

What is left is all minor — four inline notes below, plus one thing with no line to hang it on:

**The PR description is now stale.** It still says *"Defaults to 10 BDT, including existing stores without a saved value, through `array_merge()`"* and carries the whole **Deployment Note** about reviewing the setting before deploy. Both are obsolete now that the default is `0.0`. Since the description is what ends up in the merge record, worth updating before this goes in.

Nothing here blocks merge from my side.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -428,8 +573,22 @@ public function payment_cancel() {
...
+
+		if ( ! hash_equals( (string) $payment_token, (string) $transaction_id ) ) {
+			$camptix->log( 'SSLCommerz browser callback rejected: transaction ID mismatch; order status unchanged.', $attendee_id );
+			$camptix->error( __( 'We could not verify this payment callback. Please contact the event organizers before trying another payment.', 'bd-payments-camptix' ) );
+			return;
+		}
+
+		if ( ! $attendee_id ) {
+			$camptix->log( 'SSLCommerz browser callback: no active attendee found; the reservation may have expired.' );
+			$camptix->error( __( 'Sorry, but the reservation you are trying to use has been cancelled or has expired.', 'wordcamporg' ) );
```
bor0: Wrong text domain. This plugin declares `Text Domain: bd-payments-camptix` in its header, and the other 25 strings under `includes/` use it — these two new strings (here and line 624) use `'wordcamporg'`, so they will not be translated with the plugin.

```suggestion
			$camptix->error( __( 'Sorry, but the reservation you are trying to use has been cancelled or has expired.', 'bd-payments-camptix' ) );
```

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -451,8 +610,22 @@ public function payment_failed() {
...
```
bor0: Same text domain issue as line 587.

```suggestion
			$camptix->error( __( 'Sorry, but the reservation you are trying to use has been cancelled or has expired.', 'bd-payments-camptix' ) );
```

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
```diff
@@ -374,45 +508,56 @@ public function template_redirect() {
...
+
 		$payment_token  = sanitize_text_field( trim( $_REQUEST['tix_payment_token'] ?? '' ) );
-		$transaction_id = sanitize_text_field( $_REQUEST['tran_id'] ?? '' );
-		$val_id         = sanitize_text_field( $_REQUEST['val_id'] ?? '' );
+		$transaction_id = sanitize_text_field( $_POST['tran_id'] ?? '' );
+		$val_id         = sanitize_text_field( $_POST['val_id'] ?? '' );
+		$attendee_id    = $this->get_attendee_id_for_log( $payment_token );
+
+		if ( ! $attendee_id ) {
+			$camptix->log( 'SSLCommerz callback has no eligible attendee; the reservation may have timed out.' );
```
bor0: Minor: the check order here diverges from the browser callbacks. `payment_notify()` tests attendee-exists *before* the `tran_id` binding, while `payment_cancel()`/`payment_failed()` were reordered to test the binding first.

The security outcome is identical either way — both paths return without changing state — but an expired reservation carrying a mismatched token logs "no eligible attendee" on this path and "transaction ID mismatch" on the other, which makes the logs harder to read against each other. Worth matching the order.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-base.php
```diff
@@ -36,31 +36,91 @@ protected function build_callback_urls( $payment_token ) {
...
+	 * without comparing array outputs — which breaks when a subclass applies
+	 * array_filter() and strips falsey values like 0.
 	 *
-	 * @return array Sanitized data safe for logging.
+	 * @param string $key Field key to test.
+	 *
+	 * @return bool
 	 */
-	protected function prepare_transaction_for_log( $data ) {
+	protected function is_sensitive_key( $key ) {
```
bor0, line 49: Extracting this is the right fix for the probe. One loose end: `SSLCommerz` still overrides `prepare_transaction_for_log()` with its own extra keys (`value_a`-`value_d`, the `cus_`/`ship_` prefixes, `pass`, `key`, `validation_id`) but does **not** override `is_sensitive_key()`.

Field redaction is unaffected — `prepare_api_diagnostics()`'s `$scrub` still calls the override. But its *value collection* now runs through the base list only, so those extra keys' values no longer become redaction patterns for free-text diagnostics. Low impact in practice here (the affected fields are the hardcoded `'Dhaka'`/`'1209'` address values), but the two lists have now drifted apart. Cleaner to move the SSLCommerz-specific keys into an `is_sensitive_key()` override and let the base `prepare_transaction_for_log()` do the walking.

public_html/wp-content/plugins/camptix-bd-payments/readme.md
```diff
@@ -38,7 +38,8 @@ A payment gateway addon for the [CampTix](http://github.com/developer-developer/
 1. Get your store ID and password from [SSLCommerz](https://developer.sslcommerz.com/registration/)
 2. Enable SSLCommerz in Tickets > Setup > Payment
 3. Enter your store ID and password
-4. Toggle sandbox mode for testing
+4. Set the minimum transaction amount configured for your SSLCommerz store
```
bor0: Now that the default is `0.0` (disabled), this reads as a required setup step when it is optional. Something like:

```suggestion
4. Optionally set a minimum transaction amount, if your SSLCommerz store enforces one
```

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Fixed

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: fixed

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-base.php
nazmulch11: Good catch. I moved the missing-attendee check below the transaction ID binding check. A mismatched callback now logs the transaction ID mismatch even if the reservation has expired, while valid callbacks with no eligible attendee still receive the timeout log. I added a regression test for this case.

public_html/wp-content/plugins/camptix-bd-payments/readme.md
nazmulch11: Agreed. Since the default is disabled, setting a minimum is optional. I updated the setup step to reflect that.

public_html/wp-content/plugins/camptix-bd-payments/includes/gateway/class-gateway-sslcommerz.php
nazmulch11: Good catch. I moved the missing-attendee check below the transaction ID binding check, so expired reservations with mismatched IDs log the mismatch consistently. The order state remains unchanged, and I added a regression test for this case.

nazmulch11:
Thanks, @bor0, for the thorough review. I addressed the follow-ups: callback handling now checks the transaction ID binding before reporting a missing attendee, the buyer-facing messages use the plugin’s text domain, and SSLCommerz-specific sensitive fields now participate in diagnostic redaction. I also clarified that the minimum amount is optional and updated the PR description to reflect the disabled-by-default setting.
I added regression coverage for callback ordering and diagnostic redaction. PHPCS, PHP syntax checks, and git diff --check pass. I couldn’t run PHPUnit locally because the WordPress test library is missing; CI will need to verify the suite.

bor0:
Re-reviewed at `28ba929`. All five points from my last review are addressed:

- Text domain fixed on both strings.
- `payment_notify()` now checks the `tran_id` binding before the attendee lookup, so it matches the browser callbacks.
- The `prepare_transaction_for_log()` override is replaced by an `is_sensitive_key()` override, so the two key lists can no longer drift apart.
- readme step 4 reads as optional.
- Description is updated: the stale 10 BDT default and the deployment note are gone.

I re-ran the redaction chain against the new override rather than trusting the refactor, since adding keys to the child could have reopened the `0` bug:

```
emi_option => 0 collected?   no
value_a collected?           yes
cURL error 28 ... with 0 bytes received  ->  unchanged
Gateway echoed private-ref               ->  Gateway echoed [redacted]
```

Two deliberate behavior changes worth naming, both improvements: redacted fields now show as `[redacted]` rather than being dropped from the log entirely, and `array_filter()` is gone so falsey values are retained. The updated tests assert the new shape.

No new findings from me, and CI is green on 8.3, 8.4 and 8.5.

@obenland would you take another look before this goes in? Your original threads are all addressed at this commit, and I'd like a second pass on the redaction refactor in `class-gateway-base.php` specifically.

nazmulch11:
@obenland Can you have a look?

PR #1761, opened 2026-07-10
CampTix Invoices: Don't record the invoice PDF when generation fails

### LINKED ISSUE #1760: CampTix Invoices: failed PDF generation still records the invoice document, causing 404 downloads
## Summary

When invoice PDF generation fails in the CampTix Invoices plugin, the invoice is still marked as having a document: `create_invoice_document()` writes the `invoice_document` post meta unconditionally, even if wkhtmltopdf produced no file. The invoice then looks fine in the admin, but the "Download invoice" button links to a file that does not exist (404), and `send_invoice()` tries to attach the missing file to the attendee email.

## Mechanics

In `camptix-invoices/includes/class-camptix-addon-invoices.php`, `create_invoice_document()`:

```php
$tmp_path = $pdf_generator->generate_pdf_from_string( $invoice_content, $filename );
// ...
rename( $tmp_path, $invoices_dirname . '/' . $filename );

update_post_meta( $invoice_id, 'invoice_document', $filename );
```

Three failure points are silently ignored:

1. `WordCamp_Docs_PDF_Generator::generate_pdf_from_file()` runs `exec( 'wkhtmltopdf ...' )` with no error checking and returns the target path whether or not the binary produced anything.
2. `rename()` then fails (source file doesn't exist), emitting only a PHP warning.
3. `update_post_meta()` records the filename anyway.

Because nothing is logged through CampTix, there is no trace of the failure anywhere organizers or support can see — the first symptom is attendees reporting 404s on their invoice links.

This is not hypothetical: when wkhtmltopdf was temporarily unavailable on production (July 2026), every invoice created during that window ended up with `invoice_document` meta pointing at a PDF that was never written, producing network-wide invoice download 404s. Since `save_post` regenerates the document reusing the stored filename, those invoices could be healed by re-saving once the binary was restored — but the bad meta should never have been written in the first place.

## Proposed fix

In `create_invoice_document()`, only record the document once the PDF verifiably exists:

- bail (and log via `$camptix->log()`) if the generated file is missing or zero bytes;
- bail (and log) if `rename()` into `uploads/camptix-invoices/` fails;
- only then `update_post_meta()`.

The function's return value is ignored by both callers, so an early `return false` needs no caller changes, and the shared `WordCamp_Docs_PDF_Generator` class is untouched.

PR to follow.

wputasic:
Fixes #1760.

## Problem

`create_invoice_document()` records the `invoice_document` post meta unconditionally. `WordCamp_Docs_PDF_Generator` runs `exec( 'wkhtmltopdf ...' )` with no error checking and returns the target path either way, and the `rename()` return value is ignored — so when PDF generation fails, the invoice still ends up pointing at a file that was never written. The "Download invoice" button then 404s, `send_invoice()` attaches a missing file, and nothing is logged anywhere organizers or support can see. This happened network-wide when wkhtmltopdf was temporarily unavailable on production in July 2026.

## Fix

Only record the document once the PDF verifiably exists:

- bail and log via `$camptix->log()` if the generated file is missing or zero bytes;
- bail and log if the `rename()` into `uploads/camptix-invoices/` fails;
- only then write the `invoice_document` meta.

Both callers ignore the function's return value, so the early `return false` requires no caller changes, and the shared `WordCamp_Docs_PDF_Generator` class is untouched. Invoices that hit the guard can be regenerated later simply by re-saving (the `save_post` handler already regenerates, reusing the stored filename when present).

## Testing

Verified in the local Docker environment (multisite, camptix-invoices active):

- **Failure path** (wkhtmltopdf unavailable): `create_invoice_document()` returns `false`, no `invoice_document` meta is written, and the expected log entry (message + filename) appears in the CampTix network log via camptix-network-tools. On the unpatched code the same scenario writes meta pointing at a nonexistent PDF — reproducing the July incident.
- **Happy path**: PDF is generated, moved into `uploads/camptix-invoices/`, and the meta is written exactly as before.

### DIFF
--- a/public_html/wp-content/plugins/camptix-invoices/includes/class-camptix-addon-invoices.php
+++ b/public_html/wp-content/plugins/camptix-invoices/includes/class-camptix-addon-invoices.php
@@ -424,6 +424,15 @@ class CampTix_Addon_Invoices extends \CampTix_Addon {
 		$upload_dir    = wp_upload_dir();
 		$tmp_path      = $pdf_generator->generate_pdf_from_string( $invoice_content, $filename );
 
+		global $camptix;
+
+		// Only record the document once the PDF verifiably exists, so a wkhtmltopdf
+		// failure can't leave the invoice pointing at a file that was never written.
+		if ( ! file_exists( $tmp_path ) || 0 === filesize( $tmp_path ) ) {
+			$camptix->log( __( 'Invoice PDF generation failed: wkhtmltopdf produced no output.', 'wordcamporg' ), $invoice_id, array( 'filename' => $filename ) );
+			return false;
+		}
+
 		if ( ! empty( $upload_dir['basedir'] ) ) {
 			$invoices_dirname = $upload_dir['basedir'] . '/camptix-invoices';
 			if ( ! file_exists( $invoices_dirname ) ) {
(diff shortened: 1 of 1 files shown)


gedex (approved):
Thanks, @wputasic! Tested locally with this PR merged onto the `production` branch. The `production` branch already skips the meta when the PDF is missing, but a 0-byte PDF still gets saved as the invoice. This PR stops that and logs every failure. Missing file, empty file and a good PDF all behave as expected.

Issue #1812, opened 2026-07-27
Groups: front-end blocks miss focus management and use non-native controls (a11y)


wppoland:
# Groups: front-end blocks miss focus management and use non-native controls (a11y)

Looking at the Groups front-end work from an organizer's point of view (I run [wpgdynia.pl](https://wpgdynia.pl), a WordPress meetup in Gdynia, Poland), and I noticed there is currently no accessibility issue in the Groups backlog. Given that these blocks replace Meetup.com's UI for every group member, not just organizers, I wanted to flag a few concrete gaps before the mu-plugin lands.

All references below are to `wporg-groups-frontend` as it stands in #1793.

### What is already right

Worth stating so this does not read as a blanket complaint:

- The attendees modal is `display: none` when closed (`src/blocks/event-rsvp/style.css:82-94`), so it is correctly hidden from assistive tech rather than just visually offscreen.
- `role="dialog"`, `aria-modal="true"` and an `aria-label` are all present on the dialog.
- Escape and backdrop click are handled (`handleEscape`, `handleBackdropClick`).

So the dialog pattern is about half implemented. The missing half is focus.

### 1. No focus management in the RSVP modal

`src/blocks/event-rsvp/view.js` is 341 lines and contains zero references to `focus`, `activeElement` or `aria`. Concretely:

- Opening the modal does not move focus into it. A keyboard or screen reader user activates "View attendees", the dialog appears visually, and focus stays behind it on the trigger.
- Focus is not trapped, so tabbing walks straight out of the open dialog into the page behind it.
- Closing does not restore focus to the trigger.

This is WCAG 2.1 AA (2.4.3 Focus Order, 2.4.7 Focus Visible). The same applies to the event create/edit modal, although that one uses `@wordpress/components` `Modal`, which handles focus itself, so it is likely fine. Worth confirming.

### 2. `<div role="button" tabindex="0">` where a `<button>` would do

`src/blocks/event-rsvp/render.php:1084-1089` builds the attendees summary as a div with `role="button"`, `tabindex="0"` and a hand-written `handleSummaryKeydown` handler to emulate Enter/Space. A native `<button>` gives all of that for free, plus correct behaviour for voice control and for browser/AT combinations that treat ARIA roles differently. The custom keydown handler can then be deleted.

### 3. Avatar `alt` text becomes part of the control name

[...]

PR #1814, opened 2026-07-27
Groups: fix empty X-WP-Nonce on group join/leave

### LINKED ISSUE #1804: Groups: Leave/Join group always fails — REST calls send an empty X-WP-Nonce
Found while testing PR #1793 (`split/1699/wporg-groups-frontend`, split from #1699) live in the browser instead of via curl.

## Reproduction

Tested as a real "Member" (subscriber-tier) user on a group site with #1793 merged on top of `migration-branch`.

1. Log in as a subscriber-tier member of a group site.
2. Visit the group front page or `/members/`.
3. Click "Leave group" (confirms via `window.confirm`).

Nothing visibly happens — no error, member stays in the group. The same failure mode applies to "Join this group" for a logged-in non-member.

Network tab shows why:
```
DELETE .../wp-json/wporg-groups/v1/members/leave → 403
{"code":"rest_cookie_invalid_nonce","message":"Cookie check failed"}
```

## Root cause

`src/blocks/group-membership/view.js`'s `getNonce()` reads `window.wpApiSettings.nonce`, falling back to `window.GatherPress.urls.eventApiUrl` + a nonce endpoint. Neither exists on any front-end page checked (`window.wpApiSettings` and `window.GatherPress` are both `undefined` on the group front page and `/members/`). So every join/leave request sends an empty `X-WP-Nonce` header, which WordPress core rejects unconditionally. This is deterministic, not session- or environment-specific.

## Impact

Critical: nobody can join or leave a group through the front-end UI as shipped. The failure is silent — the `catch` swallows fetch errors, and a non-2xx response with `data.success` undefined just leaves the button inert — so it looks like nothing happened rather than surfacing an error.

## Fix direction

Enqueue `wp-api-fetch` (which auto-localizes `wpApiSettings.nonce` and refreshes it on 403) for the `group-membership` view script, or explicitly `wp_localize_script()` a nonce for it. Don't rely on `window.GatherPress` being present outside GatherPress's own templates.

## How to verify a fix

Log in as any group member, click "Leave group" and then rejoin, confirm the member count updates with no console/network error, and check the network tab that `X-WP-Nonce` is non-empty on both requests.

wppoland:
Fixes #1804.

# What

Join and leave on the `wporg/group-membership` block send an empty `X-WP-Nonce`, so every request is rejected with `403 rest_cookie_invalid_nonce`. Nobody can join or leave a group from the front end.

# Why

`view.js` resolved the nonce client side, from two sources that are not present on the pages where this block renders:

- `window.wpApiSettings.nonce` is localized by the `wp-api` script, which this page never enqueues. `group-membership` ships as a `viewScriptModule`, and script modules cannot pick up `wp_localize_script` data anyway.
- the fallback read `window.GatherPress.urls.eventApiUrl`, which only exists inside GatherPress's own templates, not on the group front page or `/members/`.

Both are `undefined`, so `getNonce()` fell through to its final `return ''`.

# How

The block already server-renders a `data-wp-context` payload containing `joinApi`, `leaveApi` and `loginUrl`. The nonce belongs there too, next to the endpoints it authenticates, so this adds `nonce` to that context and reads `ctx.nonce` in the two actions. That deletes `getNonce()` and its cache entirely, 22 lines, along with the dependency on globals that this block has no control over.

The nonce is only emitted for logged-in visitors. Anonymous markup is the cacheable case, and an anonymous visitor is redirected to `loginUrl` before any request is made, so there is nothing to authenticate.

I also stopped the two `catch` blocks from swallowing everything. This is what made the bug present as "the button does nothing": the response was a 403, but `data.success` was simply `undefined`, so neither branch ran and nothing reached the console. Non-2xx now throws, and the catch logs. Deliberately kept to `console.error` rather than adding user-facing error UI, since that is a design decision beyond this fix, and #1812 covers announcing state changes to assistive tech.

# Testing

Log in as a subscriber-tier member of a group site.

1. Click "Leave group" and confirm. Member count drops, page reloads, you are out.
2. Click "Join this group". Count goes up, page reloads, you are back in.
3. In the network tab, `X-WP-Nonce` is non-empty on both requests and both return 200.

Before this change both requests return `403 rest_cookie_invalid_nonce` with an empty header, and the console stays silent.

# Scope

Only `group-membership` is touched. `event-rsvp/view.js` also resolves a nonce at runtime, but through a server-provided `ctx.apiBase` (`rest_url( 'gatherpress/v1/event' )`) rather than a `window` global, so it does not share this failure and I have left it alone. `event-manage` and `group-settings` do not hand-roll nonces at all.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-membership/render.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-membership/render.php
@@ -29,6 +29,8 @@ $member_count    = $user_count['total_users'] ?? 0;
 $join_api        = rest_url( 'wporg-groups/v1/members/join' );
 $leave_api       = rest_url( 'wporg-groups/v1/members/leave' );
 $login_url       = wp_login_url( get_permalink() ?: home_url() );
+// Only logged-in visitors can join or leave, and only their markup should carry a nonce.
+$rest_nonce      = $is_logged_in ? wp_create_nonce( 'wp_rest' ) : '';
 $count_label     = sprintf(
 	_n( '%s member', '%s members', $member_count, 'wporg-groups-frontend' ),
 	number_format_i18n( $member_count )
@@ -47,6 +49,7 @@ $context = array(
 	'joinApi'       => $join_api,
 	'leaveApi'      => $leave_api,
 	'loginUrl'      => $login_url,
+	'nonce'         => $rest_nonce,
 	'loading'       => false,
 );
 
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-membership/view.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-membership/view.js
@@ -48,10 +48,14 @@ store( 'wporg/group-membership', {
 				const resp = await fetch( ctx.joinApi, {
 					method: 'POST',
 					credentials: 'same-origin',
-					headers: { 'X-WP-Nonce': await getNonce() },
+					headers: { 'X-WP-Nonce': ctx.nonce },
 				} );
 
 				const data = await resp.json();
+
+				if ( ! resp.ok ) {
+					throw new Error( data?.message || resp.statusText );
+				}
 
 				if ( data.success ) {
 					ctx.isMember = true;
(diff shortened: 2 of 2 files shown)


bor0 (approved):
Confirm it now works! 🎉 

https://github.com/user-attachments/assets/b95a931e-5581-4aea-87da-82d6da435482

bor0:
Thank you @wppoland for your contribution 🙌

raicem:
Hey @wppoland, thanks for the PR. Before merging, we resolved conflicts and pushed directly to this fork.

Issue #1799, opened 2026-07-23
Set up testing infra (agentic, phpunit, e2e)


bor0:
GatherPress updates fairly often and its changes can silently break backwards compatibility with our Groups integration without us noticing. We should have agentic testing instructions set up so this can be checked automatically/on-demand rather than relying on someone remembering to click through manually.

**Update after actually running #1793's test plan end-to-end (both curl and browser):** the curl-based test plans in #1790, #1791, #1792, #1793 are a good starting point for REST correctness, but on their own they are **not sufficient** — running the full flow through a real browser (not just the documented curl steps) as different roles (Organiser/editor, Event Organiser/author, Member/subscriber, anonymous) surfaced several bugs the curl plans never caught, including one critical privilege-escalation issue (editors could promote any member to Administrator via stock wp-admin — unrelated to any single curl-tested endpoint). None of these were visible from the REST test plan alone. So "lift the test plan as-is" needs revising to "lift it as a base, then extend it" — see updated plan below.

## Updated test plan for these instructions

**1. Environment setup**
- Check out the relevant branch, `npm ci` + `npm run build --workspace=public_html/wp-content/mu-plugins/wporg-groups-frontend`.
- Confirm GatherPress is installed and active on the group site under test.
- Create one test user per role tier (Organiser/editor, Event Organiser/author, Member/subscriber) with known passwords for browser login — not just application passwords for curl.

**2. Browser-based pass, per role** (this is the part the old plan was missing)
- Log in as each role in turn and walk the actual UI: event create/edit, RSVP, group-settings tabs (Events/Venues/Members/Design/About), member directory, join/leave.
- Explicitly check what each role *cannot* see/do, not just what it can — e.g. confirm a lower tier has no management UI rendered at all, not just that the button is disabled.
- Check wp-admin too, not just the custom front-end — some capability gaps (or leaks) only show up there (e.g. #1800, and the promote_users issue found this session).
- Log in as anonymous (logged out) and confirm public pages render correctly and every interactive action redirects to login rather than silently failing.

**3. Direct REST pass, bypassing the UI**
- Hit every registered route in the plugin's namespace directly (`GET {namespace}` lists them) as anonymous, and as each role, independent of what the UI shows — confirms the REST layer enforces permissions itself rather than relying on the UI to hide things.
- Attempt IDOR-style requests explicitly: can a lower-tier role affect another user's data by passing a different ID than their own?

**4. GatherPress-deactivated pass**
[...]

PR #1815, opened 2026-07-27
Groups: require an editing capability to read venues over REST

### LINKED ISSUE #1803: Groups: venue post data leaks via REST API despite non-public lockdown
Found while investigating a comment from @carstingaxion on #1791 about the `_gatherpress_venue` shadow taxonomy possibly bypassing the venue lockdown.

Their exact suggestion (taxonomy query var, e.g. `?_gatherpress_venue=slug`) does 404 correctly — that route is closed. But digging into it surfaced a related, more serious gap: **the venue post type's REST endpoint is fully open to unauthenticated requests**, regardless of the lockdown.

## Reproduction

`gatherpress-groups-tweaks.php` makes `gatherpress_venue` non-public via `register_post_type_args`:

```php
$args['public']             = false;
$args['publicly_queryable'] = false;
$args['has_archive']        = false;
```

This correctly blocks the traditional front-end route — `/venue/<slug>/` 301-redirects to the group root, verified working.

But `show_in_rest` was never touched (GatherPress registers it `true` by default, needed for block-editor authoring), and WordPress's default REST posts controller doesn't gate anonymous reads of `publish`-status posts behind `public`/`publicly_queryable` — only behind `show_in_rest`. Result:

```
$ curl -sk "https://events.wordpress.test/group/sunshine-coast-qld/wp-json/wp/v2/gatherpress_venues"
[{"id":34,...,"title":{"rendered":"Salty Spaces"},"content":{"rendered":"<p>A cozy beachside meeting space with ocean views.</p>"},"meta":{"gatherpress_address":"",...},...}]
```

Fully anonymous, no auth needed. Exposes venue title, description, address/geo fields, and any other registered venue meta — completely bypassing the lockdown that was specifically built to prevent this kind of exposure.

## Impact

Low-to-medium: venue address/description isn't typically highly sensitive, but it's clearly not intended to be publicly readable (that's the whole point of the lockdown), and it would matter more if an organizer ever puts something sensitive in venue notes.

## Fix direction

Don't disable `show_in_rest` outright (breaks block-editor authoring of venues). Instead gate anonymous REST reads specifically — e.g. hook `rest_request_before_callbacks` (or the collection/item permission callbacks) for the `gatherpress_venue` routes and require `edit_posts`/`read_private_posts` for unauthenticated requests, mirroring how the traditional front-end lockdown works.

## Acceptance criteria

- [ ] Anonymous `GET /wp-json/wp/v2/gatherpress_venues` and `/gatherpress_venues/{id}` return 401/403, not venue data — confirmed by re-running the exact repro curl command above and getting an auth error instead of the JSON payload.
- [ ] Authenticated users with `edit_posts`/`read_private_posts` can still read venue data via REST (block-editor venue authoring is not broken).
- [ ] Front-end `/venue/<slug>/` still 301-redirects to the group root, unchanged.
- [ ] No regression to the `gatherpress/venue` (formerly the custom `event-venue-name` block, removed in #1793) / `event-manage` blocks' own venue lookups (which run server-side, not through the public REST route, but worth confirming explicitly).

wppoland:
Fixes #1803.

# What

`GET /wp/v2/gatherpress_venues` returns venue titles, descriptions, addresses and meta to fully unauthenticated callers, bypassing the lockdown that made the post type non-public.

# Why

The existing `register_post_type_args` filter closes the front-end routes, and that part works: `/venue/<slug>/` still 301s to the group root. But it does not close the REST ones. `WP_REST_Posts_Controller` gates anonymous reads on `show_in_rest` alone:

- `get_items_permissions_check()` only checks a capability when `context=edit`, and returns `true` otherwise.
- `check_read_permission()` returns `true` for any post with status `publish`, without consulting `public` or `publicly_queryable`.

GatherPress registers the post type with `show_in_rest => true` because block-editor venue authoring needs it, so turning that off is not an option.

# How

Wrap the permission callback on the venue post type's readable REST endpoints and require `edit_posts`. `rest_authorization_required_code()` gives 401 to logged-out callers and 403 to logged-in ones without the capability.

Three details worth calling out:

- **The original callback is wrapped, not replaced.** Anyone who passes the capability gate still goes through the controller's own checks, so this only ever narrows access.
- **Only `GET` handlers are touched.** Writes already require capabilities through the post type's own caps, and re-gating them would be redundant.
- **The route prefix is derived from the post type object** (`rest_namespace`, `rest_base`) rather than hardcoded, and matched as `$base` or `$base . '/'` so a sibling route such as `/wp/v2/gatherpress_venues_extra` cannot be caught by accident.

`edit_posts` is the right threshold here rather than `read_private_posts`: it is what the Event Organiser (author) and Organiser (editor) tiers already have and what block-editor authoring requires, while Members (subscriber) get nothing, which is the intent of the lockdown.

# Testing

The repro from the issue, run anonymously:

```
$ curl -sk "https://events.wordpress.test/group/sunshine-coast-qld/wp-json/wp/v2/gatherpress_venues"
{"code":"rest_forbidden","message":"Sorry, you are not allowed to view venues.","data":{"status":401}}
```

Before this change the same request returns the full venue array. The single-item route behaves the same way.

Against the issue's acceptance criteria:

- [x] Anonymous collection and item reads return an auth error rather than venue data.
- [x] Users with `edit_posts` still read venues over REST, so block-editor venue authoring is unaffected.
- [x] `/venue/<slug>/` still 301-redirects to the group root. That path is handled by the `register_post_type_args` filter, which is untouched.
- [x] No regression to the venue rendering on events. `gatherpress/venue` output is server-rendered through the `render_block_gatherpress/venue` filter in this same file, and the event form's venue dropdown is served by `wporg-groups/v1/event-form-data`, so neither goes through the gated routes.

I also checked the route-selection and method-normalisation logic in isolation, confirming it gates the collection and item `GET` handlers, skips `POST`/`DELETE` handlers, and does not match sibling or unrelated routes. Happy to turn that into a real test once #1799 lands test infrastructure, since there is currently nowhere for it to live.

### DIFF
--- a/public_html/wp-content/mu-plugins/groups/gatherpress-groups-tweaks.php
+++ b/public_html/wp-content/mu-plugins/groups/gatherpress-groups-tweaks.php
@@ -536,3 +536,69 @@ add_filter(
 	10,
 	2
 );
+
+/**
+ * Require an editing capability to read venues over the REST API.
+ *
+ * Making the post type non-public closes the front-end routes, but it does
+ * not close the REST ones: WP_REST_Posts_Controller gates anonymous reads on
+ * `show_in_rest` alone, and grants them for any `publish` post. GatherPress
+ * needs `show_in_rest` for block-editor venue authoring, so it cannot simply
+ * be turned off, and the collection endpoint otherwise hands out venue
+ * addresses, descriptions and meta to unauthenticated callers.
+ *
+ * Wrap the read handlers' permission callbacks rather than replacing them,
+ * so the controller's own checks still run for anyone who passes this gate.
+ * Front-end venue output is server-rendered (see the
+ * `render_block_gatherpress/venue` filter above) and the event form's venue
+ * list comes from `wporg-groups/v1`, so neither depends on these routes.
+ */
+add_filter(
+	'rest_endpoints',
+	static function ( array $endpoints ): array {
+		$post_type = get_post_type_object( 'gatherpress_venue' );
+
+		if ( ! $post_type || empty( $post_type->show_in_rest ) ) {
+			return $endpoints;
+		}
+
+		$base   = '/' . ( $post_type->rest_namespace ?: 'wp/v2' ) . '/' . ( $post_type->rest_base ?: $post_type->name );
+		$routes = array_filter(
+			array_keys( $endpoints ),
+			static function ( string $route ) use ( $base ): bool {
+				return $route === $base || str_starts_with( $route, $base . '/' );
+			}
+		);
+
+		foreach ( $routes as $route ) {
+			foreach ( $endpoints[ $route ] as $index => $handler ) {
+				$methods = $handler['methods'] ?? array();
+
+				if ( is_string( $methods ) ) {
+					$methods = array_fill_keys( array_map( 'trim', explode( ',', $methods ) ), true );
+				}
+
+				// Writes already require capabilities; only reads are open.
+				if ( empty( $methods['GET'] ) ) {
+					continue;
+				}
+
+				$original = $handler['permission_callback'] ?? null;
+
+				$endpoints[ $route ][ $index ]['permission_callback'] = static function ( $request ) use ( $original ) {
+					if ( ! current_user_can( 'edit_posts' ) ) {
+						return new \WP_Error(
+							'rest_forbidden',
+							__( 'Sorry, you are not allowed to view venues.', 'wporg-groups-frontend' ),
+							array( 'status' => rest_authorization_required_code() )
+						);
+					}
+
+					return is_callable( $original ) ? $original( $request ) : true;
+				};
+			}
+		}
+
+		return $endpoints;
+	}
+);


wppoland:
Correcting one sentence in my own description above, before a reviewer has to.

I wrote that the event form's venue list comes from `wporg-groups/v1`, implying nothing in this plugin reads the gated `/wp/v2/gatherpress_venues` routes from the browser. The dropdown part is right, but it is not the whole picture: `src/blocks/event-manage/venue-editor.js` does client-side venue CRUD against `wp/v2/gatherpress_venues` via `apiFetch`, and it is imported by `group-settings`' Venues and Events tabs as well as the event modal.

I checked whether that breaks under this gate, and it does not:

- The `GET` gate requires `edit_posts`. `venue-editor` is only reachable behind `current_user_can_manage_group_settings()` (`edit_others_posts`, so editor/admin) or `current_user_can_manage_events()` (administrator/editor/author). Every one of those roles has `edit_posts`, so all of them still read venues.
- Its writes are `POST`/`PUT`, which this filter deliberately does not touch.
- Members (subscriber) lose access, which is the intent, and they never reach either surface.

So the behaviour is unchanged for everyone who can actually open the venue editor. The description was just narrower than reality, and anyone reviewing this should know the client-side CRUD path exists rather than take my word that nothing hits those routes.

raicem (approved):
Thank you @wppoland for your contribution! Looks and works great.

PR #1816, opened 2026-07-27
Groups: add GatherPress compat-test skill, PHPUnit suite, and Playwright e2e infra

bor0:
Adds three testing layers for the Groups/GatherPress front-end integration (#1799):

- **Agentic checklist skill** (`.claude/skills/groups-gatherpress-compat-test/`): environment setup, direct REST pass, per-role browser pass, GatherPress-off pass, security pass, perf check, and a known-issues appendix.
- **PHPUnit suite** ("WordCamp Groups" in `phpunit.xml.dist`): capability checks (including a regression pin for the fixed `promote_users` privilege escalation), `Members_Controller` role/IDOR rules, REST CRUD + draft flow + overnight rollover, block registration, and a GatherPress-off guard. Wires a pinned GatherPress install into the phpunit environment (local Docker + CI), since the plugin's real behavior can't be exercised without it — a missing/failed install now hard-fails the container and the test bootstraps (see below) rather than silently running the suites with the plugins under test inert.
- **Playwright e2e**: `playwright.config.js` + 3 specs (anonymous visitor, event organiser via wp-admin, member directory), plus a `workflow_dispatch`-only CI job (`.github/workflows/e2e-tests.yml`).
- **Speaker Feedback test isolation**: loading GatherPress in the shared PHPUnit bootstrap (needed for the suite above; `phpunit-bootstrap.php` loads every plugin's bootstrapper into one process regardless of which `--testsuite` runs) exposed GatherPress's `Rsvp\Query::exclude_rsvp_from_comment_query()` interfering with an unrelated Speaker Feedback test — it expands an empty `type` query var before Speaker Feedback's own exclusion filter reads it. GatherPress and Speaker Feedback aren't expected to run against the same query in production, so rather than reordering `pre_get_comments` priorities for real requests, the one affected test (`test-comment.php`) now neutralizes that specific GatherPress callback for its own scope (auto-restored by `WP_UnitTestCase` after the test).

**Review fixes** (d0029069, e924ef6f, 843601a8): the fail-open gaps and other issues @gedex flagged are addressed — see [the review](https://github.com/WordPress/wordcamp.org/pull/1816#pullrequestreview-4805111359) for details on each.

## How to test

**PHPUnit:**
```bash
docker compose -f docker-compose.phpunit.yml up --build -d
docker compose -f docker-compose.phpunit.yml exec phpunit_wp phpunit -c phpunit.xml.dist --testsuite "WordCamp Groups"
```
Expect `38 tests, 74 assertions, OK`. (The container needs outbound network access to `develop.svn.wordpress.org`/`wordpress.org` and `github.com` to install the WP test suite and GatherPress on first run — a blocking proxy will now make the container exit non-zero with a clear `ERROR:`/bootstrap `RuntimeException`, instead of a silent green run with GatherPress absent.)

Also worth a run of the Speaker Feedback suite, given the cross-plugin fix above:
```bash
docker compose -f docker-compose.phpunit.yml exec phpunit_wp phpunit -c phpunit.xml.dist --testsuite "WordCamp Speaker Feedback"
```

**Playwright e2e** (needs the regular dev stack up with GatherPress active on `sunshine-coast-qld`, and the `organiser1`/`eventorganiser1`/`member1` test users from the skill's setup step):
```bash
npm ci
npx playwright install chromium
npx playwright test
```
Expect 4/4 passing.

**Agentic checklist:**
```
/groups-gatherpress-compat-test
```

### DIFF
--- a/.docker/bin/install-test-suite.sh
+++ b/.docker/bin/install-test-suite.sh
@@ -13,6 +13,12 @@ DB_PASS="${WP_TESTS_DB_PASS:-}"
 WP_VERSION="${WP_TESTS_WP_VERSION:-latest}"
 TESTS_CONFIG="/tmp/wp/wordpress-tests-lib/wp-tests-config.php"
 
+# Same pinned release used by local dev (see the groups-gatherpress-compat-test
+# skill and PR #1793's test plan). GatherPress is gitignored/third-party, so the
+# `wporg-groups-frontend`/`groups` PHPUnit suites need it installed before they run.
+GATHERPRESS_VERSION="0.34.0-alpha.2"
+GATHERPRESS_DIR="/app/public_html/wp-content/plugins/gatherpress"
+
 db_ready() {
 	if command -v mariadb-admin >/dev/null 2>&1; then
 		mariadb-admin ping -h "$DB_HOST" --silent >/dev/null 2>&1
@@ -40,6 +46,19 @@ else
 		echo "WordPress test suite installed."
 	else
 		echo "WARNING: test-suite install failed (e.g. a download timeout). Re-run 'docker compose -f docker-compose.phpunit.yml up', or run /var/scripts/install-wp-tests.sh manually."
+	fi
+fi
+
+if [ -f "$GATHERPRESS_DIR/gatherpress.php" ]; then
+	echo "GatherPress already installed."
+else
+	echo "Installing GatherPress ${GATHERPRESS_VERSION}..."
+	if curl -sL "https://github.com/GatherPress/gatherpress/releases/download/${GATHERPRESS_VERSION}/gatherpress.${GATHERPRESS_VERSION}.zip" -o /tmp/gatherpress.zip \
+		&& unzip -q -o /tmp/gatherpress.zip -d /app/public_html/wp-content/plugins/ \
+		&& rm -f /tmp/gatherpress.zip; then
+		echo "GatherPress installed."
+	else
+		echo "WARNING: GatherPress install failed (e.g. a download timeout). The WordCamp Groups PHPUnit suite will not run correctly without it."
 	fi
 fi
 
--- a/.docker/bin/install-test-suite.sh
+++ b/.docker/bin/install-test-suite.sh
@@ -58,7 +58,8 @@ else
 		&& rm -f /tmp/gatherpress.zip; then
 		echo "GatherPress installed."
 	else
-		echo "WARNING: GatherPress install failed (e.g. a download timeout). The WordCamp Groups PHPUnit suite will not run correctly without it."
+		echo "ERROR: GatherPress install failed (e.g. a download timeout). The WordCamp Groups PHPUnit suites cannot run without it." >&2
+		exit 1
 	fi
 fi
 
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -1,0 +1,91 @@
+name: E2E Tests (Groups/GatherPress)
+
+# Manually triggered only, for now — this stands up real docker-based
+# provisioning (WordPress core checkout, TLS cert, DB import, third-party
+# plugin install) and is new/unproven infrastructure. Promote to a
+# `pull_request` trigger once it's shown to be stable.
+on:
+  workflow_dispatch: {}
+
+jobs:
+  e2e:
+    name: Playwright E2E
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Use Node.js 20.x
+      uses: actions/setup-node@v4
+      with:
+        node-version: 20.x
+        cache: npm
+
+    - name: Install dependencies
+      run: npm ci --no-audit --no-fund
+
+    - name: Install Playwright browsers
+      run: npx playwright install --with-deps chromium
+
+    - name: Build the groups frontend plugin
+      run: npm run build --workspace=public_html/wp-content/mu-plugins/wporg-groups-frontend
+
+    - name: Clone WordPress core (develop trunk) into public_html/mu
+      run: git clone --depth=1 https://github.com/WordPress/WordPress.git public_html/mu
+
+    - name: Generate a self-signed TLS cert for local hostnames
+      # Playwright's `ignoreHTTPSErrors: true` (see playwright.config.js)
+      # means this only needs to satisfy nginx, not a trust store — no
+      # mkcert/browser-trust step needed here, unlike the local dev setup
+      # documented in .docker/readme.md.
+      run: |
+        openssl req -x509 -newkey rsa:2048 -nodes -days 365 \
+          -keyout .docker/wordcamp.test.key.pem \
+          -out .docker/wordcamp.test.pem \
+          -subj "/CN=wordcamp.test" \
+          -addext "subjectAltName=DNS:wordcamp.test,DNS:*.wordcamp.test,DNS:events.wordpress.test,DNS:*.wordpress.test"
+
+    - name: Boot the docker environment
+      run: docker compose up --build -d
+
+    - name: Wait for the site to respond
+      run: |
+        for i in $(seq 1 30); do
+          if docker compose exec -T wordcamp.test wp core is-installed 2>/dev/null; then
+            exit 0
+          fi
+          sleep 2
+        done
+        echo "WordPress did not become ready in time" >&2
+        exit 1
+
+    - name: Add local hostnames to /etc/hosts
+      run: |
+        echo "127.0.0.1 wordcamp.test central.wordcamp.test seattle.wordcamp.test shinynew.wordcamp.test events.wordpress.test" | sudo tee -a /etc/hosts
+
+    - name: Install GatherPress on the groups site
+      run: |
+        docker compose exec -T wordcamp.test wp plugin install \
+          https://github.com/GatherPress/gatherpress/releases/download/0.34.0-alpha.2/gatherpress.0.34.0-alpha.2.zip \
+          --activate --url=events.wordpress.test/group/sunshine-coast-qld/
+
+    - name: Create role-tier test users
+      run: |
+        docker compose exec -T wordcamp.test wp user create organiser1 organiser1@example.test \
+          --role=editor --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
+        docker compose exec -T wordcamp.test wp user create eventorganiser1 eventorganiser1@example.test \
+          --role=author --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
+        docker compose exec -T wordcamp.test wp user create member1 member1@example.test \
+          --role=subscriber --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
+
+    - name: Run Playwright tests
+      run: npx playwright test
+
+    - name: Upload Playwright report
+      if: ${{ !cancelled() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 14
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -129,6 +129,23 @@ jobs:
         create_user eventorganiser1 author
         create_user member1 subscriber
 
+    - name: Create required pages
+      # `page-members.html` (public_html/wp-content/themes/groups-site/templates/)
+      # is a block-theme template keyed on the page slug — it ignores
+      # post_content entirely, but only applies once a published Page with
+      # that slug actually exists. Not part of the wordcamp_dev.sql fixture.
+      run: |
+        SITE_URL=events.wordpress.test/group/sunshine-coast-qld/
+        create_page() {
+          if docker compose exec -T wordcamp.test wp post list --post_type=page --name="$1" --field=ID --url="$SITE_URL" | grep -q .; then
+            echo "Page '$1' already exists."
+            return 0
+          fi
+          docker compose exec -T wordcamp.test wp post create --post_type=page \
+            --post_title="$2" --post_name="$1" --post_status=publish --url="$SITE_URL"
+        }
+        create_page members Members
+
     - name: Run Playwright tests
       run: npx playwright test
 
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -72,12 +72,18 @@ jobs:
 
     - name: Create role-tier test users
       run: |
-        docker compose exec -T wordcamp.test wp user create organiser1 organiser1@example.test \
-          --role=editor --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
-        docker compose exec -T wordcamp.test wp user create eventorganiser1 eventorganiser1@example.test \
-          --role=author --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
-        docker compose exec -T wordcamp.test wp user create member1 member1@example.test \
-          --role=subscriber --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
+        SITE_URL=events.wordpress.test/group/sunshine-coast-qld/
+        create_user() {
+          if docker compose exec -T wordcamp.test wp user get "$1" --field=ID --url="$SITE_URL" >/dev/null 2>&1; then
+            echo "User $1 already exists."
+            return 0
+          fi
+          docker compose exec -T wordcamp.test wp user create "$1" "$1@example.test" \
+            --role="$2" --user_pass=password --url="$SITE_URL"
+        }
+        create_user organiser1 editor
+        create_user eventorganiser1 author
+        create_user member1 subscriber
 
     - name: Run Playwright tests
       run: npx playwright test
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -27,6 +27,24 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
+    - name: Set PHP version
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '8.4'
+        coverage: none
+        tools: composer:v2
+      env:
+          COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Install PHP dependencies
+      # wporg-mu-plugins (installed via composer, then moved into place by a
+      # post-install script — see composer.json's installer-paths) is loaded
+      # unconditionally by load-other-mu-plugins.php; without it every
+      # request, including wp-cli, fatals.
+      run: |
+        rm composer.lock || true
+        composer install
+
     - name: Use Node.js 20.x
       uses: actions/setup-node@v4
       with:
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -62,17 +62,13 @@ jobs:
 
     - name: Wait for the site to respond
       run: |
-        # A from-scratch CI runner has to build both images, cold-boot
-        # MariaDB, and import the wordcamp_dev.sql fixture before
-        # `wp core is-installed` can succeed — there's no persisted
-        # `.docker/mariadb-database` volume like local dev has after the
-        # first run, and no depends_on/healthcheck between the two
-        # containers, so give this more headroom than a warm local re-run
-        # would need.
-        for i in $(seq 1 60); do
-          if docker compose exec -T wordcamp.test wp core is-installed 2>/dev/null; then
+        # Don't swallow stderr here — a real (non-timing) failure needs to
+        # be visible in the log, not silently retried for the full timeout.
+        for i in $(seq 1 20); do
+          if docker compose exec -T wordcamp.test wp core is-installed; then
             exit 0
           fi
+          echo "Attempt $i/20 failed, retrying in 3s..."
           sleep 3
         done
         echo "WordPress did not become ready in time" >&2
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -1,11 +1,22 @@
 name: E2E Tests (Groups/GatherPress)
 
-# Manually triggered only, for now — this stands up real docker-based
-# provisioning (WordPress core checkout, TLS cert, DB import, third-party
-# plugin install) and is new/unproven infrastructure. Promote to a
-# `pull_request` trigger once it's shown to be stable.
+# Runs on PRs that touch the Groups/GatherPress front end, plus manual
+# dispatch for ad-hoc runs. Scoped to these paths (rather than every PR)
+# because this stands up real docker-based provisioning (WordPress core
+# checkout, TLS cert, DB import, third-party plugin install).
 on:
+  pull_request:
+    paths:
+    - public_html/wp-content/mu-plugins/wporg-groups-frontend/**
+    - public_html/wp-content/mu-plugins/groups/**
+    - tests/e2e/**
+    - playwright.config.js
+    - .github/workflows/e2e-tests.yml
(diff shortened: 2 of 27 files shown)


bor0:
cc @gedex @raicem (still can't assign you as reviewers because of access but I [pinged in #meta again](https://wordpress.slack.com/archives/C02QB8GMM/p1785144428505519?thread_ts=1784727063.042709&cid=C02QB8GMM))

gedex:
> cc @gedex @raicem (still can't assign you as reviewers because of access but I [pinged in #meta again](https://wordpress.slack.com/archives/C02QB8GMM/p1785144428505519?thread_ts=1784727063.042709&cid=C02QB8GMM))

I'll review this today

gedex:
## My Tests

### PHPUnit ✅ 

```
docker compose -f docker-compose.phpunit.yml up --build -d
docker compose -f docker-compose.phpunit.yml exec phpunit_wp phpunit -c phpunit.xml.dist --testsuite "WordCamp Groups"

Installing...
Installing network...
Running as multisite...
Not running ajax tests. To execute these, use --group ajax.
Not running ms-files tests. To execute these, use --group ms-files.
Not running external-http tests. To execute these, use --group external-http.
PHPUnit 9.6.35 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.23
Configuration: phpunit.xml.dist

......................................                            38 / 38 (100%)
```

### Playwright e2e ✅ 

```
npm ci
npx playwright install chromium
npx playwright test
```

```
Running 4 tests using 4 workers

  ✓  1 [chromium] › tests/e2e/member-directory.spec.js:9:2 › member directory › lists members with correct role labels (7.0s)
  ✓  2 [chromium] › tests/e2e/anonymous.spec.js:10:2 › anonymous visitor › front page renders (7.0s)
  ✓  3 [chromium] › tests/e2e/anonymous.spec.js:15:2 › anonymous visitor › no management UI is rendered on the front page (7.1s)
  ✓  4 [chromium] › tests/e2e/event-organiser.spec.js:15:2 › event organiser (author role) › can create and publish an event via wp-admin, and it appears on the front page (13.9s)

  4 passed (16.2s)
  ```

### Agentic Checklsit ✅ 

```
/groups-gatherpress-compat-test

❯ /groups-gatherpress-compat-test

I'll work through this checklist. Starting with the automated layers first, as the skill directs.

  Listed 1 directory, ran 2 shell commands

Stack is up. Let me run the automated layers and check GatherPress state.

  Listed 1 directory, ran 2 shell commands

GatherPress 0.34.0-alpha.2 active (note: 0.34.1 available upstream). Now running the automated suites.

  Searched for 1 pattern, read 1 file, listed 1 directory, ran 1 shell command

PHPUnit: 38 tests, 74 assertions, all passing. Now the E2E layer.

  Listed 1 directory, ran 2 shell commands

Both automated layers green. Now setting up test users for the manual passes.

  Ran 3 shell commands

Credentials ready. Starting Section 2: direct REST pass — first confirming the route list.

  Ran 1 shell command

All expected routes present — plus members/notification-preference, which isn't in the checklist's expected list. Let me note that and proceed with the checks.

  Searched for 1 pattern, ran 1 shell command

✅ event-form-data → 200. Now event creation as editor, plus the overnight-rollover case.

  Ran 2 shell commands

✅ Rollover correct (ends next calendar day). Now the IDOR checks.

  Ran 1 shell command

✅ IDOR blocked (403, title unchanged); own-event edit works. Now the members endpoint and pagination caps.

  Ran 1 shell command

✅ Public, paginated, correct labels, no email/registered leaked. Now the role-management permission matrix.
```

File: .docker/bin/install-test-suite.sh [hunk 1]
```diff
@@ -43,4 +49,17 @@ else
...
+if [ -f "$GATHERPRESS_DIR/gatherpress.php" ]; then
+	echo "GatherPress already installed."
+else
+	echo "Installing GatherPress ${GATHERPRESS_VERSION}..."
+	if curl -sL "https://github.com/GatherPress/gatherpress/releases/download/${GATHERPRESS_VERSION}/gatherpress.${GATHERPRESS_VERSION}.zip" -o /tmp/gatherpress.zip \
+		&& unzip -q -o /tmp/gatherpress.zip -d /app/public_html/wp-content/plugins/ \
+		&& rm -f /tmp/gatherpress.zip; then
+		echo "GatherPress installed."
+	else
+		echo "WARNING: GatherPress install failed (e.g. a download timeout). The WordCamp Groups PHPUnit suite will not run correctly without it."
```
gedex: **Critical — fail-open: a download flake produces a green PHPUnit run.**

This script has no `set -e`, so the `WARNING` here falls straight through to `exec "$@"` and PHPUnit runs with GatherPress absent. And because both new test bootstraps also treat GatherPress as optional (see my comments on those files), the suites don't hard-fail either — the plugins under test just go inert and the run comes back green.

A missing test dependency should stop the container, not warn.

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/bootstrap.php
```diff
@@ -0,0 +1,32 @@
...
+ * Without it, this plugin's own bootstrap no-ops (see its `class_exists()`
+ * guard), so tests that need real behavior will fail with a clear reason
+ * rather than a fatal.
+ */
+function manually_load_plugins() {
+	$gatherpress_file = WP_PLUGIN_DIR . '/gatherpress/gatherpress.php';
+
+	if ( file_exists( $gatherpress_file ) ) {
+		require_once $gatherpress_file;
+	}
```
gedex: **Silently skipping GatherPress makes the plugin under test inert.**

`wporg-groups-frontend.php` gates its own `bootstrap()` on `class_exists( '\GatherPress\Core\Event\Event' )`. So when this `file_exists()` check fails, the plugin file loads but registers nothing — and every test that doesn't directly touch GatherPress state (`test-gatherpress-off-guard.php`, the capability tests, the members-controller cap checks) still passes.

The comment just below says tests will "fail with a clear reason", but there's no bootstrap-time assertion that GatherPress actually loaded, so that only holds for the subset of tests that happen to reach a GatherPress class.

gedex (commented):
9 inline comments, each with a suggested fix in-thread. Two things that have no diff line to attach to:

**1. Four fail-open GatherPress checks add up to a false green.** `install-test-suite.sh:61`, both `tests/bootstrap.php`, and `Groups_TestCase::set_up()` each skip-if-absent. Since `wporg-groups-frontend.php` self-gates on `class_exists( '\GatherPress\Core\Event\Event' )`, a download flake means a green run with the plugins under test doing nothing. Worth fixing as one set — details inline.

**2. `349f8d2e` has no real regression test.** `test-comment.php:184` only catches a revert to priority 99 because this PR happens to load GatherPress in `phpunit-bootstrap.php`. Isolate that suite and the coverage vanishes with nothing failing. Pin the `pre_get_comments` priorities, or at least name the invariant in the docstring.

**The `query.php` 99 → 1 fix itself is correct** — at priority 1, GatherPress's empty-`type` expansion can no longer look like an explicit feedback request, and `type__not_in` survives into the final SQL. Also checked and fine: bootstrap ordering, `unit-tests.yml`, `workflow_dispatch`-only e2e, and the behavioral coverage in `test-capabilities` / `test-rest` / `test-class-members-controller` / `test-blocks`.

public_html/wp-content/mu-plugins/groups/tests/bootstrap.php
```diff
@@ -0,0 +1,26 @@
...
+ * GatherPress is gitignored/third-party (not a composer dependency), so it's
+ * only present if the phpunit environment installed it first — see
+ * `.docker/bin/install-test-suite.sh` / `.github/workflows/unit-tests.yml`.
+ */
+function manually_load_plugin() {
+	$gatherpress_file = WP_PLUGIN_DIR . '/gatherpress/gatherpress.php';
+
+	if ( file_exists( $gatherpress_file ) ) {
+		require_once $gatherpress_file;
+	}
```
gedex: **Same fail-open as `wporg-groups-frontend/tests/bootstrap.php`.**

`gatherpress-groups-tweaks.php` is entirely about GatherPress post types and meta, so with GatherPress missing this suite verifies almost nothing while still reporting green. Combined with the `WARNING`-and-continue in `install-test-suite.sh`, there's no layer that turns "dependency absent" into a failure.

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/class-groups-testcase.php
```diff
@@ -0,0 +1,87 @@
...
+		// GatherPress creates its `{prefix}gatherpress_events` table on
+		// plugin activation; since these tests require the plugin file
+		// directly instead of truly activating it, the table never gets
+		// created for this fixture blog. `check_plugin_version()` is
+		// GatherPress's own public, idempotent self-heal for exactly this
+		// scenario ("site created before plugin activation") — safe to
+		// call every test.
+		if ( class_exists( '\GatherPress\Core\Setup' ) ) {
+			\GatherPress\Core\Setup::get_instance()->check_plugin_version();
+		}
```
gedex: **the per-site table self-heal is silently skipped too.**

If `\GatherPress\Core\Setup` doesn't exist then GatherPress isn't loaded at all, and the right outcome is a hard error from the bootstrap — not quietly skipping table creation and letting tests fail later with confusing missing-table errors that don't name the actual cause.

Once the bootstrap `require_once`s GatherPress unconditionally, this guard can just go.

tests/e2e/utils/login.js
```diff
@@ -0,0 +1,19 @@
...
+ * @param {import('@playwright/test').Page} page
+ * @param {string} username
+ * @param {string} password
+ */
+async function login( page, username, password ) {
+	await page.goto( 'wp-login.php' );
+	await page.locator( '#user_login' ).fill( username );
+	await page.locator( '#user_pass' ).fill( password );
+	await page.locator( '#wp-submit' ).click();
+	await page.waitForLoadState( 'networkidle' );
```
gedex, line 16: **`login()` resolves successfully even when authentication failed.**

A wrong password, a missing user, or wp-login re-rendering with an error all leave the page on `wp-login.php` and still satisfy `networkidle`. The helper returns "ok", and the caller then fails on some group-site locator — pointing whoever reads the failure at entirely the wrong thing.

Given `.github/workflows/e2e-tests.yml` currently creates its role users with `|| true`, "user was never created" is the single most likely CI failure mode, and this helper is what would disguise it.

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,91 @@
...
+          --activate --url=events.wordpress.test/group/sunshine-coast-qld/
+
+    - name: Create role-tier test users
+      run: |
+        docker compose exec -T wordcamp.test wp user create organiser1 organiser1@example.test \
+          --role=editor --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
+        docker compose exec -T wordcamp.test wp user create eventorganiser1 eventorganiser1@example.test \
+          --role=author --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
+        docker compose exec -T wordcamp.test wp user create member1 member1@example.test \
+          --role=subscriber --user_pass=password --url=events.wordpress.test/group/sunshine-coast-qld/ || true
```
gedex: **`|| true` swallows every failure, not just "user already exists".**

If user creation fails for a real reason (site not ready, GatherPress activation left the site half-configured, a typo in `--url`), the step still succeeds and the suite runs partially green: `anonymous.spec.js` and `member-directory.spec.js` never log in so they pass, while `event-organiser.spec.js` fails on a missing-user login rather than on the regression it exists to catch.

Worth making idempotent-but-strict instead of blanket-ignoring.

tests/e2e/anonymous.spec.js
```diff
@@ -0,0 +1,22 @@
...
+ * block permission checks in the groups-gatherpress-compat-test skill).
+ */
+test.describe( 'anonymous visitor', () => {
+	test( 'front page renders', async ( { page } ) => {
+		const response = await page.goto( '' );
+		expect( response.status() ).toBe( 200 );
+	} );
+
+	test( 'no management UI is rendered on the front page', async ( { page } ) => {
+		await page.goto( '/' );
```
gedex: **This test currently passes for the wrong reason.**

`playwright.config.js`'s `baseURL` is `https://events.wordpress.test/group/sunshine-coast-qld/`, and its own comment warns that specs must use relative paths *without* a leading slash. `'/'` resolves to `https://events.wordpress.test/` — the network root, not the group site — so "no Manage button" is being asserted against a page that would never render one either way.

The first test in this file (line 11) already gets this right.

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/test-gatherpress-off-guard.php
```diff
@@ -0,0 +1,57 @@
...
+	 * The main plugin file's bootstrap() must stay guarded on GatherPress
+	 * being loaded.
+	 */
+	public function test_frontend_plugin_bootstrap_is_guarded() {
+		// phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents -- reading a local plugin file, not a remote URL.
+		$source = file_get_contents(
+			dirname( __DIR__ ) . '/wporg-groups-frontend.php'
+		);
+
+		$this->assertStringContainsString(
```
gedex, line 34: **Minor — asserts source text, not behavior.**

Both tests here `file_get_contents()` the plugin and string-match the guard. Reformatting it (dropping the leading `\`, wrapping the condition across lines, switching to an early-return shape) fails the test with no behavioral change; conversely a change that actually breaks the guard while keeping this literal intact passes. The docstring acknowledges the tradeoff, so this is a note rather than a blocker.

public_html/wp-content/mu-plugins/groups/tests/test-gatherpress-groups-tweaks.php
```diff
@@ -0,0 +1,100 @@
...
+	 * the whole test run; `WP_UnitTestCase::tearDown()` unregisters all meta
+	 * keys after every test (a known core testing quirk), so by the time
+	 * this test runs the registration from bootstrap is long gone. Re-fire
+	 * just this one `init` callback's effect directly (rather than
+	 * `do_action( 'init' )`, which would also re-run block registration and
+	 * trip "already registered" `_doing_it_wrong` notices) — this must stay
+	 * in sync with the real registration in `gatherpress-groups-tweaks.php`.
+	 */
+	public function test_event_speakers_meta_registered_with_array_default() {
+		register_post_meta(
```
gedex: **Minor — this test re-registers the meta itself, then asserts its own registration.**

`register_post_meta()` is called right here with the expected args, and the assertion below then checks that registration — so what's verified is core's `register_post_meta`, not that `gatherpress-groups-tweaks.php` registers `_event_speakers` correctly. The docstring at line 61 notes these args "must stay in sync with the real registration", which is exactly the drift this shape can't catch.

.docker/bin/install-test-suite.sh
gedex:
```suggestion
		echo "ERROR: GatherPress install failed (e.g. a download timeout). The WordCamp Groups PHPUnit suites cannot run without it." >&2
		exit 1
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/bootstrap.php
gedex:
```suggestion
	if ( ! file_exists( $gatherpress_file ) ) {
		throw new \RuntimeException(
			"GatherPress is required by this suite but was not found at {$gatherpress_file}. Run `.docker/bin/install-test-suite.sh` (see also `.github/workflows/unit-tests.yml`) first."
		);
	}

	require_once $gatherpress_file;
```

public_html/wp-content/mu-plugins/groups/tests/bootstrap.php
gedex:
```suggestion
	if ( ! file_exists( $gatherpress_file ) ) {
		throw new \RuntimeException(
			"GatherPress is required by this suite but was not found at {$gatherpress_file}. Run `.docker/bin/install-test-suite.sh` (see also `.github/workflows/unit-tests.yml`) first."
		);
	}

	require_once $gatherpress_file;
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/class-groups-testcase.php
gedex:
```suggestion
		\GatherPress\Core\Setup::get_instance()->check_plugin_version();
```

tests/e2e/utils/login.js
gedex:
```suggestion
	await page.waitForLoadState( 'networkidle' );

	// Fail here, with the real reason, rather than downstream on a missing
	// group-site element, if authentication did not actually succeed.
	if ( new URL( page.url() ).pathname.endsWith( 'wp-login.php' ) ) {
		throw new Error( `login( '${ username }' ) failed — still on wp-login.php.` );
	}
```

Asserting `#wpadminbar` is attached would be stronger, but needs `expect` imported into this file — hence the dependency-free version above.

.github/workflows/e2e-tests.yml
gedex:
```suggestion
        SITE_URL=events.wordpress.test/group/sunshine-coast-qld/
        create_user() {
          if docker compose exec -T wordcamp.test wp user get "$1" --field=ID --url="$SITE_URL" >/dev/null 2>&1; then
            echo "User $1 already exists."
            return 0
          fi
          docker compose exec -T wordcamp.test wp user create "$1" "$1@example.test" \
            --role="$2" --user_pass=password --url="$SITE_URL"
        }
        create_user organiser1 editor
        create_user eventorganiser1 author
        create_user member1 subscriber
```

Still idempotent across re-runs, but a genuine `wp user create` failure now fails the step.

tests/e2e/anonymous.spec.js
gedex:
```suggestion
		await page.goto( '' );
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/test-gatherpress-off-guard.php
gedex: If you want behavioral coverage later, the observable property is "with GatherPress absent, the plugin registers nothing" — load the plugin in a suite/subprocess without GatherPress and assert no `wporg/*` blocks and no `_event_speakers` meta were registered. Not worth doing in this PR; the string check is a reasonable placeholder as long as its limits stay documented.

public_html/wp-content/mu-plugins/groups/tests/test-gatherpress-groups-tweaks.php
gedex: Cheapest fix that keeps the deliberate `do_action( 'init' )` avoidance: extract the real registration in `gatherpress-groups-tweaks.php` into a named function (e.g. `register_event_speakers_meta()`), hook *that* on `init`, and have this test call it directly. The args then live in exactly one place and drift becomes impossible.

.docker/bin/install-test-suite.sh
bor0: Fixed as suggested — install failure now exits 1 with an ERROR message (d0029069).

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/bootstrap.php
bor0: Fixed as suggested — bootstrap now throws a RuntimeException instead of silently skipping GatherPress (d0029069).

public_html/wp-content/mu-plugins/groups/tests/bootstrap.php
bor0: Fixed as suggested — same RuntimeException guard applied here (d0029069).

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/class-groups-testcase.php
bor0: Fixed as suggested — dropped the now-redundant class_exists() guard now that the bootstrap requires GatherPress unconditionally (d0029069).

tests/e2e/utils/login.js
bor0: Fixed as suggested — login() now throws if still on wp-login.php after submit (d0029069).

.github/workflows/e2e-tests.yml
bor0: Fixed as suggested — role-tier user creation is now idempotent-but-strict via a wp user get check instead of `|| true` (d0029069).

tests/e2e/anonymous.spec.js
bor0: Fixed as suggested — goto('') instead of goto('/') (d0029069).

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/test-gatherpress-off-guard.php
bor0: Agreed this is a placeholder as-is — leaving it, per your note that it's not worth doing in this PR.

public_html/wp-content/mu-plugins/groups/tests/test-gatherpress-groups-tweaks.php
bor0: Fixed per your suggestion — extracted the registration into a named register_event_speakers_meta() function hooked on init, and the test now calls it directly instead of duplicating the args (d0029069).

bor0:
@gedex thanks for the thorough review — all 9 inline comments are fixed as suggested (replied on each thread), plus the two non-inline points:

1. **Fail-open set** (`install-test-suite.sh`, both `tests/bootstrap.php`, `class-groups-testcase.php`): a missing/failed GatherPress install now hard-fails — the Docker entrypoint `exit 1`s and the PHPUnit bootstraps throw a `RuntimeException` — instead of silently letting the plugins under test go inert on a green run. Verified by removing the installed GatherPress plugin and confirming the suite now fails loudly instead of passing.
2. **`349f8d2e` regression coverage**: added a `has_filter( 'pre_get_comments', ... )` priority assertion directly in `test-comment.php`, so the priority-99-vs-1 regression is caught even if the Speaker Feedback suite runs in isolation without GatherPress loaded.

All in commit d0029069.

**Verified after the fixes:**
- `WordCamp Groups` suite: `38 tests, 74 assertions, OK`
- Full PHPUnit suite: `384 tests, 719 assertions, OK` (5 pre-existing intl-extension skips)
- `WordCamp Speaker Feedback` suite: `58 tests, 144 assertions, OK`
- Playwright e2e: `4 passed`

PR description updated to reflect the Speaker Feedback fix and the stricter fail-open behavior.

bor0:
Follow-up on the Speaker Feedback fix mentioned above: on reflection, changing `pre_get_comments`'s priority for real requests (99 → 1) to work around an interaction that only occurs because the shared PHPUnit bootstrap loads GatherPress and Speaker Feedback into the same process wasn't the right tradeoff — GatherPress and Speaker Feedback aren't expected to run against the same query in production.

Reverted `includes/query.php` back to its original priority 99 (843601a8). Instead, `test_get_feedback_exclude_other_comments` in `test-comment.php` now explicitly removes GatherPress's `Rsvp\Query::exclude_rsvp_from_comment_query()` callback for its own scope before asserting — `WP_UnitTestCase`'s hook backup/restore undoes this automatically afterward, so it's isolated to this one test.

Verified: `WordCamp Speaker Feedback` suite 58/58, full PHPUnit suite 384/384 (5 pre-existing intl-extension skips). Also confirmed the test-isolation fix is load-bearing by temporarily disabling it and reproducing the original failure (6 comments returned instead of 3).

PR description updated accordingly.

gedex (approved):

PR #1817, opened 2026-07-27
Groups: let Organisers edit group name and description

### LINKED ISSUE #1805: Groups: Group Settings "About" tab fails for Organisers — /wp/v2/settings requires manage_options
Found while testing PR #1793 (`split/1699/wporg-groups-frontend`, split from #1699) live in the browser instead of via curl.

## Reproduction

Tested as an "Organiser" (WordPress `editor` role — the plugin's own top capability tier) on a group site with #1793 merged on top of `migration-branch`.

1. Log in as a user with the `editor` role (an "Organiser" per this plugin's tier model — see `EVENT_MANAGER_ROLES` / `current_user_can_manage_group_settings()` in `inc/capabilities.php`).
2. Open group Settings → About tab.

Group name/description fields render blank instead of the site's actual `blogname`/`blogdescription`. Typing a new name and clicking Save shows "Sorry, you are not allowed to do that."

Network tab:
```
GET .../wp-json/wp/v2/settings → 403 rest_forbidden (on load, silently swallowed)
POST .../wp-json/wp/v2/settings → 403 rest_forbidden (on save, shown to user)
```

## Root cause

`src/blocks/group-settings/components/about-tab.js` calls the core `/wp/v2/settings` endpoint directly. That endpoint's permission callback requires `manage_options`, which only `administrator` has — not `editor`. But the block itself is gated by `current_user_can_manage_group_settings()` = `edit_others_posts`, which editors *do* have. So the UI shows the tab to Organisers, but every request it makes 403s for them.

The load-time failure is also silent: `about-tab.js`'s `.catch(() => setLoading(false))` swallows the error and just leaves the fields blank, with no indication anything went wrong.

## Impact

An Organiser (editor role) — the top tier this plugin expects to manage a group day-to-day — can never change their group's name/description through the UI. Only a network administrator (which Organisers usually aren't) can.

## Fix direction

Add a dedicated mu-plugin REST endpoint (or a permission override) for group name/description gated by the same `edit_others_posts` / `current_user_can_manage_group_settings()` check used for the rest of the settings UI, rather than routing through the admin-only core settings endpoint. Also surface load failures to the user instead of swallowing them silently.

## How to verify a fix

Log in as an editor-tier Organiser, open Settings → About, confirm the current group name/description are pre-filled, change them, Save, and confirm the change persists (`wp option get blogname`).

wppoland:
Fixes #1805.

# What

Settings > About renders blank for an Organiser (editor role) and saving fails with "Sorry, you are not allowed to do that." The top tier this plugin expects to run a group day to day can never change the group's name or description.

# Why

`about-tab.js` called core's `/wp/v2/settings`, whose permission callback requires `manage_options`. Only administrators have that, and Organisers are editors. Meanwhile the tab itself is gated by `current_user_can_manage_group_settings()` (`edit_others_posts`), which editors do have. So the UI offered a tab whose every request 403'd for exactly the audience it was built for.

# How

Two new routes in `wporg-groups/v1`, `GET` and `POST /group-info`, reading and writing `blogname` and `blogdescription`, both behind `manage_group_settings_permissions_check()`. That is the same check that gates the settings block, so the UI and the endpoint now agree instead of disagreeing.

Details:

- **Two fields, not a settings passthrough.** The route exposes only the two options the tab edits. Widening it later is easy; handing out a general option writer to editors is not something to do casually.
- **`POST` writes only the fields present in the request**, so a client can send one without clobbering the other. Both are `sanitize_text_field`, and `update_option` applies core's own `sanitize_option` on top.
- **`POST` returns the stored values** by delegating to the same reader, so the client sees what was actually persisted rather than an echo of its own payload.

I also stopped the load failure from being silent. `.catch( () => setLoading( false ) )` left the fields blank, which is indistinguishable from a group that genuinely has no name, and is why this presented as "renders blank" rather than as an error. Failures now surface in the existing `Notice`.

# Testing

As an editor-tier Organiser:

1. Open Settings > About. Current group name and description are pre-filled.
2. Change both, Save. Notice confirms, and `wp option get blogname` / `wp option get blogdescription` reflect the new values.
3. Reload the tab. The new values are still there.

Capability boundaries:

- `GET`/`POST /wporg-groups/v1/group-info` as a subscriber-tier member returns 403.
- Anonymous returns 401.
- Administrators are unaffected.

Before this change, step 1 shows empty fields with a silent 403 in the network tab, and step 2 fails with a visible permission error.

# Note on scope

The About tab's docblock mentions location and social links, but the component only renders name and description today, so this covers what actually exists. If those fields arrive later they will want their own storage anyway, not `/wp/v2/settings`.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
@@ -242,10 +242,20 @@ function get_group_info(): WP_REST_Response {
  *
  * Only the fields present in the request are written, so a client can send
  * one without clobbering the other.
+ *
+ * @return WP_Error|WP_REST_Response
  */
-function update_group_info( WP_REST_Request $request ): WP_REST_Response {
+function update_group_info( WP_REST_Request $request ) {
 	$title       = $request->get_param( 'title' );
 	$description = $request->get_param( 'description' );
+
+	// An empty group name is not recoverable from this UI: the next load
+	// returns the blank value, so there is nothing left to restore it from.
+	// Checked after sanitization, because `sanitize_text_field()` can empty
+	// an input that was not empty when it was sent.
+	if ( null !== $title && '' === trim( $title ) ) {
+		return new WP_Error( 'wporg_groups_empty_group_name', 'Group name is required.', array( 'status' => 400 ) );
+	}
 
 	if ( null !== $title ) {
 		update_option( 'blogname', $title );
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
@@ -15,8 +15,15 @@
  *   POST /event/{id}
  *        Updates an existing gatherpress_event.
  *
- * All routes require the `current_user_can_manage_events()` capability,
- * plus post-specific capabilities when operating on an existing event.
+ *   GET  /group-info
+ *   POST /group-info
+ *        Reads and writes the group's name and description. Exists because
+ *        core's /wp/v2/settings requires `manage_options`, which Organisers
+ *        (editors) do not have.
+ *
+ * The event routes require the `current_user_can_manage_events()` capability,
+ * plus post-specific capabilities when operating on an existing event. The
+ * group-info routes require `current_user_can_manage_group_settings()`.
  *
  * @package WordCamp\Groups\Frontend
  */
@@ -35,6 +42,7 @@ use WP_REST_Server;
 use const WordCamp\Groups\Frontend\Defaults\DESCRIPTION_BLOCK_NAMES;
 
 use function WordCamp\Groups\Frontend\Capabilities\current_user_can_manage_events;
+use function WordCamp\Groups\Frontend\Capabilities\current_user_can_manage_group_settings;
 use function WordCamp\Groups\Frontend\Defaults\extract_description_blocks;
 use function WordCamp\Groups\Frontend\Defaults\get_default_event_data;
 use function WordCamp\Groups\Frontend\Defaults\get_event_venue_post_id;
@@ -170,6 +178,83 @@ function register_routes(): void {
 			) + event_args_schema(),
 		)
 	);
+
+	// ----- Group info -----------------------------------------------------
+	//
+	// The Settings > About tab edits `blogname` and `blogdescription`. Core's
+	// /wp/v2/settings gates both behind `manage_options`, which only network
+	// administrators have, so Organisers got a 403 on every read and write.
+	// These routes expose just those two fields, behind the same capability
+	// check that gates the settings UI itself.
+
+	register_rest_route(
+		NAMESPACE_V1,
+		'/group-info',
+		array(
+			array(
+				'methods'             => WP_REST_Server::READABLE,
+				'callback'            => __NAMESPACE__ . '\get_group_info',
+				'permission_callback' => __NAMESPACE__ . '\manage_group_settings_permissions_check',
+			),
+			array(
+				'methods'             => WP_REST_Server::CREATABLE,
+				'callback'            => __NAMESPACE__ . '\update_group_info',
+				'permission_callback' => __NAMESPACE__ . '\manage_group_settings_permissions_check',
+				'args'                => array(
+					'title'       => array(
+						'type'              => 'string',
+						'required'          => false,
+						'sanitize_callback' => 'sanitize_text_field',
+					),
+					'description' => array(
+						'type'              => 'string',
+						'required'          => false,
+						'sanitize_callback' => 'sanitize_text_field',
+					),
+				),
+			),
+		)
+	);
+}
+
+/**
+ * Capability check for the group-info routes.
+ */
+function manage_group_settings_permissions_check(): bool {
+	return current_user_can_manage_group_settings();
+}
+
+/**
+ * Return the group's name and description.
+ */
+function get_group_info(): WP_REST_Response {
+	return new WP_REST_Response(
+		array(
+			'title'       => get_option( 'blogname', '' ),
+			'description' => get_option( 'blogdescription', '' ),
+		)
+	);
+}
+
+/**
+ * Update the group's name and description.
+ *
+ * Only the fields present in the request are written, so a client can send
+ * one without clobbering the other.
+ */
+function update_group_info( WP_REST_Request $request ): WP_REST_Response {
+	$title       = $request->get_param( 'title' );
+	$description = $request->get_param( 'description' );
+
+	if ( null !== $title ) {
+		update_option( 'blogname', $title );
+	}
+
+	if ( null !== $description ) {
+		update_option( 'blogdescription', $description );
+	}
+
+	return get_group_info();
 }
 
 /**
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/about-tab.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/about-tab.js
@@ -13,7 +13,6 @@ import {
 } from '@wordpress/element';
 import {
 	TextControl,
-	TextareaControl,
 	Button,
 	Notice,
 	Spinner,
@@ -23,6 +22,7 @@ import { __ } from '@wordpress/i18n';
 
 export default function AboutTab() {
 	const [ loading, setLoading ] = useState( true );
+	const [ loadFailed, setLoadFailed ] = useState( false );
 	const [ saving, setSaving ] = useState( false );
 	const [ notice, setNotice ] = useState( '' );
 	const [ noticeType, setNoticeType ] = useState( 'success' );
@@ -42,7 +42,11 @@ export default function AboutTab() {
 			} )
 			.catch( ( err ) => {
 				// Don't leave the fields silently blank: a failed load looks
-				// identical to an empty group name otherwise.
+				// identical to an empty group name otherwise. Lock the form as
+				// well, so a save can't write those blanks back over the real
+				// values — a failed read followed by a successful write is all
+				// it takes to wipe the group's name and description.
+				setLoadFailed( true );
 				setNoticeType( 'error' );
 				setNotice(
 					err.message ||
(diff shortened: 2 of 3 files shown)


public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
```diff
@@ -171,6 +179,83 @@ function register_routes(): void {
 			) + event_args_schema(),
 		)
 	);
+
+	// ----- Group info -----------------------------------------------------
+	//
+	// The Settings > About tab edits `blogname` and `blogdescription`. Core's
+	// /wp/v2/settings gates both behind `manage_options`, which only network
+	// administrators have, so Organisers got a 403 on every read and write.
```
gedex, line 187: `administrator` has `manage_options` on each subsite too. Editors don't, but they do have `edit_others_posts`, which is what `current_user_can_manage_group_settings()` gates on — so the file header's framing is the accurate one.

```suggestion
	// /wp/v2/settings gates both behind `manage_options`, which Organisers
	// (editors) do not have, so they got a 403 on every read and write.
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
```diff
@@ -171,6 +179,83 @@ function register_routes(): void {
...
+					'title'       => array(
+						'type'              => 'string',
+						'required'          => false,
+						'sanitize_callback' => 'sanitize_text_field',
+					),
+					'description' => array(
+						'type'              => 'string',
+						'required'          => false,
+						'sanitize_callback' => 'sanitize_text_field',
+					),
```
gedex, line 214: Nothing validates these args — `rest_validate_request_arg` is only injected for schema-derived args via `rest_get_endpoint_args_for_schema()`, so `'type' => 'string'` here is decorative and `{"title":""}` stores an empty `blogname`. The theme renders `wp:site-title` in every template header and as the front-page `<h1>`, so the site loses its name everywhere.

A `validate_callback` wouldn't fix it either: `has_valid_params()` runs before `sanitize_params()`, so `{"title":"<b></b>"}` would pass `trim()` and then sanitize down to `''`. The guard belongs in the callback — suggestion on `update_group_info()` below.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
```diff
@@ -171,6 +179,83 @@ function register_routes(): void {
...
+ * Return the group's name and description.
+ */
+function get_group_info(): WP_REST_Response {
+	return new WP_REST_Response(
+		array(
+			'title'       => get_option( 'blogname', '' ),
+			'description' => get_option( 'blogdescription', '' ),
+		)
+	);
+}
```
gedex, line 238: `update_option()` runs both options through `sanitize_option()`, which applies `esc_html()`, so `Portland's Rock & Roll` comes back as `Portland&#039;s Rock &amp; Roll` and shows those entities literally in `TextControl`. Front-end output is fine (`get_bloginfo()` expects the escaped form) and re-saving is stable because `esc_html()` doesn't double-encode — display bug in the edit field only, not creeping corruption.

```suggestion
function get_group_info(): WP_REST_Response {
	return new WP_REST_Response(
		array(
			'title'       => wp_specialchars_decode( get_option( 'blogname', '' ), ENT_QUOTES ),
			'description' => wp_specialchars_decode( get_option( 'blogdescription', '' ), ENT_QUOTES ),
		)
	);
}
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
```diff
@@ -171,6 +179,83 @@ function register_routes(): void {
...
+	if ( null !== $title ) {
+		update_option( 'blogname', $title );
+	}
+
+	if ( null !== $description ) {
+		update_option( 'blogdescription', $description );
+	}
+
+	return get_group_info();
 }
```
gedex: The empty `title` from the args above lands here unguarded. Checking after `get_param()` is what makes the guard see the sanitized value. Returning a `WP_Error` needs the `: WP_REST_Response` return type dropped, or it's a `TypeError`; message left untranslated to match `persist_event()` further down.

```suggestion
function update_group_info( WP_REST_Request $request ) {
	$title       = $request->get_param( 'title' );
	$description = $request->get_param( 'description' );

	if ( null !== $title ) {
		if ( '' === trim( $title ) ) {
			return new WP_Error(
				'wporg_groups_empty_group_name',
				'Group name cannot be empty.',
				array( 'status' => 400 )
			);
		}

		update_option( 'blogname', $title );
	}

	if ( null !== $description ) {
		update_option( 'blogdescription', $description );
	}

	return get_group_info();
}
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/about-tab.js
```diff
@@ -32,23 +32,32 @@ export default function AboutTab() {
...
+			.catch( ( err ) => {
+				// Don't leave the fields silently blank: a failed load looks
+				// identical to an empty group name otherwise.
+				setNoticeType( 'error' );
+				setNotice(
+					err.message ||
+						__( 'Could not load group settings.', 'wporg-groups-frontend' )
+				);
+				setLoading( false );
+			} );
```
gedex, line 56: Still falls through to the normal render with `loading` false, so a failed load leaves an editable empty form with Save enabled — one click POSTs `title: ''` over the real value. The fall-through is pre-existing and the notice here strictly improves on it; what's new is that the write now succeeds where `/wp/v2/settings` used to 403, so the loss path is live for the first time.

```suggestion
			.catch( ( err ) => {
				// Don't leave the fields silently blank: a failed load looks
				// identical to an empty group name otherwise.
				setNoticeType( 'error' );
				setNotice(
					err.message ||
						__( 'Could not load group settings.', 'wporg-groups-frontend' )
				);
				setLoadFailed( true );
				setLoading( false );
			} );
```

Plus `const [ loadFailed, setLoadFailed ] = useState( false );` alongside the other state, and `disabled: saving || loadFailed` on the Save button.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/about-tab.js
```diff
@@ -32,23 +32,32 @@ export default function AboutTab() {
...
 	}, [] );
 
 	const handleSave = async () => {
 		setSaving( true );
 		setNotice( '' );
 		try {
 			await apiFetch( {
```
gedex, line 63: The response body is ignored, so a write that didn't land still reports "Settings saved." Concretely: if `sanitize_option()` errors on unsalvageable UTF-8 it reassigns `$value = get_option( $option )` and reports via `add_settings_error()`, a no-op outside admin — the write vanishes while the field keeps the unsaved text. `update_option` returning `false` for unchanged values makes the server-side return a poor signal, so sync from the response instead; that also surfaces sanitization to the user.

```suggestion
			const saved = await apiFetch( {
```

Then after the call, before the success notice:

```js
			setForm( {
				blogname: saved.title,
				blogdescription: saved.description,
			} );
```

gedex (commented):
Right shape for the problem: `WP_REST_Settings_Controller::get_item_permissions_check()` is a bare `manage_options` with no filter, and GatherPress 0.34.0-alpha.2 has no settings REST surface of its own (its routes are geocoding, venue map, event; `class-settings.php` is admin-only). Scoping two options behind `current_user_can_manage_group_settings()` beats granting Organisers `manage_options`. Multisite cache invalidation comes free via `clean_site_details_cache` on `update_option_blogname`, and CSRF via apiFetch's `X-WP-Nonce`.

Two blockers that combine into a data-loss path: a failed GET leaves an editable empty form with Save enabled, and the POST accepts an empty `title`. Neither half is new, but pointing the tab at an endpoint Organisers can write to arms the chain — a read that fails while the next write succeeds (transient 500, timeout, dropped connection) plus one Save blanks the group name and tagline. Guarding either side breaks it; guard both.

Outside the diff, so noting it here rather than inline: `TextareaControl` at `about-tab.js:100-106` invites multi-line input that `sanitize_text_field` silently flattens, and nothing in `groups-site` renders the tagline (no `site-tagline`, `blogdescription`, or `bloginfo` in `templates/`, `parts/`, `patterns/`, `inc/`, `functions.php`, or `theme.json`), so "shown on the homepage" holds only via `wp_get_document_title()`. Either `TextControl` plus reworded help text, or add a tagline block to the theme.

No tests — consistent with the event routes in this file, but `group-info` is self-contained, so editor-can-write / subscriber-403 / empty-title-rejected / escaping round-trip would be cheap and would cover exactly the above.

wppoland:
Thanks @gedex, that data-loss chain is real and both halves are now guarded, as you suggested.

**Server** — `update_group_info()` rejects an empty or whitespace-only `title` with a 400 (`wporg_groups_empty_group_name`). The check runs *after* sanitization rather than in a `validate_callback`, because `sanitize_text_field()` can empty an input that arrived non-empty: `<script>alert(1)</script>` sanitizes to `''` and would otherwise have been written straight to `blogname`. An empty `description` stays allowed on purpose — a group may legitimately have no tagline, and unlike the name it can be retyped from this same form.

**Client** — the GET failure is now tracked in a `loadFailed` state. Both fields and Save are disabled while it stands, so the blank form can't be written back, and Save is additionally disabled whenever the name is blank.

**On the two points outside the diff:**

`TextareaControl` → `TextControl`. You're right on both counts, and they compound: `sanitize_text_field()` flattens the line breaks the textarea invites, and I confirmed nothing in `groups-site` renders the tagline (no `site-tagline`, `blogdescription` or `bloginfo` anywhere in `templates/`, `parts/`, `patterns/`, `functions.php` or `theme.json`), so the old help text promised something that doesn't happen. Went with your first option since it's the smaller change; the help text now says the tagline is used in the browser title and search results, which is what `wp_get_document_title()` actually does with it. Adding a tagline block to the theme seems worth a separate issue rather than smuggling a theme change into this PR — happy to open one.

**Tests** — added `public_html/wp-content/mu-plugins/tests/test-groups-group-info-rest.php`, 10 cases covering exactly what you listed: Organiser reads and writes, Member gets a 403, logged-out gets a 401, the three empty-title rejections (empty, whitespace-only, emptied-by-sanitization), the escaping round-trip, empty description allowed, and one field not clobbering the other.

Verified locally against the repo's own suite (`phpunit.xml.dist`, multisite, PHP 8.2):

```
OK (10 tests, 22 assertions)
```

and confirmed they actually bite — reverting just the server-side guard fails 3 of them:

```
Tests: 10, Assertions: 18, Failures: 3.
1) test_empty_title_is_rejected
2) test_whitespace_only_title_is_rejected
3) test_title_emptied_by_sanitization_is_rejected
Failed asserting that 200 is identical to 400.
```

`phpcs` is clean on both files (the one warning in `rest.php` is the pre-existing `numberposts` pagination notice on line 550, untouched here).

bor0 (approved):
Verified locally as an Organiser (editor tier): About tab loads pre-filled with no 403, edits persist correctly (wp option get confirms), Save disables correctly on an empty name (client-side no-op, does not touch the stored value), and the server rejects empty/whitespace-only titles with 400 independent of the UI.

Also confirmed directly via REST: subscriber returns 403, anonymous returns 401, organiser returns 200; omitting a field leaves the other untouched; an XSS payload in the title is stripped by sanitize_text_field(); and the organiser account does not have manage_options (the new capability-scoped gate is not a broader access leak).

All of @gedex's review points (empty-GET/empty-POST data-loss chain, TextareaControl to TextControl, missing tests) are addressed in 7998601a. CI is green across the board.

bor0:
Thanks for your contribution, @wppoland! This tests well 🎉

PR #1821, opened 2026-07-28
Groups: let organizers message all members

gedex:
Fixes #1771.

Adds a front-end **Message all members** modal for editable events and calls GatherPress's existing email route with its All Members segment. GatherPress continues to enforce permissions, group-scoped recipients, member opt-ins, and asynchronous delivery.

### Testing

1. Run `npm run build`, sign in as an organizer, and open https://events.wordpress.test/group/sunshine-coast-qld/event/revolution-101/.

   <kbd>image</kbd>

3. Send a unique message and confirm the modal displays **“Your message has been scheduled for delivery.”**

    <kbd>image</kbd>

    <kbd>image</kbd>

5. Run `docker compose exec wordcamp.test wp cron event run --due-now`.
6. Open http://localhost:1080 and confirm only opted-in Sunshine Coast members received it.

    <kbd>image</kbd>

7. Sign in as an ordinary member and confirm the action is hidden.

PR #1822, opened 2026-07-28
Groups: let organizers message event attendees

### LINKED ISSUE #1772: Organizer email: message event attendees only
Let organizers email only the people RSVPed to a specific event (reminders, location changes, cancellations), rather than the whole group.

**Update after checking GatherPress core (`~/dev/gatherpress`, v0.34.0):** this is essentially already built. GatherPress's existing "email members about this event" modal lets the sender pick **Attending**, **Waiting List**, and **Not Attending** as recipient segments (independently of "All Members"), resolved server-side from the event's RSVP comment records — including non-WordPress-user/anonymous RSVPs. Delivery is async (`wp_schedule_single_event()`), per-recipient opt-in is checked (user meta for site users, comment meta for anonymous RSVPs), and the REST route is gated by `current_user_can('edit_post', $post_id)`, matching our own capability model.

**What is actually missing:** same gap as #1771 — this only exists in wp-admin (block editor sidebar + modal), with no front-end equivalent in `wporg-groups-frontend`. Remaining scope is front-end wiring to the existing REST route, not new backend work.

**Acceptance criteria:**
- [ ] A "Message attendees" trigger is available in the front-end event-management UI, with Attending / Waiting List / Not Attending selectable independently.
- [ ] Triggering it calls GatherPress's existing email REST route with the corresponding recipient-segment flags.
- [ ] Only a user who can edit the specific event can trigger it, enforced server-side.
- [ ] Non-WordPress-user (anonymous) RSVPs are included when their status is selected, matching GatherPress's existing recipient resolution — no regression from the wp-admin behavior.

gedex:
Fixes #1772.

Stacked on #1821.

Adds a front-end **Message attendees** action with independently selectable Attending, Waiting list, and Not attending segments. It reuses GatherPress's existing email route, permissions, RSVP recipient resolution—including anonymous RSVPs—opt-in enforcement, and asynchronous delivery.

### Testing

1. Run `npm run build`, sign in as an organizer, and open a test event, for example https://events.wordpress.test/group/sunshine-coast-qld/event/revolution-101/.

    <kbd>image</kbd>

3. Open **Message attendees** and confirm all three RSVP segments can be selected independently and at least one is required.

    <kbd>image</kbd>

4. Send a unique message to Attending and confirm **“Your message has been scheduled for delivery.”**

    <kbd>image</kbd>

5. Run `docker compose exec wordcamp.test wp cron event run --due-now`.
6. Open http://localhost:1080 and confirm only opted-in attendees received it.

    <kbd>image</kbd>

7. Sign in as an ordinary member and confirm the action is hidden.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/event-modal-app.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/event-modal-app.js
@@ -819,9 +819,10 @@ const NS =
 			return null;
 		}
 
-		if ( state.mode === 'message-all' ) {
+		if ( [ 'message-all', 'message-attendees' ].includes( state.mode ) ) {
 			return h( MessageMembersModal, {
 				eventId: state.eventId,
+				recipientMode: state.mode,
 				onClose: () => setState( { open: false, mode: 'create', eventId: 0 } ),
 			} );
 		}
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/message-members-modal.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/message-members-modal.js
@@ -5,15 +5,30 @@
  */
 
 import apiFetch from '@wordpress/api-fetch';
-import { Button, Modal, Notice, TextareaControl } from '@wordpress/components';
+import { Button, CheckboxControl, Modal, Notice, TextareaControl } from '@wordpress/components';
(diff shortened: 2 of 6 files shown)


bor0 (approved):
It works!

Image

Thanks

carstingaxion:
@gedex @bor0 Am I right, that the message textarea is using the same mechanisms like GatherPress core? 

@mauteri it may help an editor to know, what kind of text, html or markdown syntax is allowed for that field to provide a nice editing experience.

Core or not, the field should get some explaining labels, about the allowed syntax(es).

Issue #1781, opened 2026-07-21
Activity dashboard: per-group and network metrics


bor0:
A dashboard showing events held, RSVPs, and attendance, at both the per-group and network-aggregate level.

No analytics/stats code exists in PR #1699, and confirmed GatherPress core itself has no built-in stats/metrics module either (searched `includes/` for anything stat/analytic/metric-related — nothing).

## Scope

- **Per-group view** (Organiser tier): events held (upcoming/past split, matching the existing archive filters), total RSVPs, attendance rate, member count over time.
- **Network-aggregate view** (network admin only): the same metrics rolled up across every group — active groups, total events/RSVPs network-wide, growth trend.
- Likely a new `wporg-groups/v1` REST endpoint (e.g. `/stats`) following the same namespace/capability pattern already established by `Members_Controller` and `rest.php`, rather than a separate system.

## Things to get right given what this session's performance review found

- This is exactly the kind of feature that can silently become an expensive, uncached, unbounded query if not careful from the start — the earlier performance review of this codebase found more than one place doing unbounded `get_users()`/per-row queries on hot pages. A stats dashboard aggregating across potentially hundreds of groups needs pagination/caching considered in the initial design, not retrofitted after it's slow.
- Network-aggregate numbers should probably be cached (transient or scheduled pre-computation) rather than computed live on every dashboard view, since it needs to touch every group's site tables.

## Acceptance criteria

- [ ] Per-group dashboard shows accurate events-held, RSVP, and attendance numbers for that group only.
- [ ] Network-aggregate dashboard is admin-gated and shows correct totals across all groups.
- [ ] Aggregate computation is cached or pre-computed rather than querying every group's site live on each page view.
- [ ] Dashboard remains responsive at a realistic scale (tens of groups, hundreds of events) — not just the current handful of test groups.

carstingaxion:
Hello Team,

Once again, I want to note existing work that could be included in a first triage session:

One little experiment plugin, that is so small and focused, that I could see this being merged into GatherPress core. But since now I haven’t even suggested this because of missing time. 

https://github.com/carstingaxion/gatherpress-at-a-glance

![](https://raw.githubusercontent.com/carstingaxion/gatherpress-at-a-glance/main/.wordpress-org/screenshot-1.png)

<hr>

And one that is probably to big to go into GatherPress core, but could be improved as a kind of recommended companion plugin, if it solves more than only my needs for statistics about existing data, other than a usage tracker.

[...]

PR #1833, opened 2026-07-30
Groups: include organised events in the my-events block

### LINKED ISSUE #1810: Groups: my-events block shows nothing for users who created events but haven't RSVP'd
Flagged during the curl/REST test pass on #1793: https://github.com/WordPress/wordcamp.org/pull/1793#issuecomment-5068168317

> **`my-events` block returns 0 chars for every role**, even `organiser1` who created 4 events during testing. Might be intentional (block might filter by RSVP-attendance rather than authorship), but worth confirming with the block's author.

Confirmed by reading `wporg-groups-frontend/src/blocks/my-events/render.php`: the block is purely RSVP-attendance based — it queries `gatherpress_rsvp` comments by the current user with an `attending` term, and returns nothing if there are none. **Creating or organizing an event does not RSVP the creator to it**, so an organiser who just created 4 events but never clicked "RSVP" on any of them correctly sees an empty block under the current logic. This isn't a rendering bug, but it may not match user expectations for a block titled "My upcoming events" — organisers likely expect to see events they're running, not just ones they've RSVP'd to.

## Open question

Should `my-events` also (or instead) surface events the current user organizes/authored, in addition to ones they've RSVP'd "attending"? If so, decide whether:
- creating an event should auto-RSVP the creator as attending, or
- the block's query should independently include events where the user is the post author, unioned with their RSVP'd events.

## Acceptance criteria

- [ ] Decide the intended behavior (RSVP-only vs. RSVP + organized-events) and document it in the block's render.php docblock.
- [ ] If organized events should be included: update the query in `my-events/render.php` to also pull events authored by the current user, deduplicated against their RSVP'd events.
- [ ] Re-run the block-registration/rendering check from #1793's test plan with a user who has created but not RSVP'd to an event, confirming the block now reflects the decided behavior.

wppoland:
Fixes #1810

Props wppoland

An organiser who created events but never clicked RSVP saw an empty `my-events` block, because the block answered purely from RSVP data and creating an event does not RSVP the creator to it.

This implements **Option B** from the issue discussion: authored events are unioned into the query, rather than auto-RSVPing the creator. That keeps "attending" meaning that the person said they are coming, which matters because the RSVP data feeds attendee counts, email targeting (#1772) and metrics (#1781), and because an organiser who creates an event they will not personally attend is normal on a group with several organisers. It also leaves authorship as a separate signal at render time, so distinguishing "organising" from "attending" entries visually stays possible later; auto-RSVPing would have closed that door. If you would rather have Option A, say so and I will swap it.

Three changes:

* The query moves to `inc/my-events.php` as `My_Events\get_upcoming_event_ids()`, so the definition of "my events" is testable on its own and the template is left with rendering. It unions attending and authored IDs, deduplicates, then filters to upcoming in one query against GatherPress's datetime table, as the block already did.
* Results are ordered by start time ascending. The block is billed as upcoming events, and the previous `post__in` ordering followed the RSVP lookup order.
* The empty case renders the heading with a short explanation instead of nothing. Whichever definition of "mine" wins, a block that disappears entirely is indistinguishable from a broken one, which is exactly how this was reported. This addresses the "worth doing regardless" point from the issue thread.

The `render.php` docblock now states the intended behaviour, per the first acceptance criterion.

### Screenshots

Not included: the visible change is one extra list entry for organisers, plus the empty-state line, both of which the tests below assert directly.

1. As an organiser on a group site, create and publish an event, and do **not** RSVP to it. Load a page containing the `wporg/my-events` block. Before this change the block was absent; now the event is listed.
2. RSVP to that same event as the same organiser. The event stays listed exactly once.
3. As a member who did not author the event, RSVP as attending. The event is listed, so the original behaviour is intact.
4. Set an event's end time in the past and confirm it drops off both paths.
5. As a member with no events at all, confirm the block now shows "My upcoming events" with an explanatory line instead of rendering nothing.
6. Create three upcoming events with different start times and confirm they list soonest first.

`public_html/wp-content/mu-plugins/tests/test-groups-my-events.php` covers all six: authored-without-RSVP, attending-without-authorship, the deduplicated both case, past events on both paths, another member's events staying out, ordering, and the empty resolution. 7 tests / 9 assertions green locally.

Verified the tests actually capture the bug: reverting `get_upcoming_event_ids()` to the old RSVP-only behaviour fails `test_authored_events_are_included_without_an_rsvp` and `test_events_are_ordered_soonest_first`, and restoring it makes them pass.

One note on the local run: `test-blocks.php`'s `test_exactly_the_expected_wporg_blocks_are_registered` fails in my checkout because the blocks register from `build/blocks`, which this clone has not built. It fails the same way on a clean `production` checkout here, so it is unrelated to this change.


wppoland:
The E2E failure on the first push was the change working, so I have folded the fix into this PR rather than leaving it red.

`event-organiser.spec.js` asserted the published event's title with a bare `getByText( title )`. With this change the title legitimately appears twice on the front page for an organiser: once in the events list, once in the my-events block. Playwright's strict mode then fails on two matches:

```
Error: strict mode violation: getByText('E2E Test Event 1785410135399') resolved to 2 elements:
    1) <a href=".../event/e2e-test-event-1785410135399/">E2E Test Event 1785410135399</a>
    2) <span class="wporg-my-events__title">E2E Test Event 1785410135399</span>
```

Element 2 is the point of #1810: that organiser published the event and never RSVP'd to it, and previously the block rendered nothing at all. So the existing assertion is now scoped to the events-list link, and a second assertion targets `.wporg-my-events__title` directly, which makes the E2E flow cover the fixed behaviour end to end.

PHP tests passed on 8.3, 8.4 and 8.5 on that same run.

bor0 (approved):
PR tests well and I like that it has good test coverage to my attempt.

Thanks for the contribution, @wppoland! 🎉

PR #1836, opened 2026-07-30
Close the one-request gap in the theme network-restriction backstop

### LINKED ISSUE #1835: Theme network-restriction backstop is delayed by one request, contradicting its docblock
`wcorg-network-theme-control.php`'s `revert_wrong_network_activation()` is described as a "hard backstop" that prevents ending up with "a broken-looking site" (see the file's own docblock, lines 6-12). In practice the revert is delayed by at least one full HTTP request, during which the restricted theme is fully live and publicly rendered — not just a documentation nit, since it means the guard doesn't do what it currently claims to do.

### Why

`switch_theme()` in WP core only fires the `switch_theme` action synchronously and writes the `theme_switched` option. The `after_switch_theme` hook this file listens on (line 24) is fired later by `check_theme_switched()`, which core hooks to `init` at priority 99 — i.e. on the *next* request after the switch, not the one that performed it. `public_html/wp-content/mu-plugins/tests/test-wcorg-network-theme-control.php` already documents this precisely (see the comment on `test_activation_on_wrong_network_is_reverted()`, ~line 100) and calls `check_theme_switched()` manually to simulate the next load — so the timing gap is understood at the test level, just not reflected in the guard's own docblock or addressed as a real-world exposure window.

Concretely: an admin activates `groups-site` (or a future network-restricted theme) on the wrong network. The `stylesheet`/`template` options update immediately. Any real visitor hitting the site before the *next* request triggers `check_theme_switched()` sees the restricted theme rendered — and even that triggering request itself serves the restricted theme, since the current request's template is resolved from the (still wrong) option earlier in the bootstrap, before `init` priority 99 runs the revert. If a page cache or CDN caches that response, the exposure window extends well past a single request.

### Suggested fix

Not a security hole on its own (activating a theme already requires admin privileges), but the backstop should either:
- Also guard at template-resolution time (e.g. a `template_redirect`/`setup_theme` check that force-reverts *before* the page renders, not just after), or
- Have its docblock updated to accurately describe the one-request-delayed nature of the current backstop, so it isn't relied on as an instantaneous guarantee.

### Where this came from

Surfaced by an automated code review of the `issue-1772-message-event-attendees` branch (PR #1822) scanning the full diff against `production`, which includes this file from the earlier #1831/#1826 work — not something introduced by #1822 itself.

bor0:
Fixes #1835.

## Summary

`revert_wrong_network_activation()`'s `after_switch_theme` hook is deferred by WordPress core to `check_theme_switched()`, hooked on `init` at priority 99 — i.e. it only runs on the *next* request after a restricted theme (e.g. `groups-site`) is switched onto the wrong network. In between, the restricted theme's `functions.php` — which may assume plugins/constants unique to its own network — loads during that one request's bootstrap, well before `init` runs. This contradicted the file's own docblock claim of preventing "a broken-looking site."

This adds a `stylesheet`/`template` filter (`prevent_wrong_network_theme_boot()`) that reports the *previous* theme for the remainder of that one request, so the restricted theme's `functions.php` never actually loads — not even once. It deliberately skips itself for the request that performs the switch (tracked via a `switch_theme` action), since a theme's `functions.php` only loads once at bootstrap, before any application code (including the switch itself) runs — so that request was never at risk, and skipping it also keeps `wp theme activate`/wp-admin's own success feedback correct for that request.

The existing `revert_wrong_network_activation()` backstop is unchanged in what it does — it still persists the corrected `stylesheet`/`template` options and shows the admin notice — just updated to read the raw option instead of the now-filtered `get_stylesheet()`.

### Testing

1. `docker compose -f docker-compose.phpunit.yml up -d`
2. `docker compose -f docker-compose.phpunit.yml exec phpunit_wp phpunit --filter Test_WCORG_Network_Theme_Control --testsuite 'WordCamp MU Plugins'` — 7 tests, all green (5 existing + 2 new).
3. Full `WordCamp MU Plugins` (161 tests) and `WordCamp Groups` (38 tests) suites also pass, confirming no regressions.
4. Manually verified against the local dev stack:
   - `wp theme activate groups-site --url=central.wordcamp.test` still reports `Success: Switched to 'WordPress Group' theme.` (the activating request's own feedback is unaffected).
   - A subsequent request confirms the theme is reverted back to `wordcamp-central-2012`, same as before this PR.
   - `wp theme activate groups-site --url=events.wordpress.test/group/sunshine-coast-qld/` (its own network) is unaffected — the theme stays active.
   - No new PHP warnings/fatals in `debug.log`, both networks render 200.

### DIFF
--- a/public_html/wp-content/mu-plugins/tests/test-wcorg-network-theme-control.php
+++ b/public_html/wp-content/mu-plugins/tests/test-wcorg-network-theme-control.php
@@ -44,11 +44,15 @@ class Test_WCORG_Network_Theme_Control extends Groups_TestCase {
 
 	/**
 	 * Restore the network global that `set_current_network()` overwrote,
-	 * so state doesn't leak into other tests.
+	 * and clear the request-scoped "did this request just switch a theme"
+	 * flag `mark_theme_switched_this_request()` sets, so neither leaks into
+	 * other tests -- PHPUnit runs the whole suite in one process, unlike
+	 * the real one-flag-per-HTTP-request lifecycle this is designed for.
 	 */
 	protected function tearDown(): void {
 		// phpcs:ignore WordPress.WP.GlobalVariablesOverride.Prohibited -- Restoring original state.
 		$GLOBALS['current_site'] = $this->original_current_site;
+		unset( $GLOBALS['wcorg_network_theme_control_switched_this_request'] );
 
 		parent::tearDown();
 	}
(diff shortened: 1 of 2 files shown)


bor0:
Checked production: `wp_get_theme( 'groups-site' )->is_allowed()` is `false` on both `central.wordcamp.org` and `events.wordpress.org` — that covers the network *and* site allowlists, and core's `wp_prepare_themes_for_js()` already filters to allowed themes. Core REST can't activate themes either. So the theme is unreachable from wp-admin regardless.

Only `wp theme activate` from a shell skips `is_allowed()`. The window this closes is: someone with production access deliberately activates the wrong theme, and one request renders before the existing backstop reverts. `groups-site/functions.php` just registers hooks at load time — no fatal — so the harm is one ugly page render.

Not worth `stylesheet`/`template` filters on every request across the whole multisite. Closing.

PR #1837, opened 2026-07-31
Groups: use native GatherPress event queries

### LINKED ISSUE #1801: Groups: use GatherPress's native event-query block variation instead of custom SQL ordering
Flagged by @carstingaxion on #1791: the custom `pre_get_posts`/`posts_join`/`posts_orderby` block in `gatherpress-groups-tweaks.php` that orders events by datetime duplicates functionality GatherPress already provides via `GatherPress\Core\Event\Query` (`get_upcoming_events()`/`get_past_events()`) and its native `gatherpress-event-query` block variation.

Reason we have it: the theme's Query Loop blocks (`front-page.html`, `archive-gatherpress_event.html`) are plain `core/query` blocks with `orderBy:"date"`, not GatherPress's registered variation, so they don't get native date-ordering for free.

Fix: rework those Query Loop blocks to use GatherPress's `gatherpress-event-query` variation (with the right `orderBy`/`gatherpress_event_query` attrs), then delete the custom SQL-join reimplementation.

Also: the table name (`$wpdb->prefix . 'gatherpress_events'`) is currently correct but hardcoded rather than derived from `GatherPress\Core\Event\Event::TABLE_FORMAT` — fragile against a future rename. Worth fixing as part of the same cleanup, or independently if the refactor above is deferred.

## Acceptance criteria

- [ ] `front-page.html` and `archive-gatherpress_event.html`'s Query Loop blocks use GatherPress's `gatherpress-event-query` block variation instead of plain `core/query` with `orderBy:"date"`.
- [ ] The custom `pre_get_posts`/`posts_join`/`posts_orderby` reimplementation in `gatherpress-groups-tweaks.php` is removed once the variation covers the same behavior.
- [ ] Upcoming vs. past ordering on both the front page and archive is visually verified correct after the switch, not just "no fatal error."
- [ ] Any remaining table-name reference uses `GatherPress\Core\Event\Event::TABLE_FORMAT` (confirmed to exist, `'%sgatherpress_events'`) instead of a hardcoded string.
- [ ] Event search/filtering (event_time query var, upcoming/past/all) on the archive still works after the refactor.

raicem:
Use the GatherPress event-query variation on group home/archive templates and remove the duplicated SQL ordering. Also derive the remaining table name from `Event::TABLE_FORMAT`.

Keep event cards aligned when featured images are absent or their files fail to load. Featured images are constrained to a 16:9 frame, and cards fill their grid row without broken media inflating the layout.

Fixes #1801

### Event card layout

The fixture below includes cards with valid featured images, cards without featured images, and cards whose seeded image files are unavailable.

Past event cards with mixed featured-image states aligned in a three-column grid

### How to test

1. Create two future and two finished events, each with different start dates.
2. Add featured images to some events and leave others without one.
3. Confirm the homepage shows future events earliest-first.
4. On `/event/`, confirm Upcoming is earliest-first, Past is newest-first, and All contains all four.
5. Confirm event cards remain aligned with valid, absent, or unavailable featured images.
6. Search for one event with All selected; confirm only the match appears.


bor0 (approved):
Tests well! Excellent simplification and improvements. Thanks!

PR #1845, opened 2026-08-03
Add recurring events support to GatherPress

gedex:
## Summary

Adds a portable GatherPress extension for storage-efficient recurring events. It can be activated on an individual WordPress site or across a multisite network; it contains no WordPress.org-specific loading or frontend assumptions.

- Stores one `gatherpress_event` series post and projects lightweight occurrence rows with WP-Cron.
- Supports weekly, monthly, and yearly recurrence rules for the initial version.
- Adds occurrence permalinks, date-aware queries, calendar links, per-occurrence RSVP/comment mapping, and generic occurrence navigation.
- Adds editor controls for creating, ending, and cancelling recurring series.
- Keeps the extension and its PHPUnit bootstrap/tests together under `plugins/gatherpress-recurring-events`.
- Leaves WordPress.org Groups activation, its organizer-facing forms, and its custom RSVP bridge to the stacked #1846.

## Why

Recurring events should remain a reusable GatherPress extension that can be tested on WordPress.org Groups without coupling its core behavior to that environment. If the experiment proves successful, this separation also makes the functionality easier to upstream into GatherPress.

The rolling occurrence projection keeps unbounded series compact: shared content remains on one GatherPress event while each occurrence has its own date, URL, RSVP state, and discussion.

## MVP constraints

Published recurrence schedules are immutable. Organizers can end a series or cancel projected occurrences, but changing an existing recurrence pattern or editing a single occurrence is intentionally deferred.

Aggregate GatherPress calendar feeds, including site-wide, event archive, venue, and taxonomy feeds, remain series-based. They include the series' master event but do not expand its projected occurrences. Calendar links for an individual occurrence still use that occurrence's date and URL.

The wp-admin events list intentionally shows one editable row per recurring series rather than one row per projected occurrence. Projected occurrences are not separate posts and do not have individual edit screens.

## Test instructions

1. From the repository root, activate GatherPress and GatherPress Recurring Events for the network:

   ```bash
   docker compose exec wordcamp.test wp plugin activate gatherpress --network --url=events.wordpress.test/group/sunshine-coast-qld/
   docker compose exec wordcamp.test wp plugin activate gatherpress-recurring-events --network --url=events.wordpress.test/group/sunshine-coast-qld/
   ```

2. Open [Add New Event](https://events.wordpress.test/group/sunshine-coast-qld/wp-admin/post-new.php?post_type=gatherpress_event).
3. Create an event with a future start date, select **Weekly** under **Repeating event**, choose one or more weekdays, configure an end condition, and publish it.
4. Edit the published event. Confirm the recurrence rule is visible but locked, shared event content remains editable, and individual future occurrences can be cancelled or used to end the series.
5. Open the event on the frontend and navigate between occurrence dates. Confirm the selected date, URL, active state, and event details update while earlier upcoming dates remain available.
6. As a logged-in member, RSVP and comment on one occurrence. Confirm that RSVP and discussion state do not appear on another occurrence.

> **FYI:** On WordPress.org Groups sites, occurrence-specific RSVP persistence is provided by stacked PR #1846, so it is expected not to persist when testing #1845 alone.

### Create a recurring event

<kbd>image</kbd>

### Published recurrence schedule is locked

<kbd>image</kbd>

### Navigate individual occurrences

<kbd>image</kbd>


public_html/wp-content/plugins/gatherpress-recurring-events/includes/class-query.php
```diff
@@ -0,0 +1,121 @@
...
+		foreach ( $posts as $index => $post ) {
+			if ( empty( $rows[ $index ]->recurrence_id ) ) {
+				continue;
+			}
+
+			$occurrence = Occurrences::get( (int) $post->ID, (string) $rows[ $index ]->recurrence_id );
+			if ( $occurrence ) {
+				$posts[ $index ]                                     = clone $post;
+				self::$contexts[ spl_object_id( $posts[ $index ] ) ] = $occurrence;
+			}
```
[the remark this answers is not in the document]
gedex: Fixed in 16be237b: the single reconstructed archive SQL query now selects the joined occurrence row, so Query::posts() reuses it instead of calling Occurrences::get() per result.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
```diff
@@ -0,0 +1,193 @@
...
+			Rule::META_PREFIX . 'count',
+		);
+
+		if ( 'publish' !== get_post_status( $object_id ) || ! Rule::is_recurring( $object_id ) || ! in_array( $meta_key, $locked, true ) || ! metadata_exists( 'post', $object_id, $meta_key ) ) {
+			return $check;
+		}
+
+		$current = get_post_meta( $object_id, $meta_key, true );
+		return maybe_serialize( $current ) === maybe_serialize( $meta_value ) ? $check : false;
+	}
```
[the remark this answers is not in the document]
gedex: Fixed in 16be237b: the end type and until date are now locked for published series. The dedicated end-series path uses a controlled internal update and refuses changes that would move an existing end later.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-occurrences.php
```diff
@@ -0,0 +1,408 @@
...
+
+final class Occurrences {
+
+	const CRON_HOOK = 'gpre_project_occurrences';
+
+	/** Ensures the per-site daily projection job is scheduled. */
+	public static function schedule_cron(): void {
+		if ( ! wp_next_scheduled( self::CRON_HOOK ) ) {
+			wp_schedule_event( time() + HOUR_IN_SECONDS, 'daily', self::CRON_HOOK );
+		}
```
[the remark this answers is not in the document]
gedex: Fixed in 16be237b: registered a deactivation hook that clears the current site’s projection cron and iterates all sites when the plugin is network-deactivated.

public_html/wp-content/plugins/gatherpress-recurring-events/plugin.php
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Plugin Name: GatherPress Recurring Events
+ * Description: Adds storage-efficient recurring occurrences to GatherPress events.
+ * Version: 0.1.0
+ * Requires PHP: 8.1
+ * Requires Plugins: gatherpress
+ * Network: true
+ *
+ * @package WordPressdotorg\GatherPress_Recurring_Events
+ */
```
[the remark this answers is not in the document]
gedex: Addressed in 16be237b: added the gpre text domain. I’m retaining the PHP 8.1 requirement intentionally: this is a portable standalone extension, it only uses PHP 8.1 features, and the WordCamp runtime’s 8.3 floor does not require the plugin to advertise a higher compatibility floor.

gedex:
All Copilot feedback has been addressed. Translation metadata was added, published end conditions are now locked with a controlled shortening path, the archive N+1 query was removed, and projection cron cleanup was added. The PHP 8.1 floor remains intentional for portability. All inline threads are resolved.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-comments.php
```diff
@@ -0,0 +1,121 @@
...
+			$table
+		);
+		$clauses['where'] .= $wpdb->prepare(
+			' AND gpre_oc.series_post_id = %d AND gpre_oc.recurrence_id = %s',
+			$occurrence->series_post_id,
+			$occurrence->recurrence_id
+		);
+
+		return $clauses;
+	}
```
gedex, line 57: `prepare_query` still tags every `WP_Comment_Query` while a context is active, and `clauses()` INNER JOINs on it. Recent-comments widgets and other listings on singular event pages get silently reduced to the current occurrence. Scope the guard to queries targeting the series post.

public_html/wp-content/plugins/gatherpress-recurring-events/includes/class-query.php
```diff
@@ -0,0 +1,118 @@
...
+namespace WordPressdotorg\GatherPress_Recurring_Events;
+
+use WP_Post;
+use WP_Query;
+
+defined( 'WPINC' ) || die();
+
+final class Query {
+
+	private static array $contexts = array();
```
gedex: `self::$contexts` still keyed on `spl_object_id` and never reset. Reused ids after GC can attach the wrong occurrence to a later cloned post; also grows unbounded in CLI/cron. Reset per batch or key on `(post ID, recurrence_id)`.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
```diff
@@ -0,0 +1,241 @@
...
+	public function save_event( int $post_id, $post ): void {
+		if ( wp_is_post_revision( $post_id ) || wp_is_post_autosave( $post_id ) ) {
+			return;
+		}
+
+		if ( 'publish' === $post->post_status && Rule::is_recurring( $post_id ) ) {
+			Occurrences::project( $post_id );
+		} elseif ( ! Rule::is_recurring( $post_id ) && 'publish' !== $post->post_status ) {
+			Database::delete_series( $post_id );
+		}
```
gedex, line 161: `elseif ( ! is_recurring && ! publish )` still needs both. Unpublishing a recurring event orphans projected rows + comment mappings; republish re-projects on top of stale state. Change to `else { Database::delete_series(...); }`.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
```diff
@@ -0,0 +1,241 @@
...
+			Rule::META_PREFIX . 'count',
+		);
+
+		if ( self::$schedule_update_allowed || 'publish' !== get_post_status( $object_id ) || ! Rule::is_recurring( $object_id ) || ! in_array( $meta_key, $locked, true ) || ! metadata_exists( 'post', $object_id, $meta_key ) ) {
+			return $check;
+		}
+
+		$current = get_post_meta( $object_id, $meta_key, true );
+		return maybe_serialize( $current ) === maybe_serialize( $meta_value ) ? $check : false;
+	}
```
gedex, line 237: Lock still only hooks `update_post_metadata`. `delete_post_meta( _gpre_frequency )` flips `is_recurring()` to false and unlocks everything. Add the same filter for `delete_post_metadata`.

public_html/wp-content/plugins/gatherpress-recurring-events/includes/class-plugin.php
```diff
@@ -0,0 +1,241 @@
...
+	 */
+	public static function update_end_condition( int $post_id, string $until ): void {
+		self::$schedule_update_allowed = true;
+
+		try {
+			update_post_meta( $post_id, Rule::META_PREFIX . 'end_type', 'until' );
+			update_post_meta( $post_id, Rule::META_PREFIX . 'until', $until );
+		} finally {
+			self::$schedule_update_allowed = false;
+		}
```
gedex: `$schedule_update_allowed` is request-wide and unscoped: while it's true the lock short-circuits for any `(object_id, meta_key)`. Hooks fired by the two `update_post_meta` calls that touch locked meta on this or any other post escape the lock too. Scope to `($post_id, $meta_key)`.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
```diff
@@ -0,0 +1,241 @@
...
+				'fields' => 'ids',
+				'number' => 0,
+			)
+		);
+		foreach ( $site_ids as $site_id ) {
+			switch_to_blog( (int) $site_id );
+			Occurrences::clear_cron();
+			restore_current_blog();
+		}
+	}
```
gedex, line 65: Network deactivate runs `get_sites( 'number' => 0 )` + per-site `switch_to_blog` in a single request. Will time out at WP.org scale and leave cron rows behind on later sites. Paginate, or defer per-site cleanup to a self-rescheduling job.

gedex:
A few things worth another look: the `pre_get_comments`/`comments_clauses` pair filters every comment query on singular event pages, not just ones targeting the series (`class-comments.php:22-56`); `Query::$contexts` is keyed on `spl_object_id` and never reset, so post-GC id reuse can misattribute context and the map grows unbounded (`class-query.php:17,91,104`); `save_event`'s cleanup elseif requires both non-recurring and non-published, so unpublishing a recurring event orphans projected rows and mappings (`class-plugin.php:138-142`); the schedule lock only hooks `update_post_metadata`, so `delete_post_meta( _gpre_frequency )` disables it entirely (`class-plugin.php:197-218`); `$schedule_update_allowed` is a request-wide unscoped flag, so cascading meta writes during `update_end_condition()` bypass the lock on any post/key (`class-plugin.php:16-17, 62-69, 212`); and network deactivate iterates every site in one request with `'number' => 0` — will time out at WP.org scale (`class-plugin.php:36-53`). Also worth a glance: `end_after`'s `<=` rejects idempotent same-date calls, `Occurrences::around()` dropping `$selected` gives past-occurrence views a misleading forward-only nav plus false "series has ended" text, `Query::posts()`'s SELECT-list regex is still fragile, and `filemtime()` runs on every enqueue with no fallback.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-comments.php
gedex: Fixed in 9a2caa0d: occurrence scoping now applies only to comment queries explicitly targeting the active series post.

public_html/wp-content/plugins/gatherpress-recurring-events/includes/class-query.php
gedex: Fixed in 9a2caa0d: occurrence contexts now use a WeakMap keyed by exact cloned post objects, avoiding ID reuse and allowing automatic cleanup.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
gedex: Fixed in 9a2caa0d: all non-published or non-recurring saves now delete projected series data.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
gedex: Fixed in 9a2caa0d: the published schedule lock now also filters metadata deletion.

public_html/wp-content/plugins/gatherpress-recurring-events/includes/class-plugin.php
gedex: Fixed in 9a2caa0d: the controlled end-condition bypass is scoped to the exact post and two permitted meta keys.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
gedex: Fixed in 9a2caa0d: network deactivation now processes sites in bounded batches of 100.

wppoland:
Read this with a GatherPress hat on, since the occurrence projection has to coexist with core's own event queries.

The overall shape reads well. Keeping occurrences in their own table and treating the `gatherpress_events` row as an immutable series seed sidesteps the failure mode where a `gatherpress_event` post exists with no row in that table and `Event\Query` simply cannot see it. Reading the seed as local time plus the stored `timezone`, rather than off the GMT columns, is also the right call for DST correctness across a recurrence.

One consequence I think is worth surfacing before this lands.

### `Query::clauses()` opts out of `fields => 'ids'`, and core's event list is exactly that

```php
if ( ! $type || 'ids' === $query->get( 'fields' ) || is_admin() ) {
	return $clauses;
}
```

I assume the guard is deliberate: `Query::posts()` clones `WP_Post` objects to attach occurrence context, and with `fields => 'ids'` the `the_posts` filter hands you plain integers, so there is nothing to clone.

The catch is that `GatherPress\Core\Event\Query::get_events_list()` sets both the query var your `type()` keys on and the field mode you bail on, in the same `$args`:

```php
$args = array(
	'post_type'             => get_post_types_by_support( 'gatherpress-event-date' ),
	'fields'                => 'ids',
	'no_found_rows'         => true,
	'posts_per_page'        => $number,
	self::EVENT_QUERY_PARAM => $event_list_type,   // gatherpress_event_query
	'order'                 => $order,
);
```

(`includes/core/classes/event/class-query.php`, in `get_events_list()`)

That method is what `Calendar\Setup` calls to build the `.ics` responses for the post type archive and for taxonomy terms. With the join skipped, those feeds resolve each series only at its master `datetime_start_gmt`, so someone subscribed to a group's calendar would receive the first occurrence of a weekly series and none of the repeats. Anything else routed through `get_events_list()` inherits the same gap.

I do not think there is an obvious cheap fix, since the clone path genuinely cannot work in ids mode, and it may well be out of scope for the MVP. If it is, it seems worth an explicit line in the plugin README, so that whoever notices the ICS discrepancy later reads it as a known limitation rather than a bug in core. If you do want it covered, the occurrence rows already hold everything needed to expand the id list at `the_posts` time; it just needs a separate branch from the clone path.

### Minor, same guard

`is_admin()` is bailed on as well, so the admin events list shows series rather than occurrences. I assume that is intentional, but it does mean the event count an organiser sees in wp-admin will not match the front end, which is the kind of thing that generates support questions.

gedex:
The `fields => 'ids'` concern is valid. Calendar feeds currently only see the original series event, and removing the guard alone wouldn’t preserve occurrence-specific dates. It as an MVP limitation. I'll update PR description and `gatherpress-recurring-events/README.md`

The wp-admin behavior is intentional. It lists editable series rather than individual occurrences, since occurrences aren’t separate posts.

gedex:
> I'll update PR description and `gatherpress-recurring-events/README.md`

PR description updated and 5369c982 updated the `gatherpress-recurring-events/README.md`.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-rule.php
```diff
@@ -0,0 +1,263 @@
...
+	 */
+	private static function matches( DateTimeImmutable $candidate, DateTimeImmutable $start, array $rule ): bool {
+		if ( $candidate < $start ) {
+			return false;
+		}
+
+		// DTSTART is always part of an RFC recurrence set, even when a BY* rule
+		// would not independently select it.
+		if ( $candidate->getTimestamp() === $start->getTimestamp() ) {
+			return true;
```
bor0, line 176: Confirmed live: creating a weekly event with `REPEAT ON = Wednesday` but a start date of Tuesday, Aug 11 produces occurrences on **both** Aug 11 (Tue) and Aug 12 (Wed) — one day apart — before settling into the correct weekly Wednesday cadence. DB rows and the front-end occurrence selector both confirm the extra occurrence.

This line is RFC5545-correct (DTSTART is always part of the recurrence set even when it doesn't independently match a BY* rule), so I'm not asking to remove it. But the editor UI (`assets/editor.js`) doesn't warn the organizer or offer to align the start date with the chosen weekday(s) when they diverge, so in practice this reads as a bug: most organizers will pick a placeholder start date first and the weekday(s) after, producing this off-schedule duplicate by default.

Could we either (a) snap/prompt to adjust the start date to the first matching weekday when the rule is set in the editor, or (b) show an explicit notice in the Repeating Event panel when the start date doesn't match the selected BYDAY, so this is a deliberate choice rather than a silent surprise?

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-rest-api.php
```diff
@@ -0,0 +1,229 @@
...
+
+	/**
+	 * Updates the current user's RSVP for an occurrence.
+	 *
+	 * @param WP_REST_Request $request REST request.
+	 * @return WP_REST_Response REST response.
+	 */
+	public static function update_rsvp( WP_REST_Request $request ): WP_REST_Response {
+		$post_id    = (int) $request->get_param( 'post_id' );
+		$occurrence = self::set_context( $post_id, (string) $request['recurrence_id'] );
```
bor0, line 131: `update_rsvp` doesn't independently verify `get_post_status( $post_id ) === 'publish'` before writing an RSVP, unlike `occurrences()` above which does. Today this is mitigated because `Plugin::save_event()` calls `Database::delete_series()` (wiping occurrence rows) on every non-publish save, so occurrence rows practically can't exist for a draft/private event under the normal lifecycle — but that's an indirect guarantee with no defense-in-depth backup if a future code path (bulk edit, direct `wp_update_post()` call, etc.) skips that hook. Since `occurrences()` already has the pattern, could we add the same `'publish' !== get_post_status( $post_id )` guard here (and in `responses()` below) for a cheap, explicit safeguard?

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-rest-api.php
```diff
@@ -0,0 +1,229 @@
...
+
+	/**
+	 * Returns occurrence-scoped RSVP responses.
+	 *
+	 * @param WP_REST_Request $request REST request.
+	 * @return WP_REST_Response REST response.
+	 */
+	public static function responses( WP_REST_Request $request ): WP_REST_Response {
+		$post_id    = (int) $request->get_param( 'post_id' );
+		$occurrence = self::set_context( $post_id, (string) $request['recurrence_id'] );
```
bor0, line 168: Same as the `update_rsvp` comment above — no independent `publish`-status check before returning RSVP response data for this occurrence. Low severity (mitigated by `delete_series()` on the save lifecycle), but worth the same explicit guard for defense-in-depth.

bor0 (changes requested):
Reviewed for security, performance, and correctness, plus live-tested in a browser (created a real weekly-recurring event through the block editor, published it, and exercised the front-end occurrence selector, per-occurrence permalinks/comments, and REST routes against a local WordCamp.org env).

**Security:** clean. No SQL injection (all dynamic queries go through `$wpdb->prepare()`), no XSS (output is consistently escaped), no CSRF, and the state-changing REST routes correctly gate on `current_user_can( 'edit_post', $post_id )` for the specific post in the URL rather than just "is logged in". Two low-severity, non-blocking notes below.

**Performance:** no blockers. `Plugin::save_event()` re-runs full occurrence projection (multiple `INSERT IGNORE` round-trips + a stale-row scan) on *every* save of a published recurring event, not only when the schedule changes — since the schedule is locked post-publish this is idempotent, but it's needless cost on routine content edits. Also `Query::posts()` re-executes the entire built `WP_Query` SQL a second time for occurrence-aware listings. Neither blocks shipping, just flagging for a follow-up.

**Functional bug (live-tested, see inline comment):** creating a weekly event with a start date that doesn't fall on the selected weekday(s) produces two occurrences one day apart in week one, rather than snapping to the chosen weekday. Verified end-to-end: editor → publish → DB rows → front-end occurrence selector all show the extra occurrence. Root cause is intentional/RFC5545-compliant (DTSTART is always part of the recurrence set), but nothing in the UI warns the organizer or aligns the start date, so this will read as a bug to real users and create unwanted duplicate/off-schedule occurrences whenever the placeholder start date doesn't already match the chosen weekday — which will be the common case.

Requesting changes for the DTSTART/weekday mismatch (real user-facing incorrect behavior) plus the two RSVP endpoint hardening asks inline. Nothing else here should block merge.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-rule.php
bor0: Fixed in 45dda021: added a non-dismissible warning Notice in the editor's Repeating Event panel (option b from above) when the start date's weekday isn't among the selected repeat days, so this is a deliberate, visible choice rather than a silent surprise. Left the RFC5545-correct DTSTART-inclusion behavior in class-rule.php as-is per your note.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-rest-api.php
bor0: Fixed in 45dda021: update_rsvp() now checks get_post_status( $post_id ) === 'publish' up front, matching the guard occurrences() already had.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-rest-api.php
bor0: Fixed in 45dda021: same guard added to responses().

bor0 (commented):
forgot to mention that I'm approving and merging :)

PR #1849, opened 2026-08-04
Groups: add global sponsors management

### LINKED ISSUE #1784: Global sponsors management
Manage sponsors (logo, description, link) shown consistently across all group sites.

No sponsor-related code exists anywhere in PR #1699 or the current `wporg-groups-frontend`/`groups-site` diff.

## Scope and the key architectural question

"Across all group sites" is the part that needs deciding first: this multisite network isolates data per-site by design (each group is its own site, with its own `wc_NN_posts` tables, per the pattern seen throughout this codebase — e.g. `Members_Controller` is always scoped to `get_current_blog_id()`). A sponsor shown consistently across every group site is **cross-site, network-level data** — nothing in the current architecture does this today, so this needs a genuinely new pattern, likely one of:

1. A sponsors CPT/dataset living on the network's main/root site, exposed via a REST endpoint other group sites fetch from (client-side or server-side) to render a shared sponsors block.
2. A `wc_global` or network-level table outside the per-site table prefix pattern.

Either way, this is a bigger architectural lift than most of the other issues here, since it's the first genuinely cross-site data requirement in this feature set.

## Scope (once the above is decided)

- Sponsor fields: logo (image), name, description, link.
- A block (matching the existing block pattern — `wporg/sponsors` alongside `wporg/event-rsvp`, `wporg/group-members`, etc.) that renders the shared sponsor list on any group site.
- Management UI: likely network-admin-only initially (a new tab in an network-level admin screen), separate from the per-group `group-settings` modal since sponsors aren't group-owned data.

## Acceptance criteria

- [ ] Architectural decision recorded: where does sponsor data live (network-level store vs. per-site with sync), and how do group sites read it.
- [ ] Sponsors added/edited in one place appear consistently across every group site without per-site duplication of the data.
- [ ] A group organiser cannot edit or remove sponsors themselves (network-admin-only, unless explicitly decided otherwise).
- [ ] Sponsor block renders correctly (logo/description/link) and gracefully handles zero sponsors (no broken empty state).

bor0:
Fixes #1784.

Adds a `wporg_sponsor` post type (logo, name, description, link) on the **events root site**, rendered by a new `wporg/sponsors` block on every group site via `switch_to_blog()` — one copy of the data, no per-site duplication. Network-admin-only: every capability maps to `manage_network`, and the admin UI is only registered on the store site.

The groups network root would have been the obvious home, but `sunrise-groups.php` redirects its `/group/files/…` uploads away, so logos 404 there. Full reasoning in the commit message.

## Testing

1. `wp --url=<events root> post create --post_type=wporg_sponsor --post_status=publish --post_title=Woo --post_excerpt="…"`, or add one via **Sponsors** in the events root wp-admin. Set a featured image (logo) and the Sponsor Website URL.
2. Visit any group site front page → the card lists sponsors, ordered by **Order** then title. The event single view shows the same list in the sidebar.
3. Add a 5th sponsor → **Show all** appears and expands the list in place.
4. Edit a sponsor and reload the group site → the change shows immediately (cache is invalidated on save/trash/delete).
5. As a group-site administrator (not a network admin), confirm there's no Sponsors menu and no way to edit one.
6. Trash all sponsors → the block disappears entirely instead of rendering an empty card.

Groups suite 65 green, full suite 454 green, `phpcs` clean.

## Screenshots

**Sponsors in wp-admin (events root)**
sponsors-admin-list

**Editing a sponsor — logo, description, link, order**
sponsors-admin-edit

**Group site front page**
sponsors-front-page

**…after "Show all"**
sponsors-front-page-expanded

**Event single view**
sponsors-event-sidebar

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
@@ -1,0 +1,432 @@
+<?php
+/**
+ * Global sponsors for the groups network.
+ *
+ * Sponsors are network-level data: the same list is shown on every group
+ * site, so it can't live in a per-site table the way events, venues and
+ * members do. The store is a `wporg_sponsor` post type on the events root
+ * site (`EVENTS_ROOT_BLOG_ID`), edited in wp-admin there by network admins,
+ * and read at render time by every group site via `switch_to_blog()`.
+ * Nothing is copied into the group sites, so there is exactly one copy of
+ * each sponsor to keep up to date (cf. the `multi-event-sponsors` plugin on
+ * central.wordcamp.org, which forks its posts out to each camp site and then
+ * has to push edits back over them).
+ *
+ * The events root rather than the groups network root, which would otherwise
+ * be the obvious home, because sponsor logos are attachments and have to be
+ * publicly fetchable: this network serves uploads through ms-files rewriting,
+ * and `sunrise-groups.php` bounces every non-admin request for the groups
+ * root site to the events root — including its own `/group/files/…` uploads,
+ * which 302 away instead of serving. `events.wordpress.org/files/…` serves
+ * normally, so the logos load with no rewrite or redirect changes needed.
+ *
+ * @package WordCamp\Groups\Frontend
+ */
+
+namespace WordCamp\Groups\Frontend\Sponsors;
+
+use WP_Post;
+
+defined( 'WPINC' ) || die();
+
+/**
+ * The sponsor post type. Registered on every group site (so `get_post_type_object()`
+ * resolves when the block reads the store site's posts) but only editable on the
+ * store site — see `register_post_type()` below.
+ */
+const POST_TYPE = 'wporg_sponsor';
+
+/**
+ * Post meta holding the sponsor's website URL.
+ */
+const URL_META_KEY = '_wporg_sponsor_url';
+
+/**
+ * Transient caching the render-ready sponsor list.
+ *
+ * Lives on the store site itself, not on the sites that read it: the store is
+ * on the events network while the group sites are on the groups network, so a
+ * *network*-scoped transient written when a sponsor is saved would never be
+ * the one a group site reads. Invalidated whenever a sponsor changes; the
+ * expiry is only a backstop for edits this plugin doesn't see (e.g. a direct
+ * database change, or an attachment being edited).
+ */
+const CACHE_KEY = 'wporg_groups_sponsors';
+
+/**
+ * How long the sponsor list stays cached, in seconds.
+ */
+const CACHE_TTL = HOUR_IN_SECONDS;
+
+/**
+ * Bootstrap the sponsors feature.
+ *
+ * Runs on both the groups network (where the block reads sponsors) and the
+ * events network (where they're edited — see
+ * `mu-plugins/events/wporg-groups-sponsors.php`). Deliberately called outside
+ * the main plugin's GatherPress guard: sponsors don't touch events, and the
+ * store site doesn't run GatherPress at all.
+ */
+function bootstrap(): void {
+	add_action( 'init', __NAMESPACE__ . '\register_post_type' );
+	add_action( 'add_meta_boxes', __NAMESPACE__ . '\add_meta_boxes' );
+	add_action( 'save_post_' . POST_TYPE, __NAMESPACE__ . '\save_sponsor_url' );
+
+	// Keep the cached list honest.
+	add_action( 'save_post_' . POST_TYPE, __NAMESPACE__ . '\flush_cache' );
+	add_action( 'deleted_post', __NAMESPACE__ . '\flush_cache_for_post', 10, 2 );
+	add_action( 'trashed_post', __NAMESPACE__ . '\flush_cache_for_post' );
+	add_action( 'untrashed_post', __NAMESPACE__ . '\flush_cache_for_post' );
+}
+
+/**
+ * The blog ID that stores the sponsors — see the file header for why it's
+ * this site.
+ *
+ * Returns 0 when the constant isn't defined (an environment that predates
+ * it), which callers treat as "no sponsors".
+ */
+function get_store_blog_id(): int {
+	return defined( 'EVENTS_ROOT_BLOG_ID' ) ? (int) EVENTS_ROOT_BLOG_ID : 0;
+}
+
+/**
+ * Run a callback in the store site's context.
+ *
+ * @param callable $callback Receives no arguments; its return value is passed through.
+ * @return mixed The callback's return value, or null when there's no store site.
+ */
+function with_store_site( callable $callback ) {
+	$store_blog_id = get_store_blog_id();
+
+	if ( ! $store_blog_id ) {
+		return null;
+	}
+
+	$switched = get_current_blog_id() !== $store_blog_id;
+
+	if ( $switched ) {
+		switch_to_blog( $store_blog_id );
+	}
+
+	try {
+		return $callback();
+	} finally {
+		if ( $switched ) {
+			restore_current_blog();
+		}
+	}
+}
+
+/**
+ * Register the sponsor post type.
+ *
+ * The admin UI is only exposed on the store site, so a group organiser never
+ * sees the post type on their own site, and every capability maps to
+ * `manage_network` so only network admins can add, edit or delete sponsors
+ * even if they somehow reach the screen.
+ */
+function register_post_type(): void {
+	$is_store = get_current_blog_id() === get_store_blog_id();
+
+	// Every primitive capability the post type can ask for, all mapped to
+	// `manage_network`. `map_meta_cap` then resolves `edit_post`/`delete_post`
+	// through these, so there's no route to editing a sponsor without being a
+	// network admin.
+	$network_admin_only = array_fill_keys(
+		array(
+			'edit_post',
+			'read_post',
+			'delete_post',
+			'edit_posts',
+			'edit_others_posts',
+			'edit_published_posts',
+			'edit_private_posts',
+			'publish_posts',
+			'read_private_posts',
+			'delete_posts',
+			'delete_others_posts',
+			'delete_published_posts',
+			'delete_private_posts',
+			'create_posts',
+		),
+		'manage_network'
+	);
+
+	\register_post_type(
+		POST_TYPE,
+		array(
+			'labels'          => array(
+				'name'               => __( 'Sponsors', 'wporg-groups-frontend' ),
+				'singular_name'      => __( 'Sponsor', 'wporg-groups-frontend' ),
+				'add_new_item'       => __( 'Add New Sponsor', 'wporg-groups-frontend' ),
+				'edit_item'          => __( 'Edit Sponsor', 'wporg-groups-frontend' ),
+				'new_item'           => __( 'New Sponsor', 'wporg-groups-frontend' ),
+				'view_item'          => __( 'View Sponsor', 'wporg-groups-frontend' ),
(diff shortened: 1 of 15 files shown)


public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
```diff
@@ -0,0 +1,432 @@
...
+	// here means "this request isn't the meta box" rather than "clear it".
+	if ( ! isset( $_POST['wporg_sponsor_url'] ) ) {
+		return;
+	}
+
+	if ( ! isset( $_POST['wporg_sponsor_url_nonce'] )
+		|| ! wp_verify_nonce( sanitize_key( $_POST['wporg_sponsor_url_nonce'] ), 'wporg_sponsor_url' )
+	) {
+		return;
+	}
```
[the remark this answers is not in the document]
bor0: Changed in bcc43e19, though the stated reason isn't right: `wp_create_nonce()` returns `substr( wp_hash( … ), -12, 10 )`, and `wp_hash()` is `hash_hmac()` — lowercase hex — so `sanitize_key()`'s lowercasing was a no-op and saving the field worked (verified through wp-admin before and after).

Still worth changing: depending on core's hashing algorithm producing lowercase output is an implementation detail. Now uses `sanitize_text_field( wp_unslash() )`, matching the rest of the repo.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
```diff
@@ -0,0 +1,438 @@
...
+
+	$url = sanitize_url( wp_unslash( $_POST['wporg_sponsor_url'] ) );
+
+	if ( '' === $url ) {
+		delete_post_meta( $post_id, URL_META_KEY );
+	} else {
+		update_post_meta( $post_id, URL_META_KEY, $url );
+	}
```
gedex: `sanitize_url()` returns `''` for anything it can't parse, and this code treats invalid input identically to empty — `delete_post_meta()` silently wipes the saved URL. Verified live: `htps://sponsor.example` (typo), `javascript:alert(1)`, and `   ` all sanitise to `''` and destroy the existing meta with no notice. Suggest deleting only when the *raw* input was empty; otherwise surface an admin notice and keep the meta.

gedex (commented):
Blockers:
- `sanitize_url()` silently destroys a saved URL on any typo (verified live).
- Empty `get_posts()` result cached for an hour — sponsors vanish network-wide on any transient read failure with no self-healing.
- Missing tests for `save_sponsor_url()` (the just-fixed function) and `correct_switched_upload_url()` (the whole logo-loading path).

Non-blocking:
- The meta `auth_callback` isn't reachable today (parent-post cap gate blocks plain admins with `403`, verified live), but a test would guard against future cap-map changes.
- Oddity: as super-admin, `POST /wp/v2/wporg_sponsor/<id>` with `meta._wporg_sponsor_url` returns `200` but doesn't apply the write, despite `show_in_rest => true` on both. Doesn't affect this PR — writes go through the classic metabox — but worth a look.
- Cover permanent-delete cache invalidation (`wp_delete_post( id, true )`) and default-limit "Show all" behaviour.
- Exercise render-time escaping with a hostile title/URL.

Housekeeping: the workflow "downgrade" in `git diff production..HEAD` is a rebase artifact — three-dot diff is empty. A rebase would tidy the PR view.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
```diff
@@ -0,0 +1,438 @@
...
+					),
+					'suppress_filters' => false,
+				)
+			);
+
+			$sponsors = array_map( __NAMESPACE__ . '\prepare_sponsor', $posts );
+
+			set_transient( CACHE_KEY, $sponsors, CACHE_TTL );
+
+			return $sponsors;
```
gedex, line 432: Empty `get_posts()` results get cached for an hour. Any transient failure (DB blip, misbehaving `pre_get_posts` filter, `EVENTS_ROOT_BLOG_ID` pointing at a missing blog) freezes sponsors as invisible network-wide until either the transient expires or an admin re-saves a sponsor. Suggest `if ( ! empty( $sponsors ) ) { set_transient(...); }`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	$url = sanitize_url( wp_unslash( $_POST['wporg_sponsor_url'] ) );
+
+	if ( '' === $url ) {
+		delete_post_meta( $post_id, URL_META_KEY );
+	} else {
+		update_post_meta( $post_id, URL_META_KEY, $url );
+	}
+}
```
gedex, line 363: `save_sponsor_url()` has no test — nonce, cap, autosave-guard, empty-URL-deletes-meta are all unexercised. Commit bcc43e19 fixed exactly this function; a regression would ship green. Please add coverage.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	$uploads = wp_get_upload_dir();
+
+	return str_replace(
+		untrailingslashit( $uploads['baseurl'] ),
+		trailingslashit( get_option( 'siteurl' ) ) . 'files',
+		$url
+	);
+}
```
gedex, line 512: `correct_switched_upload_url()` — the entire reason logos load under ms-files — has no test. `test_block_renders_sponsor_details` only checks name/description. A refactor that drops the call would 404 every sponsor logo in production while every test stays green.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
gedex:
```suggestion
	$raw = wp_unslash( $_POST['wporg_sponsor_url'] );
	$url = sanitize_url( $raw );

	if ( '' === trim( (string) $raw ) ) {
		delete_post_meta( $post_id, URL_META_KEY );
	} elseif ( '' !== $url ) {
		update_post_meta( $post_id, URL_META_KEY, $url );
	}
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
gedex:
```diff
-			set_transient( CACHE_KEY, $sponsors, CACHE_TTL );
+			if ( ! empty( $sponsors ) ) {
+				set_transient( CACHE_KEY, $sponsors, CACHE_TTL );
+			}
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
bor0: Confirmed and fixed in 61482edb — good catch, this was real data loss.

Reproduced through wp-admin before fixing: typing `htps://woocommerce.com/typo` over Woo's working link deleted the meta outright, and the sponsor then rendered unlinked on the group site front end.

Took the suggestion but kept the notice from your original comment rather than the code snippet, since a silent no-op still leaves the editor thinking their edit saved. Empty field clears as before; unparseable input keeps the stored URL and raises a dismissible warning naming the rejected value. Verified all three paths in wp-admin, and `htps://…` / `javascript:…` are now covered by tests.

sponsor-invalid-url-notice

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
bor0: Partly taken, in 61482edb — but not the suggestion as written, because never caching empties is a worse trade than the one it avoids.

"No sponsors yet" is the normal state of this feature until network admins add some, and it stays that way for every group site in the meantime. `if ( ! empty( $sponsors ) )` means that common path does a `switch_to_blog()` plus a `WP_Query` on *every page view of every group site*, indefinitely — effectively no caching at all for the case that's most frequent.

The underlying concern is fair though: an empty result is exactly what your failure modes produce, and it's indistinguishable from a genuine empty here. So empties are still cached, but for 5 minutes instead of an hour (`EMPTY_CACHE_TTL`). Bad reads self-heal quickly, the common path stays cached, and an admin re-saving still busts it immediately.

Happy to go to your version if you'd rather not carry the extra constant — the query is cheap in isolation, it's the frequency I didn't want to sign up for.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
bor0: Fair, and it found the bug in the sibling thread. Added in 61482edb:

- `test_meta_box_saves_a_valid_url`
- `test_meta_box_clears_the_url_when_the_field_is_emptied`
- `test_meta_box_keeps_the_url_when_input_is_unusable` (data provider: mistyped scheme, `javascript:`)
- `test_meta_box_ignores_a_bad_nonce`
- `test_meta_box_ignores_a_user_without_the_capability`

One gap I left deliberately: the `DOING_AUTOSAVE` guard. `define()` is process-global and PHPUnit shares the process, so a test that defines it would make every subsequent test see an autosave. `@runInSeparateProcess` re-runs the WP bootstrap and is slow and fragile here. It's a one-line `defined()` check, so I'd rather leave it uncovered than poison the suite — open to a better idea if you have one.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/sponsors.php
bor0: Agreed, and it turned out to be testable after all — I'd assumed the fixture couldn't reproduce ms-files, so I probed it instead of guessing: the test network has `ms_files_rewriting` on, `UPLOADS` defined, and every test runs switched. The real condition, exactly.

So 61482edb adds `test_sponsor_logo_url_is_loadable_from_a_group_site`, which attaches a logo on the store site and asserts the URL a group site renders. Verified it's a real guard by deleting the `correct_switched_upload_url()` call and re-running:

```
-'http://events.wordpress.test/files/2026/08/logo.png'
+'http://events.wordpress.test/wp-content/blogs.dir/47/files/2026/08/logo.png'
```

That's the URL that 404s in production, so the scenario you described now fails loudly. Also split `rebase_url()` out of `correct_switched_upload_url()` and covered it directly (trailing slashes, URLs outside the base, empty base).

gedex (commented):
All four blockers fixed. Live-verified on the running site.

- **`sanitize_url()` destructive delete** — new flow keeps existing meta on unparseable input and raises a dismissible admin notice. Verified: `htps://…` and `javascript:…` preserve the saved URL + set the notice; empty/whitespace clears; bad nonce is no-op.
- **Empty-cache poisoning** — TTL split. Verified live: empty cached 300s, non-empty 3600s. Reasonable tradeoff given fresh-network cost.
- **`save_sponsor_url()` tests** — added valid/empty/typo/js/bad-nonce/no-cap coverage.
- **`correct_switched_upload_url()` tests** — split into `rebase_url()` (4-case dataProvider) + an end-to-end logo-URL assertion that fails without the fix.

Non-blockers unchanged: REST `auth_callback` still untested (defence-in-depth, unreachable today), permanent-delete cache invalidation, default-limit "Show all", hostile-title escaping. Workflow "downgrade" remains a rebase artifact (three-dot diff empty).

**Good to merge.**

PR #1852, opened 2026-08-04
Groups: Improve accessibility and focus management in event RSVP block

### LINKED ISSUE #1812: Groups: front-end blocks miss focus management and use non-native controls (a11y)

raicem:
Fixes #1812

### Changes Proposed
- **Focus Management:** Added focus trap, automatic focus placement on open, and focus restoration to the trigger button on close for the RSVP modal (`view.js`).
- **Native Control:** Replaced `<div role="button" tabindex="0">` on the attendee summary with a native `<button type="button">` (`render.php`).
- **Decorative Avatars:** Marked avatar image stack as decorative (`alt=""`, `aria-hidden="true"`) to prevent verbose avatar image descriptions from polluting the button name (`render.php`).
- **Screen Reader Announcements:** Added a live region (`role="status"`, `aria-live="polite"`) to announce RSVP status changes and errors to assistive tech (`render.php`, `view.js`).
- **Tests:** Added unit and E2E accessibility tests for the RSVP block.

### Testing instructions

1. Build the Groups frontend and open the future test event.
2. Tab to **View attendees**. Confirm its name excludes avatar names.
3. Press Enter or Space. Focus should enter the dialog and remain trapped inside.
4. Close with Escape, Close, and the backdrop. Focus should return to the trigger.
5. Log in and RSVP, then cancel. Confirm the count, attendee list, button label, and screen-reader announcements update correctly.


public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/style.css
```diff
@@ -15,6 +22,11 @@
 	opacity: 0.85;
 }
 
+.wporg-event-rsvp__summary:focus-visible {
+	outline: 2px solid currentColor;
```
bor0: It was fun to dig into this on the pair call, apparently it's the same as

```suggestion
	outline: 2px solid;
```

But GPT claims explicitness improves readability 🤷

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/view.js
```diff
@@ -3,38 +3,43 @@
  *
  * Handles the avatar-stack click → full-screen modal, RSVP toggling,
  * and attendee list display.
- *
- * @package WordCamp\Groups\Frontend
  */
 
-import { store, getContext, getElement } from '@wordpress/interactivity';
+// WordPress provides this as a script-module dependency.
+// eslint-disable-next-line import/no-extraneous-dependencies
```
bor0: Maybe we can find a way to remove this `eslint-disable`

bor0 (approved):
only a few non-blocking comments, good to merge

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/view.js
```diff
@@ -229,20 +299,31 @@ async function sendRsvp( ctx, status, retry = false ) {
...
-			ctx.apiBase + '/rsvp-responses?post_id=' + ctx.postId
-		);
+		const resp = await fetch( ctx.apiBase + '/rsvp-responses?post_id=' + ctx.postId );
 		const data = await resp.json();
 
 		if ( data.success && data.data?.attending?.records ) {
-			const list = document.querySelector(
-				'.wporg-event-rsvp__attendee-list'
-			);
+			const list = document.querySelector( '.wporg-event-rsvp__attendee-list' );
```
gedex: `refreshAttendees` uses `document.querySelector` here and on L355 (`__avatars`), while `openRsvpModal` correctly scopes via `.closest('.wp-block-wporg-event-rsvp')`. With this PR formalizing multi-block support, an RSVP in block B rewrites block A's list/avatars. Scope to the current block. *(Pre-existing, but relevant now.)*

gedex (commented):
No high-severity blockers. Native `<button>` + decorative avatars + `aria-live` live region are correct. Focus trap, `activeModalState` two-modal handoff, and `previousBodyOverflow` restore all check out. Error paths tighten cleanly (throw on `!ok || !success`, optimistic rollback preserved, 403 nonce retry intact). Tests cover the key a11y flows.

Only the `refreshAttendees` scoping is worth addressing — see inline.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/view.js
raicem: Addressed in 788664d4e98052b6b57949a34f915edac0ac7786.

Issue #1780, opened 2026-07-21
Export event and RSVP history (CSV/JSON)


bor0:
Let organizers and network admins export event and RSVP history, per group and network-wide.

No export code exists today — confirmed nothing in `class-members-controller.php` or `rest.php` implements export of any kind. Related: #1779 (member list export, same "not now" bucket) covers the member side; this covers events/RSVPs.

## Scope

- **Per-group export** (Organiser/editor tier, matching `current_user_can_manage_group_settings()`): all of that group's events, with RSVP counts/status per event.
- **Network-wide export** (network admin only): same data across every group, for the network activity picture this feeds into #1781.
- **Formats:** CSV (for spreadsheet use) and JSON (for programmatic use), consistent with the "(CSV/JSON)" already in scope for #1779.
- **Event fields:** title, date/time, venue (name — physical or online per #1808), RSVP counts by status (attending/waiting_list/not_attending, matching GatherPress's own segment names), organiser.
- **RSVP fields:** attendee (WordPress user for logged-in RSVPs), status, RSVP timestamp. For anonymous RSVPs (if #1778 keeps that path open), export a non-identifying token rather than raw email, unless GDPR/consent basis for full export is explicitly confirmed.

## Open questions to resolve before implementation

- Does a network-wide export include member/attendee emails, or aggregate counts only? This has real privacy implications and should be decided alongside #1774 (deliverability/GDPR) rather than assumed.
- Should exports be synchronous (small groups) or generated async with a download-when-ready notification (large/network-wide exports)?

## Acceptance criteria

- [ ] Organiser-tier export produces correct CSV and JSON for that group's events + RSVP breakdown, scoped only to that group's site.
- [ ] Network-wide export (admin only) aggregates correctly across every group with no cross-group data leakage into the wrong export.
- [ ] Export access is enforced server-side (capability-checked), not just a hidden UI control.
- [ ] Privacy question above is explicitly answered and reflected in what the export actually includes.
- [ ] Large exports (network-wide, many groups/events) don't time out the request — confirmed against a realistic data volume, not just the current handful of test events.

gedex:
My plan is to split this into two PRs.

* PR 1 — per-group export: organizer-tier REST endpoint returning CSV/JSON of the group's events + RSVP breakdown, synchronous download, with an Export option in the group settings UI. Recurring events export per-occurrence rows.

[...]

PR #1853, opened 2026-08-04
Groups: custom registration questions per event

### LINKED ISSUE #1813: Groups: custom registration fields per event
Organizer perspective, in case it is useful for scoping the MVP. I run [WordUp Gdynia](https://wpgdynia.pl), a WordPress meetup plus a yearly conference in Poland, and I do not see this in the backlog.

### The problem

Our stack today is not "Meetup.com" as a single thing. It is four things running at once:

- Meetup.com for discovery and the event listing
- [Evenea](https://evenea.pl) (a Polish ticketing platform) for the actual registration form
- our own WordPress site for the conference registration section
- Sender.net for the newsletter

That split is the part worth flagging. A lot of groups have already routed around Meetup's own RSVP, because a yes/no RSVP does not collect what an organizer needs. So the bar for this migration is not "match Meetup RSVP", it is "match what organizers already assembled around it". If it is not met, organizers keep the external form.

And this is the failure mode I would like to avoid: **the moment an organizer needs one extra question at signup, they move registration to an external form, and from then on the RSVP data on the group site is wrong.** Not incomplete, wrong. The group page shows 12 going while the real list lives in Evenea with 40. Every downstream feature then inherits that: the attendee email in #1772, the export in #1780, the metrics in #1781, the reminders in #1773.

### What we actually collect at signup

Concrete, from our own events:

- company or organization name, for name badges
- dietary requirements, for catering counts
- t-shirt size, for conference editions
- invoice details (company name, tax ID, address), a legal requirement here for anything paid
- a per purpose GDPR consent checkbox, event attendance separate from newsletter signup
- occasionally a free-text question, for example "what would you like covered in the Q&A"

None of these are exotic. Most meetup organizers I know collect at least one of them.

### Suggested scope

A per event set of custom registration fields, defined by the organizer when creating or editing the event, collected at RSVP time, and visible in the attendee list and the export.

Even a deliberately small version would remove most of the reason to leave the site:

- field types: text, textarea, select, checkbox
- per field: label, required flag, and whether it is visible to other attendees or organizers only
- values stored against the RSVP, included in #1780's export
- a sensible cap on field count, so this does not turn into a form builder

Group level defaults (fields that apply to every event of a group, so an organizer sets "dietary requirements" once) would be a natural follow up, but are not needed first.

### Related, separate question

Worth raising although it probably deserves its own issue: **paid events**. Free meetups and paid conferences are different products. GatherPress RSVP covers the first well. The second needs ticketing, invoices and local payment methods (here that means BLIK and Przelewy24), and WordCamp.org already has CampTix in this same codebase doing exactly that. If group sites and WordCamp sites end up with two unrelated registration stacks, organizers who run both (fairly common) will keep an external tool regardless, and the Meetup.com saving just moves to a different vendor. I am happy to split that into its own issue if that is preferred, I did not want to bundle two decisions into one thread.

Happy to help test either with a real Polish group.

bor0:
Fixes #1813.

Organizers can add up to 5 text questions to an event. Attendees answer them at RSVP time; only people who can edit the event see the answers. Required questions block the RSVP, client- and server-side.

RSVP posts to a new `wporg-groups/v1` endpoint instead of GatherPress's, so the status and the answers are saved in one request. Questions are stored in post meta on the event, answers in comment meta on the RSVP comment — cancelling an RSVP drops them.

Not included: non-text field types, per-field visibility, group-level defaults, export (#1780).

## Testing instructions

1. As an organizer, open an event and click **Edit this event**. Under **Registration questions**, add "Dietary requirements" (tick Required) and "Company or organization". Save.

Screenshot 2026-08-04 at 20 14 43

2. Reopen the edit modal — both questions and the Required flag come back.
3. As a different member, open the event and click **RSVP**. The modal opens with both questions instead of RSVPing immediately.
4. Click **Attend** with the required field empty → "Please answer the required questions.", no RSVP recorded.

Screenshot 2026-08-04 at 20 15 54

5. Fill both in, click **Attend** → RSVP goes through, count increments, questions hide.
6. As the organizer, reload and open the attendee list → the answers show under the attendee's name.

Screenshot 2026-08-04 at 20 19 12

7. As another non-organizer member, open the attendee list → no answers visible.
8. As the attendee, **Cancel RSVP**, reload → the stored answers are gone.
9. Regression: on an event with no questions, RSVP is still one click with no modal.

Server-side guard (should return 400 `wporg_groups_missing_answers`):

```js
const n = await fetch('/wp-json/gatherpress/v1/event/nonce').then(r=>r.json());
await fetch('/wp-json/wporg-groups/v1/event/<EVENT_ID>/rsvp', {method:'POST', credentials:'same-origin',
  headers:{'Content-Type':'application/json','X-WP-Nonce':n.nonce},
  body: JSON.stringify({status:'attending', answers:{}})}).then(r=>r.json());
```

Unit tests: `phpunit --group=groups` (14 new tests).

## Screenshots

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
@@ -15,6 +15,12 @@
  *   POST /event/{id}
  *        Updates an existing gatherpress_event.
  *
+ *   POST /event/{id}/rsvp
+ *        RSVPs the current user to an event, together with their answers to
+ *        the event's custom registration questions. Wraps GatherPress's own
+ *        RSVP save so the answers and the RSVP are written in one request and
+ *        required questions are enforced server-side.
+ *
  *   GET  /group-info
  *   POST /group-info
  *        Reads and writes the group's name and description. Exists because
@@ -47,6 +53,11 @@ use function WordCamp\Groups\Frontend\Capabilities\current_user_can_manage_group
 use function WordCamp\Groups\Frontend\Defaults\extract_description_blocks;
 use function WordCamp\Groups\Frontend\Defaults\get_default_event_data;
 use function WordCamp\Groups\Frontend\Defaults\get_event_venue_post_id;
+use function WordCamp\Groups\Frontend\RSVP_Questions\get_missing_required;
+use function WordCamp\Groups\Frontend\RSVP_Questions\get_questions;
+use function WordCamp\Groups\Frontend\RSVP_Questions\sanitize_answers;
+use function WordCamp\Groups\Frontend\RSVP_Questions\save_answers;
+use function WordCamp\Groups\Frontend\RSVP_Questions\save_questions;
 
 const NAMESPACE_V1 = 'wporg-groups/v1';
 
@@ -106,6 +117,36 @@ function register_routes(): void {
 					},
 				),
 			) + event_args_schema(),
+		)
+	);
+
+	register_rest_route(
+		NAMESPACE_V1,
+		'/event/(?P<id>\d+)/rsvp',
+		array(
+			'methods'             => WP_REST_Server::CREATABLE,
+			'callback'            => __NAMESPACE__ . '\save_rsvp',
+			'permission_callback' => __NAMESPACE__ . '\rsvp_permissions_check',
+			'args'                => array(
+				'id'      => array(
+					'type'              => 'integer',
+					'required'          => true,
+					'sanitize_callback' => 'absint',
+				),
+				'status'  => array(
+					'type'              => 'string',
+					'required'          => true,
+					'sanitize_callback' => 'sanitize_key',
+					'validate_callback' => static function ( $param ) {
+						return in_array( $param, array( 'attending', 'not_attending' ), true );
+					},
+				),
+				'answers' => array(
+					'type'     => 'object',
+					'required' => false,
+					'default'  => array(),
+				),
+			),
 		)
 	);
 
@@ -333,6 +374,92 @@ function publish_existing_event_permissions_check( WP_REST_Request $request ): b
 }
 
 /**
+ * Capability check for the RSVP route — any logged-in visitor may RSVP to a
+ * published event, exactly as with GatherPress's own RSVP endpoint.
+ */
+function rsvp_permissions_check(): bool {
+	return is_user_logged_in();
+}
+
+/**
+ * POST /event/{id}/rsvp
+ *
+ * RSVP the current user, storing their answers to the event's custom
+ * registration questions on the same request.
+ *
+ * This exists rather than posting answers to GatherPress's RSVP endpoint and
+ * then storing them in a follow-up call, because a required question has to be
+ * able to *block* the RSVP. Splitting it in two would leave an RSVP recorded
+ * with the answers rejected — exactly the "the RSVP data is wrong" failure the
+ * feature is meant to prevent.
+ *
+ * @return WP_Error|WP_REST_Response
+ */
+function save_rsvp( WP_REST_Request $request ) {
+	$event_id = (int) $request->get_param( 'id' );
+	$status   = (string) $request->get_param( 'status' );
+	$post     = get_post( $event_id );
+
+	if ( ! $post || Event::POST_TYPE !== $post->post_type || 'publish' !== $post->post_status ) {
+		return new WP_Error( 'wporg_groups_invalid_event', 'Invalid event ID', array( 'status' => 404 ) );
+	}
+
+	$event = new Event( $event_id );
+
+	if ( ! $event->rsvp ) {
+		return new WP_Error( 'wporg_groups_rsvp_unavailable', 'RSVP is not available for this event.', array( 'status' => 400 ) );
+	}
+
+	if ( $event->has_event_past() ) {
+		return new WP_Error( 'wporg_groups_event_past', 'This event has already happened.', array( 'status' => 400 ) );
+	}
+
+	$questions = get_questions( $event_id );
+	$answers   = sanitize_answers( $questions, $request->get_param( 'answers' ) );
+
+	if ( 'attending' === $status ) {
+		$missing = get_missing_required( $questions, $answers );
+
+		if ( $missing ) {
+			return new WP_Error(
+				'wporg_groups_missing_answers',
+				sprintf(
+					/* translators: %s: comma-separated list of question labels. */
+					__( 'Please answer: %s', 'wporg-groups-frontend' ),
+					implode( ', ', $missing )
+				),
+				array( 'status' => 400 )
+			);
+		}
+	}
+
+	// Group sites are open to join, and RSVPing is the point at which someone
+	// becomes a member — same as GatherPress's own endpoint and our
+	// /members/join route.
+	$user_id = get_current_user_id();
+	if ( ! is_user_member_of_blog( $user_id ) ) {
+		add_user_to_blog( get_current_blog_id(), $user_id, 'subscriber' );
+	}
+
+	$record     = $event->rsvp->save( $user_id, $status );
+	$comment_id = (int) ( $record['comment_id'] ?? 0 );
+
+	if ( $comment_id ) {
+		// Answers belong to an attendance, not to a declined invitation — a
+		// cancelled RSVP drops them.
+		save_answers( $comment_id, 'attending' === $record['status'] ? $answers : array() );
+	}
+
+	return new WP_REST_Response(
+		array(
+			'success'   => (bool) $comment_id,
+			'status'    => $record['status'] ?? 'no_status',
+			'responses' => $event->rsvp->responses(),
+		)
+	);
+}
+
+/**
  * Whether the current user can create a GatherPress event on this site.
  */
 function current_user_can_create_event(): bool {
@@ -470,7 +597,36 @@ function event_args_schema(): array {
 			'default'           => 0,
 			'sanitize_callback' => 'absint',
 		),
+		// Custom registration questions. Deliberately has no `default` — an
+		// absent parameter means "leave the existing questions alone", which
+		// an empty-array default would turn into "delete them all".
+		'rsvp_questions'    => array(
+			'type'     => 'array',
+			'required' => false,
+			'items'    => array(
+				'type'       => 'object',
+				'properties' => array(
+					'id'       => array( 'type' => 'string' ),
+					'label'    => array( 'type' => 'string' ),
+					'required' => array( 'type' => 'boolean' ),
+				),
+			),
+		),
 	);
+}
+
+/**
+ * Write the event's custom registration questions, if the request carried any.
+ *
+ * @param int             $event_id Saved event post ID.
+ * @param WP_REST_Request $request  The create/update/draft request.
+ */
+function maybe_save_rsvp_questions( int $event_id, WP_REST_Request $request ): void {
+	$questions = $request->get_param( 'rsvp_questions' );
+
+	if ( is_array( $questions ) ) {
+		save_questions( $event_id, $questions );
+	}
 }
 
 /**
@@ -535,6 +691,7 @@ function get_event_form_data( WP_REST_Request $request ): WP_REST_Response {
 	// `undefined` checks.
 	$fields['featured_image_id']  = $fields['featured_image_id'] ?? 0;
 	$fields['featured_image_url'] = $fields['featured_image_url'] ?? '';
+	$fields['rsvp_questions']     = $is_editing ? get_questions( $event_id ) : array();
 
 	$venues = array_map(
 		static function ( $post ) {
@@ -687,6 +844,8 @@ function save_draft( WP_REST_Request $request ): WP_REST_Response {
 		set_post_thumbnail( $saved_id, $featured_image_id );
 	}
 
+	maybe_save_rsvp_questions( $saved_id, $request );
+
 	return new WP_REST_Response(
 		array(
 			'id'           => $saved_id,
@@ -829,6 +988,8 @@ function persist_event( int $event_id, WP_REST_Request $request ) {
 		// Explicit clear on edit.
 		delete_post_thumbnail( $saved_id );
 	}
+
+	maybe_save_rsvp_questions( $saved_id, $request );
 
 	return new WP_REST_Response(
 		array(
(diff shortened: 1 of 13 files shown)


public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/render.php
```diff
@@ -48,6 +55,13 @@
 	update_meta_cache( 'user', $attendee_user_ids );
 }
 
+// Custom registration questions. Answers hold things attendees wouldn't post
+// publicly (dietary needs, accessibility requirements), so they're only
+// rendered for organizers.
+$questions        = get_questions( $event_post_id );
+$can_view_answers = $questions && current_user_can_view_answers( $event_post_id );
+$own_answers      = ( $questions && $is_login ) ? get_user_answers( $event_post_id, get_current_user_id() ) : array();
```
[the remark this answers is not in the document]
bor0: 5c3f190

bor0:
Reassigning to @gedex so that Cem can tackle design and recording

gedex:
@bor0 can you resolve conflicts here?

gedex (commented):
Seven high-severity items inline, three themes:

- **Answers get silently destroyed** — the waiting-list downgrade (`rest.php:450`) and delete-on-empty in `save_answers` are both reachable in normal use, and both lose data behind a success response.
- **Required isn't required** — GatherPress's own RSVP route is still open, and an attendee who already RSVP'd has no way to answer a question added afterwards.
- **Errors are unrecoverable** — the bare `catch` in `doToggleRsvp` flattens every server rejection to one generic string, dead-ending both the `"0"` bug and the changed-questions race.

Test gaps to close alongside: no coverage for the `waiting_list` path, a repeat `attending` submission, or the `"0"` answer value.

Non-blocking: `add_user_to_blog()`'s return is discarded at `rest.php:440` (diverges from `/members/join`); and a failed `/event-form-data` load leaves the modal on `EMPTY_FORM` with Save enabled — real, but pre-existing and it would wipe description and venue too.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
```diff
@@ -332,6 +373,92 @@ function publish_existing_event_permissions_check( WP_REST_Request $request ): b
...
+		add_user_to_blog( get_current_blog_id(), $user_id, 'subscriber' );
+	}
+
+	$record     = $event->rsvp->save( $user_id, $status );
+	$comment_id = (int) ( $record['comment_id'] ?? 0 );
+
+	if ( $comment_id ) {
+		// Answers belong to an attendance, not to a declined invitation — a
+		// cancelled RSVP drops them.
+		save_answers( $comment_id, 'attending' === $record['status'] ? $answers : array() );
```
gedex: `Rsvp::save()` downgrades `attending` → `waiting_list` when the attendance cap is hit (`class-rsvp.php:370`), so this writes `array()` and deletes the answers. `check_waiting_list()` later promotes the user to attending through a path that never touches `ANSWERS_META` — attending with no answers, unrecoverably. Key off the requested status, not `$record['status']`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
```diff
@@ -332,6 +373,92 @@ function publish_existing_event_permissions_check( WP_REST_Request $request ): b
...
+
+	if ( $event->has_event_past() ) {
+		return new WP_Error( 'wporg_groups_event_past', 'This event has already happened.', array( 'status' => 400 ) );
+	}
+
+	$questions = get_questions( $event_id );
+	$answers   = sanitize_answers( $questions, $request->get_param( 'answers' ) );
+
+	if ( 'attending' === $status ) {
+		$missing = get_missing_required( $questions, $answers );
```
gedex, line 433: `POST /gatherpress/v1/event/rsvp` is still registered with `permission_callback = is_user_logged_in()` (`class-rest-api.php:206`), so any logged-in member can attend with zero answers — and `check_waiting_list()` promotions bypass this too. Gate that route via `rest_endpoints` (precedent: `mu-plugins/groups/gatherpress-groups-tweaks.php:463`), or move enforcement onto `Rsvp::save()`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
```diff
@@ -332,6 +373,92 @@ function publish_existing_event_permissions_check( WP_REST_Request $request ): b
...
+	$status   = (string) $request->get_param( 'status' );
+	$post     = get_post( $event_id );
+
+	if ( ! $post || Event::POST_TYPE !== $post->post_type || 'publish' !== $post->post_status ) {
+		return new WP_Error( 'wporg_groups_invalid_event', 'Invalid event ID', array( 'status' => 404 ) );
+	}
+
+	$event = new Event( $event_id );
+
+	if ( ! $event->rsvp ) {
```
gedex: `Event::__construct` always assigns `$this->rsvp` for a `gatherpress_event`, so this never fires. The real gate is `$event->rsvp->is_enabled()`, never called — with RSVP disabled per-event, `save()` returns `comment_id => 0` and this still returns HTTP 200 `success:false` with no `message`, so the client throws `Error("OK")` and the answers drop with nothing logged. Return a `WP_Error` when `$comment_id === 0`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rsvp-questions.php
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * @param array $questions Sanitized question list.
+ * @param array $answers   Sanitized answer map.
+ * @return string[] Labels of the missing answers, in question order.
+ */
+function get_missing_required( array $questions, array $answers ): array {
+	$missing = array();
+
+	foreach ( $questions as $question ) {
+		if ( $question['required'] && empty( $answers[ $question['id'] ] ) ) {
```
gedex: `empty("0") === true` in PHP but `'0'` is truthy in JS, so a required question answered `0` passes the client check and 400s on the server. Same at :225, where a `"0"` answer is never shown to organizers. Use `! isset( $answers[ $question['id'] ] ) || '' === $answers[ $question['id'] ]`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rsvp-questions.php
```diff
@@ -0,0 +1,267 @@
...
+}
+
+/**
+ * Store the answers on an RSVP, or clear them when there are none.
+ *
+ * @param int   $comment_id RSVP comment ID.
+ * @param array $answers    Sanitized answer map.
+ */
+function save_answers( int $comment_id, array $answers ): void {
+	if ( empty( $answers ) ) {
```
gedex: Reachable outside cancellation: a second tab rendered before the RSVP posts `answers: {}` on Attend, and with no required question nothing blocks it — the stored answers are wiped behind a 200 and a success notice.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/view.js
```diff
@@ -244,18 +276,41 @@ async function doToggleRsvp( ctx, actionElement ) {
...
+		// Organizers see the answers inline in the attendee list, which the
+		// client-side refresh can't rebuild — reload so their view stays
+		// complete.
+		if ( ctx.canViewAnswers ) {
+			window.location.reload();
+			return;
+		}
+
 		refreshAttendees( ctx, actionElement );
 	} catch {
```
gedex: `sendRsvp` throws `data.message` (e.g. "Please answer: Dietary requirements"); this replaces it with the generic `rsvpError`. Any rejection the client-side check didn't predict — the `"0"` case, or a question set changed since page render — dead-ends the user with "please try again", which never works. Surface `err.message` for `wporg_groups_*` errors.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/render.php
```diff
@@ -254,6 +287,42 @@ class="wporg-event-rsvp__modal-close"
 							}
 							?>
 						</p>
+
+						<?php if ( $questions ) : ?>
+							<?php $error_id = 'wporg-event-rsvp-questions-error-' . $event_post_id; ?>
+							<fieldset class="wporg-event-rsvp__questions" data-wp-class--is-hidden="state.isAttending">
```
gedex: Hiding the fieldset on `state.isAttending` leaves Cancel as the only control. A required question added to an event that already has attendees gets blanks from all of them with no way to comply; and correcting a typo means cancelling first, which deletes the answers (`rest.php:450`) and, on a capped event, hands the seat to the waiting list.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
bor0: Confirmed and fixed in 79f8a252 — keyed off the requested status now. `check_waiting_list()` promotes via `save( $userId, 'attending', … )` which never touches `ANSWERS_META`, so the deletion really was permanent. Test: `test_waiting_list_rsvp_keeps_its_answers`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rsvp-questions.php
bor0: Confirmed and fixed in 79f8a252. Worse than it first looks: `sanitize_answers()` stored `"0"` correctly (it only skipped `''`), so the answer was saved and *then* reported missing and hidden from organizers. Replaced both sites with an explicit `has_answer()`. Test: `test_zero_is_a_valid_answer`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
bor0: Confirmed both halves and fixed in 79f8a252. `Event::__construct` assigns `$this->rsvp` for anything supporting `gatherpress-event-date`, and the handler already 404s non-events, so the guard was unreachable — replaced with `is_enabled()`, plus a `WP_Error` when `$comment_id === 0`. Test: `test_rsvp_disabled_event_returns_an_error`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rsvp-questions.php
bor0: Confirmed and fixed in 79f8a252. Needed more than guarding the empty case: the client omitted blank inputs, so "cleared this field" and "never rendered it" were indistinguishable. Blanks are now sent explicitly and `merge_answers()` only touches submitted keys, so an omitted question keeps its stored answer and a submitted blank clears it. Tests: `test_omitted_questions_keep_their_stored_answers`, `test_submitted_blank_clears_the_answer`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/view.js
bor0: Fair — this was my call during the production merge, and I weighted it wrong. I avoided `err.message` to keep raw server text off the screen, but dead-ending on the "0" case is worse. Fixed in 79f8a252: the error carries its `code`, and messages are surfaced only for `wporg_groups_*`; everything else keeps the generic wording. Tests cover both branches.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/render.php
bor0: Agreed, fixed in 79f8a252. The fieldset no longer hides while attending and there's a Save answers button that re-sends `attending` (idempotent in `Rsvp::save()`), so no cancel round trip — which, as you note, would have deleted the answers and surrendered the seat on a capped event. Test: `saves edited answers without changing attendance`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
bor0: Confirmed — the claim was overstated. Fixed in 79f8a252 by gating `/gatherpress/v1/event/rsvp` through `rest_endpoints`, following the `gatherpress-groups-tweaks.php` precedent you pointed at: the permission callback is wrapped rather than replaced, and only the harmful case is refused (attending an event with unanswered required questions). Cancelling, events without questions, and users whose stored answers already satisfy them pass through. The `check_waiting_list()` half is moot once the waiting-list fix lands — promotion re-saves an RSVP that already validated. Tests: `test_gatherpress_route_is_guarded`, `test_gatherpress_route_guard_is_registered`.

wppoland:
Reviewed this from the accessibility side, since it lands in the same `event-rsvp` files #1852 just touched. Thanks for picking it up so quickly after I filed #1813.

**The #1852 work survives intact.** The native `<button type="button">`, the `role="status"` / `aria-live="polite"` notice, the decorative avatars and the focus handling are all still in place, and no `role="button"` div has crept back in. The new question markup is careful in its own right: `<label for>` correctly bound to each input, `required` paired with `aria-required`, the `*` marker properly `aria-hidden` so it is not read as "star", and a `fieldset` with a screen-reader-text legend.

One gap worth closing, and it is a small one.

### The field in error is never identified

When validation fails you set `ctx.rsvpNotice` as well as `ctx.questionsError`, and `rsvpNotice` is bound to the polite live region, so the message genuinely is announced. I checked that before writing this, because at first glance the error paragraph itself is not a live region and I assumed it was silent. It is not. That half is fine.

What is missing is the other half of WCAG 3.3.1, identifying *which* item is in error:

- nothing sets `aria-invalid`. `grep -c aria-invalid` returns 0 for both `render.php` and `view.js`
- every input carries the same `aria-describedby`, pointing at one shared generic paragraph, so all of them describe as "Please answer the required questions" whether or not that particular field is the empty one
- focus is not moved to the offending field

With up to five questions, a screen reader user is told that something is missing but not which one, and the shared description actively works against narrowing it down.

The information is already in hand at the moment of detection:

```js
block?.querySelectorAll( '.wporg-event-rsvp__question-input' ).forEach( ( input ) => {
	const value = input.value.trim();
	if ( ! value && input.required ) {
		missingRequired = true;   // the offending input is right here
	}
```

Returning those inputs alongside `missingRequired`, setting `aria-invalid="true"` on them and clearing it once answered, then calling the `scheduleFocus()` helper you already have on the first of them, would close this without introducing any new machinery.

Not a blocker from my side.

wppoland:
Sent the accessibility point from my earlier comment as a patch instead: #1864, based on this branch so it merges inside your PR rather than alongside it.

It returns the offending inputs from `collectAnswers()`, flags them `aria-invalid`, focuses the first, and clears the flag on each submit. Two tests added; both fail on this branch without the source change and pass with it. Take it or leave it, whichever is less work for you.

PR #1854, opened 2026-08-05
Add group archiving controls

### LINKED ISSUE #1783: Archive/close a group, retain historical data
Let network admins archive or close a group while preserving its events, RSVPs, and member records (hidden from active listings, not deleted).

No archive endpoint exists today in `wporg-groups-frontend` or `class-members-controller.php`.

## Scope

WordPress multisite already has a native "archived" state for sites (the `archived` column on `wp_blogs`, toggleable via `wp site archive`) — an archived site becomes inaccessible on the front end while its database/content stays fully intact. This issue is likely about **exposing that existing core capability through a front-end-friendly network-admin action**, rather than inventing a new archive mechanism from scratch:

- A network-admin-only action (in the network dashboard, not per-group) to archive/close a group.
- On archive: the group site becomes inaccessible per core's existing archived-site behavior; events/RSVPs/members remain in the database untouched.
- The group stops appearing in any group directory/listing (if one exists or is added — check against #1782's creation flow and whatever "browse all groups" surface exists).
- An explicit "reactivate" path back to active, for groups that were archived by mistake or want to resume.

## Open questions

- Should organizers themselves be able to request archival (subject to network-admin approval), or is this purely network-admin-initiated?
- Does archiving affect the emailing features (#1771/#1772) — should archived-group members stop receiving emails immediately?

## Acceptance criteria

- [ ] Archiving a group uses (or is clearly justified as not using) WordPress core's existing site-archive mechanism, rather than building a parallel one.
- [ ] After archiving: events, RSVPs, and member records are all still present and queryable in the database (verified by direct DB/WP-CLI check, not just "site returns 404").
- [ ] Archived groups are excluded from any active-groups listing/directory.
- [ ] A network admin can reactivate an archived group and the site becomes fully functional again with all prior data intact.
- [ ] Decision recorded on whether organizers can request archival themselves or it's purely network-admin-initiated.

gedex:
## What

- Add a Network Admin **Groups** screen for archiving and reactivating group sites.
- Use WordPress multisite's native archived-site flag so events, RSVPs, and memberships are preserved.
- Exclude archived groups from active group queries and add regression coverage for archive safety and validation.

## Why

Inactive groups need to be removed from public access without deleting their historical data. Fixes #1783.

## Testing

1. Sign in as a network administrator and open https://wordcamp.test/wp-admin/network/sites.php?page=wporg-groups.
2. Find **Sunshine Coast, QLD**, confirm its status is **Active**, click **Archive**, accept the confirmation, and verify the success notice and **Archived** status.
3. Open https://events.wordpress.test/group/sunshine-coast-qld/ and confirm the archived group is unavailable.
4. Return to the Groups screen, click **Reactivate**, and confirm the group loads again with its existing events, RSVPs, and members intact.
5. Run `docker compose -f docker-compose.phpunit.yml run --rm --entrypoint phpunit phpunit_wp --testsuite 'WordCamp Groups'`.

## Screenshots

**[Network Admin Groups](https://wordcamp.test/wp-admin/network/sites.php?page=wporg-groups) — list of active and archived states**

<kbd>image</kbd>

**[Sunshine Coast group](https://events.wordpress.test/group/sunshine-coast-qld/) — archived group**

<kbd>image</kbd>

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-archive.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-archive.php
@@ -89,7 +89,22 @@ function register_page(): void {
 }
 
 /**
+ * Whether the current user may archive or reactivate groups.
+ *
+ * Extracted so the capability is named in one place and can be asserted
+ * directly in tests, matching `Groups\Messaging\current_user_can_send_messages()`.
+ *
+ * @return bool
+ */
+function current_user_can_archive_groups(): bool {
+	return current_user_can( 'manage_sites' );
+}
+
+/**
  * Archive or reactivate a group using WordPress core's site status.
+ *
+ * Callers are responsible for the capability check; use
+ * `current_user_can_archive_groups()`. Both entry points in this file do.
  *
  * @param int  $site_id  Site ID to update.
  * @param bool $archived Whether the group should be archived.
@@ -127,7 +142,7 @@ function update_group_archive_status( int $site_id, bool $archived ) {
  * Process an archive/reactivate request from Network Admin.
  */
 function handle_update(): void {
-	if ( ! current_user_can( 'manage_sites' ) ) {
+	if ( ! current_user_can_archive_groups() ) {
 		wp_die(
 			esc_html__( 'Sorry, you are not allowed to manage groups.', 'wordcamporg' ),
 			'',
@@ -166,7 +181,7 @@ function handle_update(): void {
  * Render the Groups management screen.
  */
 function render_page(): void {
-	if ( ! current_user_can( 'manage_sites' ) ) {
+	if ( ! current_user_can_archive_groups() ) {
 		wp_die( esc_html__( 'Sorry, you are not allowed to manage groups.', 'wordcamporg' ) );
 	}
 
--- a/public_html/wp-content/mu-plugins/wporg-groups-archive.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-archive.php
@@ -18,6 +18,7 @@ defined( 'WPINC' ) || die();
 
 const PAGE_SLUG     = 'wporg-groups';
 const UPDATE_ACTION = 'wporg_groups_update_archive_status';
+const PER_PAGE      = 50;
 
 add_action( 'network_admin_menu', __NAMESPACE__ . '\\register_page' );
 add_action( 'network_admin_edit_' . UPDATE_ACTION, __NAMESPACE__ . '\\handle_update' );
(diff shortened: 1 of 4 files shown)


public_html/wp-content/mu-plugins/wporg-groups-archive.php
```diff
@@ -0,0 +1,223 @@
...
+function get_group_sites( bool $include_archived = false ): array {
+	$args = array(
+		'network_id'   => GROUPS_NETWORK_ID,
+		'number'       => 0,
+		'site__not_in' => array( GROUPS_ROOT_BLOG_ID ),
+		'deleted'      => 0,
+		'spam'         => 0,
+		'orderby'      => 'domain',
+		'order'        => 'ASC',
+	);
```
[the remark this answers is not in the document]
gedex: Good catch. Fixed in 0542606f: the Network Admin screen now fetches 50 groups per page with number/offset, renders pagination controls, and uses stable site-ID ordering. I also added a count helper and regression coverage for page slices and totals. The full WordCamp Groups suite passes (54 tests, 121 assertions).

public_html/wp-content/mu-plugins/wporg-groups-archive.php
```diff
@@ -0,0 +1,272 @@
...
+	return (int) get_sites( $args );
+}
+
+/**
+ * Register the Groups management screen in Network Admin.
+ */
+function register_page(): void {
+	add_submenu_page(
+		'sites.php',
+		__( 'Groups', 'wordcamporg' ),
```
bor0, line 83: Can we unify this with https://github.com/WordPress/wordcamp.org/pull/1848/changes#diff-5572b0281673eedc6f6049217c1902e821215c2be0dc12aa7cc1ba249f0bd40bR32

bor0: I see some other potential for unifying, happy to address after this gets merged. Created #1859

wppoland:
One hardening observation, offered as a suggestion rather than a defect, because the feature is correctly gated as shipped.

### The state-changing helper is unguarded, and nothing tests the denial

`update_group_archive_status()` in `mu-plugins/wporg-groups-archive.php` goes from its site-validity checks straight to `wp_update_site( $site_id, array( 'archived' => ... ) )` with no `current_user_can()` and no docblock telling callers they must do one.

To be explicit, since "no capability check" would be alarming if left there: the feature is not exposed. `handle_update()` checks `current_user_can( 'manage_sites' )` at line 130 before any write, and `check_admin_referer()` at 141, and the admin screen is gated again at 169. I verified that ordering at this PR's head rather than assuming it. So there is no way in today for a group administrator to archive a site.

What is missing is the second layer, and the test that proves it. The sibling feature in the same directory does both:

```php
// groups/network-messaging.php:109
function current_user_can_send_messages(): bool {
	return current_user_can( CAPABILITY );
}
```

```php
// groups/tests/test-network-messaging.php:625
public function test_group_administrators_cannot_send_messages() {
	...
	$this->assertFalse( Messaging\current_user_can_send_messages() );
```

and `group-site-provisioning.php` guards its own entry points inline the same way (lines 56 and 91).

All five tests in `groups/tests/test-group-archive.php` call `update_group_archive_status()` or the query helpers directly, and none calls `wp_set_current_user()` or asserts a refusal. So the property the whole feature rests on, that only network administrators can archive a group, is currently enforced in exactly one place and asserted in none. If a second caller is ever added, for example a REST route or a WP-CLI command, nothing fails.

Extracting a `current_user_can_archive_groups()` helper the way messaging does, calling it from both `handle_update()` and the helper, and adding one `assertFalse` test for a group administrator would close it cheaply and match the convention already established next door.

wppoland:
Sent the capability point from my earlier comment as a patch: gedex/wordcamp.org#1, based on this branch so it merges inside your PR.

One correction to what I suggested there. I proposed guarding `update_group_archive_status()` itself, then backed that out: `network-messaging.php` keeps its writer (`queue_message()`) unguarded and only exposes `current_user_can_send_messages()` as a named predicate, and a guard inside the writer would have broken your three existing tests, since neither `Groups_TestCase` nor `test-group-archive.php` establishes a current user. So the patch only names the capability, routes both entry points through it, and adds the two missing assertions.

I could not run the PHP suite locally (no `vendor/` in my checkout), so CI is the verification rather than my word.

bor0 (approved):
Tests well! 🎉

PR #1857, opened 2026-08-05
Groups: network-admin messaging for organisers and group segments

### LINKED ISSUE #1775: Network level: broadcast message to all group organizers
Let network admins/Program Managers send a message to every group organizer network-wide.

No existing tool covers this (not in GatherPress, not in wordcamp.org's own `wcpt` plugin) — new, network-admin-only feature.

**Scope:**
- Network Admin screen (capability: `manage_network` or similar), with a compose form.
- Recipients: all `editor`/`administrator` users across every site on the Groups network (`GROUPS_NETWORK_ID`), deduped by email.
- Send via a scheduled/batched job, not a blocking request.

**Acceptance criteria:**
- [ ] A Program-Manager-level user can compose and send a message from Network Admin.
- [ ] Every group organiser (editor/administrator role) across all Groups-network sites receives it, with no duplicates.
- [ ] Sending to a large number of recipients does not time out the request.
- [ ] Non-Program-Manager users cannot access the screen or trigger a send.

**Out of scope for v1:** opt-out preferences, delivery tracking/read receipts, rich-text composition.

Related: #1776 (same surface, segmented to specific groups — likely worth building together).

### LINKED ISSUE #1776: Network level: segmented messaging to specific groups
Let network admins message members of one or more selected groups, rather than broadcasting to everyone.

No existing tool covers this. Distinct from #1775 in two ways: audience is all members (any role), not organizers-only, and scope is a selected subset of groups, not all of them — needs a group-selection UI.

**Scope:**
- Network Admin screen (capability: `manage_network` or similar), with a group-selector (multi-select of Groups-network sites) and a compose form.
- Recipients: all users (any role) on the selected site(s) (`get_users(['blog_id' => $site_id])` per selected site), deduped by email.
- Send via a scheduled/batched job, not a blocking request.

**Acceptance criteria:**
- [ ] A Program-Manager-level user can select one or more groups and compose/send a message from Network Admin.
- [ ] Every member (any role) of each selected group receives it, with no duplicates across overlapping group selections.
- [ ] Sending to a large number of recipients does not time out the request.
- [ ] Non-Program-Manager users cannot access the screen or trigger a send.

**Out of scope for v1:** opt-out preferences, delivery tracking/read receipts, rich-text composition.

**Note:** shares its Network Admin screen shell, site-enumeration primitive, and send mechanism with #1775 — worth building both against the same underlying tool (audience filter: organizers-only vs. all members; group filter: all vs. selected) rather than as two separate admin screens.

bor0:
Adds a Network Admin screen (Settings → Message Groups) for emailing group people network-wide: organisers only or all members, all groups or a selected few. Sends run on cron in batches of 50, deduped by email.

Fixes #1775
Fixes #1776

## Testing

https://events.wordpress.test/group/wp-admin/network/settings.php?page=wporg-groups-messaging

1. Groups network → Network Admin → Settings → **Message Groups**.

screenshot-1785924173223-1

2. "Organisers only" + "All groups" → only editors/admins across every group site get it.
3. "All members" + one selected group → everyone on that site, nobody else.

screenshot-1785924205210-2

4. Nothing sends on submit; run the batch with `wp cron event run wporg_groups_message_batch --url=<groups-network-url>`.

screenshot-1785924269843-3

5. As a group organiser (not a network admin), the screen returns "Sorry, you are not allowed to access this page."
6. `phpunit --testsuite "WordCamp Groups"`

### DIFF
--- a/public_html/wp-content/mu-plugins/groups/network-messaging.php
+++ b/public_html/wp-content/mu-plugins/groups/network-messaging.php
@@ -374,63 +374,98 @@ function process_batch(): void {
 	}
 
 	$processed = 0;
+	$crashed   = false;
 
-	while ( $processed < BATCH_SIZE ) {
-		if ( empty( $job['queue'] ) ) {
-			if ( empty( $job['pending_sites'] ) ) {
-				break;
+	/*
+	 * The job was already removed from `JOBS_OPTION` by the claim above, and
+	 * isn't written back until the commit below -- so a `Throwable` escaping
+	 * this loop (a rogue `pre_wp_mail`/`wp_mail` filter, a transport that
+	 * exhausts the request instead of returning cleanly) must not skip that
+	 * commit, or the job -- and every recipient still in it -- disappears
+	 * with nothing sent, re-queued, or recorded. `send_message()` already
+	 * turns an ordinary transport failure into `false`, so this is only
+	 * reached by something that couldn't be handled that way.
+	 */
+	try {
+		while ( $processed < BATCH_SIZE ) {
+			if ( empty( $job['queue'] ) ) {
+				if ( empty( $job['pending_sites'] ) ) {
+					break;
+				}
+
+				++$processed;
+
+				$site_id = (int) $job['pending_sites'][0];
+				$batch   = get_site_recipients( $site_id, $job['audience'], BATCH_SIZE, $job['site_offset'] );
+
+				if ( empty( $batch ) ) {
+					array_shift( $job['pending_sites'] );
+					$job['site_offset'] = 0;
+				} else {
+					$job['site_offset'] += count( $batch );
+					$job['queue']        = $batch;
+				}
+
+				continue;
 			}
+
+			$recipient = array_shift( $job['queue'] );
+			$key       = strtolower( $recipient['email'] );
 
 			++$processed;
 
-			$site_id = (int) $job['pending_sites'][0];
-			$batch   = get_site_recipients( $site_id, $job['audience'], BATCH_SIZE, $job['site_offset'] );
-
-			if ( empty( $batch ) ) {
-				array_shift( $job['pending_sites'] );
-				$job['site_offset'] = 0;
-			} else {
-				$job['site_offset'] += count( $batch );
-				$job['queue']        = $batch;
+			if ( isset( $job['sent'][ $key ] ) ) {
+				continue;
 			}
 
-			continue;
+			/*
+			 * Only mark the address done once the mail is actually away. Marking
+			 * first meant a transient transport failure retired the recipient for
+			 * good — no send, no retry, no trace. Left unmarked, the same person
+			 * gets another chance if they turn up again via another group, and
+			 * the failure is at least on the record either way.
+			 */
+			if ( send_message( $recipient, $job ) ) {
+				$job['sent'][ $key ] = true;
+
+				++$job['sent_count'];
+			} else {
+				$job['failed_count'] = ( $job['failed_count'] ?? 0 ) + 1;
+
+				Logger\log(
+					'groups_message_send_failed',
+					array(
+						'job'       => $job['id'],
+						'recipient' => $recipient['email'],
+					)
+				);
+			}
 		}
-
-		$recipient = array_shift( $job['queue'] );
-		$key       = strtolower( $recipient['email'] );
-
-		++$processed;
-
-		if ( isset( $job['sent'][ $key ] ) ) {
-			continue;
-		}
+	} catch ( \Throwable $error ) {
+		$crashed = true;
 
 		/*
-		 * Only mark the address done once the mail is actually away. Marking
-		 * first meant a transient transport failure retired the recipient for
-		 * good — no send, no retry, no trace. Left unmarked, the same person
-		 * gets another chance if they turn up again via another group, and
-		 * the failure is at least on the record either way.
+		 * Not just `compact( 'error' )`: `redact_keys()` only special-cases
+		 * `Exception`, not `Throwable` generally, so a plain `Error` (e.g. a
+		 * `TypeError`) would fall through to an `(array)` cast instead --
+		 * mangled keys for its private/protected properties that `wp_json_encode()`
+		 * silently drops, per that function's own docblock.
 		 */
-		if ( send_message( $recipient, $job ) ) {
-			$job['sent'][ $key ] = true;
-
-			++$job['sent_count'];
-		} else {
-			$job['failed_count'] = ( $job['failed_count'] ?? 0 ) + 1;
-
-			Logger\log(
-				'groups_message_send_failed',
-				array(
-					'job'       => $job['id'],
-					'recipient' => $recipient['email'],
-				)
-			);
-		}
+		Logger\log(
+			'groups_message_batch_crashed',
+			array(
+				'job'     => $job['id'],
+				'error'   => array(
+					'class'   => get_class( $error ),
+					'message' => $error->getMessage(),
+					'file'    => $error->getFile(),
+					'line'    => $error->getLine(),
+				),
+			)
+		);
 	}
 
-	$finished = empty( $job['queue'] ) && empty( $job['pending_sites'] );
+	$finished = ! $crashed && empty( $job['queue'] ) && empty( $job['pending_sites'] );
 
 	$jobs = with_jobs_lock(
 		function () use ( $job, $finished ) {
--- a/public_html/wp-content/mu-plugins/groups/network-messaging.php
+++ b/public_html/wp-content/mu-plugins/groups/network-messaging.php
@@ -242,7 +242,15 @@ function schedule_next_batch(): void {
 }
 
 /**
- * Send up to `BATCH_SIZE` emails from the oldest queued job.
+ * Do up to `BATCH_SIZE` units of work on the oldest queued job.
+ *
+ * A unit is one recipient taken off the queue or one per-site recipient
+ * lookup — not one email sent. Bounding on emails instead would leave a run
+ * unbounded whenever the work doesn't produce mail: someone who organises a
+ * dozen groups is skipped as a duplicate, and a group with no members yields
+ * nothing, so a run could walk any number of sites and queries before hitting
+ * a send-based limit. Counting the work itself is what keeps a single cron
+ * run inside the time limit.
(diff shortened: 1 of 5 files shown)


public_html/wp-content/mu-plugins/groups/network-messaging.php
```diff
@@ -0,0 +1,579 @@
...
+	);
+}
+
+/**
+ * Handle the compose form submission.
+ */
+function handle_form_post(): void {
+	if ( ! current_user_can_send_messages() ) {
+		wp_die( 'You do not have permission to send messages to groups.', 403 );
+	}
```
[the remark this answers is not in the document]
bor0: Not quite — `wp_die()` handles this. When the second argument is an integer, core promotes it to the response code (`wp-includes/functions.php`):

```php
} elseif ( is_int( $title ) ) {
	$args  = array( 'response' => $title );
	$title = '';
}
```

So both call sites already return a 403. Leaving them as-is.

public_html/wp-content/mu-plugins/groups/network-messaging.php
```diff
@@ -0,0 +1,579 @@
...
+		}
+
+		$job['sent'][ $key ] = true;
+
+		if ( send_message( $recipient, $job ) ) {
+			++$job['sent_count'];
+		}
+
+		++$sent;
+	}
```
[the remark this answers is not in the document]
bor0: Good catch, fixed in 015c2a2d.

The same hole applied to a group with no members: the loop shifted a site off `pending_sites` and ran a `get_users()` query without counting that either, so a run could walk the whole network before reaching a send-based limit.

The counter now increments for each dequeued recipient and each per-site lookup, so a run does at most `BATCH_SIZE` units of work whether or not it produces mail. Added a test that seeds a queue of 60 already-sent duplicates and expects 0 emails with 10 left queued — it fails against the old logic (the job drained in one run).

public_html/wp-content/mu-plugins/groups/network-messaging.php
```diff
@@ -0,0 +1,589 @@
...
+
+		if ( isset( $job['sent'][ $key ] ) ) {
+			continue;
+		}
+
+		$job['sent'][ $key ] = true;
+
+		if ( send_message( $recipient, $job ) ) {
+			++$job['sent_count'];
+		}
```
gedex: `$job['sent'][ $key ] = true;` runs before `send_message()`, so any `wp_mail()` failure marks the address as sent with no log and no retry. Move the mark inside `if ( send_message(...) )`.

public_html/wp-content/mu-plugins/groups/network-messaging.php
```diff
@@ -0,0 +1,589 @@
...
+ * Email a single recipient.
+ *
+ * @param array{email: string, name: string} $recipient Recipient.
+ * @param array                              $job       Job being processed.
+ * @return bool Whether the mail was handed off successfully.
+ */
+function send_message( array $recipient, array $job ): bool {
+	$to = $recipient['name']
+		? sprintf( '%s <%s>', $recipient['name'], $recipient['email'] )
+		: $recipient['email'];
```
gedex, line 504: wp_mail splits `$to` on commas, so a display_name like `attacker@evil.test, Jane` adds an unauthorised recipient. Pass `$to` as an array to skip the split:

```php
wp_mail( array( $to ), ... );
```

public_html/wp-content/mu-plugins/groups/network-messaging.php
```diff
@@ -0,0 +1,589 @@
...
+		'sent'          => array(),
+		'sent_count'    => 0,
+		'author'        => get_current_user_id(),
+		'created'       => time(),
+	);
+
+	$jobs   = get_jobs();
+	$jobs[] = $job;
+
+	update_site_option( JOBS_OPTION, $jobs );
```
gedex: `queue_message()` and `process_batch()` both do unlocked read-modify-write on the same option. Cron reads `[A]`, admin writes `[A, B]`, cron writes back `[]` — B is lost. `doing_cron` only covers cron/cron. Guard with a `wp_cache_add()` lock or use per-job keys.

gedex (commented):
Nice work overall — the batch bounding, cap/nonce checks, escaping, and test coverage are all in good shape. A few things worth sorting before this merges, though:

1. **Silent send loss (line 302):** dedupe is marked before `wp_mail()` runs, so a transient SMTP failure permanently drops the recipient with no log. Moving the mark inside the success branch should do it.

2. **Address injection (line 328):** wp_mail splits `Name <email>` on commas, so a crafted display_name can smuggle extra recipients into a broadcast. Passing `$to` as an array skips the split.

3. **JOBS_OPTION race (208, 313):** admin-post and cron both RMW the same option without a lock — an overlapping submit during a batch can silently drop a queued job. A short `wp_cache_add()` lock (or per-job keys) would cover it.

Happy to take another look once these are in.

public_html/wp-content/mu-plugins/groups/network-messaging.php
gedex:
```diff
-		$job['sent'][ $key ] = true;
-
 		if ( send_message( $recipient, $job ) ) {
+			$job['sent'][ $key ] = true;
 			++$job['sent_count'];
 		}
```

public_html/wp-content/mu-plugins/groups/network-messaging.php
gedex:
```diff
 	return wp_mail(
-		$to,
+		array( $to ),
 		$job['subject'],
 		$job['body'],
 		array( 'Content-Type: text/plain; charset=UTF-8' )
 	);
```

public_html/wp-content/mu-plugins/groups/network-messaging.php
bor0: Fixed in 8c65927e — the mark moved inside the success branch, plus a `Logger\log()` on the failure path and a `failed_count` surfaced on the compose screen, so it isn't silent in the UI either.

One side effect worth calling out: a failed address is no longer in the dedupe map, so it gets another attempt if the same person turns up via another selected group. Each attempt still costs a unit against `BATCH_SIZE`, so a run stays bounded.

Verified locally on the dev stack: forced `wp_mail()` to fail, and the summary came back `sent_count => 0, failed_count => 2`, the screen showed *"2 message(s) could not be sent — see the error log"*, and `debug.log` got one entry per failure with the job ID and recipient. Added a regression test that fails against the old code.

public_html/wp-content/mu-plugins/groups/network-messaging.php
bor0: Fixed in 8c65927e.

Verified locally end-to-end: set a group editor's `display_name` to `evil@example.test, Jane`, then sent a broadcast. MailCatcher received one message with one envelope recipient:

```
To: "evil@example.test, Jane" <organiser1@example.test>
```

PHPMailer quotes the phrase, so the comma is inert in the header too. On the pre-fix code that same send delivered to `evil@example.test` as well. Added a regression test that applies core's own split rule to whatever `wp_mail()` was handed and asserts it yields exactly one address.

public_html/wp-content/mu-plugins/groups/network-messaging.php
bor0: Fixed in 8c65927e with a `wp_cache_add()` lock, via a `with_jobs_lock()` helper.

`process_batch()` claims its job under the lock and then sends unlocked — holding it across `BATCH_SIZE` `wp_mail()` calls would block the compose form for however long the transport takes. Two details that weren't obvious:

- The lock group is registered global (`wp_cache_add_global_groups`). `process_batch()` runs under whatever blog is current, and non-global cache groups are keyed per blog, so otherwise two runs would take two different locks and not exclude each other at all.
- The batch's write-back takes the lock with a steal fallback, because at that point its progress only exists in memory and dropping the write would re-send everything the run just sent. `queue_message()` doesn't steal — it gives up and shows a "queue was busy, try again" notice.

Documented in the code: this is only a real cross-process mutex with a persistent object cache (memcached in production). Without one it degrades to the unsynchronised behaviour it replaces, no worse.

To be straight about the evidence here: this one is covered by a regression test that queues a second job from inside the send loop, which fails against the old code (the job is lost). I didn't reproduce the race with two genuinely concurrent processes.

gedex:
After sending the message via https://events.wordpress.test/group/wp-admin/network/settings.php?page=wporg-groups-messaging and run following command:

```
docker compose exec wordcamp.test wp cron event run wporg_groups_message_batch --url='https://events.wordpress.test/group'
Executed the cron event 'wporg_groups_message_batch' in 0.001s.
Success: Executed a total of 1 cron event.
```

I still didn't get the email. Maybe, I missed a thing here.

bor0:
@gedex can you try again after 7267fc5? It contains some error handling improvements, I'm curious what we'll see after it.

gedex:
So, each time I tested sending a message with a unique subject and message, it didn't get sent immediately but old emails that weren't showing up before are sent. I did following to delete queue option and remove `wporg_groups_message_batch` event:

```
docker exec wordcamptest-wordcamp.test-1 bash -c "mysql -h wordcamp.db -uroot -pmysql wordcamp_dev -e 'DELETE FROM wc_sitemeta WHERE meta_key LIKE \"wporg_groups_message_%\"'" \
&& docker exec wordcamptest-wordcamp.test-1 wp cron event delete wporg_groups_message_batch --url=https://events.wordpress.test/group/ --allow-root 2>/dev/null;
```

`wp cron event run wporg_groups_message_batch` runs **one batch** (up to 50 work-units on the **oldest** job) and returns. If more jobs remain, the code reschedules another event and stops — it doesn't drain everything in one call. So a queue of N jobs needs cron to fire N-ish times before mailcatcher shows everything.

Also worth knowing (relevant to the updated PR): the lock is in the object-cache group `wporg-groups-messaging`, key `jobs_lock`. There's no persistent object cache in this dev env, so the lock is per-request memory and can't get "stuck" across requests — no cleanup needed there.

gedex:
### Sending

| Test A (organisers only and all groups) | Test B (members and Hobart only) |
| -- | -- |
| <kbd>image</kbd> | <kbd>image</kbd> |

### Sent emails

image

gedex (approved):

Issue #1858, opened 2026-08-06
Update designs


raicem:
So far we were not intentional with designs for https://events.wordpress.org/group/internal-testing-group/ and other pages such as events. Without spending too much time, let's go through these pages and make improvements to the information hierarchy.

Later we can ask for feedback and improvements from a designer.

bor0:
Feedback from @annezazu to consider as part of this issue:

> This is so freaking cool to see. First of all, great work getting this moving. Some feedback:

- [ ] The login form is so verbose and not the same design as the official one. Can we have a similar approach to the official login page but have the message be around events ("Log in to your WordPress.org account to register for events and join the global WordPress community.")

<img width="1564" height="1488" alt="Image" src="https://github.com/user-attachments/assets/a166d5ef-731f-4a2e-8829-a397f16fee2f" />

- [x] The individual event looks solid. I would test what it would look like without an image and perhaps consider adding a backup default image. We need to think about empty states.
- [x] What does it look like if someone is an organizer (use a z instead of s in the interface) and attendee?
- [x] The blue in the create event modal and message attendees doesn't match the blue of the site. Update that for the various pop up modals. Minor but it matters.
- [x] On hover, the blue button shows blue on blue.

<img width="1204" height="582" alt="Image" src="https://github.com/user-attachments/assets/5ca6766d-0cbc-4b15-8124-4cff797a3d18" />

- Nice work with the sponsor setup.

> Overall, this definitely needs an initial clean up from design folks but, functionality wise, this is looking solid.

raicem:
> the login form is so verbose and not the same design as the official one.
 
Currently logging in takes us to [this URL](https://login.wordpress.org/?redirect_to=https%3A%2F%2Fevents.wordpress.org%2Fgroup%2Finternal-testing-group%2Fwp-admin%2F&locale=en_US). The copy is different from the official `https://login.wordpress.org/` but overall, to my eyes, they look similar. Not the same as Anne's copy but not super verbose compared `login.wordpress.org`

<img width="1726" height="1086" alt="Image" src="https://github.com/user-attachments/assets/ca104ae3-e68e-459b-97b9-d76392dffa8f" />

@bor0 could I be missing something here? I guess worse case we can get rid of the title "WordCamp is part of WordPress.org" and align the copy a bit better with Anne's suggestion.

bor0:
Some additional feedback from @annezazu:

> I tested this on mobile when I saw you all create a test event. Two issues to note:

[...]

PR #1860, opened 2026-08-07
Groups: improve front-page information hierarchy

raicem:
## What changed

Refreshes Groups pages with a compact location-aware hero, responsive member/organizer sidebar, news/event/sponsor sections, and consistent event action buttons.

## Test

1. View a group home logged out, as a member, and as an organizer on desktop/mobile; verify the hero location, sections, and sidebar states.
2. In **Group settings → About**, save online and in-person locations; verify each appears in the hero.
3. Join/leave the group, toggle email preferences, create/edit an event, and RSVP/un-RSVP; verify the buttons, modals, and states.

Part of #1858

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/page-content/block.json
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/page-content/block.json
@@ -11,10 +11,19 @@
 		"slug": {
 			"type": "string",
 			"default": ""
+		},
+		"heading": {
+			"type": "string",
+			"default": ""
+		},
+		"headingLevel": {
+			"type": "number",
+			"default": 2
 		}
 	},
 	"supports": {
-		"html": false
+		"html": false,
+		"anchor": true
 	},
 	"render": "file:./render.php",
 	"editorScript": "file:./index.js"
--- a/public_html/wp-content/themes/groups-site/assets/css/custom.css
+++ b/public_html/wp-content/themes/groups-site/assets/css/custom.css
@@ -56,13 +56,6 @@
    Front-page hero
    ========================================================================== */
 
-/* The hero carries the group's own description (`blogdescription`), which is
-   often a single short line. Cap its measure so it stays readable rather than
-   stretching the full 720px content width.
-
-   When it's unset the hero simply omits it — no placeholder. Organisers are
-   already prompted by `wporg/group-settings`, whose trigger button switches to
-   "Set up your group" on exactly this condition and opens the About tab. */
 .groups-site-hero .wp-block-site-tagline {
 	max-width: 52ch;
 	text-wrap: pretty;
@@ -140,10 +133,6 @@
 	margin: 0 !important;
 	font-size: var(--wp--preset--font-size--normal);
 }
-
-/* The separator is only needed when organiser controls render above the
-   preference. Ordinary members see the preference as the sidebar's first
-   and only section. */
 
 .groups-site-sidebar:has(.wporg-event-manage) .groups-site-sidebar-preference {
 	margin-top: var(--wp--preset--spacing--10) !important;
--- a/public_html/wp-content/themes/groups-site/assets/css/custom.css
+++ b/public_html/wp-content/themes/groups-site/assets/css/custom.css
@@ -68,31 +68,96 @@
 	text-wrap: pretty;
 }
 
-/* Page-level spacing for the email preference in the member's own section. */
-.groups-site-my-preference {
-	margin-top: var(--wp--preset--spacing--20);
+.groups-site-hero .wp-block-wporg-group-location,
+.groups-site-hero .wp-block-wporg-group-membership {
+	padding-top: var(--wp--preset--spacing--10);
+}
+
+.groups-site-hero .wp-block-wporg-group-membership {
+	margin-block-start: 0;
 }
 
 /* ==========================================================================
-   Front-page lower sections ("About this group", sponsors)
-
-   Constrained and spaced here rather than by wrapper groups in the template.
-   Both blocks render nothing when they have no content, and a wrapper would
-   leave its top margin behind as a gap on every group site that hasn't
-   written an About page or been given a sponsor.
+   Front-page content and sidebar
    ========================================================================== */
 
-.groups-site-about,
-.groups-site-sponsors {
-	max-width: 768px;
-	margin-top: var(--wp--preset--spacing--70);
+.groups-site-home-layout {
+	align-items: flex-start !important;
 }
 
-/* The rule between the events sections and About/Sponsors has nothing to
-   separate when both of those render empty, which is the common case for a
-   brand-new group. Drop it when it ends up last in the main column. */
-main > .groups-site-content-separator:last-child {
-	display: none;
+.groups-site-sidebar-column {
+	order: 2;
+}
+
+.groups-site-main-column-wrapper {
+	order: 1;
+}
+
+.groups-site-sidebar {
+	box-sizing: border-box;
+	width: 100%;
+	padding-left: var(--wp--preset--spacing--40);
+	border-left: 1px solid var(--wp--preset--color--light-grey-1);
+}
+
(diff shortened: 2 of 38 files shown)


wppoland:
Checked this mainly for accessibility regressions, since it touches `src/blocks/event-rsvp/render.php` and #1852 recently landed there. Good news on that front: the native `<button type="button">`, the `role="status"` / `aria-live="polite"` notice and the decorative avatars are all intact, and I found no skipped heading levels inside the block `render.php` files.

One thing looks unintended though.

### The About section is no longer rendered on the front page

Production `templates/front-page.html` has:

```html
<!-- wp:group {"className":"groups-site-section","anchor":"about",...} -->
<div id="about" class="wp-block-group groups-site-section" ...>
	<h2 class="wp-block-heading has-heading-3-font-size" ...>About this group</h2>
	<!-- wp:wporg/page-content {"slug":"about"} /-->
</div>
```

At this PR's head, `front-page.html` contains no `wporg/page-content` block and no `about` anchor at all. I grepped the whole theme and the mu-plugins for a replacement call site and found none: `wporg/page-content` now appears only in its own `block.json`, its `render.php` and `tests/test-blocks.php`, with no template, template part, pattern or `render_block()` reference anywhere.

The block itself has not been removed, and the PR actually extends it in the same change, adding `heading` / `headingLevel` attributes and heading output to `page-content/render.php`. As things stand that new code is unreachable, which is what makes me think the removal was accidental rather than a deliberate cut.

To be clear about the blast radius, since it is smaller than it first looks: the About content is not lost site-wide. The About page still renders standalone through `templates/page.html`, and it is still reachable from the nav, so this is a front page regression rather than content disappearing. The old hero button pointing at `#about` no longer has a target either.

Given the PR description says it "organise[s] events, news, About content, and sponsors", I suspect the intent was to move the About block into the new structure and pass it the new `heading` attribute, and that it simply got dropped in the reshuffle.

Happy to be told it was deliberate and that About is meant to live only on its own page now, in which case the new heading attributes in `page-content/render.php` are the leftover rather than the omission.

raicem:
Hey @wppoland! Thank you for checking out the PR. Indeed the about page removal was intended. I'm exploring what would be the best way to present information on the homepage and for this minute I was thinking linking to the About page from the header would be a good clean up. 

I think I'll walk back from this idea and bring it back in a future PR. IMHO, it's useful information for new visitors of the group. I don't want to make this PR any larger, so I'll do it later. It'll have to happen with some hero component changes as I believe it takes up too much useful space. So if I can make it smaller, we'll have more above the fold room for events and about page content.

public_html/wp-content/themes/groups-site/assets/css/custom.css
```diff
@@ -57,6 +57,81 @@ body.error404 .site-content-container h1 {
...
+/* ==========================================================================
+   Front-page content and sidebar
+   ========================================================================== */
+
+.groups-site-home-layout {
+	align-items: flex-start !important;
+}
+
+.groups-site-sidebar-column {
+	order: 2;
```
gedex, line 82: The sidebar column is first in the DOM but `order: 2` moves it right above 960px, so keyboard/screen-reader order is nav → Membership/Settings → main content, the reverse of the visual order.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/page-content/block.json
```diff
@@ -11,10 +11,19 @@
 		"slug": {
 			"type": "string",
 			"default": ""
+		},
+		"heading": {
```
gedex, line 15: `heading`, `headingLevel`, and `anchor` are added here but nothing renders `wporg/page-content` any more — 77c9902d dropped the About block from `front-page.html`, and no template or pattern references it. Unreachable attributes.

gedex (commented):
Looks like the About section got lost in the shuffle. 84bf42ec added `heading`/`headingLevel`/`anchor` to `wporg/page-content` so it could carry its own heading, then 77c9902d pulled the block out of `front-page.html`. Nothing references it now, and nothing links to `/about/` either (provisioning only makes Members), so any About content a group wrote is stranded, organisers lose the prompt to create it, and those new attributes are dead. Worth re-adding to the main column — or if it's an intentional cut, drop the block changes and say so.

Also, a stale country code can quietly wipe a saved location. `get_location()` returns `null` when the code no longer resolves via `wcorg_get_country_name_from_code()`, and since the About tab builds its form from that response, the radio comes back unselected and the next save of any field posts `location: null` → `clear_location()` drops type, city, and country. Rare, but silent and unrecoverable. Returning the stored values instead of collapsing to `null` would cover it.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/page-content/block.json
gedex:
```diff
--- a/public_html/wp-content/themes/groups-site/templates/front-page.html
+++ b/public_html/wp-content/themes/groups-site/templates/front-page.html
@@
 				<!-- /wp:group -->
 
+				<!-- wp:wporg/page-content {"slug":"about","heading":"About this group","anchor":"about","className":"groups-site-about"} /-->
+
 				<!-- wp:wporg/group-news {"limit":3,"lock":{"move":true,"remove":true}} /-->
```

gedex:
Looks better with right sidebar, IMO. Thanks!

|  | desktop | mobile |
| -- | -- | -- |
| anon | <kbd>anon-desktop</kbd> | <kbd>anon-mobile</kbd> |
| organizer | <kbd>organiser-desktop</kbd> | <kbd>organiser-mobile</kbd> |

public_html/wp-content/themes/groups-site/assets/css/custom.css
raicem: Fixed in d456edd8 (PR #1884).

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/page-content/block.json
raicem: I did this on purpose, but a bad call as I also [mentioned](https://github.com/WordPress/wordcamp.org/pull/1860#issuecomment-5239809129) to @wppoland. Reverted to basically your diff in efd68cd6 (PR #1884).

raicem (commented):
Thanks @gedex! Both fixed in the follow-up PR #1884.

About section: you and @wppoland were right, I'm walking back from this decision in the next PR. It's back with efd68cd6.

Stale country code is fixed in 6d0789a0.

bor0:
## Design + code review

Tested locally (Docker, `sunshine-coast-qld` test group) as logged-out/member/organiser on desktop (1440px) and mobile (390px): hero + location, sidebar, event grid, single-event + RSVP flow, and the new Settings → About location fields (in-person and online) all worked cleanly and looked great, including responsive stacking on mobile.

📸 Screenshot gallery: https://claude.ai/code/artifact/5fe112b9-08f9-41eb-b4a9-faa85720765f

Code review of the diff (rest.php, group-location.php, the new blocks, button-style consolidation) looked solid — proper escaping/sanitization, sensible REST validation, good test coverage. CI is fully green (lint, JS, PHP 8.3–8.5, Playwright).

**However**, checking current HEAD (`ff92a09c`) against @gedex's review comments above: both issues he flagged are still present on this branch — the fixes only exist in the still-open #1884:

- **About section is orphaned** — `wporg/page-content` was pulled from `front-page.html` and nothing links to `/about/` anymore. Confirmed: no reference to "about" remains anywhere in the current theme templates.
- **Stale country code can silently wipe a saved location** — `get_location()` returns `null` when the stored country code no longer resolves via `wcorg_get_country_name_from_code()`, so the About-tab form loses the selection, and the next save of *any* field (even just the group name) posts `location: null` and clears it.

Given these are acknowledged-but-unfixed, I'd hold off merging to production until #1884's fixes land here — everything else is in good shape.

bor0 (approved):
Good to merge after ~merging #1884~ cbbbe00

PR #1864, opened 2026-08-08
Identify which registration question is unanswered

wppoland:
Follows up my review comment on #1853 with the actual patch, since it is small enough that a branch is more use than a description. Targets `add/1813-rsvp-registration-questions` so it lands inside your PR rather than alongside it. Merge it, cherry-pick it, or ignore it entirely.

### What it does

`collectAnswers()` already knows which input is empty at the moment it decides `missingRequired`, it just throws that away. This returns the offending inputs, marks them `aria-invalid="true"`, and moves focus to the first via the `scheduleFocus()` helper already in the file. The flag is cleared on every submit so it never outlives the problem.

The change sits in `submitRsvp()`, which is the shared path for both the Attend toggle and the Save answers button, so both routes are covered by the one edit.

### Why

The message itself was never the gap. `ctx.rsvpNotice` is bound to the `role="status"` live region, so "Please answer the required questions" is announced properly. What was missing is the other half of WCAG 2.1 3.3.1: the error is described, but the item in error is not identified. With up to five questions, and every input sharing one `aria-describedby` pointing at the same generic paragraph, a screen reader user is told something is missing and then has to hunt for it.

### Tests

Two added to `tests-js/event-rsvp.test.js`, in the existing questions describe block:

- `identifies which required question is unanswered` asserts `aria-invalid` lands on the required empty input and not on the optional one
- `clears the invalid flag once the required question is answered`

I confirmed both fail against your branch without the source change and pass with it, and the file's full suite is green at 16 tests.

Nothing else in the PR is touched, and none of the #1852 accessibility work is affected.

Props motylanogha

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/view.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/view.js
@@ -228,23 +228,46 @@ function scheduleFocus( element ) {
  * the block so a page listing several events reads the right one.
  *
  * @param {Element} block The `.wp-block-wporg-event-rsvp` element.
- * @return {{answers: Object, missingRequired: boolean}} Collected answers.
+ * @return {{answers: Object, missingRequired: boolean, missingInputs: Element[]}} Collected answers.
  */
 function collectAnswers( block ) {
 	const answers = {};
-	let missingRequired = false;
+	const missingInputs = [];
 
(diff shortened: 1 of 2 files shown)


gedex:
### Tested

* Empty submission sets `aria-invalid="true"`.
* Focus moves to the unanswered required field.
* Error message appears.
* Answering clears `aria-invalid` and RSVP succeeds.
* No browser console issues observed.

image

image

image

gedex (approved):

PR #1865, opened 2026-08-09
Persist a recurrence rule chosen before publication

wppoland:
Follows up my review comment on #1846 with a patch for the first of the two problems, based on `feature/gatherpress-recurring-events-frontend` so it merges inside your PR.

### The fix

`persist_event()` computes `$schedule_editable` while the post is still a draft, publishes it, and only then fires `wporg_groups_frontend_event_saved`. By that point `lock_published_schedule()` applies, so each rule field the organiser just changed is short-circuited to `false`. `persist_recurring_event_rule()` does not check the return values, so the request reports success and the choice is silently dropped, unrecoverable because the schedule locks on publication.

Rather than reorder `persist_event()`, this reuses the mechanism you already have. `Plugin::with_schedule_unlocked()` sits next to the `$schedule_updates_allowed` allowlist that `update_end_condition()` uses, and `save_recurring_event()` runs the persist through it. `$schedule_editable` is precisely the record that the caller was entitled to write the schedule as of before the transition, which is when the organiser made the choice, so honouring it seems right.

The locked-key list also moves into `Plugin::locked_schedule_meta_keys()`, so the lock and the unlock cannot drift apart. `with_schedule_unlocked()` saves and restores any existing allowlist entry rather than clobbering it, so it nests safely.

This also fixes selecting "Does not repeat" at publish time, which the same guard was blocking via the `delete_post_meta()` loop.

### Not covered here

The second problem from my comment, where editing a published series' date or time moves the `gatherpress_events` row through `Event::save_datetimes()`'s direct `$wpdb` write while the mirroring `gatherpress_datetime` meta stays blocked, is left alone. That one needs a decision from you about whether the anchor should be movable at all after publication, and I did not want to encode a guess.

### Testing

I could not run the PHP suite locally: this checkout has no `vendor/`, and the groups bootstrap wants a multisite database plus GatherPress installed, which its own error message points at `.docker/bin/install-test-suite.sh` for. So please treat CI as the verification rather than my word. Both changed files pass `php -l`.

Props motylanogha


wppoland:
Added the tests I should have included in the first push (56b9907):

- `test_schedule_unlock_persists_writes_on_a_published_series` asserts the write is blocked normally, lands through `with_schedule_unlocked()`, and that the callback's return value is passed back
- `test_schedule_unlock_is_scoped_and_released` asserts the lift does not leak to another post and that the lock is back afterwards

They sit next to your existing `test_end_condition_bypass_is_scoped` and reuse `create_published_recurring_event()`. Still not run locally for the reason in the description, so CI remains the verification.

wppoland:
Lint PHP went red on the last push (f6f71c0). `phpcs-changed` attributed the pre-existing missing docblock on `update_end_condition()` to this branch, because the new `with_schedule_unlocked()` lands directly above it. Added the docblock; `BASE_REF=... php .github/bin/phpcs-branch.php` is clean locally now. No behaviour change.

wppoland:
Both Copilot points were right, addressed in 9938c1b.

**Orphaned docblock.** Removed. It was a leftover from f6f71c0, where I added the missing docblock to `update_end_condition()` without deleting the detached one that had drifted above `locked_schedule_meta_keys()`.

**Nesting.** The more interesting one, and the fix is a bit broader than the comment asked for. `update_end_condition()` replaced the whole allowlist slot and then unset it, so calling it inside `with_schedule_unlocked()` cancelled the outer lift for everything that followed in the same callback. That is exactly the silent-loss failure this PR exists to fix, just reintroduced one level down.

Rather than duplicate the save-and-restore logic in both methods, both now delegate to a private `with_schedule_keys_allowed( $post_id, $keys, $callback )` that merges into whatever entry is already there and restores it afterwards. `with_schedule_unlocked()` passes the full locked list, `update_end_condition()` passes its two keys, so its bypass stays as narrow as before. Two hand-rolled copies would only need to drift once to bring the bug back.

Added `test_schedule_unlock_survives_a_nested_end_condition_write`, which runs an end-condition write inside the lift and then asserts a following schedule write still lands, and that the lock is back afterwards. Your existing `test_end_condition_bypass_is_scoped` still covers the narrowness of the standalone bypass.

On verification, same caveat as before and I would rather repeat it than let it pass silently: I got `composer install` working and the multisite bootstrap running locally, but this suite needs the multi-network fixture that `.docker/bin/install-test-suite.sh` builds (`WORDCAMP_ROOT_BLOG_ID` and friends), and WP bails out during bootstrap against a plain multisite install. So both changed files pass `php -l` and `phpcs-branch.php` is clean, but CI is the real verification for the tests, not my word.

Props motylanogha

wppoland:
Re @bor0's 0.35.0 rebase note: checked this PR against 0.35.0 rather than assuming, and nothing here needs changing.

- No file overlap with #1873. It touched the test-suite installer, the CI workflows, a Groups test and a theme template; this PR touches the recurring-events plugin and its tests.
- `gatherpress_datetime` is still the event datetime meta key in 0.35.0 (`Event\Meta`), so `locked_schedule_meta_keys()` is still naming a real key. The other entries are `Rule::META_PREFIX` keys, which are ours and unaffected by the GatherPress version.
- `Event::save_datetimes()` is byte-identical between 0.34.1 and 0.35.0, so the second problem I described above and deliberately left out of this PR — the direct `$wpdb` write moving the `gatherpress_events` row while the mirroring meta stays blocked — still holds exactly as written, and still needs your call rather than a guess from me.

Not rebasing: this branch is based on `feature/gatherpress-recurring-events-frontend`, not production, so it picks up the new pin whenever #1845 does. Rebasing across @gedex's branch from underneath would just make his life harder. Happy to re-verify once #1845 moves.

Props motylanogha

bor0:
Thanks for this, @wppoland — great catch and a clean fix.

While reviewing #1846 I found the same issue independently and ended up applying this exact mechanism (`Plugin::with_schedule_unlocked()` / `locked_schedule_meta_keys()`, wired into `save_recurring_event()`) directly to that branch, which has since merged into `production` (`f5abd28c`). So the fix itself is already shipped.

I've rebased this PR on top of current `production` to clear the resulting conflicts, and restored the two regression tests from your `56b9907e` commit (`test_schedule_unlock_persists_writes_on_a_published_series`, `test_schedule_unlock_is_scoped_and_released`) — those weren't carried over when I applied the fix, and they're good coverage for the mechanism now in production. That test addition is the only remaining diff here; everything else is already merged. Full suite passes (610 tests) and CI is green.

Merging this for the test coverage. Thanks again for digging into this.

PR #1873, opened 2026-08-10
Groups: update GatherPress pin to 0.35.0

### LINKED ISSUE #1843: Upgrade GatherPress from 0.34.1 to 0.35.0
We're pinned to GatherPress 0.34.1 network-wide (`gatherpress/gatherpress.php` in `wcorg-network-plugin-control.php`). 0.35.0 is currently in beta (`0.35.0-beta.1`, tagged 2026-07-30) with its milestone due **2026-08-10** and all 44 tracked issues already closed, so it's close to a stable release.

Goal: stay on the latest GatherPress and pick up its fixes, while making sure our Groups/GatherPress integration (`mu-plugins/wporg-groups-frontend`, `mu-plugins/groups/gatherpress-groups-tweaks.php`, `groups-site` theme) doesn't regress.

## Notable changes in 0.35.0 (from the [beta.1 changelog](https://github.com/GatherPress/gatherpress/releases/tag/0.35.0-beta.1))

- **Minimum PHP raised to 8.1** (from 7.4) — not a blocker, we run PHP 8.3.
- **Minimum WordPress raised to 7.0** — confirm our network is on 7.0+ before/at upgrade time.
- RSVP system rebuilt on a provider-registry architecture; new `gatherpress_loaded` action.
- RSVP settings vocabulary changed (`enabled`/`disabled` instead of `on`/`off`), settings page slugs gain a `_settings` suffix, values migrated automatically via the GatherPress Alpha migration routine.
- `gatherpress_geocode_street_line` filter removed, replaced by `gatherpress_formatted_address` — we don't currently use either, confirmed via grep.
- GatherPress Icon block removed in favor of core `core/icon`; Venue Map block dimensions/width handling reworked (Default Width setting removed).
- Venue map now mounted via the Interactivity API.

None of these directly touch code we grepped for in this repo, but the RSVP/settings rework and venue map changes are exactly the kind of thing that can silently break our custom REST endpoints and blocks — worth the full compat pass below rather than assuming it's fine.

## Plan

- [x] Confirm 0.35.0 has gone stable on WordPress.org (not just beta) before rolling out network-wide.
- [x] Confirm network sites are on WP 7.0+.
- [x] Update the pinned/tested version where applicable and deploy to a staging/testing group site first.
- [x] Run PHPUnit (`docker compose -f docker-compose.phpunit.yml up`), covering the `WordCamp Groups` suite.
- [x] Run E2E (`npx playwright test` / `e2e-tests.yml` workflow).
- [x] Run the full manual `groups-gatherpress-compat-test` checklist (`.claude/skills/groups-gatherpress-compat-test/SKILL.md`): per-role browser pass, direct REST pass, GatherPress-off guard pass, security/perf sanity checks.
- [x] Specifically re-verify RSVP flows and venue map/venue creation given the RSVP and Venue Map rework in this release.
- [x] Roll out network-wide once all of the above pass clean.

## Acceptance criteria

- [x] GatherPress is upgraded from 0.34.1 to 0.35.0 network-wide.
- [x] Automated (PHPUnit + E2E) and manual compat-checklist passes are all green, with results noted on this issue.
- [x] No regressions in event creation/editing, RSVP, venue creation/map rendering, or member management across Organiser/Event Organiser/Member roles.

bor0:
Fixes #1843 

## Summary

Bumps the pinned GatherPress version from 0.34.1 to 0.35.0 and fixes the one breaking change that affects this repo.

- GatherPress 0.35.0 removes the `gatherpress/icon` block in favor of the new WP 7.0 core `Icon` block ([GatherPress#1931](https://github.com/GatherPress/gatherpress/commit/ae7ce4e3)). The `groups-site` theme's `single-event.html` template used `gatherpress/icon` for the venue address/phone/website and online-event indicators, so those are swapped for `core/icon` using GatherPress's own mapping (location → map-marker, phone → mobile, admin-site-alt3 → external, video-alt2 → capture-video) to avoid blank/broken blocks after the bump.
- Bumped every place that pins the GatherPress release zip for local dev and CI, since the plugin is gitignored/third-party and not tracked in this repo: `.github/workflows/e2e-tests.yml`, `.github/workflows/unit-tests.yml`, `.docker/bin/install-test-suite.sh` (was actually still on `0.34.0-alpha.2`), and the `groups-gatherpress-compat-test` skill's manual-install instructions.
- Fixed a `wporg-groups-frontend` PHPUnit test (`test_missing_required_answer_blocks_rsvp`) that assumed `Rsvp::get()` always returns an array — 0.35.0 changes it to return `null` when no identity resolves. Production code already handles this via `?? 'no_status'`; the test now does the same.

Everything else in the `Rsvp`, `Event`, `Venue`, and `Venue\Setup` APIs that `mu-plugins/groups`, `mu-plugins/wporg-groups-frontend`, and the `groups-site` theme depend on is unchanged or backward compatible between 0.34.1 and 0.35.0 (confirmed via a full diff of the two tags).

## ⚠️ Required rollout step: run GatherPress Alpha's migration after upgrading

GatherPress 0.35.0 ships breaking changes to **stored content**, not just code: the old `gatherpress/icon` block, old `gatherpress/venue-map` width/height attributes, and old RSVP settings option values (`rsvp_mode`: `all_on`/`per_event_on`/`per_event_off` → `enabled`/`per_event_enabled`/`per_event_disabled`) all need migrating in the database. GatherPress ships this as a **separate companion plugin, [GatherPress Alpha](https://github.com/GatherPress/gatherpress-alpha)**, version-locked to core.

After deploying this PR to any environment, install/update GatherPress Alpha to 0.35.0 and run its fixer once:
```
wp gatherpress alpha fix
```
(or via wp-admin: Events → Settings → Alpha → Apply Updates). I hit this directly in testing — a dev-seeded venue's address/phone/website/map data disappeared entirely after the bump, which looked like a GatherPress regression until I ran `wp gatherpress alpha fix` and it fully resolved (screenshots in the PR discussion). This only needs to run once per site; it does **not** touch this repo's own template file (`single-event.html`), which is why that still needed the manual fix above — Alpha only rewrites block content stored in the database, not theme template files.

We don't have a `rsvp_mode`/`all_on`/`per_event_on`/`per_event_off` reference anywhere in `mu-plugins/groups` or `mu-plugins/wporg-groups-frontend`, so the settings-value rename itself doesn't need any code change here — just the one-time Alpha run per site for any content/data already on disk.

## Notes for reviewers

- GatherPress 0.35.0 raises its minimum requirements to **PHP 8.1** and **WP 7.0**.
- `mu-plugins/groups/gatherpress-groups-tweaks.php`'s `render_block_gatherpress/venue` filter (hooked at priority 20) still works correctly, but its code comment explaining *why* priority 20 is needed is now stale — GatherPress's own competing priority-10 callback was removed in 0.35.0 (`Blocks\Venue::__construct()` is now empty). Not fixed here since it's a comment-only cleanup, not a behavior change.
- The `gatherpress/venue-map` block had a large rewrite (Google Maps provider, new REST API, Interactivity API port). Verified working end-to-end on a real venue (OSM/Leaflet provider) after running the Alpha migration above.

## Test plan

- [x] Run the `wporg-groups-frontend` and `groups` PHPUnit suites against GatherPress 0.35.0 — 137/137 passing after the one test fix above
- [x] Run the Playwright E2E suite (`npx playwright test`) — 9/9 passing. (The first parallel run threw 3 timeouts; they all passed in seconds when re-run serially, so that was resource contention from `fullyParallel: true` against a single local container, not a regression. The 4th failure, `/members/` 404ing, was this dev site never having a "Members" page provisioned — created one, unrelated to GatherPress.)
- [x] Run the `groups-gatherpress-compat-test` skill checklist (per-role browser pass, direct REST pass, GatherPress-off pass, security/perf sanity checks) — all green, including a no-fatals pass with GatherPress physically removed
- [x] Visually confirm the venue address/phone/website and online-event icons render correctly on a group's single-event page — confirmed via Claude in Chrome after running the GatherPress Alpha migration
- [x] Visually confirm the venue map block still renders correctly — confirmed (interactive Leaflet/OSM map with marker, zoom controls, attribution)
- [x] Verify venue creation end-to-end (RSVP/Venue Map rework in this release touches this path) — created a venue through the front-end editor as an Event Organiser; address geocoded correctly, static map pre-generated, structured address fields (state/postcode/country) populated, and the new venue was immediately selectable on the event
- [x] Confirm production PHP ≥ 8.1 and WP core ≥ 7.0 before merging
- [x] Confirm GatherPress 0.35.0 has gone stable on WordPress.org (not just beta) — confirmed via the plugin API, stable tag is 0.35.0 as of today

### DIFF
--- a/.claude/skills/groups-gatherpress-compat-test/SKILL.md
+++ b/.claude/skills/groups-gatherpress-compat-test/SKILL.md
@@ -44,7 +44,7 @@ docker compose ps   # expect wordcamp.test + wordcamp.db running
 docker compose exec wordcamp.test wp plugin list --url=events.wordpress.test/group/sunshine-coast-qld/ | grep gatherpress
 # If missing/inactive:
 docker compose exec wordcamp.test wp plugin install \
-  https://github.com/GatherPress/gatherpress/releases/download/0.34.0-alpha.2/gatherpress.0.34.0-alpha.2.zip \
+  https://github.com/GatherPress/gatherpress/releases/download/0.35.0/gatherpress.0.35.0.zip \
   --activate --url=events.wordpress.test/group/sunshine-coast-qld/
 ```
 
--- a/.claude/skills/groups-gatherpress-compat-test/SKILL.md
+++ b/.claude/skills/groups-gatherpress-compat-test/SKILL.md
@@ -48,6 +48,24 @@ docker compose exec wordcamp.test wp plugin install \
   --activate --url=events.wordpress.test/group/sunshine-coast-qld/
 ```
 
+**After bumping the pinned GatherPress version, also install/update
+[GatherPress Alpha](https://github.com/GatherPress/gatherpress-alpha)
+(version-locked to core) and run its one-time migration.** GatherPress ships
+breaking *content* migrations (old `gatherpress/icon` block usage, old
+`gatherpress/venue-map` width/height attributes, renamed RSVP settings
+values) through this companion plugin rather than an automatic upgrade
+routine. Skipping this step is exactly what it looks like when a venue's
+address/phone/website/map silently vanish from the front end after a version
+bump — it reads like a GatherPress regression but is actually just unmigrated
+data:
+
+```bash
+docker compose exec wordcamp.test wp plugin install \
+  https://github.com/GatherPress/gatherpress-alpha/releases/download/0.35.0/gatherpress-alpha.0.35.0.zip \
+  --activate --url=events.wordpress.test/group/sunshine-coast-qld/
+docker compose exec wordcamp.test wp gatherpress alpha fix --url=events.wordpress.test/group/sunshine-coast-qld/
+```
+
 Create one test user per role tier, with **both** a browser login password
 and a REST application password (the browser pass needs the former, the
 REST pass needs the latter):
--- a/.claude/skills/groups-gatherpress-compat-test/SKILL.md
+++ b/.claude/skills/groups-gatherpress-compat-test/SKILL.md
@@ -25,9 +25,50 @@ interaction, cross-plugin capability leaks, exploratory checks):
   `e2e-tests.yml` GitHub Action manually — it's `workflow_dispatch`-only).
   Covers anonymous front-page rendering, an author creating an event
   end-to-end through the real browser UI, and the member directory.
+  **If a spec times out waiting on a UI interaction (e.g. a sidebar
+  panel/button that should obviously be there), re-run with
+  `--workers=1` before concluding it's a regression.** `fullyParallel: true`
+  against a single local docker container is a real, repeatable source of
+  90s timeouts that have nothing to do with GatherPress — confirmed during
+  the 0.35.0 bump: 3 "failures" all passed in under 10s each serially.
 
 If either automated suite fails, stop and fix that before doing the manual
 pass below — don't duplicate debugging effort across layers.
+
+## 0. Before bumping the pinned version
+
+- **Confirm the target version is actually stable**, not a beta/RC:
+  `curl -s https://api.wordpress.org/plugins/info/1.0/gatherpress.json | python3 -c "import json,sys; print(json.load(sys.stdin)['version'])"`.
+- **Diff GatherPress core itself** for breaking API changes before touching
+  any code here. Clone/pull `github.com/GatherPress/gatherpress` locally and
+  run `git diff <old-tag> <new-tag>` (and `git log <old-tag>..<new-tag>
+  --oneline` for the human-readable summary). Specifically check every class
+  this integration touches directly —
+  `GatherPress\Core\{Rsvp\Rsvp,Event\Event,Venue\Venue,Venue\Setup,User,Setup,Blocks\Setup,Blocks\Event_Query}`
+  (grep `GatherPress\\\\` across `mu-plugins/groups`,
+  `mu-plugins/wporg-groups-frontend`, and `themes/groups-site` to get the
+  current list) — for renamed/removed constants, changed return types
+  (`Rsvp::get()` went from always-`array` to `array|null` in 0.35.0 and broke
+  a test that indexed it directly), and classes marked `final` (harmless
+  unless something here extends one — check with
+  `grep -rn "extends.*GatherPress"`).
+- **Grep theme templates for GatherPress block usage**, not just PHP:
+  `grep -rn "wp:gatherpress" public_html/wp-content/themes/groups-site`.
+  GatherPress Alpha's migration (see below) only rewrites block content
+  *stored in the database* — it does not touch theme `.html` template
+  files. If a version bump removes/renames a block GatherPress itself used
+  to ship (e.g. `gatherpress/icon` → core `core/icon` in 0.35.0), any
+  in-repo template using it needs a manual, matching edit or it silently
+  renders blank.
+- **Find every place the version string is pinned** — there is no single
+  source of truth, so grep for the old version across the whole repo before
+  declaring the bump done:
+  `grep -rln "<old-version>" . --include="*.yml" --include="*.php" --include="*.md" --include="*.sh" | grep -v "plugins/gatherpress"`.
+  As of the 0.35.0 bump these were: `.github/workflows/e2e-tests.yml`,
+  `.github/workflows/unit-tests.yml`, `.docker/bin/install-test-suite.sh`
+  (which was *already* stale at a different, older version than the two
+  CI workflows — don't assume they're in sync), and this skill's own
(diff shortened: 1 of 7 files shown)


wppoland:
@bor0 two things worth knowing before the production run, both found while testing 0.35.0 + Alpha 0.35.0 on a clean 0.34.1 site this morning. Neither blocks the upgrade, but the first one changes what "Success" means on a network this size.

**1. Alpha's multisite loop stops at 100 sites.** `Setup::fix()` calls bare `get_sites()`, and `WP_Site_Query` defaults to `number => 100`, `orderby => id`, `order => ASC`. So one invocation migrates the 100 oldest blog IDs, then prints `Success: Fixes applied successfully.` regardless. wordcamp.org is well past 100 sites, so the run you posted has almost certainly covered a small fraction of the network.

Filed as GatherPress/gatherpress-alpha#66 with a suggested batching fix, which is the same `number`/`offset` shape `Plugin::clear_all_crons()` already uses in the recurring-events plugin.

Worth knowing: `--url` does not narrow it. `is_multisite()` is still true, so the network branch runs either way — the flag neither restricts it to one site nor extends it past 100.

Until it is fixed, a way to cover the network is to drive the loop yourself, e.g.

```
wp site list --field=url | xargs -n1 -I{} wp gatherpress alpha fix --url={}
```

Each per-site run still enters the network branch and re-walks its own first 100, so it is wasteful, but every site does end up with `gatherpress_alpha_last_version` set. Worth spot-checking a few high-blog-ID sites afterwards:

```
wp option get gatherpress_alpha_last_version --url=<a recently created camp>
```

**2. The gap between the plugin update and the Alpha run is not behaviourally neutral.** GatherPress/gatherpress#2135, filed this morning: 0.35.0 renamed the RSVP mode values, and `Rsvp::is_enabled()` matches the new names with no fallback for the old ones. A site whose stored `rsvp_mode` is still `per_event_on` or `per_event_off` falls through to `return true` before the per-event meta is read, so RSVP reports enabled on every event, including ones where the organiser explicitly turned it off. Verified both directions against 0.34.1, which does read the meta for those modes.

It only affects sites that have actually saved RSVP settings — an absent option takes the new default and is fine. So the sites at risk are exactly the ones where an organiser touched that screen. To see whether any exist before you start:

```
wp site list --field=url | xargs -n1 -I{} sh -c 'wp option get gatherpress_settings --format=json --url=$1 2>/dev/null | grep -q "per_event_o" && echo "$1"' _ {}
```

If that comes back empty, the window is a non-issue for you and item 1 is the only thing to act on. If it does not, those sites are the ones to migrate first.

**The rest of the path came through clean**, on a genuine 0.34.1 site with content authored through the editor: icon blocks all convert, venue-map height migrates and renders at the right size, venue layout keeps `contentSize`/`wideSize`/`justifyContent`, re-running the fix is byte-identical, no PHP notices, no JS console errors.

On the rebase ask: my only open PR here is #1865, which is based on `feature/gatherpress-recurring-events-frontend` rather than production, so it follows whenever #1845 rebases. I will re-verify it against 0.35.0 once that happens rather than rebasing across gedex's branch underneath him.

Props motylanogha

bor0:
Thanks @wppoland. I updated the base branch of #1845

wppoland:
@bor0 one narrow follow-up, since your reply covered the rebase but not the other half, and this one has a user-visible consequence if it went unnoticed.

Did the Alpha run end up covering the whole network? `Setup::fix()` calls bare `get_sites()`, which `WP_Site_Query` caps at 100 by default (`orderby => id, order => ASC`), so a single invocation migrates the 100 oldest blog IDs and still prints `Success: Fixes applied successfully.` Filed upstream as GatherPress/gatherpress-alpha#66, no response there yet.

If only the first 100 ran, the sites past that keep their 0.34.x markup: `gatherpress/icon` blocks with the block now removed from core, and venue maps whose stored numeric height no longer applies. Both are visible on the front end rather than silent.

Cheapest way to tell, on any site you know is recent:

```
wp option get gatherpress_alpha_last_version --url=<a recently created camp>
```

`0.35.0` means it ran there. Nothing, or an older version, means it did not. If a spot check comes back empty I would look at driving the loop per-site until #66 is fixed:

```
wp site list --field=url | xargs -n1 -I{} wp gatherpress alpha fix --url={}
```

Wasteful, since each run re-walks its own first 100, but it does reach every site.

Not urgent if you already checked, just did not want to assume it had been seen.

bor0:
@wppoland yep! We only have two testing groups so far:

- https://events.wordpress.org/group/internal-testing-group/
- https://events.wordpress.org/group/desire/

Thanks for double-checking this 🙇‍♂️

PR #1880, opened 2026-08-11
Address Copilot follow-ups on #1879

bor0:
## Summary
- Adds a regression test that creates a new multisite site and verifies both GatherPress recurring-events occurrence tables get installed via the `wp_initialize_site` hook (`Plugin::on_site_create()`), per [this Copilot comment](https://github.com/WordPress/wordcamp.org/pull/1879#discussion_r3759783768).
- Stops the redirected PHPUnit `error_log` from silently swallowing real errors: a shutdown function re-reads the log file after the run and fails the build if it contains anything other than a recognized `WordCamp\Logger\log()` entry, per [this Copilot comment](https://github.com/WordPress/wordcamp.org/pull/1879#discussion_r3759783818).

## Test plan
- [ ] CI (`unit-tests.yml`) passes, including the new `test_new_site_gets_occurrence_tables_installed` test
- [ ] Confirm CI still passes cleanly (no false positives from the new error-log shutdown check)

### DIFF
--- a/phpunit-bootstrap.php
+++ b/phpunit-bootstrap.php
@@ -12,8 +12,51 @@ if ( file_exists( __DIR__ . '/vendor/autoload.php' ) ) {
  * `error_log` destination configured, PHP writes those to stderr, which CI
  * captures inline with the test output. Route them to a file instead so the
  * `Running unit tests` step only shows PHPUnit's own output.
+ *
+ * This redirects every `error_log()` call in the process, not just the
+ * intentional `WordCamp\Logger\log()` ones -- but ordinary PHP warnings/
+ * notices raised during a test are already turned into failures by
+ * `convertWarningsToExceptions`/`convertNoticesToExceptions` in
+ * phpunit.xml.dist, so they never reach `error_log()` in the first place.
+ * What can still land in this file is either an expected logger entry, or
+ * something that bypassed PHPUnit's handler (e.g. a fatal error). The
+ * shutdown function below re-reads the file once the run ends and fails the
+ * build if it contains anything that isn't a recognized `WordCamp\Logger`
+ * entry, so a real error can't silently accumulate in a file CI never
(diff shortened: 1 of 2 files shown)


phpunit-bootstrap.php
```diff
@@ -12,8 +12,51 @@
...
+ * What can still land in this file is either an expected logger entry, or
+ * something that bypassed PHPUnit's handler (e.g. a fatal error). The
+ * shutdown function below re-reads the file once the run ends and fails the
+ * build if it contains anything that isn't a recognized `WordCamp\Logger`
+ * entry, so a real error can't silently accumulate in a file CI never
+ * publishes.
  */
-ini_set( 'error_log', sys_get_temp_dir() . '/wordcamp-phpunit-error.log' );
+$wordcamp_phpunit_error_log = sys_get_temp_dir() . '/wordcamp-phpunit-error.log';
+ini_set( 'error_log', $wordcamp_phpunit_error_log );
```
gedex, line 29: Nit: the file is never truncated, so a stale non-logger line will fail later runs until it's deleted.

gedex (approved):
LGTM. The shutdown check turns the redirected `error_log` from a silent sink into a real build gate, and the new test genuinely fails if the `wp_initialize_site` hook is removed. One non-blocking nit inline.

gedex (approved):

phpunit-bootstrap.php
gedex:
```suggestion
ini_set( 'error_log', $wordcamp_phpunit_error_log );

// Start each run with an empty log, so a stale entry can't fail a later run.
file_put_contents( $wordcamp_phpunit_error_log, '' );
```

bor0: Good catch — applied your suggested fix in 8b6140a8.

PR #1885, opened 2026-08-12
Groups: cover the nested schedule-unlock case

wppoland:
Follow-up to #1865. One test, no production code.

@bor0 restored two regression tests from `56b9907e` during the rebase, which was the right pair for the mechanism as it stood in that commit. The third one came later, in `9938c1b`, alongside the Copilot review fix — so it was outside the range that got carried over.

The fix from that commit **is** in production: `Plugin::with_schedule_keys_allowed()` exists and `update_end_condition()` delegates to it rather than replacing and clearing the shared allowlist slot. Only the test that guards it is missing, which means a future simplification of `update_end_condition()` back to a direct `self::$schedule_updates_allowed[ $post_id ] = …` / `unset()` pair would reintroduce the silent-loss bug with a green suite.

### What it asserts

An end-condition write nested inside `with_schedule_unlocked()` leaves the surrounding lift intact, so schedule writes that follow it still land, and the lock is back afterwards.

### Verification, and its limit

I could not run this suite: it needs the multi-network fixture `.docker/bin/install-test-suite.sh` builds, and WP bails during bootstrap against a plain multisite install, so please treat CI as the verification rather than my word. `php -l` is clean and `phpcs-branch.php` against `production` reports nothing.

What I could check is that the test is not vacuous. Simulating the allowlist bookkeeping for both shapes of `update_end_condition()` against this exact call sequence:

```
end_condition_fixed    -> write after nested call allowed: YES  | lock released after: yes
end_condition_old      -> write after nested call allowed: NO   | lock released after: yes
```

So the third assertion (`frequency` ends up `monthly`) fails against the pre-fix shape and passes against what is in production now — which is what makes it a regression test rather than a restatement.

Props motylanogha

### DIFF
--- a/public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
+++ b/public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
@@ -172,6 +172,34 @@ final class Test_GatherPress_Recurring_Events extends WP_UnitTestCase {
 		$this->assertSame( 'weekly', get_post_meta( $post_id, Rule::META_PREFIX . 'frequency', true ) );
 	}
 
+	/**
+	 * A nested end-condition write does not cancel the surrounding lift.
+	 *
+	 * Both entry points share one allowlist slot, so an inner lift that
+	 * replaced and then cleared that slot would re-lock the schedule for the
+	 * rest of the outer callback, silently dropping the writes that follow it
+	 * -- the same silent loss the lift exists to prevent, one level down.
+	 */
+	public function test_schedule_unlock_survives_a_nested_end_condition_write(): void {
+		$post_id = $this->create_published_recurring_event();
+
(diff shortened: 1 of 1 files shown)


wppoland:
The red `PHP Tests (8.4)` is infrastructure, not this change — flagging so nobody has to dig.

It failed 1m10s in, during `install-wp-tests.sh`, before PHPUnit ran at all:

```
svn co --quiet https://develop.svn.wordpress.org/tags/7.0.3/tests/phpunit/includes/ ...
svn: E170013: Unable to connect to a repository at URL '...'
svn: E160056: Item index 8 too large in revision 63227
```

`PHP Tests (8.3)` and `PHP Tests (8.5)` both pass on the same commit, and they run the identical suite, so the test itself is green on two of the three PHP versions. A re-run of the 8.4 job should clear it; I cannot trigger one from a fork.

bor0:
Thanks for bringing these back @wppoland! 🙇

bor0 (approved):

PR #1887, opened 2026-08-12
Fix PHP deprecation/warning notices from error monitoring

bor0:
## Summary
Addresses several recurring PHP 8.1+ warnings/deprecations reported by error monitoring.

- **camptix-invoices**: `rename()` was failing with "Operation not permitted" on checkout, because the PDF generator writes to `/tmp` (a different filesystem/mount than the uploads volume) and `rename()` cannot do a cross-device move. Falls back to `copy()` + `unlink()`.
- **wc-post-types**: `is_single_cpt_post()` triggered "Undefined array key 'post_type'" when `$wp_query->query` did not have that key set (e.g. `/speaker/0/`). Added it to the `isset()` check.
- **csstidy** (`class.csstidy-print.php`): "Uninitialized string offset 0" when a selector token was an empty string. Guarded with `?? ''`.
- **lessc** (`lessc.inc.php`):
  - `preg_match(..., null, ...)` deprecated-flags warning (the single most frequent line in the error log) fixed at all 3 call sites by passing `0` instead of `null`.
  - Added `#[\AllowDynamicProperties]` to `lessc` and `lessc_parser` to silence "Creation of dynamic property" deprecations.

Out of scope: the `wordpress/mcp-adapter` trait deprecation and `wporg-mu-plugins/pub-sync` `fputcsv()` deprecation are not part of this repo (separate vendor/private mu-plugins repos).

## Test plan
- [x] `php -l` on all 4 changed files
- [ ] Confirm checkout flow still generates/attaches invoice PDFs correctly (camptix-invoices)
- [ ] Confirm Customizer custom CSS preview/save still works for LESS-processed CSS (lessc/csstidy)
- [ ] Confirm speaker archive/single pages render correctly (wc-post-types)

### DIFF
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/module-custom-css/custom-css/csstidy/class.csstidy-print.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/module-custom-css/custom-css/csstidy/class.csstidy-print.php
@@ -239,7 +239,7 @@ class csstidy_print { // phpcs:ignore
 					if ( $this->parser->get_cfg( 'lowercase_s' ) ) {
 						$token[1] = strtolower( $token[1] );
 					}
-					$out .= ( $token[1][0] !== '@' ) ? $template[2] . $this->htmlsp( $token[1], $plain ) : $template[0] . $this->htmlsp( $token[1], $plain );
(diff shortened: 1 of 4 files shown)


obenland (approved):
Thank you!

PR #1897, opened 2026-08-14
Groups: use three-letter weekday labels in recurrence picker

### LINKED ISSUE #1893: It is impossible to translate the acronym for Tuesday and Thursday if they are written as T in Italian and Spanish.
### Describe the bug

Hello, we're having a problem with the untranslatability of the days of the week. Please modify the text strings for the days of the week when they are expressed with a single letter. In Italian and Spanish, it's not possible to translate T for Tuesday and Thursday because in these languages, the days have different initials. Italian: Tuesday M, Thursday G. In Spanish: Tuesday M, Thursday J. I suggest using Tu and Th instead of T. This way it will be possible to translate and have two different strings for these two days of the week.
These are the strings to be midified:
https://github.com/WordPress/wordcamp.org/blob/production/public_html/wp-content/mu-plugins/gatherpress-recurring-events/assets/editor.js#L80
Line 80 & 82
in https://translate.wordpress.org/projects/meta/wordcamp

### Expected behavior

I suggest using Tu and Th instead of T. This way it will be possible to translate and have two different strings for these two days of the week.

### Screenshots / Screencasts

Image

gedex:
Fixes #1893.

### Problem
The single-letter buttons in the repeating-event weekday picker used `__()` with source strings `'M'`, `'T'`, `'W'`, `'T'`, `'F'`, `'S'`, `'S'`. gettext dedupes by source, so `T` collapsed into one GlotPress entry (Tue + Thu) and `S` into another (Sat + Sun). English works; Italian (Martedì/Giovedì), Spanish (Sábado/Domingo), and any language where those days start with different letters can't be represented — a translator can only supply one string for both meanings.

### Fix
Render the three-letter `dayLabels` (`Mon`, `Tue`, `Wed`, `Thu`, `Fri`, `Sat`, `Sun`) already defined for the ordinal-weekday select as the button content. Those source strings are already distinct in English, so gettext dedup is a non-issue and translators get seven independent entries with no new i18n plumbing to reason about. Drop the parallel `dayButtonLabels` map since its `.short` values were pure indirection back to `dayLabels`; tooltip and `aria-label` still read a separate `dayFullLabels` map for the full name.

Adjust the button style from 30px circles to auto-width pills (min 32px, 6px horizontal padding, 15px radius) so three-character labels fit. Switch the row to `flex-start` with `flex-wrap: wrap` + `rowGap` so longer translated abbreviations wrap gracefully instead of getting pushed to opposite edges of the sidebar.

### Screenshots

_Verified locally against `events.wordpress.test/group/sunshine-coast-qld/` on this branch and on `production`. In English the new pills wrap to 6 + 1 (Mon–Sat on row 1, Sun on row 2); wrap is intentional and left-packed._

**Before** (circles, `M T W T F S S`, collision on T and S):

<kbd>before-weekday-picker</kbd>

**After** (pills, `Mon Tue Wed Thu Fri Sat Sun`):

<kbd>weekday-picker</kbd>

### Deferred — tracked in #1898
This PR unblocks the string-collision problem the Polyglots hit (the actual bug in #1893), but the mu-plugin's translations don't load at runtime at all: no `wp_set_script_translations()` call, no shipped `.pot`/`.po`/`.json`, no `load_muplugin_textdomain()` for the `gpre` domain. So translated weekday values still won't reach the browser until that wiring lands. **Pre-existing across the whole `gatherpress-recurring-events` plugin; not introduced or widened here.** Follow-up: #1898.

### Alternative considered
An earlier version of this PR used `_x()` with per-day contexts. Dropped because (a) three-letter labels make the collision go away without inventing our own context conventions, (b) core's convention is `'Monday initial'`/`'Sunday initial'` (see `wp-includes/class-wp-locale.php`) — my initial `'Monday abbreviation'` naming was inverted from core, and (c) it's the smaller and easier-to-review change. Note also that neither approach would have rendered translated labels without the #1898 wiring — the `_x()` version would have been just as inert as the current `__()` version.

### Testing
- Local: open a `gatherpress_event` at `events.wordpress.test/group/sunshine-coast-qld/wp-admin/`, set frequency to Weekly — buttons render as pills per the screenshots above. Tooltip and aria-label continue to show the full day name.

### DIFF
--- a/public_html/wp-content/mu-plugins/gatherpress-recurring-events/assets/editor.js
+++ b/public_html/wp-content/mu-plugins/gatherpress-recurring-events/assets/editor.js
@@ -20,19 +20,20 @@
 	const weekdayButtonsStyle = {
 		display: 'flex',
 		alignItems: 'center',
-		justifyContent: 'space-between',
+		justifyContent: 'flex-start',
+		flexWrap: 'wrap',
 		gap: '4px',
+		rowGap: '6px',
 	};
 	const weekdayButtonStyle = {
 		display: 'inline-flex',
 		alignItems: 'center',
 		justifyContent: 'center',
-		width: '30px',
-		minWidth: '30px',
+		minWidth: '32px',
 		height: '30px',
-		padding: 0,
+		padding: '0 6px',
 		border: 0,
-		borderRadius: '50%',
+		borderRadius: '15px',
 		background: '#f0f0f0',
 		color: 'var(--wp-admin-theme-color, #3858e9)',
 		fontSize: '13px',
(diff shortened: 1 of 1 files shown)


bor0 (approved):
Tested locally (weekly recurrence picker) — pills render/wrap correctly, tooltip and aria-label still show full day names. This fixes the GlotPress string-collision from #1893. Translations won't actually render until #1898 lands, but that's correctly out of scope here. CI green.

PR #1900, opened 2026-08-14
Groups: add formal ownership-transfer workflow

### LINKED ISSUE #1785: Formal ownership-transfer workflow
Ordinary co-organizer role assignment (Member/Event Organiser/Organiser) works today via the group settings Members tab. This issue is about the additional network-admin-approved handoff of the group owner/admin role itself, which has no workflow yet.

Confirmed in `class-members-controller.php`'s `update_member_role()`: administrators are explicitly excluded from the plugin's own role management —

```php
if ( in_array( 'administrator', $user->roles, true ) ) {
    return new \WP_Error(
        'cannot_manage_administrator',
        __( 'Site administrators must be managed in wp-admin.', 'wporg-groups-frontend' ),
        ...
```

So today, an owner who wants to hand off their group has no supported path at all — not even through wp-admin, unless a network admin manually does it. This issue is about building that formal process.

## Scope

- **Initiate:** current owner (or a network admin, for abandoned-group cases) starts a transfer to a specific candidate — likely required to already be a member of the group, at Organiser tier or above.
- **Accept:** candidate explicitly accepts (prevents an owner from silently offloading responsibility onto someone who didn't agree to it, and prevents a hostile takeover attempt).
- **Approve:** network admin gives final sign-off before the swap actually happens — this is the "formal" part distinguishing it from ordinary role changes.
- **Execute:** on approval, old owner is demoted to Organiser (editor), new owner promoted to Administrator on that site.

## Open questions

- What happens if the current owner is unresponsive/inactive (abandoned group) — can a network admin force a transfer without the current owner's initiation step?
- Does this need an audit trail (who transferred what, when) given it's a sensitive, security-relevant action? Given this session's finding of a real privilege-escalation bug in the *ordinary* role-management path, a formal admin-role transfer flow should be reviewed carefully for the same class of issue (relying on a broad core capability rather than a scoped one) before shipping.

## Acceptance criteria

- [ ] A current owner can initiate a transfer to a specific candidate member.
- [ ] The candidate must explicitly accept before anything changes.
- [ ] A network admin must approve before the swap executes — no path bypasses this approval step.
- [ ] On execution: old owner's role and new owner's role update correctly and atomically (no window where a site has zero or two administrators unexpectedly).
- [ ] Abandoned-group handling (unresponsive current owner) is explicitly decided and implemented, or explicitly out of scope.
- [ ] The new workflow is reviewed for the same class of capability-leak issue found elsewhere in this codebase (a broad core capability grant that also unlocks unrelated wp-admin functionality) before merging.

bor0:
## Summary
Closes #1785. Group owners had no supported way to hand off their group (`Members_Controller::update_member_role()` refuses to touch anyone holding `administrator`). Adds a four-stage workflow instead:

- **Initiate** — the current owner, or a network admin (for an unresponsive owner), nominates a candidate (must be an existing Organiser/`editor`).
- **Accept** — the candidate must explicitly agree before anything changes.
- **Approve** — a network admin gives final sign-off from a new "Ownership Transfers" Network Admin screen.
- **Execute** — roles swap: new owner promoted to `administrator` before the old owner is demoted to `editor`, so the site is never briefly without an admin. Post-swap state is verified; any inconsistency blocks and logs instead of reporting false success.

## How to review this PR

Suggested order: **`group-ownership-transfer.php`** (state machine, capability checks, Network Admin screen — where the actual logic lives) → **`class-ownership-transfer-controller.php`** (thin REST wrapper the front-end calls) → **`ownership-transfer-panel.js`** (UI) → **`group-ownership-transfer-notifications.php`** (email side-effects, hooked on the state machine's actions).

A few decisions worth knowing going in:
- **Split across two mu-plugins.** The state machine lives in the always-loaded `mu-plugins/groups/`, not `wporg-groups-frontend` (which only bootstraps when GatherPress is active) — the Network Admin approval screen has to work regardless.
- **No new core capability grants.** Initiate = `administrator` role check or `is_super_admin()`; approve = `manage_sites`. This is deliberate — avoids the class of bug fixed elsewhere in this plugin (a broad core capability grant that also unlocks unrelated wp-admin functionality).
- **Promote-then-demote, not a DB transaction.** `execute_transfer()` promotes the new owner before demoting the old one (no zero-admin window), then re-reads and verifies — see that function's docblock for why a raw SQL transaction isn't used instead.
- **Per-site locking.** Every transition runs under `with_transfer_lock()`, a `wp_cache_add()` mutex mirroring `Groups\Messaging\with_jobs_lock()`, so concurrent requests can't race the read-then-write against site meta.

## Test plan
- [x] `phpunit --testsuite "WordCamp Groups"` — 223/223 passing (state-machine + REST-layer + notification tests, including decline/cancel/reject and permission-gate cases)
- [x] `phpunit --testsuite "WordCamp MU Plugins"` — 171/171 passing (no regressions)
- [x] `phpcs` clean on all new/changed PHP
- [x] `npm run build` compiles the new frontend panel with no errors
- [x] Manual browser pass of the full happy path on a local multisite group (steps below)

### Manual test steps

Performed against the local dev stack (`sunshine-coast-qld` group at `https://events.wordpress.test/group/sunshine-coast-qld/`), using three separate logged-in sessions:

1. **Set up test users.** Via `wp user set-role`, promoted `organiser1` (existing `editor`) to `administrator` on the group site to act as the current owner, leaving `organiser` (`editor`) as the nominee.
2. **Initiate, as the owner.** Logged in as `organiser1`. On the group's front end, opened **Settings → Members**. A new "Group ownership" panel appears above the member list, with a "Transfer ownership to" dropdown listing only editor-tier members (`organiser`; author/subscriber-tier members correctly excluded). Selected `organiser` and clicked "Request transfer" → panel updated to "Transfer from organiser1 to organiser: Awaiting candidate acceptance," and a "Cancel transfer" action appeared for the owner.
3. **Accept, as the candidate.** Logged out, logged in as `organiser`. Opened the same Members tab — the panel showed "Accept ownership" / "Decline" buttons (and, confirmed separately, that a third user could not accept on the candidate's behalf: `not_the_candidate` error). Clicked "Accept ownership" → panel updated to "Accepted — awaiting network admin approval."
4. **Approve, as a network admin.** Logged out, logged in as `admin` (super admin). Navigated to **Network Admin → Groups → Ownership Transfers** (new submenu). The pending transfer listed correctly: group name, current owner, nominated owner, who initiated it and when, and "Accepted — awaiting your approval." Clicked "Approve" (behind a confirm-before-executing prompt, matching the existing Archive screen's pattern) → "Transfer approved and completed."
5. **Verified the result.** Via `wp user list --fields=ID,user_login,roles` on the group site: `organiser` is now `administrator`, `organiser1` is now `editor` — no other role changes. The "Recently decided transfers" section on the same screen lists the completed transfer (from/to/outcome/date).

Decline, cancel, reject, and the various permission-gate rejections (wrong user accepting, non-owner initiating, non-network-admin approving, ineligible candidate roles) were exercised via the new PHPUnit suite rather than manually clicked through, since they're straightforward negative-path variations of the same flow already covered end-to-end above.

#### Screenshots

| Description | Screenshot |
| --- | --- |
| Owner's view (`organiser1`): the new "Group ownership" panel in Settings → Members, with the transfer-to dropdown before initiating — only editor-tier members are offered as candidates | screenshot-1786711823378-4 |
| After requesting the transfer: the "Cancel transfer" action appears for the owner while it's pending (this is the button whose bug the manual pass caught and fixed) | screenshot-1786711932852-7 |
| Candidate's view (`organiser`): "Accept ownership" / "Decline" buttons on the same pending transfer | screenshot-1786712335275-14 |
| After the candidate accepts: "Accepted — awaiting network admin approval" | screenshot-1786712344155-15 |
| Network Admin → Groups → Ownership Transfers, right after clicking "Approve" (behind the confirm-before-executing prompt) | screenshot-1786712404127-17 |
| The completed transfer listed under "Recently decided transfers" on the same screen, after approval executed the role swap | screenshot-1786712466893-19 |

### DIFF
--- a/public_html/wp-content/mu-plugins/groups/group-ownership-transfer-notifications.php
+++ b/public_html/wp-content/mu-plugins/groups/group-ownership-transfer-notifications.php
@@ -1,0 +1,268 @@
+<?php
+/**
+ * Email notifications for the group-ownership transfer workflow.
+ *
+ * Plain-text `wp_mail()` only, matching `Groups\Messaging\send_message()` —
+ * no templating system. There are no magic-link tokens: accept/decline/
+ * approve are all gated by the acting user's logged-in identity (via REST or
+ * the Network Admin screen), so these emails just point people at the
+ * group's home URL rather than carrying any auth of their own.
+ *
+ * Kept as its own file, separate from the state machine in
+ * `group-ownership-transfer.php`, mirroring how `wporg-groups-frontend`
+ * keeps `inc/notifications.php` separate from the controllers it supports.
+ *
+ * Only loaded on the Groups network, via
+ * `load-other-mu-plugins.php::wcorg_include_network_only_plugins()`.
+ *
+ * @package WordCamp\Groups
+ */
+
+namespace WordCamp\Groups\Ownership_Transfer\Notifications;
+
+use WordCamp\Groups\Ownership_Transfer as Transfer;
+
+defined( 'WPINC' ) || die();
+
+add_action( 'wporg_groups_ownership_transfer_initiated', __NAMESPACE__ . '\notify_candidate_nominated', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_accepted', __NAMESPACE__ . '\notify_admins_awaiting_approval', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_declined', __NAMESPACE__ . '\notify_initiator_declined', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_cancelled', __NAMESPACE__ . '\notify_candidate_cancelled', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_executed', __NAMESPACE__ . '\notify_parties_executed', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_rejected', __NAMESPACE__ . '\notify_parties_rejected', 10, 3 );
+
+/**
+ * Get every network admin's email address and display name.
+ *
+ * @return array<int, array{email: string, name: string}>
+ */
+function get_network_admin_emails(): array {
+	$recipients = array();
+
+	// `site_admins` is shared, mutable network state; a plugin or a stale
+	// `update_site_option()` call elsewhere on the network can leave it in a
+	// shape `get_super_admins()` doesn't expect. Cast defensively rather
+	// than let a malformed option value turn a missed approval-notice email
+	// into a fatal error.
+	foreach ( (array) get_super_admins() as $login ) {
+		$user = get_user_by( 'login', $login );
+
+		if ( $user && is_email( $user->user_email ) ) {
+			$recipients[] = array(
+				'email' => $user->user_email,
+				'name'  => $user->display_name,
+			);
+		}
+	}
+
+	return $recipients;
+}
+
+/**
+ * Send a single plain-text email.
+ *
+ * @param array{email: string, name: string} $recipient Recipient.
+ * @param string                             $subject   Message subject.
+ * @param string                             $body      Message body.
+ * @return bool Whether the mail was handed off successfully.
+ */
+function send_notification( array $recipient, string $subject, string $body ): bool {
+	if ( ! is_email( $recipient['email'] ) ) {
+		return false;
+	}
+
+	$to = $recipient['name']
+		? sprintf( '%s <%s>', $recipient['name'], $recipient['email'] )
+		: $recipient['email'];
+
+	// Passed as an array, exactly as in `Groups\Messaging\send_message()`, so a
+	// display name containing a comma can't smuggle in an extra recipient.
+	$sent = wp_mail(
+		array( $to ),
+		$subject,
+		$body,
+		array( 'Content-Type: text/plain; charset=UTF-8' )
+	);
+
+	if ( ! $sent ) {
+		trigger_error(
+			sprintf(
+				'Failed to send an ownership-transfer notification ("%1$s") to %2$s.',
+				$subject, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Not HTML output; relayed to Slack by this repo's error handler.
+				$recipient['email'] // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Ditto.
+			),
+			E_USER_WARNING
+		);
+	}
+
+	return $sent;
+}
+
+/**
+ * Build the `{name, email}` pair for a user, for use with `send_notification()`.
+ *
+ * @param int $user_id User ID.
+ * @return array{email: string, name: string}|null
+ */
+function recipient_for_user( int $user_id ): ?array {
+	$user = get_userdata( $user_id );
+
+	if ( ! $user || ! is_email( $user->user_email ) ) {
+		return null;
+	}
+
+	return array(
+		'email' => $user->user_email,
+		'name'  => $user->display_name,
+	);
+}
+
+/**
+ * Notify the candidate that they've been nominated.
+ */
+function notify_candidate_nominated( int $site_id, array $pending ): void {
+	$recipient = recipient_for_user( (int) $pending['to_user_id'] );
+
+	if ( ! $recipient ) {
+		return;
+	}
+
+	$group_name = get_blog_option( $site_id, 'blogname' );
+	$group_url  = get_home_url( $site_id, '/' );
+
+	send_notification(
+		$recipient,
+		sprintf( __( 'You\'ve been nominated to take over "%s"', 'wordcamporg' ), $group_name ),
+		sprintf(
+			/* translators: 1: group name, 2: group URL. */
+			__( "You've been nominated to become the owner of the WordPress group \"%1\$s\".\n\nTo accept or decline, visit the group's Members settings:\n%2\$s", 'wordcamporg' ),
+			$group_name,
+			$group_url
+		)
+	);
+}
+
+/**
+ * Notify network admins that a transfer needs their approval.
+ */
+function notify_admins_awaiting_approval( int $site_id, array $pending ): void {
+	$group_name = get_blog_option( $site_id, 'blogname' );
+	$from_user  = get_userdata( (int) $pending['from_user_id'] );
+	$to_user    = get_userdata( (int) $pending['to_user_id'] );
+
+	$body = sprintf(
+		/* translators: 1: group name, 2: current owner, 3: nominated owner, 4: URL of the approval screen. */
+		__( "A group ownership transfer is awaiting your approval.\n\nGroup: %1\$s\nCurrent owner: %2\$s\nNominated owner: %3\$s\n\nReview it here:\n%4\$s", 'wordcamporg' ),
+		$group_name,
+		$from_user ? $from_user->display_name : __( '(deleted user)', 'wordcamporg' ),
+		$to_user ? $to_user->display_name : __( '(deleted user)', 'wordcamporg' ),
+		network_admin_url( 'admin.php?page=' . Transfer\MENU_SLUG )
+	);
+
+	foreach ( get_network_admin_emails() as $recipient ) {
+		send_notification(
+			$recipient,
+			sprintf( __( 'Approval needed: ownership transfer for "%s"', 'wordcamporg' ), $group_name ),
+			$body
+		);
+	}
+}
+
+/**
+ * Notify the initiator that the candidate declined.
+ */
+function notify_initiator_declined( int $site_id, array $pending ): void {
+	$recipient = recipient_for_user( (int) $pending['initiated_by'] );
+
+	if ( ! $recipient ) {
+		return;
+	}
+
+	$group_name = get_blog_option( $site_id, 'blogname' );
+
+	send_notification(
+		$recipient,
+		sprintf( __( 'Ownership transfer declined for "%s"', 'wordcamporg' ), $group_name ),
+		sprintf(
+			/* translators: %s: group name. */
+			__( 'The candidate you nominated to take over "%s" has declined the transfer.', 'wordcamporg' ),
+			$group_name
+		)
+	);
+}
+
+/**
+ * Notify the candidate that their pending nomination was cancelled.
+ */
+function notify_candidate_cancelled( int $site_id, array $pending ): void {
+	$recipient = recipient_for_user( (int) $pending['to_user_id'] );
+
+	if ( ! $recipient ) {
+		return;
+	}
+
+	$group_name = get_blog_option( $site_id, 'blogname' );
+
+	send_notification(
+		$recipient,
+		sprintf( __( 'Ownership transfer cancelled for "%s"', 'wordcamporg' ), $group_name ),
+		sprintf(
+			/* translators: %s: group name. */
+			__( 'Your nomination to take over "%s" has been cancelled.', 'wordcamporg' ),
+			$group_name
+		)
+	);
+}
+
+/**
+ * Notify both parties that the transfer executed.
+ */
+function notify_parties_executed( int $site_id, array $pending ): void {
+	$group_name = get_blog_option( $site_id, 'blogname' );
+	$group_url  = get_home_url( $site_id, '/' );
+
+	$body = sprintf(
+		/* translators: 1: group name, 2: group URL. */
+		__( "Ownership of \"%1\$s\" has been transferred.\n\n%2\$s", 'wordcamporg' ),
+		$group_name,
+		$group_url
+	);
+
+	foreach ( array( $pending['from_user_id'], $pending['to_user_id'] ) as $user_id ) {
+		$recipient = recipient_for_user( (int) $user_id );
+
+		if ( $recipient ) {
+			send_notification(
+				$recipient,
+				sprintf( __( 'Ownership transfer completed for "%s"', 'wordcamporg' ), $group_name ),
+				$body
+			);
+		}
+	}
+}
+
+/**
+ * Notify both parties that a network admin rejected the transfer.
+ */
+function notify_parties_rejected( int $site_id, array $pending, string $reason ): void {
+	$group_name = get_blog_option( $site_id, 'blogname' );
+
+	$body = sprintf(
+		/* translators: 1: group name, 2: optional reason. */
+		__( 'The proposed ownership transfer for "%1$s" was rejected by a network admin.%2$s', 'wordcamporg' ),
+		$group_name,
+		$reason ? "\n\n" . sprintf( /* translators: %s: reason given. */ __( 'Reason: %s', 'wordcamporg' ), $reason ) : ''
+	);
+
+	foreach ( array( $pending['from_user_id'], $pending['to_user_id'] ) as $user_id ) {
+		$recipient = recipient_for_user( (int) $user_id );
+
+		if ( $recipient ) {
+			send_notification(
+				$recipient,
+				sprintf( __( 'Ownership transfer rejected for "%s"', 'wordcamporg' ), $group_name ),
+				$body
+			);
+		}
+	}
+}
--- a/public_html/wp-content/mu-plugins/groups/group-ownership-transfer-notifications.php
+++ b/public_html/wp-content/mu-plugins/groups/group-ownership-transfer-notifications.php
@@ -118,12 +118,40 @@ function recipient_for_user( int $user_id ): ?array {
 }
 
 /**
+ * Report a notification that had no one to go to.
+ *
+ * A transfer whose emails silently go nowhere looks, to everyone involved,
+ * exactly like a transfer nobody has got round to yet: the candidate never
+ * learns they were nominated, or the network admins never learn something is
+ * waiting on them. `send_notification()` already warns when handing a message
+ * to `wp_mail()` fails; this covers the other half -- never getting that far
+ * because there was no usable address to send to.
+ *
+ * @param string $event   Which notification was being sent.
+ * @param int    $site_id Group site ID.
+ * @param string $who     Who the message was meant for.
+ */
+function warn_no_recipient( string $event, int $site_id, string $who ): void {
+	trigger_error(
+		sprintf(
+			'Skipped the "%1$s" ownership-transfer notification for site %2$d: no usable email address for %3$s.',
+			$event, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Not HTML output; relayed to Slack by this repo's error handler.
+			$site_id, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Ditto.
+			$who // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Ditto.
+		),
+		E_USER_WARNING
+	);
+}
+
+/**
  * Notify the candidate that they've been nominated.
  */
 function notify_candidate_nominated( int $site_id, array $pending ): void {
 	$recipient = recipient_for_user( (int) $pending['to_user_id'] );
 
 	if ( ! $recipient ) {
+		warn_no_recipient( 'nominated', $site_id, sprintf( 'the candidate (user %d)', (int) $pending['to_user_id'] ) );
+
 		return;
 	}
 
@@ -146,6 +174,14 @@ function notify_candidate_nominated( int $site_id, array $pending ): void {
  * Notify network admins that a transfer needs their approval.
  */
 function notify_admins_awaiting_approval( int $site_id, array $pending ): void {
+	$recipients = get_network_admin_emails();
+
+	if ( empty( $recipients ) ) {
+		warn_no_recipient( 'awaiting approval', $site_id, 'any network admin' );
+
+		return;
+	}
+
 	$group_name = get_blog_option( $site_id, 'blogname' );
 	$from_user  = get_userdata( (int) $pending['from_user_id'] );
 	$to_user    = get_userdata( (int) $pending['to_user_id'] );
@@ -159,7 +195,7 @@ function notify_admins_awaiting_approval( int $site_id, array $pending ): void {
 		network_admin_url( 'admin.php?page=' . Transfer\MENU_SLUG )
 	);
 
-	foreach ( get_network_admin_emails() as $recipient ) {
+	foreach ( $recipients as $recipient ) {
 		send_notification(
 			$recipient,
 			sprintf( __( 'Approval needed: ownership transfer for "%s"', 'wordcamporg' ), $group_name ),
@@ -175,6 +211,8 @@ function notify_initiator_declined( int $site_id, array $pending ): void {
 	$recipient = recipient_for_user( (int) $pending['initiated_by'] );
 
 	if ( ! $recipient ) {
+		warn_no_recipient( 'declined', $site_id, sprintf( 'the initiator (user %d)', (int) $pending['initiated_by'] ) );
+
 		return;
 	}
 
@@ -198,6 +236,8 @@ function notify_candidate_cancelled( int $site_id, array $pending ): void {
 	$recipient = recipient_for_user( (int) $pending['to_user_id'] );
 
 	if ( ! $recipient ) {
+		warn_no_recipient( 'cancelled', $site_id, sprintf( 'the candidate (user %d)', (int) $pending['to_user_id'] ) );
+
 		return;
 	}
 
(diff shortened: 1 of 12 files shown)


public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/class-ownership-transfer-controller.php
```diff
@@ -0,0 +1,331 @@
...
+	public function view_permissions_check() {
+		if ( ! is_user_logged_in() ) {
+			return new WP_Error(
+				'rest_not_logged_in',
+				__( 'You must be logged in.', 'wporg-groups-frontend' ),
+				array( 'status' => 401 )
+			);
+		}
+
+		if ( ! is_user_member_of_blog() ) {
```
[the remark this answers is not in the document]
bor0: Confirmed and fixed in 5cca90c2. `view_permissions_check()` (which `initiate_permissions_check()` and every other route delegate to) now exempts super admins from the membership requirement:

```php
if ( ! is_super_admin() && ! is_user_member_of_blog() ) {
	return new WP_Error( 'not_a_member', ... );
}
```

This was actually needed on every route, not just initiate — the front-end panel's `GET` call would have hit the same 403 before ever reaching the initiate form, so a non-member super admin couldn't even see the UI to start the abandoned-group flow. Added `test_super_admin_can_view_state_without_membership()` covering both `view_permissions_check()` and `initiate_permissions_check()`.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,863 @@
...
+	if ( get_current_blog_id() !== $site_id ) {
+		return new WP_Error( 'wrong_site_context', 'execute_transfer() must be called with the target site as the current blog.' );
+	}
+
+	$old_owner = get_userdata( (int) $pending['from_user_id'] );
+	$new_owner = get_userdata( (int) $pending['to_user_id'] );
+
+	if ( ! $old_owner || ! $new_owner ) {
+		return new WP_Error( 'transfer_user_missing', __( 'One of the users in this transfer no longer exists.', 'wordcamporg' ), array( 'status' => 400 ) );
+	}
```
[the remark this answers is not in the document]
bor0: Confirmed and fixed in 5cca90c2. `execute_transfer()` now re-validates immediately before mutating roles: re-checks the old owner still holds `administrator`, and reuses `validate_candidate()` on the new owner (covers removed/demoted/already-promoted). Added `test_cannot_execute_if_candidate_no_longer_a_member()` and `test_cannot_execute_if_owner_no_longer_administrator()`.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,863 @@
...
+	$record = array(
+		'from_user_id' => $from_user_id,
+		'to_user_id'   => $to_user_id,
+		'status'       => STATUS_PENDING_ACCEPTANCE,
+		'initiated_by' => $initiated_by,
+		'initiated_at' => time(),
+		'accepted_at'  => null,
+	);
+
+	set_pending_transfer( $site_id, $record );
```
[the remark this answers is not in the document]
bor0: Confirmed and fixed in 5cca90c2. Added a per-site `wp_cache_add()`-based lock (`with_transfer_lock()`) around every transition — initiate/accept/decline/cancel/execute/reject — mirroring `Groups\Messaging\with_jobs_lock()`'s identical reasoning, scoped to one site instead of one global job queue. A transition that can't acquire the lock returns a `transfer_busy` WP_Error instead of racing the read-then-write. Added `test_concurrent_transition_is_refused_while_locked()`.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,863 @@
...
+/**
+ * Render the Ownership Transfers screen.
+ */
+function render_page(): void {
+	if ( ! current_user_can_approve() ) {
+		wp_die( esc_html__( 'Sorry, you are not allowed to decide group ownership transfers.', 'wordcamporg' ) );
+	}
+
+	$pending_rows = get_sites_with_pending_transfers();
+	$recent_rows  = get_recent_decided_transfers();
```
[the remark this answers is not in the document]
bor0: Confirmed and fixed in 5cca90c2. Both listing functions now query `wp_blogmeta` directly for sites carrying the pending/history meta key (`get_group_sites_with_meta_key()`) instead of iterating every group site via `Archive\get_group_sites( true )`. Scales with sites that actually have transfer activity rather than total network size. Added `test_listing_functions_only_include_sites_with_transfer_activity()`, which also asserts an untouched group site never appears.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/style.scss
```diff
@@ -429,6 +429,51 @@
 	}
 }
 
+// === Ownership transfer panel (inside Members tab) ===
+
+.wporg-ownership-transfer {
+	margin-bottom: 20px;
+	padding-bottom: 16px;
+	border-bottom: 1px solid #e0e0e0;
```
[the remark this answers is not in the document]
bor0: Verified — `#656a71`/`#e0e0e0` do map to `charcoal-4`/(a close but not exact match for) the border token per `docs/design/groups-site.md`. I'm leaving this one as-is for this PR though: the exact same raw-hex pattern (`#656a71`, `#d9d9d9`/`#e0e0e0`) is already used 14 other times throughout this same pre-existing file (lines 72–428, none of it touched by this PR). Converting only my 2 new lines would leave the file inconsistent in both directions rather than fixing anything — this reads as a pre-existing, file-wide design-token migration gap that predates this PR, not something introduced here. Happy to do it as a follow-up cleanup PR across the whole file if the team wants it.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/ownership-transfer-panel.js
```diff
@@ -0,0 +1,219 @@
...
+					history.map( ( entry, index ) =>
+						h(
+							'li',
+							{ key: index },
+							sprintf(
+								/* translators: 1: previous owner, 2: new/nominated owner, 3: outcome. */
+								__( '%1$s → %2$s: %3$s', 'wporg-groups-frontend' ),
+								entry.fromUserName,
+								entry.toUserName,
+								entry.status
```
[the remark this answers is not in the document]
bor0: Confirmed and fixed in 5cca90c2. Added a `FINAL_STATUS_LABELS` map alongside the existing `STATUS_LABELS` (which already covered the pending statuses) and used it for history rows, with a fallback to the raw value for anything unrecognised.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,863 @@
...
+							$site_id   = (int) $row['site']->blog_id;
+							$entry     = $row['entry'];
+							$from_user = get_userdata( (int) $entry['from_user_id'] );
+							$to_user   = get_userdata( (int) $entry['to_user_id'] );
+							?>
+							<tr>
+								<td><?php echo esc_html( get_blog_option( $site_id, 'blogname' ) ?: $row['site']->domain . $row['site']->path ); ?></td>
+								<td><?php echo esc_html( $from_user ? $from_user->display_name : __( '(deleted user)', 'wordcamporg' ) ); ?></td>
+								<td><?php echo esc_html( $to_user ? $to_user->display_name : __( '(deleted user)', 'wordcamporg' ) ); ?></td>
+								<td><?php echo esc_html( $entry['final_status'] ?? '' ); ?></td>
```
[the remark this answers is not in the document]
bor0: Confirmed and fixed in 5cca90c2. Added `get_final_status_label()` (PHP-side equivalent of the JS `FINAL_STATUS_LABELS` map) and used it for the wp-admin "Outcome" column. Covered by `test_get_final_status_label_covers_every_final_status()`.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer-notifications.php
```diff
@@ -0,0 +1,268 @@
...
+use WordCamp\Groups\Ownership_Transfer as Transfer;
+
+defined( 'WPINC' ) || die();
+
+add_action( 'wporg_groups_ownership_transfer_initiated', __NAMESPACE__ . '\notify_candidate_nominated', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_accepted', __NAMESPACE__ . '\notify_admins_awaiting_approval', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_declined', __NAMESPACE__ . '\notify_initiator_declined', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_cancelled', __NAMESPACE__ . '\notify_candidate_cancelled', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_executed', __NAMESPACE__ . '\notify_parties_executed', 10, 2 );
+add_action( 'wporg_groups_ownership_transfer_rejected', __NAMESPACE__ . '\notify_parties_rejected', 10, 3 );
```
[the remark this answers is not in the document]
bor0: Confirmed and fixed in 5cca90c2. Added `test-group-ownership-transfer-notifications.php` following the exact `pre_wp_mail`-capture pattern from `test-notifications.php`/`test-network-messaging.php`: one focused test per transition hook asserting exactly the intended recipients (and no one else — e.g. `test_accept_notifies_network_admins_only()` explicitly asserts the owner/candidate are *not* among the recipients), plus a reason-in-body check for reject and a `wp_mail()`-failure/warning test.

raicem (approved):
Thanks for the PR @bor0! Tested manually and reviewed all the code.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,1093 @@
...
+			array( 'status' => 409 )
+		);
+	}
+
+	$candidate_check = validate_candidate( $new_owner->ID, $old_owner->ID, $site_id );
+	if ( is_wp_error( $candidate_check ) ) {
+		return $candidate_check;
+	}
+
+	$new_owner->set_role( 'administrator' );
```
gedex, line 784: promotion result isn't checked before 641 demotes. first write misses, second lands → single-owner group has zero admins.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,1093 @@
...
+
+	// Re-read both users; `set_role()` above may have been intercepted by
+	// another plugin's hook, so the in-memory objects aren't trustworthy.
+	$old_owner_after = get_userdata( (int) $pending['from_user_id'] );
+	$new_owner_after = get_userdata( (int) $pending['to_user_id'] );
+
+	$old_is_admin = $old_owner_after && in_array( 'administrator', $old_owner_after->roles, true );
+	$new_is_admin = $new_owner_after && in_array( 'administrator', $new_owner_after->roles, true );
+
+	if ( $old_is_admin || ! $new_is_admin ) {
```
gedex: doesn't check the old owner actually ended up `editor`, so zero-roles passes as success. and this branch bails without undoing either write or clearing the record — candidate keeps `administrator`, retry hits `candidate_already_administrator` from 635, reject doesn't revoke it.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,1093 @@
...
+
+	if ( ! $acquired ) {
+		return new WP_Error(
+			'transfer_busy',
+			__( "This group's ownership transfer is being updated by another request. Please try again.", 'wordcamporg' ),
+			array( 'status' => 409 )
+		);
+	}
+
+	try {
```
gedex, line 158: 10s lock with the mail fan-out inside it. expires mid-section, then `finally` frees someone else's lock. network-messaging.php:347 sends outside the lock on purpose.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,1093 @@
...
+
+	$sites = array();
+
+	foreach ( $site_ids as $site_id ) {
+		$site_id = (int) $site_id;
+		$site    = get_site( $site_id );
+
+		// Same exclusions as `Archive\get_group_sites()`: only sites on this
+		// network, never the placeholder root, never a deleted site.
+		if ( ! $site || GROUPS_ROOT_BLOG_ID === $site_id || GROUPS_NETWORK_ID !== (int) $site->network_id || $site->deleted ) {
```
gedex: `Archive\get_group_sites()` also filters `spam`/`archived` (wporg-groups-archive.php:54). archived group still gets an Approve button. test:502's fixture has no transfer meta so it can't catch this.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,1093 @@
...
+ * the existing Archive screen paginates at `wporg-groups-archive.php::PER_PAGE`
+ * precisely because it doesn't scale otherwise.
+ *
+ * @param string $meta_key Blog meta key to look for.
+ * @return \WP_Site[]
+ */
+function get_group_sites_with_meta_key( string $meta_key ): array {
+	global $wpdb;
+
+	$site_ids = $wpdb->get_col(
```
gedex, line 1121: `last_error` unchecked, so a failed query renders "No transfers are awaiting approval".

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,1093 @@
...
+
+/**
+ * Move a decided transfer from "pending" into the capped history list.
+ *
+ * @param int    $site_id      Group site ID.
+ * @param array  $pending      The pending record being decided.
+ * @param string $final_status One of 'declined', 'cancelled', 'completed', 'rejected'.
+ * @param array  $extra        Extra fields to merge in (e.g. `decided_by`, `reason`).
+ */
+function finalize_transfer( int $site_id, array $pending, string $final_status, array $extra = array() ): void {
```
gedex: meta write returns dropped, caller still returns true. at 682 the roles are already swapped, so a failed delete → "approved and completed" with the row still sitting there.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,1093 @@
...
+		static function () use ( $site_id, $user_id ) {
+			return decline_transfer_unlocked( $site_id, $user_id );
+		}
+	);
+}
+
+/**
+ * @see decline_transfer() -- runs under that function's per-site lock.
+ */
+function decline_transfer_unlocked( int $site_id, int $user_id ) {
```
gedex, line 614: no status guard, unlike 456. candidate can decline a record already at `pending_approval`.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
```diff
@@ -0,0 +1,1093 @@
...
+		'rejected'  => __( 'Rejected', 'wordcamporg' ),
+	);
+
+	return $labels[ $status ] ?? $status;
+}
+
+/**
+ * Process an approve/reject decision from the Network Admin screen.
+ */
+function handle_decision(): void {
```
gedex, line 1212: no test touches this. execute/reject have no cap check of their own, so dropping 898-900 or the nonce at 906 stays green.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer-notifications.php
```diff
@@ -0,0 +1,268 @@
...
+	return $sent;
+}
+
+/**
+ * Build the `{name, email}` pair for a user, for use with `send_notification()`.
+ *
+ * @param int $user_id User ID.
+ * @return array{email: string, name: string}|null
+ */
+function recipient_for_user( int $user_id ): ?array {
```
gedex, line 107: null recipient is a bare return, no `trigger_error` unlike 87. same at 162 when there are no admin emails at all.

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/class-ownership-transfer-controller.php
```diff
@@ -0,0 +1,339 @@
...
+	 * abandoned-group path (`Transfer\current_user_can_initiate()`'s
+	 * `is_super_admin()` branch) actually works for a network admin who
+	 * isn't personally a member of the group in question — otherwise this
+	 * check would 403 them before that override is ever consulted, on
+	 * every route including the initial `GET` the front-end panel needs
+	 * just to render the initiate form.
+	 *
+	 * @return true|WP_Error
+	 */
+	public function view_permissions_check() {
```
gedex, line 136: membership only; every other settings route uses `current_user_can_manage_group_settings()` (rest.php:280). members can read the rejection `reason`.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/ownership-transfer-panel.js
```diff
@@ -0,0 +1,228 @@
...
+export default function OwnershipTransferPanel() {
+	const [ loading, setLoading ] = useState( true );
+	const [ state, setState ] = useState( null );
+	const [ notice, setNotice ] = useState( '' );
+	const [ noticeType, setNoticeType ] = useState( 'success' );
+	const [ busy, setBusy ] = useState( false );
+	const [ candidateId, setCandidateId ] = useState( '' );
+	const [ fromUserId, setFromUserId ] = useState( '' );
+
+	const fetchState = useCallback( () => {
```
gedex, line 82: falsy body leaves state null with no notice → panel just disappears. 94 destructures with no defaults, no boundary at members-tab.js:128, so a missing `history` kills the whole tab.

gedex (commented):
things worth to fix before merge:

`execute_transfer_unlocked()` — a half-completed swap can leave the candidate holding `administrator` on a transfer that reports failure and is then rejected, with the record stuck where 627/635 reject every retry.

`with_transfer_lock()` — 10s TTL wrapped around the `wp_mail()` fan-out, so the lock can expire mid-section and the `finally` can free another request's lock.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
gedex:
```suggestion
	$old_is_candidate = $old_owner_after && in_array( CANDIDATE_ROLE, $old_owner_after->roles, true );

	if ( $old_is_admin || ! $new_is_admin || ! $old_is_candidate ) {
```

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
gedex:
```suggestion
		if ( ! $site || GROUPS_ROOT_BLOG_ID === $site_id || GROUPS_NETWORK_ID !== (int) $site->network_id || $site->deleted || $site->spam || $site->archived ) {
```

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
bor0: Confirmed and fixed in e0aebdc2. `execute_transfer_unlocked()` now re-reads the candidate right after `set_role( 'administrator' )` and bails with `transfer_promotion_failed` if it didn't land — before the demotion runs at all. Nothing is mutated, the record stays pending, and it can just be retried. `test_failed_promotion_does_not_demote_the_old_owner()` asserts no `set_user_role` fires for the old owner in that case.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
bor0: Confirmed and fixed in e0aebdc2. Took your suggestion for the `$old_is_candidate` check, and added the rollback for the bail-out: `roll_back_execution()` restores the old owner to `administrator` first (so the group is never briefly admin-less during the rollback either), then puts the candidate back on `CANDIDATE_ROLE`, and verifies both. So the candidate no longer keeps `administrator`, a retry no longer hits `candidate_already_administrator`, and the error message says whether the rollback succeeded. `test_demotion_that_lands_nowhere_is_rolled_back()` covers the zero-roles case end to end.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
bor0: Confirmed and fixed in e0aebdc2. The `do_action` for each transition moved out of the `*_unlocked()` worker and into the public wrapper, which fires it after `with_transfer_lock()` returns — same reasoning as network-messaging.php:347. The workers now return the record so the wrapper has something to pass to the hook. The lock itself also stores a unique token and only deletes it if `wp_cache_get()` still returns that token, so an expired lock can't be freed out from under whoever holds it now.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
bor0: Confirmed and fixed in e0aebdc2 — took your suggestion verbatim. Also added `test_archived_and_spammed_groups_are_excluded_from_listings()`, which uses a fixture that does have transfer meta (the gap you pointed out in test:502): it initiates a real transfer, flips `archived` then `spam`, asserts the group drops off the approvals screen each time, and comes back when the flag clears.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
bor0: Confirmed and fixed in e0aebdc2. `get_group_sites_with_meta_key()` now checks `$wpdb->last_error` and returns a `transfer_query_failed` WP_Error (plus a warning); both listing functions propagate it, and `render_page()` shows an error notice and swaps the empty-state text instead of "No transfers are awaiting approval." `test_a_failed_lookup_is_reported_rather_than_read_as_empty()` breaks the query through the `query` filter and asserts all three surfaces report it.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
bor0: Confirmed and fixed in e0aebdc2. `finalize_transfer()` now returns bool, and writes history *before* clearing the pending record — if the history write fails it returns early without deleting, so the worst case is a still-pending transfer rather than a decided one that vanished. Every caller checks it. At the execute site specifically it is deliberately not rolled back (the roles really are swapped; undoing that because bookkeeping failed is worse) — instead it warns loudly and returns an error saying the ownership change is done but the leftover request needs rejecting to clear. `test_a_decision_that_cannot_be_saved_is_reported()` covers the reject path.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
bor0: Confirmed and fixed in e0aebdc2. `decline_transfer_unlocked()` now has the same `STATUS_PENDING_ACCEPTANCE` guard as accept, returning `transfer_not_awaiting_acceptance` with a hint to ask a network admin to reject instead. `test_decline_is_refused_once_the_transfer_has_been_accepted()` asserts the accepted record survives untouched.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer.php
bor0: Confirmed and fixed in e0aebdc2. Added three tests driving `handle_decision()` directly: `test_handle_decision_requires_manage_sites()` (non-network-admin → WPDieException), `test_handle_decision_requires_a_valid_nonce()` (which also asserts the transfer is still pending and neither role changed), and `test_handle_decision_rejects_an_unknown_decision()`. Dropping either guard now fails.

public_html/wp-content/mu-plugins/groups/group-ownership-transfer-notifications.php
bor0: Confirmed and fixed in e0aebdc2. Added `warn_no_recipient()` and called it from all six notifiers — the four single-recipient bare returns, both `foreach` skips, and the no-network-admins case. Two tests: one for a nomination whose candidate has no usable address, one for an acceptance on a network with no reachable admins (the case where the transfer would otherwise just sit there waiting on an approval nobody was told about).

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/class-ownership-transfer-controller.php
bor0: Confirmed and fixed in e0aebdc2. `view_permissions_check()` now requires `current_user_can_manage_group_settings()` on top of membership, matching rest.php:280. The candidate holds `CANDIDATE_ROLE` (`editor`) by definition, so accept/decline are unaffected; super admins still short-circuit above it for the abandoned-group path. `test_member_without_settings_access_cannot_view_state()` asserts a subscriber gets `rest_forbidden` and an editor does not.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/ownership-transfer-panel.js
bor0: Confirmed and fixed in e0aebdc2. `assertUsable()` throws on a falsy/non-object body in both `fetchState()` and `runAction()`, so it lands in the existing error notice instead of leaving `state` null with nothing rendered. The destructure at 94 now has defaults, plus an `asList()` pass over the three arrays — a default only covers `undefined`, and an explicit `null` would still take the Members tab down given there is no boundary at members-tab.js:128.

PR #1904, opened 2026-08-18
Payments Network: batch the network-wide payment index aggregation

bor0:
## Summary

`Payment_Requests_Dashboard::aggregate()` (run daily via the `wordcamp_payments_aggregate` cron event) iterates all ~3000 sites on the network via `switch_to_blog()` in a single request to rebuild the payments index table. This has been showing up in prod as `E_USER_WARNING: Peak memory usage at 90.6%` / `92.2%`, with `current_action() === 'switch_blog'`, e.g.:

```
Peak memory usage at 90.6%. Current action/filter: switch_blog.
Stack Trace: do_action_ref_array('wordcamp_payments_aggregate'), ... Payment_Requests_Dashboard::aggregate, switch_to_blog, do_action('switch_blog'), ...
```

The code already had a comment flagging this ("This may need to be refactored in the future to use batching to avoid out-of-memory-issues").

- Split the run into batches of 250 blogs (`LIMIT`/`OFFSET`), tracked via an `$offset` arg.
- When a batch is full, self-reschedule the next batch via `wp_schedule_single_event( time(), ... )` instead of continuing the loop in the same process — the next batch runs as its own `wp-cron.php` request, so peak memory per request stays bounded.
- Only truncate the index table on the first batch (`$offset === 0`), not on every batch.

## Test plan

- [ ] `php -l` passes on the changed file (done locally).
- [ ] No existing test coverage for this class; none added, consistent with existing scope.
- [ ] Manually trigger `?wcp-debug-network` (or `do_action('wordcamp_payments_aggregate')` via WP-CLI) on a multisite install with several dozen+ sites and confirm:
  - the index table is truncated once, not per batch
  - all blogs eventually get indexed across multiple batches
  - a follow-up single event is scheduled/consumed correctly until done

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments-network/includes/payment-requests-dashboard.php
+++ b/public_html/wp-content/plugins/wordcamp-payments-network/includes/payment-requests-dashboard.php
@@ -117,6 +117,13 @@ class Payment_Requests_Dashboard {
 	 *
 	 * Runs in batches of self-rescheduled single events to keep peak memory bounded, since a single request
 	 * iterating all sites was pushing peak memory usage above 90%.
+	 *
+	 * Each blog's rows are replaced as the run reaches that blog, rather than the whole table being emptied
+	 * up front. Spreading a rebuild over many requests rules out the empty-table approach: `generate_payment_report()`
+	 * builds the actual wire/SEPA payment files straight out of this index, so a run that emptied the table
+	 * and refilled it over the following minutes would let an export taken in that window silently come out
+	 * short -- or completely empty. Replacing per blog keeps the index continuously readable, and makes
+	 * indexing a blog idempotent, so a resumed batch can safely re-cover ground it already did.
 	 */
 	public static function aggregate( $after_blog_id = 0 ) {
 		global $wpdb, $wp_object_cache;
@@ -127,10 +134,10 @@ class Payment_Requests_Dashboard {
 		require_once WP_PLUGIN_DIR . '/wordcamp-payments/includes/payment-request.php';
 		WCP_Payment_Request::register_post_statuses();
 
-		// A fresh run is about to empty the index, which makes any continuation still queued from an
-		// earlier run that never finished stale: it would index its remaining blogs into a table it no
-		// longer owns, duplicating every row this run's own batches also insert. Retire those first,
-		// before this run claims the state option for itself.
+		// Retire any continuation still queued from an earlier run that never finished, before this run
+		// claims the state option for itself. Two chains walking the network at once would interleave
+		// their per-blog delete/insert pairs, and a delete landing between the other chain's delete and
+		// its inserts leaves that blog indexed twice.
 		if ( 0 === $after_blog_id ) {
 			self::abandon_stalled_chain();
 		}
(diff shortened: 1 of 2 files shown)


obenland:
> Persist the cursor/run state independently and resume it from a recurring watchdog

That would be my feedback, too. Making it resumable at any point.

bor0:
Parking this for now — shifting focus elsewhere, and I'd rather not land a change of this shape into the payments export path while it isn't getting attention. Closing rather than leaving it open so it's not sitting in anyone's review queue. The branch (`fix/payments-aggregate-batch-memory`) is preserved and this reopens with everything intact.

Writing down the state of it so whoever picks it up (probably me) doesn't have to re-derive it.

## What's in here

The PR is small by line count — ~196 lines of production code in one file — but it bundles five independent changes to the job that builds the index `generate_payment_report()` turns into wire/SEPA files:

1. **Idempotent per-blog refresh.** `TRUNCATE` → per-blog `DELETE` + `prune_orphaned_rows()`.
2. **Batching.** Keyset pagination on `blog_id` + self-rescheduled continuations. The actual memory fix.
3. **Durable cursor.** New `wcp_network_aggregate_run` site option.
4. **Watchdog.** New hourly network cron to resume a stalled chain.
5. **Dropping `LIMIT 3000`.** Trunk only indexes the 3000 most-recently-updated blogs; this indexes all of them.

Only (2) is the stated goal of the PR.

## Why it should be split before it lands

Suggested order, each independently deployable and revertable:

- **First: per-blog refresh (1).** Same single request, same memory profile, no cron changes, no new options — behaviorally almost a no-op. But it fixes a real pre-existing bug on its own: today an export taken while `aggregate()` is between the `TRUNCATE` and the refill reads a partially-filled table and silently comes out short. It's also the prerequisite that makes an interrupted batch harmless, which is what makes (2) safe.
- **Then: batching + cursor (2, 3),** landing against an index that's already idempotent.
- **Then: watchdog (4).**

## Open items to resolve before it lands

- **`LIMIT 3000` is an unannounced scope change.** Nothing in the title or description says the index starts covering sites it never has. If the network is over 3000 blogs, rows appear in the dashboard and in exports that weren't there before — a functional change riding along inside a memory fix. Needs to be its own reviewed decision, or split out. **First thing to check: how many sites are actually in the network?** If it's under 3000 this is just dead-code removal and none of this matters.
- **`watchdog()` has no attempt cap.** It reschedules the same cursor indefinitely. If a batch reliably dies — one blog with a pathological post count — the failure mode goes from "fails once a day" to "retries hourly, forever." Wants a max-attempts or at minimum a log line.
- **`AGGREGATE_STALL_TIMEOUT` is load-bearing with no in-batch heartbeat.** `save_run_state()` only writes at batch start and end, so telling "running now" from "died" rests entirely on the timeout. A batch that legitimately ran past an hour would get a second chain scheduled alongside it — the exact double-indexing the code warns about. PHP's max execution time makes this unreachable in practice, so it holds, but it holds by accident rather than by design. Worth either a periodic heartbeat inside the loop or a comment making the dependency on request timeout explicit.
- **`abandon_stalled_chain()` uses `_get_cron_array()`,** a private core function. Minor, but flagging it.

## What looks right

The subtle parts are correct as far as I can tell: the int cast on the cursor so `wp_next_scheduled()` actually matches (`0f93e0b6`), recording run state *before* doing any work so a batch that dies mid-request is still recoverable, and retiring a stale chain before a fresh run claims the state option. Test coverage of the interesting cases — mid-run export completeness, `save_post` racing the run, re-run dedup, orphan pruning — is there and worth keeping.

**Note that the underlying problem is still live in production:** the aggregate run pushes peak memory above 90%. Parking this doesn't fix that.

PR #1902, opened 2026-08-18
Groups: per-group event + RSVP history export (CSV/JSON) (#1780)

gedex:
Per-group half of #1780: Organisers can export event + RSVP history as CSV or JSON.

New **Export** tab in the settings modal (Organiser-only, enforced server-side): pick columns, events, and a date range — all default to everything. Anonymous RSVPs become a token, never a name/email. Formula injection escaped. Recurring RSVPs attributed per-occurrence.

**Not here:** network-wide export (PR 2).

**Review order:** `inc/export.php` → `tests/test-export.php` → `export-tab.js`

## Test instructions

As an Organiser (editor, e.g. `organiser1`) on a group site:

1. Homepage → **Settings** → **Export** tab.
2. Click **Download CSV** with everything default → opens in a spreadsheet, 16 columns, one row per RSVP, zero-RSVP events still listed.
3. Pick a couple of columns (e.g. *Event title*, *RSVP status*) → download → only those columns.
4. Pick one event, set dates to *Past events* / a custom range → download → only matching rows.
5. **Download JSON** → parses; RSVPs nested under events.
6. If the group has an anonymous RSVP: attendee shows as `anonymous-…`, no name/username anywhere in the file.

Permissions (should all fail):

7. Log in as author or subscriber → no Settings button; direct `GET /wp-json/wporg-groups/v1/export` → 403.
8. Logged out → same URL → 401.

## Screenshots

<kbd>image</kbd>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1905, opened 2026-08-18
Groups: network-wide event + RSVP-count export (#1780)

gedex:
Network-admin half of #1780 (per-group half: #1902 — **stacked on it, review/merge that first**).

New Network Admin screen (Groups → Export Events, `manage_network` only): exports every group's event history with per-status RSVP counts as CSV or JSON. **Aggregate only — no attendee names, logins, or emails**, locked in by a no-PII test. Runs on cron in batches (10 sites/run, `network-messaging.php`'s claim/commit pattern) so it can't time out; single-job model, download when ready from a stored artifact. Failed sites are reported in the summary, never silently missing.

**Review order:** `network-export.php` → `tests/test-network-export.php`

## Test instructions

Sign in as the seeded super admin — `admin` / `password` — at
<https://events.wordpress.test/group/wp-login.php>
(note it holds only `subscriber` on the group sites; the `manage_network` this screen needs comes from being a super admin of the groups network).

Then open the export screen:
<https://events.wordpress.test/group/wp-admin/network/admin.php?page=wporg-groups-export>
(also reachable via Network Admin → **Groups** → **Export Events**)

1. **Start export** → "queued" notice.
2. Reload — small networks finish in one cron tick; larger ones show "X of Y sites processed" with a **Cancel export** button.
3. **Download CSV** → 12 columns (group, event, dates, `recurrence_id`, `occurrence_status`, venue, 3 counts). Recurring series get one row per occurrence, each with its own date and counts; cancelled occurrences are marked, not hidden. **Download JSON** → groups nested over events.
4. Check the file: no attendee names, logins, or emails anywhere.
5. Start another export while one is running → "already running" notice; **Cancel export** clears it and leaves the last completed export downloadable.

Should all fail — sign in as `organiser1` / `password` (editor on a group site):

6. The screen above → 403; both actions posted directly to `https://events.wordpress.test/group/wp-admin/admin-post.php` (`action=wporg_groups_start_network_export`, `action=wporg_groups_cancel_network_export`) → 403.
7. Download with a stale or bogus nonce → 403:
   `https://events.wordpress.test/group/wp-admin/admin-post.php?action=wporg_groups_download_network_export&format=csv&_wpnonce=bogus`

## Screenshots

<kbd>image</kbd>


public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,781 @@
...
+
+			if ( empty( $job ) ) {
+				return null;
+			}
+
+			// Claiming removes the job for the duration of the batch, so a
+			// concurrent start can't resurrect stale progress. It also means
+			// a concurrent start would begin a *new* run — which is why the
+			// commit below uses `$steal` rather than giving up.
+			delete_site_option( JOB_OPTION );
```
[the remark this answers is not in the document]
gedex: fixed in d7f65c22

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,781 @@
...
+
+/**
+ * Handle the start-export form submission.
+ */
+function handle_start_export(): void {
+	if ( ! current_user_can_export() ) {
+		wp_die( 'You do not have permission to export group data.', 403 );
+	}
+
+	check_admin_referer( FORM_ACTION );
```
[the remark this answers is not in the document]
gedex: fixed in d7f65c22

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,781 @@
...
+ * @return array[]|WP_Error Aggregate rows, or the collector's error.
+ */
+function collect_site_rows( int $site_id ) {
+	// The switch sits inside the try: it pushes the switched-blog stack
+	// before firing its action, so a Throwable from a `switch_blog` listener
+	// would otherwise leave the process stuck on the wrong blog.
+	try {
+		switch_to_blog( $site_id );
+
+		$data = \WordCamp\Groups\Frontend\Export\collect_export_data();
```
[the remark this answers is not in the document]
gedex: pushing back — per-site cost is identical to the per-group export that already runs synchronously in one request (#1902), sites are collected one at a time so memory is bounded by the largest single site, and group sites are tens of events. If one outgrows that, the per-group export breaks first and both get fixed together.

gedex (commented):
Suite is green locally (14 tests, 64 assertions). The privacy design holds up — `test_no_pii_in_artifact_or_output` genuinely substantiates the aggregate-only claim.

The lease protocol does not establish ownership. `:377` guards on an export-wide ID, so a slow batch commits stale state and clears another run's lease; it needs a per-batch token. Related: `:298` drops the batch without rescheduling.

Three ways to lose a run with no signal: unchecked artifact write (`:383`), a fatal wedging the job forever with the previous artifact already deleted (`:776` / `:182`), and `(string) wp_json_encode()` (`:548`). The memory pressure behind the fatal case is self-inflicted (`:418`).

"Never silently missing" is not quite true yet: the CSV carries no failed-sites marker, `site_count` counts failures, and a site deleted between queue time and processing vanishes quietly.

Mutation-tested 12 production changes; 8 survived. The two that matter most are the download nonce action and cross-batch rows. Both are blocked by `redirect_with_notice()` having no test interception, so that is worth fixing first.

Checked and fine: the `admin_url`/`admin_post` routing (matches shipped `network-messaging.php:709`), the `switch_to_blog`-inside-`try` reasoning at `:412-414`, `handle_download()`'s header ordering, `render_page()`'s array access, and the `$crashed` branch itself.

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,808 @@
...
+	$finished = ! $crashed && empty( $job['pending_sites'] );
+
+	with_job_lock(
+		function () use ( $job, $finished ) {
+			$current = get_job();
+
+			// Only commit onto our own job. An expired lease means another
+			// run may have taken over (or an admin restarted the export);
+			// their copy of the progress is the durable one, not ours.
+			if ( empty( $current ) || $current['id'] !== $job['id'] ) {
```
gedex: `$job['id']` is the same for every batch, so this guard can't tell whose lease it is. An overrunning run passes it and commits stale state: the cursor rewinds past batches that committed while it was collecting, and `:385` clears the live lease of whichever run is collecting right now, so the next tick re-claims the same sites. Replacing this guard with `if ( false )` keeps all 14 tests green.

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,808 @@
...
+				return;
+			}
+
+			if ( $finished ) {
+				record_artifact( $job );
+				delete_site_option( JOB_OPTION );
```
gedex: `update_site_option()`'s return in `record_artifact()` is ignored and this delete runs regardless. A failed artifact write loses the whole run: `render_summary()` returns early at `:719` and the screen is identical to never having run an export. `:182` already deleted the previous artifact.

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,808 @@
...
+		<h1>Export Events</h1>
+
+		<?php render_notice(); ?>
+
+		<p>
+			Download every group's event history with its RSVP breakdown as CSV or JSON.
+			The export contains aggregate RSVP counts only — no attendee names or emails.
+		</p>
+
+		<?php if ( $job ) : ?>
```
gedex, line 1107: No cancel control in this branch, `$job['created']` is written at `:169` and never read, and `JOB_OPTION` is deleted only at `:383`. An uncatchable fatal commits nothing, so the same site is re-claimed after every lease expiry and dies again — "Exporting… N of M" indefinitely, new runs refused at `:176`, and the last artifact already gone.

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,808 @@
...
+ * @return array[]|WP_Error Aggregate rows, or the collector's error.
+ */
+function collect_site_rows( int $site_id ) {
+	// The switch sits inside the try: it pushes the switched-blog stack
+	// before firing its action, so a Throwable from a `switch_blog` listener
+	// would otherwise leave the process stuck on the wrong blog.
+	try {
+		switch_to_blog( $site_id );
+
+		$data = \WordCamp\Groups\Frontend\Export\collect_export_data();
```
gedex: This runs `posts_per_page => -1`, fetches all approved RSVP comments (`number => 0`), resolves every attendee through `get_users_by_id()`, and builds a full per-RSVP record. `:427-438` then keeps 10 fields per event and drops the rest. A group with ~30k RSVPs allocates 30k arrays plus a `WP_User` per attendee in an unattended cron process to produce three integers per event. It also materialises attendee PII inside the network-wide process the file header says keeps identities out.

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,808 @@
...
+	}
+
+	$date     = substr( (string) $artifact['generated_gmt'], 0, 10 );
+	$filename = sprintf( 'groups-network-events-%s.%s', $date, $format );
+
+	if ( 'json' === $format ) {
+		return array(
+			'filename'     => $filename,
+			'content_type' => 'application/json; charset=utf-8',
+			'body'         => (string) wp_json_encode( build_network_json( $artifact ) ),
```
gedex: `wp_json_encode()` returns `false` on failure and there is no `strict_types` in this tree, so `(string) false` is `''` — HTTP 200, `application/json`, `Content-Disposition: attachment`, zero bytes. `json_encode` fails for the whole document on one invalid UTF-8 byte, so a single latin1-in-utf8 event title in one group empties the entire network export while the CSV of the same artifact downloads fine. `build_network_csv(): string` at `:568` coerces a `false` from `stream_get_contents()` the same way.

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,808 @@
...
+
+			$job['claimed_until'] = time() + CLAIM_LEASE;
+
+			update_site_option( JOB_OPTION, $job );
+
+			return $job;
+		}
+	);
+
+	if ( empty( $job ) ) {
```
gedex: `with_job_lock()` returns `null` for "lock not acquired", "no job", and "another run holds the lease" alike, and this treats all three as nothing-to-do — no log, and no `schedule_next_batch()`. The cron event is consumed before dispatch, so losing the 1s lock race to a concurrent start-form submission leaves the export with no scheduled work until an admin loads the screen. Without a persistent object cache `wp_cache_add` never contends, so this will not show up in tests.

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,808 @@
...
+ *
+ * @param int $site_id Group site ID.
+ * @return array[]|WP_Error Aggregate rows, or the collector's error.
+ */
+function collect_site_rows( int $site_id ) {
+	// The switch sits inside the try: it pushes the switched-blog stack
+	// before firing its action, so a Throwable from a `switch_blog` listener
+	// would otherwise leave the process stuck on the wrong blog.
+	try {
+		switch_to_blog( $site_id );
```
gedex, line 565: No check that the site still exists. `get_group_site_ids()` is evaluated once at queue time and batches run minutes to hours later, so a site deleted or archived in between points `$wpdb` at dropped tables; `get_posts()` does not check for query errors and `get_event_dates( array() )` returns early, so the collector returns an empty `events` array rather than a `WP_Error`. That group lands in the export with zero events and no `failed_sites` entry — the outcome `:325-329` says it prevents.

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,808 @@
...
+ *
+ * @param array $job Completed job.
+ */
+function record_artifact( array $job ): void {
+	update_site_option(
+		EXPORT_OPTION,
+		array(
+			'generated_gmt' => gmdate( 'Y-m-d H:i:s' ),
+			'author'        => (int) $job['author'],
+			'site_count'    => count( $job['sites'] ),
```
gedex: This counts every queued site, including the ones that failed, so the summary at `:730` asserts complete coverage over a file that does not have it. `build_network_csv()` also only sees `$artifact['rows']`, so failed sites reach the JSON download (`:629-635`) and the admin screen but not the CSV. Dropping `'failed_sites' => $failed` from `build_network_json()` keeps the suite green too.

public_html/wp-content/mu-plugins/groups/tests/test-network-export.php
```diff
@@ -0,0 +1,524 @@
...
+		$this->expectException( 'WPDieException' );
+		$this->expectExceptionMessage( 'permission' );
+
+		Network_Export\handle_download();
+	}
+
+	/**
+	 * Even a network admin can't start or download without a valid nonce.
+	 */
+	public function test_handlers_require_a_valid_nonce() {
```
gedex: Neither handler has a passing-path test. Changing `check_admin_referer( DOWNLOAD_ACTION )` to `FORM_ACTION` at `network-export.php:501` keeps all 14 tests green even though it 403s every Download CSV/JSON click — `render_summary()` mints `DOWNLOAD_ACTION` nonces at `:750-751`. The `format` allowlist at `:504-508` is in the same blind spot, since tests only ever call `get_download_payload()` directly. This test also counts `$died === 2` without distinguishing a permission death from a nonce death. The reason those success paths are unreachable: `redirect_with_notice()`'s bare `exit` has no interception in `Groups_TestCase`, so removing the nonce check kills the PHPUnit process after 4 tests instead of failing one.

public_html/wp-content/mu-plugins/groups/tests/test-network-export.php
```diff
@@ -0,0 +1,524 @@
...
+		Network_Export\process_batch();
+
+		$job = Network_Export\get_job();
+		$this->assertCount( 3, $job['pending_sites'], 'One run must stop at the batch boundary.' );
+		$this->assertNotFalse( wp_next_scheduled( Network_Export\CRON_HOOK ), 'The remainder must be rescheduled.' );
+
+		$this->drain_export();
+
+		$this->assertSame( array(), Network_Export\get_job() );
+		$this->assertIsArray( get_site_option( Network_Export\EXPORT_OPTION ) );
```
gedex: This is the only post-drain assertion — no row count, no per-site attribution. Injecting `$job['rows'] = array();` into the resume branch at `network-export.php:385` (so every batch but the last discards its work) keeps the suite green, which is the regression d7f65c22 exists to prevent. The fixture cannot distinguish batches either: `:371` repeats the same two site IDs 13 times.

public_html/wp-content/mu-plugins/groups/tests/test-network-export.php
```diff
@@ -0,0 +1,524 @@
...
+		$this->create_site_rsvp( $brisbane, $event_id, 'attending' );
+
+		Network_Export\queue_export( array( $brisbane ) );
+		$this->drain_export();
+
+		$csv = Network_Export\get_download_payload( 'csv' );
+		$this->assertStringStartsWith( 'text/csv', $csv['content_type'] );
+		$this->assertMatchesRegularExpression( '/^groups-network-events-\d{4}-\d{2}-\d{2}\.csv$/', $csv['filename'] );
+
+		$lines = array_filter( explode( "\n", trim( preg_replace( '/^\xEF\xBB\xBF/', '', $csv['body'] ) ) ), 'strlen' );
```
gedex: This strips the BOM as setup and never asserts it was there, so deleting the `fwrite()` at `network-export.php:572` passes. RFC 4180 quoting is unexercised too — no cell here contains a comma, quote, newline, or backslash, and the parser passes the matching escape, so replacing `fputcsv( $handle, $line, ',', '"', '' )` at `:585` with the default-escape call also stays green. The sibling covers both at `wporg-groups-frontend/tests/test-export.php:344-359`.

public_html/wp-content/mu-plugins/groups/tests/test-network-export.php
```diff
@@ -0,0 +1,524 @@
...
+		$brisbane = $this->group_sites['brisbane'];
+		$this->create_site_event( $brisbane, '=HYPERLINK("https://evil.example")' );
+
+		Network_Export\queue_export( array( $brisbane ) );
+		$this->drain_export();
+
+		$body = Network_Export\get_download_payload( 'csv' )['body'];
+
+		$this->assertStringContainsString( "'=HYPERLINK", $body );
+		$this->assertStringNotContainsString( "\n\"=HYPERLINK", $body );
```
gedex: This negative assertion can never fail — `event_title` is column 4 of `CSV_COLUMNS`, so its cell is never preceded by a newline. Escaping is also only checked on `event_title`, never `group_name` or `venue`, which flow through the same branch at `:580-582`.

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: fixed in 930a5eb1 — per-claim token now, verified by mutating the guard

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: fixed in 930a5eb1

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: fixed in 930a5eb1 — cancel button, auto-abandon after 3 claims, and starting a run no longer deletes the previous artifact

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: fixed in 930a5eb1 — counts come from one grouped SQL query now, no comments or users loaded

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: fixed in 930a5eb1. one nuance: wp_json_encode() repairs invalid UTF-8 rather than failing, so a latin1 title doesn't empty the file — but false is real for other encode failures and both coercions are gone

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: fixed in 930a5eb1 — claim returns claimed/none/leased, and a lost lock reschedules

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: fixed in 930a5eb1

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: count fixed in 930a5eb1 (exported-of-queued). leaving CSV as pure data — a flat table has no metadata row; failed sites stay in the JSON and on the screen

public_html/wp-content/mu-plugins/groups/tests/test-network-export.php
gedex: fixed in 930a5eb1 — validator split out of the handler, nonce action and format allowlist both tested, redirect intercepted for the success paths

public_html/wp-content/mu-plugins/groups/tests/test-network-export.php
gedex: fixed in 930a5eb1 — asserts per-batch row accumulation now, fails if a resume drops its rows

public_html/wp-content/mu-plugins/groups/tests/test-network-export.php
gedex: fixed in 930a5eb1 — BOM asserted, and a backslash-before-quote cell catches the default-escape mode

public_html/wp-content/mu-plugins/groups/tests/test-network-export.php
gedex: fixed in 930a5eb1 — asserts parsed cells across group_name, event_title and venue

public_html/wp-content/mu-plugins/groups/network-export.php
```diff
@@ -0,0 +1,1465 @@
...
+				'order'                  => 'ASC',
+				// Only IDs and titles are exported; priming meta and term
+				// caches for every event would be pure overhead here.
+				'no_found_rows'          => true,
+				'update_post_meta_cache' => false,
+				'update_post_term_cache' => false,
+			)
+		);
+
+		if ( empty( $events ) ) {
```
raicem, line 778: Could we preserve group metadata when a site has no events? Returning [] means the group produces no CSV row. This makes an empty group indistinguishable from one omitted by the export.

Please lmk what you think

raicem (approved):
Works as described! Thank you @gedex! I have left one comment but it's not a blocker.

public_html/wp-content/mu-plugins/groups/network-export.php
gedex: I'd rather skip empty groups. The tool exports events, and a group with no events has nothing to export. The full list of groups is already on the parent page this screen (`admin.php?page=wporg-groups`), so that's where you'd go to see which groups exist. We can add group metadata to an export later as its own PR, rather than blank rows here.

PR #994, opened 2023-07-31
WC Post Types, Blocks: Add support for private speakers & sessions

ryelle:
Fixes #717 — This is a simpler approach to the issue described, where it's difficult to style the speakers & sessions without making them public. By enabling private sessions & speakers to be used in blocks, this lets organizers (who have permissions to see private content) view and edit the content, without showing sessions & speakers to site visitors, and without notifying the wporg profiles code.

The intention to only show private sessions and not _all_ sessions is that all sessions are submitted as drafts, so this lets the organizer team update and privately publish the accepted talks, while leaving the rest alone.

### Screenshots

Private speakers now show up in the sessions list table, with updated label.

Screenshot 2023-07-31 at 1 11 23 PM

Private sessions appear in the Speaker Sessions block, which can be used on speaker posts.

Screenshot 2023-07-31 at 1 13 53 PM

And the main use for this, private sessions will show up in the schedule. In the following screenshot, the "Friday keynote" is a privately published regular session (the rest are public custom sessions).

Screenshot 2023-07-31 at 1 07 18 PM

### How to test the changes in this Pull Request:

1. Add some sessions and speakers
2. Publish them privately
3. The should be visible on the frontend when logged in, but not when logged out

PR #1910, opened 2026-08-18
WC Post Types, Blocks: List only what the viewer can read

mcliwanow:
## Why

Four places in the session and speaker listings print content the viewer cannot read. Each turns stored ids or a status list into titles and permalinks without the check a query or a REST controller would apply.

**The session listings ask for a status they don't check.** The Speaker Sessions block and the `wcb_session` REST collection query for `publish` and `private` without passing `perm`, and `get_posts()` doesn't set one, so `WP_Query` folds the private-status branch into the readable one and unsets it before the `read_private_posts` test runs.

In the REST collection that surfaces as pagination that doesn't add up: the controller filters unreadable posts out of the body, but `found_posts` comes from the unfiltered query, so a logged out request to a camp with one published and one private session returns `X-WP-Total: 2` next to a single-item body. The block has no per-item filter at all, so it lists private sessions with their times and a link to each.

**The speaker listings don't resolve their ids at all.** The Session Speakers block and the `session_speakers` REST field read `_wcpt_speaker_id` and pass each id straight to `get_the_title()` and `get_permalink()`. No query, so no status defaults, and no post type check. Speaker records come from the Call for Speakers form as drafts, so a published session pointing at an unpublished speaker is ordinary while a schedule is being built. Both are public output: the block is on the network's default `single-wcb_session` template with `isLink` on, and the field is served in the anonymous `view` context. Doing it one id at a time is also expensive: 151 queries for one collection request over 30 sessions and 90 speakers.

The four sibling controllers that read this data correctly all hard-code `'publish'` (`speakers/controller.php:141`, `sessions/controller.php:217`, `organizers/controller.php:94`, `sponsors/controller.php:92`), and the classic-theme path already uses a single `post__in` query. None of this needs an organiser to opt in: `theme-templates/bootstrap.php` injects both templates for any camp on a block theme, and new camps get `twentytwentyfour` (`wordcamp-new-site.php:692`).

## What changed

**Session statuses are gated on capability.** The block and `prepare_session_query_args()` add `private` only when the viewer holds the post type's `read_private_posts`. That also settles the REST counting, so `X-WP-Total` and the page count agree with the body.

**Speaker ids are resolved through a query and then checked.** Both use `post__in` the way `add_speaker_info_to_session_posts()` already does, which enforces the post type in SQL, then skip what the viewer cannot read. Meta and term caches are not primed, since only the title, slug, permalink and status are read. That took the field from 151 queries to 91, mostly by not priming caches nothing reads rather than by the batching.

**All four use the same check** that `WP_REST_Posts_Controller::check_read_permission()` applies: published posts are readable by everyone, `read_post` decides the rest. Published has to come first, since logged out visitors hold no capabilities and a bare `current_user_can( 'read_post', … )` drops published posts too.

**Smaller things:** the Session Speakers block returns early when nothing is readable, so it no longer renders a byline with no names under it, and `post_status => 'any'` leaves out trashed speakers that `get_post()` would have printed. Note that `perm => 'readable'` is not sufficient for the session half, since it falls back to matching `post_author` and a session with no author (0) matches a logged out visitor (also 0).

## Testing

None of these four paths had coverage. This adds 19 tests across four files: the statuses the session query asks for, and the output of each listing path.

Four guard mistakes that are easy to make here, each failing against a version of this patch that made it: scoping by `post_author` lists a session with no author, checking `read_post` without allowing for published first drops published posts, checking only the status prints a stray non-speaker id, and skipping the early return leaves the byline behind. Against the pre-change code the session tests fail for the logged out and subscriber cases, and the speaker tests fail with draft speakers, private speakers and an unrelated draft post in the output. Full suite is 650 tests, with two `Test_Error_Handling` failures a macOS temp path produces locally and CI doesn't.

Also verified by hand on a local network install over HTTPS. Logged out, each listing returns only what is published, with `X-WP-Total` matching the body; as an administrator all of it comes back, so organisers keep the visibility #994 added. The private session's permalink and the draft speaker's both still 404. I haven't measured how many private sessions or unpublished speakers exist across the network, so that rests on these cases rather than on production counts.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/session-speakers/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/session-speakers/controller.php
@@ -40,6 +40,29 @@ function render( $attributes, $content, $block ) {
 		return '';
 	}
 
+	// The ids come straight from post meta, so check them before dereferencing.
+	// Mirrors WP_REST_Posts_Controller::check_read_permission(): published posts
+	// are readable by everyone, including logged out visitors, who hold no
+	// capabilities at all. The post type check keeps a stray id from printing
+	// unrelated content.
+	$speaker_ids = array_filter(
+		$speaker_ids,
+		function ( $speaker_id ) {
+			$speaker = get_post( $speaker_id );
+
+			if ( ! $speaker instanceof \WP_Post || 'wcb_speaker' !== $speaker->post_type ) {
+				return false;
+			}
+
+			return 'publish' === $speaker->post_status
+				|| current_user_can( 'read_post', $speaker->ID );
+		}
+	);
+
+	if ( empty( $speaker_ids ) ) {
+		return '';
+	}
+
 	$byline  = ! empty( $attributes['byline'] ) ? $attributes['byline'] : false;
 	$classes = array_filter( array(
 		isset( $attributes['textAlign'] ) ? 'has-text-align-' . $attributes['textAlign'] : false,
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/session-speakers/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/session-speakers/controller.php
@@ -40,9 +40,8 @@ function render( $attributes, $content, $block ) {
 		return '';
 	}
 
-	// Resolve the ids through a query, the way add_speaker_info_to_session_posts()
-	// does for the classic themes, so that the post type is enforced and the
-	// posts are fetched together rather than one lookup at a time.
+	// Resolve the ids the way add_speaker_info_to_session_posts() does, so the
+	// post type is enforced in SQL and the posts are fetched in one query.
 	$speakers = get_posts( array(
 		'post_type'      => 'wcb_speaker',
 		'post__in'       => array_map( 'absint', $speaker_ids ),
@@ -55,9 +54,8 @@ function render( $attributes, $content, $block ) {
 		'update_post_term_cache' => false,
 	) );
 
-	// Mirrors WP_REST_Posts_Controller::check_read_permission(): published posts
-	// are readable by everyone, including logged out visitors, who hold no
-	// capabilities at all.
+	// Mirrors WP_REST_Posts_Controller::check_read_permission(). Logged out
+	// visitors hold no capabilities, so published has to be allowed for first.
 	$speakers = array_filter(
 		$speakers,
 		function ( $speaker ) {
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/session-speakers/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/session-speakers/controller.php
@@ -40,26 +40,34 @@ function render( $attributes, $content, $block ) {
 		return '';
 	}
 
-	// The ids come straight from post meta, so check them before dereferencing.
+	// Resolve the ids through a query, the way add_speaker_info_to_session_posts()
+	// does for the classic themes, so that the post type is enforced and the
+	// posts are fetched together rather than one lookup at a time.
+	$speakers = get_posts( array(
+		'post_type'      => 'wcb_speaker',
+		'post__in'       => array_map( 'absint', $speaker_ids ),
+		'post_status'    => 'any',
+		'posts_per_page' => -1,
+		'orderby'        => 'post__in',
+		'no_found_rows'  => true,
+		// Only the title, slug, permalink and status are used below.
+		'update_post_meta_cache' => false,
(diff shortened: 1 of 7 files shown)


bor0 (approved):
Tested on my sandbox on a pair call with Michal and Herman, works well.

| Description | Pre-patch | Post-patch |
| ----------- | --------- | ---------- |
| Schedule screen | Image | Image |
| Submission screen | Image | Image |

bor0:
To get potentially affected private speakers

First, we found all WordCamp sites for this year:

```php
<?php
// Run on central. Prints the site URL of every camp starting on/after $since.
$since = strtotime( '2026-01-01' );

$camps = get_posts( array(
	'post_type'      => 'wordcamp',
	'post_status'    => 'any',
	'posts_per_page' => -1,
	'meta_query'     => array(
		array(
			'key'     => 'Start Date (YYYY-mm-dd)',
			'value'   => $since,
			'compare' => '>=',
			'type'    => 'NUMERIC',
		),
	),
) );

foreach ( $camps as $camp ) {
	$blog_id = get_wordcamp_site_id( $camp );

	if ( ! $blog_id ) {
		continue; // No site provisioned yet.
	}

	echo get_site_url( $blog_id ) . "\n";
}
```

Then list speakers for specific site

```php
<?php
/**
 * Read-only audit: what the session/speaker read checks actually change on this site.
 *
 * Only counts content that renders publicly today. A private session attached
 * solely to an unpublished speaker never appears, because that speaker's page
 * 404s; an unpublished speaker attached solely to a private session never
 * appears either.
 */
$site = home_url( '/' );

$published_speakers = array_flip( get_posts( array(
	'post_type'      => 'wcb_speaker',
	'post_status'    => 'publish',
	'posts_per_page' => -1,
	'fields'         => 'ids',
) ) );

// 1) Private sessions listed on a published speaker's page.
$leaky_sessions   = array();
$private_sessions = get_posts( array(
	'post_type'      => 'wcb_session',
	'post_status'    => 'private',
	'posts_per_page' => -1,
	'fields'         => 'ids',
	'meta_key'       => '_wcpt_speaker_id',
) );
foreach ( $private_sessions as $session_id ) {
	foreach ( get_post_meta( $session_id, '_wcpt_speaker_id' ) as $speaker_id ) {
		if ( isset( $published_speakers[ (int) $speaker_id ] ) ) {
			$leaky_sessions[] = $session_id;
			break;
		}
	}
}

// 2) Speakers listed on a published session's page, in its REST field, and in
//    the schedule, that are not published themselves.
$published_sessions = get_posts( array(
	'post_type'      => 'wcb_session',
	'post_status'    => 'publish',
	'posts_per_page' => -1,
	'fields'         => 'ids',
) );

$referenced = array();
foreach ( $published_sessions as $session_id ) {
	foreach ( get_post_meta( $session_id, '_wcpt_speaker_id' ) as $speaker_id ) {
		$referenced[ (int) $speaker_id ] = true;
	}
}

$unpublished = array();
$not_speaker = array();
foreach ( array_keys( $referenced ) as $speaker_id ) {
	$speaker = get_post( $speaker_id );

	if ( ! $speaker || 'wcb_speaker' !== $speaker->post_type ) {
		$not_speaker[] = $speaker_id;
	} elseif ( 'publish' !== $speaker->post_status ) {
		$unpublished[] = $speaker_id . ':' . $speaker->post_status;
	}
}

if ( ! $leaky_sessions && ! $unpublished && ! $not_speaker ) {
	return;
}

printf(
	"%s  private_sessions=%d  unpublished_speakers=%d  non_speaker_ids=%d\n",
	$site,
	count( $leaky_sessions ),
	count( $unpublished ),
	count( $not_speaker )
);

if ( $leaky_sessions ) {
	echo '    private sessions: ' . implode( ', ', array_slice( $leaky_sessions, 0, 20 ) ) . "\n";
}
if ( $unpublished ) {
	echo '    unpublished speakers: ' . implode( ', ', array_slice( $unpublished, 0, 20 ) ) . "\n";
}
if ( $not_speaker ) {
	echo '    non-speaker ids: ' . implode( ', ', array_slice( $not_speaker, 0, 20 ) ) . "\n";
}
```

Then finally:

```
wporgdev@bor0.dev.dca.wordpress.org:/home/wordcamp/public_html$ while read -r url; do bin/wp-cli/wp eval-file audit-session-visibility.php --url="$url"; done < sites.txt
```

We only had this:

```
https://kampala.wordcamp.org/2026/  private_sessions=0  unpublished_speakers=5  non_speaker_ids=0
    unpublished speakers: 5511:trash, 5530:trash, 5534:trash, 5732:trash, 3028:draft
```

But this event is past, so all good to deploy tomorrow on the code chat.

PR #1911, opened 2026-08-18
Groups: send the publish notification after the event's data is saved

### LINKED ISSUE #1909: Groups: new-event email shows "Date: —" instead of the event date
The "new event published" email that goes out to all group members renders the date as GatherPress's em-dash placeholder instead of the actual event date:

```html
<!-- Date & Time -->
<p style="text-align: center;">
    Date: —		</p>

<!-- Venue -->

```

The venue block is empty for the same reason, and a featured image would be missing too.

Image

## Cause

`schedule_new_event_notification()` (`mu-plugins/wporg-groups-frontend/inc/notifications.php`) hooks `transition_post_status` and calls `Rest_Api::send_emails()` synchronously from there. `transition_post_status` fires from *inside* `wp_insert_post()` — before the caller has written any of the event's own data.

In the frontend create path, `persist_event()` (`mu-plugins/wporg-groups-frontend/inc/rest.php`) does:

1. `wp_insert_post( [ 'post_status' => 'publish', ... ] )` — fires the transition, sends the email
2. `Event::save_datetimes( ... )` — writes `gatherpress_datetime_*` meta
3. `sync_event_venue_terms()` / `set_post_thumbnail()`
4. `do_action( 'wporg_groups_frontend_event_saved' )` — recurrence meta and occurrence projection

So the email template's `Event::get_display_datetime()` reads datetime meta that does not exist yet and falls back to `Event::DATETIME_PLACEHOLDER` (`—`).

Not recurrence-specific — every event created through the frontend form is affected; I just noticed it on a recurring one. A wp-admin publish has the same shape, since the block editor's REST meta update also lands after `wp_insert_post()` returns.

## Steps to reproduce

1. Create and publish a new event from a group site's frontend form.
2. Open the caught mail (`http://localhost:1080` locally).
3. The body shows `Date: —` and no venue.

Confirmed on the event's post meta afterwards: `gatherpress_datetime_start` etc. are present, so the data saves fine — it's purely the send timing.

## Notes toward a fix

GatherPress core has no action or filter fired after an event's datetimes are saved (zero `do_action()` calls in `includes/core/classes/event/`), so there's nothing upstream to hook.

GatherPress does hit this exact problem itself and solves it by deferring to `shutdown` — see `Event\Setup::set_datetimes()` / `resolve_pending_datetimes()`, and the comment there ("This hook fires from inside `wp_insert_post()`, and every caller writes the meta afterwards"). Queueing the notification on `transition_post_status` and sending it on `shutdown` at a priority after that would fix every publish path. A narrower alternative is reordering `persist_event()` to insert as `draft` and flip to `publish` last, but that only covers the frontend path and has to stay ahead of `wporg_groups_frontend_event_saved`, which the recurring extension expects to run against an already-published post.

Existing unit coverage doesn't catch this because the tests call `schedule_new_event_notification()` directly rather than going through `wp_insert_post()`.

av3nger:
## Description

The "new event published" email that goes out to all group members rendered GatherPress's em-dash placeholder instead of the event date (`Date: —`), with an empty venue block and no featured image.

## Issue

The `schedule_new_event_notification()` hooks `transition_post_status` and called `Rest_Api::send_emails()` synchronously from there. Core fires that hook from inside `wp_insert_post()` - before the caller writes any of the event's own data. In the frontend path, `persist_event()` inserts with `post_status => 'publish'` first and writes datetimes / venue terms / online link / thumbnail afterwards; the wp-admin path has the same shape, since the editor's REST meta update also lands after `wp_insert_post()` returns. The email template therefore read datetime meta that did not exist yet.

## Fix

`transition_post_status` now only records the event ID (same guards as before). The actual send runs from a `shutdown` callback registered at `PHP_INT_MAX`. This is the same approach the #1909 suggests, and it fixes every publish path:

- **Frontend REST**: all of `persist_event()`'s writes complete long before shutdown.
- **wp-admin**: GatherPress resolves the editor's `gatherpress_datetime` meta in its own *default-priority* shutdown callback (`Event\Setup::resolve_pending_datetimes()`), so ours must run after it - that's what the `PHP_INT_MAX` priority guarantees (GatherPress registers at priority 10).

Fixes #1909.

Props vanyukov, bor0

1. On a group site, create and publish a new event from the frontend form - **select or create a venue**, and set a featured image if you want to check that too. (The email only prints a venue line when the event has one, so an event without a venue can't distinguish "broken" from "none set".)
3. Open the caught mail at http://localhost:1080. The body must show the real event date and the venue name - not `Date: —` with an empty venue block.
4. The sharper variant (what actually distinguishes this fix): save an event as a draft with one date, then publish it later with a **changed** date and venue. The caught mail must show the *new* date/venue, not the draft's.
5. Repeat via wp-admin: create and publish an event in the block editor. Same expectation - the mail carries the date the editor set (this path relies on GatherPress's own shutdown datetime resolver running before our sender).
6. Edit the published event and save - no second email may arrive.


gedex:
> The schedule_new_event_notification() hooks transition_post_status and called Rest_Api::send_emails() synchronously from there

Will this synchronously sending emails scale when there are ton of members on the group?

gedex:
When I tested this via "Create event" button

<kbd>image</kbd>

the emails were sent with proper datetimes:

<kbd>image</kbd>

however, when I tested it via the create event settings modal:

<kbd>image</kbd>

the emails don't have the datetimes set:

<kbd>image</kbd>

public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/notifications.php
```diff
@@ -119,6 +119,46 @@ function schedule_new_event_notification( string $new_status, string $old_status
 		return;
 	}
 
+	pending_notification_queue( $post->ID );
+	add_action( 'shutdown', __NAMESPACE__ . '\send_pending_new_event_notifications', PHP_INT_MAX );
+}
+
+/**
+ * Request-static queue of event IDs awaiting their first-publish notification.
```
gedex, line 127: What is request-static means here?

av3nger: The `static $pending` below (line 134). I can replace it with just "Event IDs queued..." if that's easier to read

av3nger:
@gedex , I am not able to replicate the issue on my local setup. With the setting modal I still get correct emails

> Will this synchronously sending emails scale when there are ton of members on the group?

No, it won't. But it's a pre-existing issue, not introduced with this PR. Can be adjusted to use a batching pattern, similar to mu-plugins/groups/network-messaging.php, but I would do it as a separate task.

gedex:
> @gedex , I am not able to replicate the issue on my local setup. With the setting modal I still get correct emails

sorry i didn't mentioned this before. I was testing it on `production` branch. On the test, the issue happened only when sending emails via settings modal.

PR #1914, opened 2026-08-19
WordCamp Base: tidy i18n and output handling in attachment templates

KokkieH:
## What

Housekeeping pass over the two `wordcamp-base` attachment templates.

- **Markup out of translatable strings.** Several msgids wrapped their text in `<span>` markup, or consisted of nothing but markup and a placeholder. That markup now lives in the templates, so the msgids carry only translatable text.
- **Consistent escaping at each output.** Every interpolated value now goes through the helper that matches its context — `esc_url()` for hrefs, `esc_attr()` for attribute values, `esc_html()` for element text, `absint()` for the image dimensions.
- **Translator comments** added for all four remaining placeholder strings.

No change to the rendered markup, classes, or visible output.

## Why

These templates predate the current coding standards, so they mixed markup into msgids and passed a number of values straight to output. Keeping markup out of translatable strings is what the i18n guidelines ask for, and it leaves the remaining strings much easier to translate — `By` and `Published` instead of `<span class="%1$s">By</span> %2$s`.

## Translation impact

Three msgids are retired and four short ones added:

| Retired | Replaced by |
| --- | --- |
| `<span class="meta-nav">&larr;</span> %s` | *(nothing — it contained no words)* |
| `<span class="%1$s">By</span> %2$s` | `By` |
| `<span class="%1$s">Published</span> %2$s` | `Published` |
| `Continue reading %s <span class="meta-nav">&rarr;</span>` | `Continue reading %s` |

Existing locale translations for the retired strings will orphan until the new ones are translated. That is two words of actual copy, so the churn is small.

`Return to %s`, `View all posts by %s`, `Full size is %s pixels`, `Link to full-size image`, `Pages:` and `Edit` are all unchanged.

## Testing

These templates take a little setup to reach, because attachment pages are turned off network-wide. Steps 1–3 are local-only changes; step 7 reverts them.

1. **Activate the theme** on a local test site:
   ```
   wp theme activate wordcamp-base --url=<site>
   ```

2. **Turn off the Coming Soon page**, which otherwise replaces the theme's templates entirely and you will see the splash page instead of the attachment page:
   ```
   wp option patch update wccsp_settings enabled off --url=<site>
   ```

3. **Let attachment pages render.** `mu-plugins/wcorg-misc.php:81` forces the option off for every site, which makes core 301 the attachment page to the file URL. Comment that line out locally:
   ```php
   // add_filter( 'pre_option_wp_attachment_pages_enabled', '__return_zero' );
   ```
   Then confirm the underlying option is on — legacy sites already have it set, new installs do not:
   ```
   wp option get wp_attachment_pages_enabled --url=<site>   # expect 1
   wp option update wp_attachment_pages_enabled 1 --url=<site>   # only if it returned 0
   ```

4. **Find an image attachment that has a parent post:**
   ```
   wp post list --post_type=attachment --fields=ID,post_title,post_name,post_parent --url=<site>
   ```
   Its URL is the parent post's permalink plus the attachment's `post_name`, e.g. `/<year>/<parent-slug>/<attachment-slug>/`.

5. **Compare the page before and after.** Attachment pages are disabled on every production site, so there is no live page to compare against — render the `production` branch's copy of these two files locally instead:
   ```
   curl -s '<attachment-url>' > /tmp/after.html
   git checkout production -- public_html/wp-content/themes/wordcamp-base/attachment.php \
                              public_html/wp-content/themes/wordcamp-base/loop-attachment.php
   curl -s '<attachment-url>' > /tmp/before.html
   git checkout HEAD -- public_html/wp-content/themes/wordcamp-base/attachment.php \
                        public_html/wp-content/themes/wordcamp-base/loop-attachment.php
   diff -w /tmp/before.html /tmp/after.html   # expect no output
   ```
   Or check by eye — these are the elements the diff touches:
   - the `&larr; <parent post title>` link above the image, including its `title` attribute
   - the `By <author>` byline — author link `href` and `title`
   - the `Published <date>` meta, including the `<abbr class="published" title="…">` wrapper
   - the `Full size is <width> &times; <height> pixels` link
   - `Edit` and `Pages:` links, where they appear

6. **Check two edge cases:**
   - a non-image attachment (e.g. a PDF), which should still render its filename link
   - an author whose display name contains an apostrophe or an ampersand, to confirm the byline still reads correctly

7. **Revert the local setup:** restore `wcorg-misc.php:81`, and set `wccsp_settings.enabled` back to `on` if the site had it on.

Verified locally following the steps above: the rendered page is unchanged from the `production` branch apart from insignificant whitespace, and `php -l` is clean on both files.

Note that CI lints only changed lines, so this PR brings these lines into scope for the coding-standards sniffs for the first time.

### DIFF
--- a/public_html/wp-content/themes/wordcamp-base/attachment.php
+++ b/public_html/wp-content/themes/wordcamp-base/attachment.php
@@ -52,23 +52,23 @@ get_header(); ?>
 								esc_html( get_the_date() )
 							);
 
-							if ( wp_attachment_is_image() ) {
-								$metadata = wp_get_attachment_metadata();
-								
-								if ( ! empty( $metadata['width'] ) && ! empty( $metadata['height'] ) ) {
-									echo ' <span class="meta-sep">|</span> ';
+						if ( wp_attachment_is_image() ) {
+							$metadata = wp_get_attachment_metadata();
 
-									/* translators: %s - a link reading "<width> × <height>" */
-									printf( esc_html__( 'Full size is %s pixels', 'wordcamporg' ),
-										sprintf( '<a href="%1$s" title="%2$s">%3$s &times; %4$s</a>',
-											esc_url( wp_get_attachment_url() ),
-											esc_attr__( 'Link to full-size image', 'wordcamporg' ),
-											absint( $metadata['width'] ),
-											absint( $metadata['height'] )
-										)
-									);
-								}
+							if ( ! empty( $metadata['width'] ) && ! empty( $metadata['height'] ) ) {
+								echo ' <span class="meta-sep">|</span> ';
+
+								/* translators: %s - a link reading "<width> × <height>" */
+								printf( esc_html__( 'Full size is %s pixels', 'wordcamporg' ),
+									sprintf( '<a href="%1$s" title="%2$s">%3$s &times; %4$s</a>',
+										esc_url( wp_get_attachment_url() ),
+										esc_attr__( 'Link to full-size image', 'wordcamporg' ),
+										absint( $metadata['width'] ),
+										absint( $metadata['height'] )
+									)
+								);
 							}
+						}
 						?>
 						<?php edit_post_link( esc_html__( 'Edit', 'wordcamporg' ), '<span class="meta-sep">|</span> <span class="edit-link">', '</span>' ); ?>
 					</div><!-- .entry-meta -->
--- a/public_html/wp-content/themes/wordcamp-base/attachment.php
+++ b/public_html/wp-content/themes/wordcamp-base/attachment.php
@@ -36,7 +36,7 @@ get_header(); ?>
 						<?php
 							printf(
 								'<span class="meta-prep meta-prep-author">%1$s</span> <span class="author vcard"><a class="url fn n" href="%2$s" title="%3$s">%4$s</a></span>',
-								esc_html__( 'By', 'wordcamporg' ),
+								esc_html_x( 'By', 'post author byline', 'wordcamporg' ),
 								esc_url( get_author_posts_url( get_the_author_meta( 'ID' ) ) ),
 								/* translators: %s - author display name */
 								sprintf( esc_attr__( 'View all posts by %s', 'wordcamporg' ), esc_attr( get_the_author() ) ),
@@ -47,7 +47,7 @@ get_header(); ?>
 						<?php
 							printf(
 								'<span class="meta-prep meta-prep-entry-date">%1$s</span> <span class="entry-date"><abbr class="published" title="%2$s">%3$s</abbr></span>',
-								esc_html__( 'Published', 'wordcamporg' ),
+								esc_html_x( 'Published', 'post publication date', 'wordcamporg' ),
 								esc_attr( get_the_time() ),
 								esc_html( get_the_date() )
 							);
--- a/public_html/wp-content/themes/wordcamp-base/attachment.php
+++ b/public_html/wp-content/themes/wordcamp-base/attachment.php
@@ -18,12 +18,13 @@ get_header(); ?>
 					<p class="page-title">
 						<a
 							href="<?php echo esc_url( get_permalink( $post->post_parent ) ); ?>"
-							title="<?php esc_attr( printf( __( 'Return to %s', 'wordcamporg' ), get_the_title( $post->post_parent ) ) ); ?>"
+							title="<?php
+								/* translators: %s - title of parent post */
+								printf( esc_attr__( 'Return to %s', 'wordcamporg' ), esc_attr( get_the_title( $post->post_parent ) ) );
+							?>"
 							rel="gallery">
-							<?php
-							/* translators: %s - title of parent post */
-							printf( __( '<span class="meta-nav">&larr;</span> %s', 'wordcamporg' ), get_the_title( $post->post_parent ) );
-							?>
+							<span class="meta-nav">&larr;</span>
+							<?php echo esc_html( get_the_title( $post->post_parent ) ); ?>
 						</a>
 					</p>
 				<?php endif; ?>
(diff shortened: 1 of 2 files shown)


mcliwanow (commented):
The escaping direction is right throughout, and the translator comments are all placed where xgettext needs them (on the line immediately preceding their gettext call). `esc_attr__()` on the format plus `esc_attr()` on the value is not double-escaping, since `_wp_specialchars()` defaults to `$double_encode = false`. `absint()` on the dimensions and `esc_url()` on both hrefs are plainly better than what was there.

Nothing below blocks a merge. Edited after checking this on a local site: the fourth section corrects a claim I got wrong the first time round, and adds two notes on the testing steps.

## `loop-attachment.php` is unreachable, and so are two of its siblings

Worth knowing while reviewing this, because it means half the diff is cosmetic.

`wordcamp-base/attachment.php:15` inlines the loop itself and never calls `get_template_part( 'loop', 'attachment' )`, which is the only way WordPress reaches a `loop-*.php` file. `twentyten/attachment.php:21` does call it, which is presumably where our copy came from. No theme in the repo lists `Template: wordcamp-base` either, so no child theme reaches it.

It is not a one-off. The same is true of `loop-page.php` and `loop-single.php`, because `page.php:20` and `single.php:15` also inline their loops where `twentyten/page.php:26` and `twentyten/single.php:21` call `get_template_part()`. Only `loop.php` is live, via the six `get_template_part( 'loop', $slug )` calls that all fall back to it.

So three orphans from the original fork, and I would leave all three alone here. Deleting one of them inside an i18n pass makes the theme less consistent, not more, and `wordcamp-central-2012/loop-attachment.php` is a live example of the convention still working elsewhere in the repo. Keeping the spare in sync with `attachment.php` is a reasonable thing for this PR to be doing.

## Since the spare is being kept in sync, sync the guard too

`attachment.php:58` guards with `! empty( $metadata['width'] ) && ! empty( $metadata['height'] )`. `loop-attachment.php:60` has no guard and goes straight into `absint( $metadata['width'] )`. `wp_get_attachment_metadata()` returns `false` when an attachment has no metadata, so that is a PHP 8 warning on all three versions we run in CI.

Pre-existing, and unreachable today, but the PR rewrites those exact lines and the whole value of the spare is that it matches the live copy.

## `By` and `Published` want `_x()`

Two bare words with no context is about the hardest thing you can hand a translator. `Published` also already exists in this textdomain as `_x( 'Published', 'post', 'wordcamporg' )` at `public_html/wp-content/plugins/camptix/camptix.php:3037`, so a context-free `Published` will sit next to it in GlotPress with nothing to tell them apart.

Since the PR is already paying the retirement cost on these msgids, this is the moment to add context rather than retire them a second time later.

## Correction to what I wrote above, and what the description should say

I claimed a parent post title containing a double quote broke out of the `title` attribute. That is wrong, and I have now checked it on a local site rather than reasoning about it. `get_the_title()` runs through `wptexturize`, which turns `"` into `&#8220;`/`&#8221;`, `'` into `&#8217;` and a bare `&` into `&#038;` before anything reaches output. For the title `Bob's "Great" Post & More`, `production` and this branch render byte-identical.

There is a genuinely broken attribute in the diff, just not that one. It is the author byline. With a display name of `O'Brien & Sons "Test"`, `production` emits:

```html
title="View all posts by O'Brien &#38; Sons "Test"">O'Brien &#38; Sons "Test"</a>
```

The quote closes the attribute early. `production` had `sprintf( esc_attr__( 'View all posts by %s' ), get_the_author() )`, so `esc_attr__()` escaped the format string and the name went in raw. Display names get no wptexturize, unlike titles, so the straight quote survives. `esc_attr( get_the_author() )` on this branch fixes it.

The parent title fix is real too, for a different character class. wptexturize leaves `<` and `>` alone, so `production` writes raw markup into both the attribute and the link text:

```
title="Return to Angle <b>bold</b> and <script>x</script> Post"
<span class="meta-nav">&larr;</span> Angle <b>bold</b> and <script>x</script> Post
```

kses strips `<script>` on save for anyone without `unfiltered_html`, which is nobody here, but it keeps `<b>`. So this branch does change one visible thing: a title containing `<b>` renders bold on `production` and renders as literal tags here. In the attribute that is unambiguously right. In visible link text it is a choice worth making on purpose, and it is the same choice a `loop.php` follow-up would face.

Either way, "No change to the rendered markup, classes, or visible output" is not true, and the two escaping fixes are the most useful thing in the PR.

## Two notes on the testing steps

Step 5 gives a false pass. The recipe curls, runs `git checkout production -- …`, then curls again straight away. With `opcache.validate_timestamps=On` and `opcache.revalidate_freq=2`, which is what the local container ships, both fetches serve the same bytecode and `diff` prints nothing. I got a clean empty diff on my first run with a parent title full of quotes and angle brackets, and only saw the real differences after putting a couple of seconds between the checkout and the fetch. Worth a note in the step.

Step 6 tests an author display name with an apostrophe or an ampersand. Neither of those breaks anything. The double quote is the one that does, and it is the case that demonstrates the fix.

Two things I checked that are clean: a non-image (PDF) attachment renders its filename link with no "Full size" block and no PHP warnings, and an image with `_wp_attachment_metadata` emptied hits the guard with no warnings. `loop-attachment.php` was never loaded once during any of it, as expected.

## Not this PR

`esc_attr( get_the_time() )` inside `<abbr class="published" title="">` is a localized time of day where hAtom wants ISO 8601. Broken since Twenty Ten, out of scope here, worth an issue if anyone cares.

bor0:
Nice catch on the `esc_attr(printf())` no-op. Worth flagging: the same unescaped-title-in-more-link pattern this PR fixes is still present in `wordcamp-base/loop.php:153-154` and `wordcamp-base-v2/content.php:27-28` (`the_title( '<span>...', '</span> ', false )` passed straight into `the_content()` with no `esc_html()`). Unlike `attachment.php`, `loop.php` is reached from `archive.php`/`author.php`/`category.php`/`index.php`/`search.php`/`tag.php` — so if `wordcamp-base` is active directly on any site, those are live listing pages rather than a template blocked by the attachment-pages mu-plugin. Might be worth a quick follow-up PR applying the same fix there.

mcliwanow (approved):
Changes test well and escaping looks good too. Thanks

PR #1915, opened 2026-08-20
Groups: design updates v3 — shared header band, single event-card source, front-page and archive polish

raicem:
## Summary

Refresh the Groups site presentation and simplify its event content.

- Add a calmer home-page hero with an optional featured image, location/member-count meta, and cleaner sidebar spacing.
- Simplify page navigation with group back links and add a dedicated news-post layout with author/date, featured image, and comments.
- Use one event-card presentation on the home page and event archive, with clearer filters, past-event styling, and useful empty states.
- Refine membership states: linked member count, role badges for organizers/event organizers, and no empty "My upcoming events" section.
- Remove the unused event-card helper, patterns, CSS, and tests.

## Test plan

- [x] `phpunit --group groups` - 256/256 passing
- [x] PHPCS clean on changed PHP
- [x] Smoke-tested the home page, events archive, news post, static page, and search results

### Browser flows

Use the seeded Sunshine Coast group with an upcoming event that has an image, an event without an image, a past event, news posts with and without featured images, and subscriber/author/editor accounts.

1. **Logged-out visitor**
   - Open the group home. Check the hero, location, member count, Join button, and upcoming cards. The member count should link to `/members/`; cards should show date, excerpt, RSVP count, venue, and either the event image or placeholder. "My upcoming events" should not appear. Click Join and verify the visitor is sent to login rather than seeing a silent failure.
   - Open `/event/`. Use the search and Time filters. Confirm the selected filter is visible, past cards are muted, and an empty search result offers a link back to all upcoming events.
   - Open an event, a news post, About, Members, and search results. Confirm the group back link/title treatment. On news posts, check the author/date, featured-image behavior, comments, and nested replies.

2. **Member (subscriber)**
   - Log in with no upcoming RSVP. Confirm there is no "Member" badge, the Leave group and email-preference controls are present, and "My upcoming events" is absent.
   - RSVP to an upcoming event and return home. Confirm "My upcoming events" appears and links to the event.
   - Accept the Leave group confirmation, verify Join this group returns, then rejoin the group before continuing.

3. **Event organizer (author)**
   - Confirm the membership area shows "Event Organizer", Leave group, and event-management controls.
   - Create or open an upcoming event through the front-end flow. Confirm it appears in "My upcoming events" and the home/archive cards show its date, RSVP count, venue, image/placeholder, and excerpt.

4. **Group organizer (editor or administrator)**
   - Confirm the membership area shows "Organizer", does not show Leave group, and retains event-management and group-settings controls.
[...]

PR #1917, opened 2026-08-20
Blocks: build the Latest Posts container by element, not by string

mcliwanow:
## Why

Two things in `mu-plugins/blocks/source/hooks/latest-posts/controller.php` operate on a proxy for what they actually care about.

**`render()` rewrites markup by string content.** It builds the live-update container with `str_replace()` on `wp-block-latest-posts ` and `ul class=`. Both get rewritten wherever they occur, including inside attribute values, and this block puts post content into two: the featured image's alt text into `alt`, the post title into `aria-label` when `addLinkToFeaturedImage` is on. `esc_attr()` escapes `& < > " '`, so it leaves either needle intact.

The replacement string carries a `"` of its own. When it lands in a value, the attribute ends early and the rest is parsed as markup:

| Alt text stored on the image | What the page renders today |
| --- | --- |
| `Speaker at the podium` | `alt="Speaker at the podium"` |
| `ul class=x foo=bar z` | `alt="div data-attributes="%7B…%7D" class=x foo=bar z"` |

One attribute in the database, four tokens in the browser. The class rewrite above it has the same shape but no quote, so it only adds stray classes.

**`safelist_block_renderer()` keys on the path, and discards the response.** It runs on `rest_request_after_callbacks`, which `WP_REST_Server::respond_to_request()` applies unconditionally at `class-wp-rest-server.php:1318`, after skipping the route callback for a failed permission check at `:1273`. The filter calls `$handler['callback']` itself and returns that, dropping whatever `$response` held.

It matches `$request->get_route()`, the URL path, while the controller reads the block from `$request['name']`, and `get_parameter_order()` puts JSON, POST and GET ahead of `URL`. A `name` in the query string outranks the path capture. Requesting the Latest Posts path while naming a post-scoped block returns that post's content, whatever its `post_status`, to a caller with no session. `WordCamp_Coming_Soon_Page::disable_rest_endpoints()` sets its `rest_cannot_access` on `rest_request_before_callbacks`, so it is dropped the same way and an unlaunched site answers like a launched one.

Both reach `central.` too: `blocks.php:19` gates only on `class_exists( 'WordCamp_Post_Types_Plugin' )`, and PHP binds that class before the early `return` at `wc-post-types.php:15`.

## What changed

- `render()` uses `WP_HTML_Tag_Processor`: first tag, only if it is a `UL`, then `add_class()` and `set_attribute()`. Neither can address anything but a real tag, and `set_attribute()` escapes.
- The `<ul>` to `<div>` swap stays: `render-frontend.js` renders the polled markup into this element, and that markup has been through this filter too, so it arrives as a `<div>`. It is anchored with `\A` and `\z`, where an attribute value cannot match.
- `is_swappable_container()` gates the whole rewrite. The swap renames the first `<ul` and the last `</ul>`, so it walks the `<ul>`/`</ul>` nesting to confirm those are the same element. Markup it cannot handle, a sibling list or content outside the container, is returned untouched rather than half-rewritten or left as a `<ul>` wearing `has-live-update`.
- `safelist_block_renderer()` compares `$request->get_param( 'name' )`, so it tests the block that will render rather than the path that was requested.
- It returns `$response` untouched when `post_id` is set, which is what `WP_REST_Block_Renderer_Controller::get_item_permissions_check()` gates on and the live block never sends.
- It replaces only a `WP_Error` of code `block_cannot_read`. Anything else, including the Coming Soon lockdown, is returned as it arrived.
- `WordCamp_Coming_Soon_Page::disable_rest_endpoints()` is registered on `rest_request_after_callbacks` as well as before. `before_callbacks` only sets a response, and anything hooked after it can replace that response, so the lockdown now decides at both points.

## Test plan

- [x] `phpunit --group=blocks` 24/24, `--group=coming-soon-page` 13/13
- [x] `phpcs-changed` against `production` clean
- [x] Manual browser pass on a local multisite camp site (steps below)

Two new suites. `mu-plugins/tests/test-blocks-latest-posts.php`, twelve tests, drives `rest_do_request()` and the real `render_block` hook rather than the internals; seven fail against `production`. `wordcamp-coming-soon-page/tests/test-rest-lockdown.php`, four tests, covers both registration points and a filter that replaces the response.

Each guard is covered on its own: removing any one of them turns a test red.

Compatibility, since this rewrites the markup of a shipping block:

- The container still emits `<div data-attributes="…" class="… wp-block-latest-posts … has-live-update is-loading">`, so `render-frontend.js`'s `.wp-block-latest-posts.has-live-update` selector still matches and `getAttributesFromData()` still finds `data-attributes`. Only class order changes, and nothing keys on it.
- The anonymous request the front end makes still returns 200, published posts only. Its shape came from `rendererPath()` in `@wordpress/server-side-render`: `context=edit`, block in the path, no `post_id`.
- Without `liveUpdateEnabled` the hook returns core's markup byte for byte. Markup with a sibling list, or content before or after the container, comes back byte for byte too. A list nested inside the container still swaps.
- Full suite is clean: 852 tests, 1942 assertions, no failures. Earlier runs had pre-existing `Test_Groups_Blocks` failures, since fixed on `production`.

### Manual test steps

Against the local docker env, on `seattle.wordcamp.test/2023/`. Every request below is sent with no cookie. Paste the blocks into one terminal so the shell variables carry over.

**Set up the fixtures.** One draft to try to read, one image whose alt text carries the rewrite's needle, one published post that owns the image and whose title carries the same needle, and one post holding a Latest Posts block with live update on.

```bash
docker compose exec -T wordcamp.test sh -c 'cat > /tmp/pr1917.php' <<'PHP'
<?php
$draft = wp_insert_post( array(
	'post_type' => 'post', 'post_status' => 'draft',
	'post_title' => 'PR1917 draft title', 'post_content' => 'PR1917-CANARY-BODY',
) );

$upload = wp_upload_dir();
$file   = $upload['path'] . '/pr1917.png';
file_put_contents( $file, base64_decode( 'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAACklEQVR4nGNgAAAABQABDQottAAAAABJRU5ErkJggg==' ) );
$image = wp_insert_attachment( array(
	'post_mime_type' => 'image/png', 'post_title' => 'pr1917', 'post_status' => 'inherit',
), $file );
require_once ABSPATH . 'wp-admin/includes/image.php';
wp_update_attachment_metadata( $image, wp_generate_attachment_metadata( $image, $file ) );
update_post_meta( $image, '_wp_attachment_image_alt', 'ul class=x foo=bar z' );

$carrier = wp_insert_post( array(
	'post_type' => 'post', 'post_status' => 'publish', 'post_title' => 'ul class=t baz=qux z',
) );
update_post_meta( $carrier, '_thumbnail_id', $image );

wp_insert_post( array(
	'post_type' => 'post', 'post_status' => 'publish', 'post_title' => 'PR1917 venue',
	'post_content' => '<!-- wp:latest-posts {"displayFeaturedImage":true,"addLinkToFeaturedImage":true,"liveUpdateEnabled":true} /-->',
) );

echo "DRAFT=$draft\n";
PHP

SITE="https://seattle.wordcamp.test/2023"
BR="$SITE/?rest_route=/wp/v2/block-renderer"
eval $( docker compose exec -T wordcamp.test wp --url=seattle.wordcamp.test/2023/ eval-file /tmp/pr1917.php | tr -d '\r' )
echo "draft id: $DRAFT"
```

**1. Control.** Core's own route for that draft. `401` on `production` and here, so core's own check is working either way.

```bash
curl -sk -o /dev/null -w '%{http_code}\n' "$BR/core/post-content&context=edit&post_id=$DRAFT"
```

**2. The same parameters sent to the Latest Posts path,** naming a different block in the query string. The block that renders is read from `name`, and the path does not determine it. `401` here. The request shape is in the ticket.

**3. The other shapes,** including a JSON body, which outranks the query string. All `401`. Also in the ticket.

**4. The request the live block actually makes still works.** `200`, and the rendered list holds published posts only.

```bash
curl -sk -o /dev/null -w '%{http_code}\n' "$BR/core/latest-posts&context=edit&attributes%5BpostsToShow%5D=2"
```

**5. The container rewrite.** Load the venue post logged out and look at the two values the block puts into attributes. Each should come back as one intact attribute. On `production` both end at an injected quote, and what follows becomes further attributes on the `` and the `<a>`.

```bash
curl -sk "$SITE/pr1917-venue/" | grep -o 'alt="[^"]*"' | head -1
curl -sk "$SITE/pr1917-venue/" | grep -o 'aria-label="ul[^"]*"' | head -1
```

Expected here:

```
alt="ul class=x foo=bar z"
aria-label="ul class=t baz=qux z"
```

**6. Coming Soon.** Turn it on and re-run steps 2 and 4. Both return `403`, matching `/wp/v2/posts`.

```bash
CS() { docker compose exec -T wordcamp.test wp --url=seattle.wordcamp.test/2023/ eval "\$s=get_option('wccsp_settings',array()); \$s['enabled']='$1'; update_option('wccsp_settings',\$s);"; \
       docker compose exec -T wordcamp.test pkill -USR2 -f 'php-fpm: master' >/dev/null 2>&1; sleep 2; }

CS on
curl -sk -o /dev/null -w 'posts:  %{http_code}\n' "$SITE/?rest_route=/wp/v2/posts"
curl -sk -o /dev/null -w 'block:  %{http_code}\n' "$BR/core/latest-posts&context=edit&attributes%5BpostsToShow%5D=2"
CS off
```

**Clean up.**

```bash
docker compose exec -T wordcamp.test wp --url=seattle.wordcamp.test/2023/ eval '
foreach ( get_posts( array( "post_type" => array( "post", "attachment" ), "post_status" => "any", "s" => "pr1917", "numberposts" => -1 ) ) as $p ) { wp_delete_post( $p->ID, true ); }
foreach ( get_posts( array( "post_type" => "post", "post_status" => "any", "s" => "ul class=", "numberposts" => -1 ) ) as $p ) { wp_delete_post( $p->ID, true ); }'
```

`opcache.revalidate_freq` is 2 in the container, so give it a couple of seconds after switching branches before re-testing. To compare against `production` without leaving the branch:

```bash
git checkout origin/production -- public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
docker compose exec -T wordcamp.test pkill -USR2 -f 'php-fpm: master'
# ... run the steps ...
git checkout HEAD -- public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
```

Unrelated to this change, and worth its own fix: the live-update script throws `Block type 'core/latest-posts' is not registered` before it ever polls, because the front-end bundle does not depend on `wp-blocks`. That is identical on `production`, which is why step 4 is a curl rather than watching the block refresh.


bor0:
Read through the change and ran it locally against `production` in the same checkout, on `seattle.wordcamp.test/2023/`. Everything in the description reproduces, including the before/after on each of the manual steps. Two things worth doing before this goes out, then some smaller notes.

### Verified

- The renderer route behaves as described on both branches: the shapes in the test plan change from a successful response to a rejection, and the request the live block actually makes is unaffected on either branch.
- Same for the Coming Soon pass. On `production` the route answers while `/wp/v2/posts` is already refusing; here they agree.
- The container rewrite, rendering the same post through `do_blocks()`:
  ```
  production    alt="div data-attributes="       aria-label="div data-attributes="
  this branch   alt="ul class=x foo=bar z"       aria-label="ul class=t baz=qux z"
  ```
  One thing worth adding to the description: in that case `production` also emits the container **without** `has-live-update`, because the class rewrite matched its needle inside an attribute value first. So the old approach doesn't only corrupt the markup, it can quietly switch off the feature it's there to turn on. That makes the change a fix for the block as well as for the markup.
- The guards line up with core rather than approximating it. `(int) get_param( 'post_id' ) > 0` is the same condition as `get_item_permissions_check()`, and `block_cannot_read` is the code core emits on both of its branches — I checked the bundled copy rather than assuming. `core/latest-posts` hardcodes `post_status => 'publish'`, and `attributes` is schema-validated before it reaches the query, so the callback that gets re-run is a narrow one.
- The new tests earn their keep. With just the two source files reverted to `production` and the test files kept, 7 of the blocks tests and 1 of the Coming Soon tests go red — matching the claim in the description.
- Compatibility, all exercised through the real filter rather than by reading: live update off is byte-identical, a non-default `orderBy` is byte-identical, a sibling list and trailing content come back untouched, a nested list still swaps with one `<ul>` left, and `data-attributes` round-trips through `rawurldecode()`/`json_decode()`. The rendered class list is `wp-block-latest-posts has-live-update is-loading`, so the front-end selector still matches.
- `--group=blocks` 24/24, `--group=coming-soon-page` 13/13.

### Two before merge

1. **The branch is five commits behind `production`** — the merge base is #1913. The two `Test_Groups_Blocks` failures on a full run are exactly that gap; they're fixed by #1915, so they're pre-existing against the merge base but not against current `production`, where a full run is clean. `git merge-tree` reports no conflicts, so merging should be uneventful and the deploy would then run a green suite.

2. **`is_swappable_container()` is the one piece of the new code that still reasons about markup with a regex**, and it gates everything else. It is safe as written — `esc_attr()` escapes `<`, so `<ul` and `</ul` can't appear literally inside an attribute value, which is exactly why the old bug needed the quote-less `ul class=` needle rather than a tag. But the docblock explains the nesting walk without stating that dependency, in the one function whose reason for existing is that string matching wasn't trustworthy. Worth a sentence, particularly because the failure mode is silent: a mis-parse just means live update stops attaching, with nothing logged.

### Smaller

- The safelist means the route will render a published-post listing with a caller-chosen `postsToShow`, and the block's schema declares that as `number` with no maximum, with `displayPostContent` + `full_post` reachable. Nothing there isn't already public, so it's a cost question rather than a disclosure one — but this branch is the first place that access is deliberate rather than incidental, so it may be worth a bound.
- `blocks.php:19` still gates only on `class_exists( 'WordCamp_Post_Types_Plugin' )`, so the hook still loads on `central`. Harmless now the guards hold; noting it only because the description raises it and doesn't change it.
- The `PHP_INT_MAX` registration does what the comment says. `init()` derives the flag from `current_user_can( 'edit_posts' )`, so the after pass covers the same set as the before pass and organisers keep their REST access — no new lockout to worry about.
- Step 5 of the manual plan didn't work for me as written: the venue post rendered without the block on my setup, so I compared through `do_blocks()` instead. Same evidence either way, and quite possibly just my local site — but if it reproduces for you, that step could use a fallback.

mcliwanow:
✅  "The branch is five commits behind production". But this might be outdated later too 😉 

✅  "Worth a sentence, particularly because the failure mode is silent: a mis-parse just means live update stops attaching, with nothing logged."

re `has-live-update`:

I don't think it holds on a rendered page. Core emits the container as:

```
class="wp-block-latest-posts__list wp-block-latest-posts is-layout-flow wp-block-latest-posts-is-layout-flow"
```

There's a space after `wp-block-latest-posts`, so the old str_replace needle matches and production does add the classes. Those layout classes come from block supports, and `do_blocks()` alone doesn't append them. Probably the same reason step 5 didn't render the block for you. I'd leave it out of the description.

✅ Description updated.

> The safelist means the route will render a published-post listing with a caller-chosen postsToShow, and the block's schema declares that as number with no maximum, with displayPostContent + full_post reachable. Nothing there isn't already public, so it's a cost question rather than a disclosure one — but this branch is the first place that access is deliberate rather than incidental, so it may be worth a bound.

I checked the network: 1,809 sites, median 12 published posts, p99 90, max 610 (lima.wordcamp.org/2020, 8.2 MB of content). So one request tops out at that site's archive rather than an unbounded number. Modest, but 8 MB uncached per request is worth closing, so I've bounded it at 100 in the safelist rather than deferring.

mcliwanow:
Discussed with @bor0 on Slack and I'm going to merge the PR as is

PR #1918, opened 2026-08-20
WC QBO: skip REST route registration when unconfigured

KokkieH:
## Summary

The QBO integration authenticates its two REST routes (`/expense`, `/paid_invoices`) with an HMAC over the request, keyed on `WORDCAMP_QBO_HMAC_KEY`. When that secret isn't provisioned, the routes are still registered even though no caller can produce a signature the server will accept — so the endpoints exist but can never be used successfully, and nothing reports the misconfiguration.

This brings the plugin in line with its sibling `wordcamp-qbo-client`, which already returns early when the secret is absent (`wordcamp-qbo-client.php:45-47`).

## Changes

- **`wordcamp-qbo.php`** — `rest_api_init()` returns early when the secret is unset, so the routes aren't registered, and logs `qbo_hmac_key_missing` so the misconfiguration is visible rather than silent. Hooked here rather than in `plugins_loaded()` so the log only fires on REST requests instead of on every page load.
- **`wordcamp-qbo.php`** — `is_valid_request()` returns early when the secret is unset.
- **wcorg-misc.php** — the `wordcamp_qbo_options` filter returns early when `WORDCAMP_QBO_HMAC_KEY` is undefined, mirroring the `wordcamp_qbo_client_options` filter directly below it. Without this, an undefined constant is an uncaught `Error` on PHP 8 that takes down every request on the site, not just REST ones. `categories_map` is left unset too, which is unobservable: it's read only by the `/expense` route callback, and that route isn't registered without a key.

No behaviour change on a host where the secret is provisioned — the new guards are no-ops there.

## Testing

1. **Baseline.** `.docker/wp-config.php` ships a local value for the secret, so the routes should be registered:
   ```bash
   curl -sk 'https://central.wordcamp.test/wp-json/wordcamp-qbo/v1/paid_invoices?invoice_ids=1' \
     -o /dev/null -w '%{http_code}\n'
   ```
   Expect **401** — the route exists and permission was denied (no `Authorization` header).

2. **Simulate an unprovisioned host.** Create `public_html/wp-content/mu-plugins/sandbox-functionality.php` (git-ignored):
   ```php
   <?php
   add_filter( 'wordcamp_qbo_options', function( $options ) {
       $options['hmac_key'] = '';
       return $options;
   }, 20 );
   ```

3. **Re-run the request from step 1.** Expect **404** with `rest_no_route` — the routes are no longer registered.

4. **Confirm the misconfiguration is reported:**
   ```bash
   docker compose logs wordcamp.test | grep qbo_hmac_key_missing
   ```
   Expect one entry per REST request, naming the requested route.

5. **Remove `sandbox-functionality.php`** and re-run the request from step 1. Expect **401** again.

6. **Regression check — a legitimately signed request still reaches the callback.** Sign the same request with the configured secret, the way `wordcamp-qbo-client` does:
```bash
SIG=$(docker compose exec -T wordcamp.test wp eval '
  $payload = json_encode( array( "get", strtolower( esc_url_raw( home_url( "/wp-json/wordcamp-qbo/v1/paid_invoices" ) ) ), "", array( "invoice_ids" => "1" ) ) );
  echo hash_hmac( "sha256", $payload, WORDCAMP_QBO_HMAC_KEY );
' --url=central.wordcamp.test | tr -d '\r\n')

curl -sk -H "Authorization: wordcamp-qbo-hmac $SIG" \
  'https://central.wordcamp.test/wp-json/wordcamp-qbo/v1/paid_invoices?invoice_ids=1' \
  -o /dev/null -w '%{http_code}\n'
```

Expect **500**, not 401. The 500 is what confirms the request passed the permission callback and reached the route callback, which then fails because no QuickBooks OAuth tokens are stored in the local network options. That failure is pre-existing and unrelated to this change — running the same two commands on `production` branch gives the same result.

`php -l` passes on both files. `phpcs`/PHPUnit weren't run locally (no `vendor/` in this checkout), so CI is the first lint pass.

## Notes

`prime_classes_cache()` (`wordcamp-qbo.php:249-251`) notes an intent to eventually remove `wordcamp-qbo-client` and network-activate `wordcamp-qbo`. That would delete the one half of this integration that currently handles an absent secret, and would register these routes on every site in the network rather than just central — so this is useful groundwork for that refactor.

### DIFF
--- a/public_html/wp-content/mu-plugins/wcorg-misc.php
+++ b/public_html/wp-content/mu-plugins/wcorg-misc.php
@@ -295,11 +295,13 @@ add_filter( 'change_locale', function() {
 
 // WordCamp.org QBO Integration.
 add_filter( 'wordcamp_qbo_options', function( $options ) {
-	// Secrets. The plugin withholds its REST routes when the key is absent, so leaving this unset is safer
-	// than fataling on an undefined constant.
-	if ( defined( 'WORDCAMP_QBO_HMAC_KEY' ) ) {
-		$options['hmac_key'] = WORDCAMP_QBO_HMAC_KEY;
+	// Secrets. Matches the `wordcamp_qbo_client_options` filter below; the plugin withholds its REST routes
+	// when the key is absent, so bailing is safer than fataling on an undefined constant.
+	if ( ! defined( 'WORDCAMP_QBO_HMAC_KEY' ) ) {
+		return $options;
 	}
+
+	$options['hmac_key'] = WORDCAMP_QBO_HMAC_KEY;
 
 	// WordCamp Payments to QBO categories mapping.
 	$options['categories_map'] = array(
--- a/public_html/wp-content/mu-plugins/wcorg-misc.php
+++ b/public_html/wp-content/mu-plugins/wcorg-misc.php
@@ -295,8 +295,11 @@ add_filter( 'change_locale', function() {
 
 // WordCamp.org QBO Integration.
 add_filter( 'wordcamp_qbo_options', function( $options ) {
-	// Secrets.
-	$options['hmac_key'] = WORDCAMP_QBO_HMAC_KEY;
+	// Secrets. The plugin withholds its REST routes when the key is absent, so leaving this unset is safer
+	// than fataling on an undefined constant.
+	if ( defined( 'WORDCAMP_QBO_HMAC_KEY' ) ) {
+		$options['hmac_key'] = WORDCAMP_QBO_HMAC_KEY;
+	}
 
 	// WordCamp Payments to QBO categories mapping.
 	$options['categories_map'] = array(
--- a/public_html/wp-content/plugins/wordcamp-qbo/wordcamp-qbo.php
+++ b/public_html/wp-content/plugins/wordcamp-qbo/wordcamp-qbo.php
@@ -72,6 +72,18 @@ class WordCamp_QBO {
 	 * Runs during rest_api_init.
 	 */
 	public static function rest_api_init() {
+		/*
+		 * The routes below are authenticated solely by an HMAC signature. Every input to that signature comes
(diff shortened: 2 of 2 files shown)


public_html/wp-content/mu-plugins/wcorg-misc.php
```diff
@@ -295,8 +295,11 @@ function wcorg_switch_to_blog_locale( $new_blog_id, $prev_blog_id, $context ) {
 
 // WordCamp.org QBO Integration.
 add_filter( 'wordcamp_qbo_options', function( $options ) {
-	// Secrets.
-	$options['hmac_key'] = WORDCAMP_QBO_HMAC_KEY;
+	// Secrets. The plugin withholds its REST routes when the key is absent, so leaving this unset is safer
+	// than fataling on an undefined constant.
+	if ( defined( 'WORDCAMP_QBO_HMAC_KEY' ) ) {
+		$options['hmac_key'] = WORDCAMP_QBO_HMAC_KEY;
```
mcliwanow: (non-blocking) that works, based on what I see in how filter `wordcamp_qbo_options` is used (and how `categories_map` is used) we could also do sth similar to what `wordcamp_qbo_client_options()` does:

```php
	if ( ! defined( 'WORDCAMP_QBO_HMAC_KEY' ) ) {
		return $options;
	}
```

PR body mentiones that we keep `categories_map` assigned on purpose but not sure we even want to have it set when hmac_key is empty

mcliwanow (approved):
Tests as described, I also checked usages of filter `wordcamp_qbo_options` and of const `WORDCAMP_QBO_HMAC_KEY` and changes make sense. Thanks!

public_html/wp-content/mu-plugins/wcorg-misc.php
KokkieH: Makes sense. Applied in 2a841a1

PR #1921, opened 2026-08-21
WCPT, Events: validate the Event Subtype filter and the Campus Connect intake

mcliwanow:
## Summary

The Event Subtype filter reads `type` from the request in two places and passes it straight on. Both readers assume it is one of the five keys `get_event_subtypes()` defines, and neither checks. An unrecognised value is an undefined array key in `alter_views()`, and the same value gets spliced into the view links that `WP_List_Table` builds with `esc_url()`.

`create_campus_connect_tracker()` stores form values as they arrive and applies no submission cap. The other two Grunion converters sanitise in `validate_data()` and call `is_rate_limited()`; this one did neither.

## Changes

- **`wordcamp-admin.php`** routes both readers through one accessor that checks the value against the known list. They also disagreed about the source, `$_GET` in one and `$_REQUEST` in the other, so a POSTed `type` filtered the query while the header showed no filter. Both read `$_GET` now, which is where the subtype links put it. An unrecognised value renders the unfiltered list with "All Events" active, instead of zero rows and a view labelled off an undefined key. The subtype label is escaped where it is spliced.
- **`metabox-status.php`** had the same undefined key, indexing the list with the stored `event_subtype` meta. The edit-capable dropdown also offered no option for a value outside the list, so a plain Update silently rewrote it to the first entry. The status dropdown in the same view already injects the stored value to avoid that; the subtype one now does too.
- **`class-event-admin.php`** applies the same list on save. Storing whatever arrived is how a value outside the list got there in the first place.
- **`jetpack-form-to-wcpt-application.php`** sanitises the field values where every caller reads them, flattens an array value instead of storing the literal `Array`, and applies the same 3 per address per hour cap as the application form. The cap counts by post status and this network never registers the `wordcamp` ones, so they are registered before the check; without that the query filters on statuses that cannot match and the cap never trips. It also restores the current blog on the insert failure path, which returned while still switched to central.

The cap is per address, and campuses tend to NAT behind one, so a fourth organiser from
the same university within an hour gets the thank-you message with no tracker created.
Nothing is lost when that happens: the feedback post is written before the hook runs, so
the submission is still in Jetpack Forms responses, and the skip is logged with that
feedback post id.

Sanitising changes little in practice. The metabox escapes already, and the vendored Jetpack sanitises `$field->value` before the hook fires, so what it adds today is newline and tab collapsing plus the array handling. Existing rows are unaffected.

### Event Subtype filter

1. **Tabs still filter.** On `edit.php?post_type=wordcamp`, click through the five subtypes and All Events. Each highlights its own tab, relabels the status views, and filters the list. Add `&post_status=wcpt-closed` to one and check the status carries through.

2. **An unrecognised value no longer filters to nothing.** Load `edit.php?post_type=wordcamp&type=not-a-subtype` and expect the full list with "All Events" active. On `production` it gives zero rows and a view labelled off an undefined key.

3. **The status metabox.** Set the meta outside the list, open the post, expect the raw slug rather than a warning, then set it back:
   ```bash
   wp post meta update <id> event_subtype not-a-subtype --url=central.wordcamp.test
   ```

### Campus Connect intake

The handler dispatches on an exact `entry_permalink` match against the production URL, so a local form cannot reach it. Steps 1 and 2 are local-only, and step 6 removes them.

1. **Create a form page** at `campusconnect/organize` on the events site, fields labelled Name, Email, Username, Campus, City, Country, Date, Number of Attendees. Name before Username, the lookup is substring based. The module is off on that site by default:
   ```bash
   wp jetpack module activate contact-form --url=events.wordpress.test
   ```

2. **Dispatch the handler.** In `public_html/wp-content/mu-plugins/0-sandbox-functionality.php`, which is git-ignored:
   ```php
   <?php
   add_action( 'grunion_after_feedback_post_inserted', function ( $post_id, $fields, $is_spam, $entry_values ) {
       if ( $is_spam || 'https://events.wordpress.test/campusconnect/organize/' !== ( $entry_values['entry_permalink'] ?? '' ) ) {
           return;
       }
       ( 'WordPress_Community\Applications\JetpackIntegration\create_campus_connect_tracker' )( $post_id, $fields, $is_spam, $entry_values );
   }, 10, 4 );
   ```
   Calling the handler beats filtering the permalink, since Jetpack builds the form's `action` from it and a local submission would then POST to the real events.wordpress.org.

3. **Submit once.** Expect a "WordPress Campus Connect …" tracker at Needs Vetting on central.

4. **Submit three more times.** The fourth still shows the thank-you message, because the cap cannot surface an error there, but no fourth tracker appears:
   ```bash
   docker compose logs wordcamp.test | grep campus_connect_application_rate_limited
   ```
   Delete the trackers to reset the hour.

5. **Sanitising.** Submit a campus name containing a tab and an inline tag. `Venue Name` reads as plain single-line text, while `_application_data` keeps the tab, since that record renders through `nl2br()`.

6. **Remove the setup:** delete `0-sandbox-functionality.php` and the two pages.

### Automated

`phpcs` clean on all seven files. `WordCamp Post Type` is 53 tests, 18 of them added here across two files covering the accessor, the query filter and the rate limit. The full run is 750 tests with 4 failures, and the same 4 fail on `production` in the same checkout: two `Test_Error_Handling::test_is_third_party_file` cases that assert on core paths and pass in CI, and two `Test_Groups_Blocks` cases that need the blocks built first.


bor0:
Reviewed this and ran it locally alongside `production` in the same checkout. The description holds up on every point I checked. A few notes, none blocking.

### Verified

- `--group wcpt` is 53/53. The full suite is 768 tests with 0 failures in my checkout — I don't hit the 4 mentioned, since I have the Groups blocks built.
- `phpcs` issue counts are identical per file before and after, so nothing new was introduced.
- Walked the five subtype tabs and All Events, with and without `&post_status=`. Active tab, regenerated views and the status carry-through all match `production`.
- Both metabox branches with a stored value outside the list. On `production` the read-only branch logs `Undefined array key` and renders an empty label; here it renders the stored slug. On the edit-capable side `production` offers no matching option, so a plain Update rewrites the meta to the first entry — this branch leaves it alone.
- Called `create_campus_connect_tracker()` directly on the events blog rather than building the form pages. Five calls produce three trackers and then the cap holds; `production` produces five. The blog is restored on all three return paths. `Venue Name` arrives as single-line text while `_application_data` keeps the original whitespace, as described.

### Compatibility

Nothing existing gets rejected: `metabox_save()` and the Campus Connect handler are the only two writers of `event_subtype`, and neither can emit a value outside the list. Rows that already hold a stray value survive an Update here and don't on `production`. `find_first_field_matching_label()` has no callers outside its own file, so narrowing the return type isn't an API change. `$_REQUEST` → `$_GET` only affects a POSTed `type`, and `edit.php`'s `posts-filter` form is `method="get"`. Meetups share the metabox, but `Meetup_Admin::get_event_subtypes()` has a single key that the metabox already defaults to. `wcor-mailer.php:315` re-enters `metabox_save()` with the same `$_POST`, so both calls agree.

### Notes

1. `( new \WordCamp_Loader() )` builds a whole second loader to reach one method, re-registering `wp_insert_post_data`, `pre_get_posts`, `cron_schedules` and the rewrite/query-var filters on a second instance. I went through each one — all idempotent or guarded by a post-type check, so there's no functional problem — but a direct `register_post_status()` loop would avoid the surprise.

2. The cap is per address, and Campus Connect's audience is university networks that NAT behind a single one. Four organizers from one campus within an hour means the fourth sees the thank-you message, no tracker appears, and the only trace is the log line. The feedback post is written before the hook so nothing is actually lost — probably worth telling the Community team that Jetpack Forms responses are the fallback.

3. `find_first_field_matching_label()` still calls `str_contains( $field->attributes['label'], $needle )` with no fallback, and Jetpack's default for the `label` attribute is `null`, which is a deprecation on 8.1+. The loop just above already handles that case with `?? $field_id`, so it's inconsistent within the one function now. Cheap to align.

4. `test_counting_depends_on_registered_statuses()` can't fail for the reason its name gives — the harness always registers the statuses, so it reduces to `test_limited_at_the_threshold` plus an assertion about the fixture. Fine as documentation, but it wouldn't catch a regression in what it describes.

5. `_application_submitter_ip_address` is still written straight from `$_SERVER['REMOTE_ADDR']` without unslashing. Pre-existing, but it moved from informational to load-bearing for the cap.

Logging the feedback post id rather than the tracker id is a nice touch — that's what makes a capped submission recoverable.

mcliwanow:
Thanks, this is a thorough pass. Especially useful that you ran it against `production` in
the same checkout and compared phpcs counts per file, and I had not checked
`wcor-mailer.php:315` re-entering `metabox_save()` at all.

Four of the five notes are in. One I think is not right, with evidence below.

**1, the second loader.** Agreed, and changed. It is a `register_post_status()` loop now.
One thing that came out of doing it: passing only `label` is not equivalent.
`register_post_status()` infers `internal => true` when `public`, `internal`, `protected`
and `private` are all unset, so the statuses would have come out `internal` on the events
network and `public` on central. The loop passes `public => true` explicitly to keep the
two the same. Verified the cap still trips on the 4th submission and that the status object
comes out `public => true, internal => false`, matching central.

**4, the test.** You are right, and it is the second time I have put a weak test in that
spot. It now removes the statuses from `$wp_post_statuses`, asserts the cap stops counting,
restores them, and asserts counting comes back. The middle assertion is the one that does
the work: if unregistering made no difference it would fail.

**5, `wp_unslash`.** Correct that `wp_magic_quotes()` slashes `$_SERVER` (`load.php:1290`),
so it should be unslashed. I have left it out of this PR deliberately rather than half-doing
it: `is_rate_limited()` reads the same raw `$_SERVER['REMOTE_ADDR']` for the query, and
`class-wordcamp-application.php:212` writes it the same way, so unslashing only the Campus
Connect write would leave the reader and the other writer inconsistent for no functional
gain, since an IP contains nothing `add_magic_quotes()` escapes. Filing it as one small
change covering all three sites. Say the word if you would rather have it here.

**2, the shared address.** Fair, and your framing is better than mine. The feedback post is
written before the hook, so a capped submission is recoverable from Jetpack Forms responses
rather than lost. I will make sure the Community team knows that is the fallback, and I have
added it to the PR description.

**3, the null label.** I do not think this one holds. The default at
`class-contact-form-field.php:112` is indeed `null`, but it never reaches
`$field->attributes`: `Contact_Form_Shortcode::__construct()` does
`$this->attributes = $this->unesc_attr( $attributes )`, and `unesc_attr()` casts with
`(string) $value`. Both the shortcode and block paths run that constructor.

Checked it rather than just reading: a form with one labelled and one unlabelled field
gives `label='Campus'` and `label=''`, and `str_contains()` over both raises no deprecation
on 8.3.

The inconsistency you are pointing at is real though. The loop above uses
`$field->attributes['label'] ?? $field_id` and this one does not, which reads as if one of
them knows something the other does not. Happy to align them for that reason alone if you
think it is worth it, I just did not want to add a fallback implying a case that cannot
happen.

bor0 (approved):

PR #1924, opened 2026-08-21
WordCamp Base: consistent output escaping in the post listing templates

KokkieH:
## What

Follow-up to #1914, extending the same treatment across the rest of `wordcamp-base` and `wordcamp-base-v2`. Two recurring habits are cleaned up:

- **Markup inside translatable strings.** Msgids that wrapped their text in `<span>` markup, or consisted of markup plus a placeholder, now keep only the translatable text; the markup lives in the template.
- **Values and labels reaching output unescaped.** Interpolated values go through the escaper matching their context, and translated labels are escaped where they are printed.

One commit per class of change, so they can be read independently:

| Commit | Change |
| --- | --- |
| 1 | The "continue reading" more link in `loop.php` and `content.php` — escape the interpolated post title, move the arrow markup into the template |
| 2 | Post and comment navigation labels, plus the `Posted in` / `Tagged` entry-utility labels — markup out of the msgid, translator contexts added |
| 3 | The author bio in `single.php`, `loop-single.php` and `author.php` — escape the display name, and use `esc_html__()` rather than `esc_attr__()` for element text |
| 4 | The shared `twentyten_posted_on()` / `twentyten_posted_in()` helpers in both themes |
| 5 | The remaining translated labels — bare `_e()`, and the labels handed to `edit_post_link()`, `comments_popup_link()`, `edit_comment_link()` and `wp_link_pages()` |

## Why

These themes predate the current coding standards. Keeping markup out of translatable strings is what the i18n guidelines ask for, and it leaves far simpler strings behind — `Older posts` instead of `<span class="meta-nav">&larr;</span> Older posts`.

Two notes for reviewers:

- **`twentyten_posted_in()` is live; `twentyten_posted_on()` is not.** `twentyten-overrides.php` defines `twentyten_posted_on()` and is loaded first by `lib/utils/functions.php`, so the `function_exists()`-guarded copy in `twentyten-functions.php` never runs. There is no override for `twentyten_posted_in()`, so that one is the copy in use. Both are updated, since the idiom shouldn't survive a refactor of the loader.
- **`twentyten_posted_in()` is filtered, not escaped.** Its three strings carry an `<a href … title … rel="bookmark">` that a translator has to keep and reposition, so escaping them would render the anchor as text. The composed sentence goes through `wp_kses_post()` instead, which keeps the link. Verified that the category list and the permalink anchor come through with `href`, `title` and `rel` intact.
- Commit 5 changes no msgids at all — it only changes which function prints them.

## Translation impact

Twelve msgids retired, nine added. The retired strings were markup, or markup plus a placeholder; what replaces them is the text that was inside.

| Retired | Replaced by |
| --- | --- |
| `<span class="meta-nav">&larr;</span> Older posts` | `Older posts` |
| `Newer posts <span class="meta-nav">&rarr;</span>` | `Newer posts` |
| `<span class="meta-nav">&larr;</span> Older Comments` | `Older Comments` |
| `Newer Comments <span class="meta-nav">&rarr;</span>` | `Newer Comments` |
| `&larr; Older Comments` | `Older Comments` |
| `Newer Comments &rarr;` | `Newer Comments` |
| `Continue reading <span class="meta-nav">&rarr;</span>` | `Continue reading` |
| `Continue reading %s <span class="meta-nav">&rarr;</span>` | `Continue reading %s` *(already in the textdomain from #1914)* |
| `View all posts by %s <span class="meta-nav">&rarr;</span>` | `View all posts by %s` *(already in the textdomain)* |
| `<span class="%1$s">Posted in</span> %2$s` | `_x( 'Posted in', 'post categories label' )` |
| `<span class="%1$s">Tagged</span> %2$s` | `_x( 'Tagged', 'post tags label' )` |
| `<span class="%1$s">Posted on</span> %2$s <span class="meta-sep">by</span> %3$s` | `_x( 'Posted on', 'post publication date label' )` + `_x( 'by', 'post author separator' )` |
| `%1$s <span class="meta-sep">by</span> %2$s` | `_x( 'by', 'post author separator' )` |

The two v2 comment-navigation strings collapse into the same msgids as the `wordcamp-base` ones, which is why nine strings replace twelve. Existing locale translations for the retired strings will orphan until the new ones are translated; the surviving text is short and mostly single words. Following #1914, the bare labels get a translator context so they are not ambiguous in GlotPress.

## Testing

`wordcamp-base` renders on ordinary page types, so most of this needs no special setup. Steps 3 and 6 are the two exceptions.

1. **Activate the theme** on a local test site and turn off the Coming Soon page, which otherwise replaces the theme's templates entirely:
   ```
   wp theme activate wordcamp-base --url=<site>
   wp option patch update wccsp_settings enabled off --url=<site>
   ```

2. **Walk the page types** and confirm each renders unchanged, with no PHP notices:
   ```
   for p in "/" "/?s=test" "/category/uncategorized/" "/author/<user>/" "/no-such-page-xyz"; do
     curl -s "<site>$p" | grep -iE 'fatal error|Warning:|Notice:' && echo "NOTICE on $p"
   done
   ```
   Check by eye: the `Posted on <date> by <author>` byline, the `Posted in` / `Tagged` entry-utility labels, the `Leave a comment` link, and the 404 and no-results copy.

3. **The more link needs a `<!--more-->` tag**, since without one `the_content()` never emits one. Add it to a post, then load the posts index — not an archive, as `loop.php:144` renders excerpts for `is_archive()` and `is_search()`:
   ```
   wp post list --post_type=post --fields=ID,post_title --url=<site>
   # insert <!--more--> into the post content, then:
   curl -s '<site>/' | grep -o 'Continue reading.*meta-nav[^<]*<[^>]*>'
   ```
   Expect `Continue reading <span class="screen-reader-text">Title</span> <span class="meta-nav">&rarr;</span>`.

4. **Check a single post** for `twentyten_posted_in()`, whose sentence is now filtered rather than printed directly. The category link and the trailing permalink anchor must both survive:
   ```
   curl -s '<site>/<post-slug>/' | grep -o 'This entry was posted.*permalink</a>'
   ```

5. **Check the author bio** on a single post and on `/author/<user>/` — the `About <name>` heading and the `View all posts by <name> &rarr;` link. Set a display name containing a quote, an ampersand and a `<code>` tag; the name should render as its text content rather than as markup, and the arrow should still appear.

6. **Paginated comments** exercise the comment navigation labels, which need `page_comments` on and more comments than the page size:
   ```
   wp option update page_comments 1 --url=<site>
   wp option update comments_per_page 1 --url=<site>
   ```
   Then load a post with two or more comments and confirm `&larr; Older Comments` / `Newer Comments &rarr;` render with their arrows.

7. **Repeat 1–5 for `wordcamp-base-v2`** (`wp theme activate wordcamp-base-v2`), where the more link also renders on archives since `content.php:19` only excludes `is_search()`. (**Note:** this theme is not loading any styles for me locally, but still possible to verify the copy hasn't changed.)

Verified locally: `php .github/bin/phpcs-branch.php` reports no findings, `php -l` passes on all 32 changed files, and the six page types above render with no PHP notices. The entity-bearing strings (`&larr; WordPress.org`, `can&rsquo;t`, `Search &hellip;`) were checked specifically — `esc_html()` defaults to `$double_encode = false`, so they pass through unchanged rather than becoming `&amp;larr;`.


bor0:
Read through all 32 files and ran the branch locally against `production` in the same checkout, on a site with `wordcamp-base` activated. One thing to fix before merge, plus some scope notes.

### `image.php:117` breaks the trackback link

`wordcamp-base-v2/image.php:117` applies `esc_html_e()` to a msgid that contains an anchor, so it renders as visible markup rather than a link:

```
Trackbacks are closed, but you can &lt;a class=&quot;comment-link&quot; href=&quot;#respond&quot; ...&gt;post a comment&lt;/a&gt;.
```

I grepped both themes for escapers applied to strings containing markup, and this is the only hit — and lines 113 and 115 right above it carry the same pattern and were left as `printf( __( … ) )`. So within four lines one HTML-bearing string is escaped and the other two aren't. `wp_kses_post()` would match what you already chose for `twentyten_posted_in()`.

### Three spots the treatment didn't reach

1. **`twentyten_continue_reading_link()`** in `twentyten-functions.php` (both themes) still has `__( 'Continue reading %s <span class="meta-nav">&rarr;</span>' )` and the unescaped `get_the_title()` — the same thing commit 1 fixes in `loop.php` and `content.php`. It isn't dead code: it's on `excerpt_more` and `get_the_excerpt`, so it renders on exactly the archive and search path that step 3 tells reviewers to avoid. This also means the translation table is slightly off — that msgid is listed as retired, but this copy keeps it alive in the POT.

2. **`wordcamp-base-v2/content-single.php:40-50`** builds the same `This entry was posted in …` sentence and prints it with a bare `printf()` and an unescaped `get_permalink()`, while the `twentyten-functions.php` copy got `wp_kses_post()`. Same sentence handled two ways in one theme.

3. **`archive.php`** in both themes — only the `else` branch was escaped. `Daily / Monthly / Yearly Archives: <span>%s</span>` still keep markup in the msgid with an unescaped `get_the_date()`.

None of those block this PR if the intent is to land the rest first, but 1 is worth folding in, since the excerpt path is the more commonly rendered one.

### Verified

- The entity claim holds — `&larr;`, `&hellip;` and `&rsquo;` all come through `esc_html()` byte for byte.
- `twentyten_posted_in()` through `wp_kses_post()` is byte-identical to `production` output, with `href`, `title` and `rel` intact, including for a title carrying quotes and an ampersand.
- The author-link fix is doing real work: on `production` the attribute renders as `title="View all posts by Tom &amp; "`, truncated at the quote. On this branch the full name is there.
- The override load order is as described in **both** themes, so the note about `twentyten_posted_on()` being the unused copy applies to v2 as well.
- `phpcs` findings drop on every one of the 32 files (563 → 462 in total, none regress), and `php -l` is clean on all of them.
- Six page types (home, search, category, author, 404, single) render with zero PHP notices on both branches, and the only output differences are the intended escaping changes.

### Nits

- Step 3 expects `<span class="screen-reader-text">`, but v2's `content.php` uses `assistive-text`, and step 7 says to repeat steps 1-5 there.
- The more link loses one space (`</span>  <span` becomes `</span> <span`) because the old `the_title()` `$after` argument already carried a trailing space. Invisible once rendered — flagging it only so it doesn't read as accidental in a later diff.

I didn't get to the paginated comment navigation (step 6) or the v2 attachment page, since both needed fixture changes I didn't want to leave behind locally. The `image.php` finding doesn't depend on either — I confirmed that one at the function level.

mcliwanow:
Read through all five commits. One thing I'd fix before merge, then a couple worth discussing.

### `image.php:117` escapes an anchor

```php
esc_html_e( 'Trackbacks are closed, but you can <a class="comment-link" href="#respond" title="Post a comment">post a comment</a>.', 'wordcamporg' );
```

Output on the branch:

```
Trackbacks are closed, but you can &lt;a class=&quot;comment-link&quot; href=&quot;#respond&quot;&gt;post a comment&lt;/a&gt;.
```

The two branches right above it carry the same kind of anchor and stayed `_e()`, so I assume this one is a slip. Renders on attachment pages when comments are open and pings are closed. I checked every other `esc_html*()` you added for a tag in the msgid and this is the only one. The `&rsquo;` and `&larr;` strings are fine, confirmed they pass through unchanged.

### `wp_kses_post()` in `twentyten_posted_in()` holds up, the `esc_url()` inside it isn't doing anything

I ran the composed sentence through the new path:

```
esc_url out:  https://example.org/?p=1&#038;post_type=x&#038;foo=bar
after kses:   <a href="https://example.org/?p=1&amp;post_type=x&amp;foo=bar" title="Permalink to Bob&apos;s &quot;quoted&quot; &amp; titled post" rel="bookmark">permalink</a>
```

Nothing double-encodes, `href`/`title`/`rel` all survive, so your verification holds. `wp_kses_hair()` runs on `WP_HTML_Tag_Processor` now, so it decodes each attribute and re-encodes with `strtr()`, which is why `&#038;` comes back as `&amp;` and `&#039;` as `&apos;`. Cosmetic.

The `esc_url()` on `%3$s` is redundant though: kses already runs `wp_kses_bad_protocol()` on `href` and re-encodes the value regardless. Not wrong, but the block now escapes its arguments three different ways before handing them all to one kses call (raw HTML for `%1$s` and `%2$s`, `esc_url()` for `%3$s`, `esc_attr()` for `%4$s`), and the comment saying the result "is filtered rather than escaped" doesn't quite match that. Either drop the `esc_url()`, or say in the comment that kses is doing the work and the rest is extra.

Separate and pre-existing, but you're rewriting that exact call: `sprintf( $posted_in, ... )` takes its format string from `__()`, so a locale that mistypes a `%` throws a `ValueError` and fatals the single template. The old `printf` had the same problem. Up to you whether to guard it here.

### The bare label fragments

This is the one I'd push back on. `_x( 'by', 'post author separator' )` gives a translator a preposition with no sentence around it, and plenty of languages have nothing to put there. Same for `Posted on`, and to a lesser degree `Posted in` and `Tagged`. Splitting also takes away word order: date-then-author is now hardcoded in the template.

The pattern core themes use keeps the placeholder and passes the markup as the value:

```php
$byline = sprintf(
	/* translators: %s: post author link */
	esc_html__( 'by %s', 'wordcamporg' ),
	$author_link
);
```

Markup stays out of the msgid, translator gets a sentence. The catch is that `meta-prep` and `meta-sep` currently wrap only the label, so this moves the spans, and `wordcamp-base` is a parent theme with per-site child CSS. Might not be worth the risk. Your call, I just don't want the fragments landing in GlotPress unexamined.

### Smaller

- `Older Comments` and `Newer Comments` are already in the `wordcamporg` textdomain (`campsite-2017/comments.php:53` and `:57`), so those two won't orphan. Nine added is really seven new. `Continue reading %s` and `View all posts by %s` check out as carried over from #1914.
- `the_title( '<span…>', '</span> ', false )` returns nothing when the title is empty, `'<span…>' . esc_html( get_the_title() ) . '</span> '` always emits the span. Untitled posts get an empty screen-reader span in the more link now. Tiny, but it's a behavior change in `loop.php` and v2 `content.php`.
- Testing step 3 expects `screen-reader-text`, v2's `content.php` uses `assistive-text`. Only matters if someone runs the steps literally on v2 in step 7.

The two `echo esc_html_e()` fixes in the options classes are real bugs closed, and splitting this by class of change made a 32-file diff actually reviewable. Thanks for that 🙏

KokkieH:
Thanks both — pushed in db9f0455.

**`image.php`** was a real slip in the `_e()` sweep. Fixed with `wp_kses_post()`, and lines 113/115 get the same treatment so all three trackback branches match. I re-ran the audit myself rather than take "only hit" on trust: grepping for any escaper applied to a msgid containing a tag returns that one line and nothing else.

**Byline labels** — good push, and I've gone further than the core pattern: the msgids are restored exactly as they were and filtered with `wp_kses_post()` instead. No fragments in GlotPress, no orphaned translations, and the DOM is byte-identical to `production`, so no child-theme CSS risk either. `wp_kses_post()` leaves `<span class="%1$s">Posted in</span> %2$s` unchanged and still strips `on*`, so `printf()` substitutes as before. Same approach as `twentyten_posted_in()`, which makes the theme consistent rather than adding a third convention.

That also shrinks the translation impact a lot — **12 retired, 5 added, 3 genuinely new**. You were right that `Older Comments` / `Newer Comments` already exist (campsite-2017). `Continue reading` and `Older posts` turned out to be the `jetpack` textdomain rather than `wordcamporg`, so those two are still new.

**`twentyten_continue_reading_link()`** — agreed, and the more important of the three. Fixed in both themes; it was also what kept `Continue reading %s <span…>` alive in the POT, so the table was wrong on that too. Also took the empty-title point: the more link no longer emits an empty screen-reader span. `content-single.php` and `archive.php` are in as well.

**`esc_url()` in `posted_in`** — kept, comment rewritten to say kses is doing the work and the rest is belt-and-braces.

Left alone: the lost space, and `sprintf( $posted_in, … )` fatalling on a mistyped `%` in a translation. The latter wants validating at catalogue import rather than at the sink, so it feels like its own issue.

`phpcs-branch.php` clean, `php -l` clean, eight page types render with no notices. PR description still needs updating for the new table and the `assistive-text` nit in step 3 — doing that next.

bor0 (approved):

PR #1926, opened 2026-08-21
Instamojo: confirm payments with the gateway before settling orders

bor0:
## Summary

Settle a CampTix order only on a payment Instamojo confirms was credited against the payment request created for that order — the same server-side confirmation the Razorpay sibling gained in #1795.

- Checkout already hands Instamojo a `redirect_url` and `webhook` carrying the order's payment token, and the API echoes both back on a fetched payment request. That echo arrives over an authenticated channel, so it is used to tie a payment to an order, rather than the token in the incoming request — which the sender chooses, and which the webhook signature does not cover.
- `payment_return()` records a result only for a confirmed payment. Anything else is left alone and shown the ticket page; the webhook settles orders in practice.
- `payment_notify()` rejects a signature that does not verify instead of recording the order as pending, compares with `hash_equals()`, and treats an unconfigured salt as unverifiable rather than comparing against a keyless hash.
- A result Instamojo cannot confirm — because it could not be reached — is answered 5xx so the webhook is retried, rather than 4xx, which would tell it not to and leave a paid order unsettled.
- The signing payload is built from `wp_unslash( $_POST )`: WordPress escapes `$_POST` before any handler sees it, but Instamojo signed the unescaped values.
- Pre-existing undefined-index notices (`mac`, `payment_id`, `payment_request_id`) fixed, since a rejected webhook now reaches them routinely.

Nothing is written at checkout, so there is **no deploy window** — the binding already exists for every order ever created, including any sitting on a checkout page during deploy.

## How to review this PR

Suggested order: `payment_notify()` → `payment_return()` → the two helpers at the bottom.

- **Why the echoed URL rather than stored post meta.** A meta binding would only cover orders created after deploy, leaving in-flight ones unverifiable. The echo covers everything retroactively. Verified against 6 real payment requests on the two sites currently transacting — all bound to their own order, all rejecting a foreign token, none missing.
- **Only the token is compared, not the URL**, so a site domain change does not break the binding.
- **No amount comparison.** `payment_checkout()` creates the request with `'amount' => $order['total']` and `'allow_repeated_payments' => false`, so Instamojo charges the amount we set and a payment bound to that request cannot be for less. An amount check would re-verify our own figure. It would also have to be "covers" and not "equals": production has sites where the buyer is charged the gateway fee plus GST on top, so a 500 order is credited as 521.24.
- **An unconfirmed return no longer errors.** Earlier iterations refused with `wp_die()`. Production shows the webhook lands first and the return finds the order already settled, so that path barely runs — falling through to the ticket page is smaller, and means a missing `payment_request_id` on the redirect is harmless rather than fatal.

## Test plan

- [x] phpunit — 784/784 passing (14 in `CampTix Indian Payments`; 6 pre-existing skips from the missing `intl` extension)
- [x] phpcs — `.github/bin/phpcs-branch.php` clean; no new sniff types in the gateway
- [x] Mutation-tested — every check below fails the suite when removed
- [x] End-to-end against a stubbed Instamojo (below)
- [x] Verified against live production data (below)

### Automated

```bash
docker exec wordcamporg-phpunit_wp-1 bash -lc 'cd /app && phpunit'
BASE_REF=production php .github/bin/phpcs-branch.php
```

The suite is new, so it is worth confirming it holds rather than merely passes. Each edit below turns it red:

| Edit | Failures |
| --- | --- |
| `wp_die( 'Invalid signature.' … )` → record the order as pending instead | 3 |
| `payment_request_owns_token()` → `return true` after the guard clause | 3 |
| Drop `! $instamojo_salt ||` from the signature condition | 1 |
| `'Credit' === $payment->status` → `true` | 1 |
| Return's `if ( owns_token && get_credit_payment )` → `if ( true )` | 2 |
| Sign over `$_POST` instead of `wp_unslash( $_POST )` | 1 |
| Drop the `is_wp_error( $request )` retry branch | 1 |
| Drop `$payment_data['transaction_id'] = $credit->payment_id` | 1 |

### Manual test steps

The dev container has no outbound DNS, so stub the API. Save as `public_html/wp-content/mu-plugins/zzz-instamojo-stub.php` and **delete it afterwards**. `req_down` deliberately fails, to stand in for Instamojo being unreachable:

```php
<?php
add_filter( 'pre_http_request', function ( $pre, $args, $url ) {
	if ( false === strpos( $url, 'instamojo.com' ) ) {
		return $pre;
	}

	if ( false !== strpos( $url, 'req_down' ) ) {
		return new WP_Error( 'http_request_failed', 'Connection timed out.' );
	}

	$map  = array(
		'req_large' => array( 'tok_large', '5000.00' ),
		'req_small' => array( 'tok_small', '1.00' ),
		'req_quote' => array( 'tok_quote', '500.00' ),
		'req_txn'   => array( 'tok_txn', '500.00' ),
	);
	$site = 'https://seattle.wordcamp.test/2023/';

	foreach ( $map as $id => $info ) {
		if ( false === strpos( $url, $id ) ) {
			continue;
		}
		list( $tok, $amt ) = $info;
		return array(
			'response' => array( 'code' => 200, 'message' => 'OK' ),
			'headers'  => array(),
			'body'     => wp_json_encode( array( 'success' => true, 'payment_request' => array(
				'id'           => $id,
				'status'       => 'Completed',
				'amount'       => $amt,
				'redirect_url' => $site . '?tix_action=payment_return&tix_payment_token=' . $tok . '&tix_payment_method=instamojo',
				'webhook'      => $site . '?tix_action=payment_notify&tix_payment_token=' . $tok . '&tix_payment_method=instamojo',
				'payments'     => array( array( 'payment_id' => 'MOJO_' . $id, 'status' => 'Credit', 'amount' => $amt ) ),
			) ) ),
		);
	}

	return array( 'response' => array( 'code' => 404, 'message' => 'Not Found' ), 'headers' => array(), 'body' => '' );
}, 10, 3 );
```

Enable the gateway on a test camp with an `Instamojo-salt` of `dev_salt_123` in `camptix_options['payment_options_instamojo']`, and create five draft orders with `tix_payment_method` set to `instamojo`: `tok_large` (5000), `tok_small` (1), `tok_quote` (500), `tok_down` (500), `tok_txn` (500). Sign each body as Instamojo does — `hash_hmac( 'sha1', implode( '|', $body_sorted_by_key ), 'dev_salt_123' )`, `mac` excluded.

**Webhook — over real HTTP.** Use `curl`, not a browser `fetch`: with `redirect: 'manual'` a fetch reports an opaque status and will misreport these.

| # | Request | Expected | Observed |
| --- | --- | --- | --- |
| 1 | Invalid signature, naming `tok_large` | 403, order untouched | 403, `tok_large` draft |
| 2 | Valid signature for `req_small`, URL names `tok_large` | 400, neither order settled from the other | 400, `tok_large` draft |
| 3 | Valid signature for `req_small`, URL names `tok_small` | 200, order completed | 200, `tok_small` publish |
| 4 | Valid signature, `buyer_name=D'Souza` | 200, order completed | 200, `tok_quote` publish |
| 5 | Valid signature for `req_down` (Instamojo unreachable) | 503, so it is retried; order untouched | 503, `tok_down` draft |
| 6 | Valid signature, no `payment_request_id` | 400, no notice, no API call | 400, `tok_down` draft |

**Return.** This dev site gates front-end pages behind `wp-login.php`, so a plain request never reaches `payment_return()` — it is the gate that answers. These were driven through the method directly, inside a full WordPress bootstrap, with the superglobals a real return would carry:

```bash
wp eval --url=<site> '
$_REQUEST["tix_payment_token"] = "tok_txn"; $_GET["tix_payment_token"] = "tok_txn";
$_REQUEST["payment_request_id"] = "req_txn";
$_REQUEST["payment_id"] = "MOJO_supplied_by_the_browser";
$GLOBALS["camptix"]->get_payment_method_by_id( "instamojo" )->payment_return();'
```

| # | Request | Expected | Observed |
| --- | --- | --- | --- |
| 7 | Return for `tok_large` naming `req_small` | not settled | `tok_large` draft |
| 8 | Return for `tok_down`, Instamojo unreachable | not settled | `tok_down` draft |
| 9 | Return for `tok_txn` naming its own `req_txn`, plus a made-up `payment_id` | settled, recording Instamojo's payment id | `tok_txn` pending, `tix_transaction_id` = `MOJO_req_txn` |

Then remove the stub, delete the fixtures, and restore `camptix_options`.

### Verified against production

Read-only audits over all 1,644 sites, before and during this work:

- 40 sites have the gateway enabled; 17,225 historical Instamojo orders — the blast radius is real, not theoretical.
- The binding the change depends on was checked against 6 real payment requests on the two sites currently transacting: all 6 bound to their own order, all 6 rejected a foreign token, none missing the token.
- `payments[].status` is `Credit` on paid orders, confirmed against the live API.
- Two sites have the gateway enabled with no salt and no API key (1420, 1627). Both will get 503s on webhooks. 1627 has never taken a payment; 1420's orders ended some time ago with 0 draft/pending, so nothing is in flight. Their credentials should be restored or the gateway disabled, independently of this PR.
- 2,979 payment tokens across two blogs are not md5. Neither blog uses Instamojo, and `sanitize_text_field( wp_unslash() )` leaves every one of them unchanged.

#### Screenshots

| Description | Screenshot |
| --- | --- |
| Final state after all nine cases: the two orders that own their payment are settled (`Small Order` published, `Txn Check` pending), and the three that do not are untouched |  |


KokkieH:
Code review of this branch. All findings are in `public_html/wp-content/plugins/campt-indian-payment-gateway/inc/instamojo/class-camptix-payment-method-instamojo.php`.

### 1. Signature verification breaks on slashed input — line 171 (high)

`$data = $_POST;` uses the slashed superglobal. WordPress runs `wp_magic_quotes()` on every request, so any webhook field containing `'`, `"` or `\` (`buyer_name`, `purpose`, address fields) arrives with a backslash inserted. `hash_hmac( "sha1", implode( "|", $data ), $instamojo_salt )` on line 191 therefore hashes different bytes than Instamojo signed, and the signature never verifies.

Before this PR that mismatch fell into the `else` branch and recorded `PAYMENT_STATUS_PENDING`, so the order was at least visible to an organizer. Now line 203 responds `403 Invalid signature.` and records nothing: a buyer named e.g. `D'Souza` pays, the order stays `draft`, and CampTix's abandoned-order cleanup eventually cancels it.

Fix: build the signing payload from `wp_unslash( $_POST )`.

### 2. A failed API fetch is indistinguishable from a real mismatch — line 220 (medium)

`get_payment_request()` returns `false` on any transport failure — timeout, Instamojo outage, missing or invalid `Instamojo-Api-Key`/`Auth-Token` (the PR description itself names two production sites in that state). `payment_request_owns_token( false, … )` then returns `false`, so a genuine, correctly signed `Credit` webhook is answered `400 Unknown payment request.` and the order is never settled. A 4xx also tells Instamojo not to retry.

Worth distinguishing "the API did not answer" (respond 5xx so the webhook is retried, and/or record `PENDING` as before) from "the request confirmedly belongs to another order" (400).

### 3. Unguarded array access on `payment_request_id` — line 208 (low)

`$this->get_payment_request( $data['payment_request_id'] )` is still an unguarded access, even though the sibling `mac`/`payment_id` notices were fixed. A signed webhook that omits `payment_request_id` (non-RAP payloads, or a body Instamojo signs for another event type) now reaches this line routinely, raises an undefined-array-key warning, and then makes an API call to `…/payment-requests//`. Suggest `$data['payment_request_id'] ?? ''` plus an early bail.

### 4. Blocking API call on every abandoned-checkout return — line 143 (low)

When `payment_request_id` is absent from the redirect, `payment_return()` calls `get_payment_request( '' )`, issuing a blocking outbound HTTP request to `…/payment-requests//` — for every abandoned checkout return, and for anyone replaying a known payment token. The `isset()` guard prevents the notice but not the call; bail out before the API call when the id is empty.

### 5. `transaction_id` from a browser-supplied value — line 131 (low)

In the confirmed-return branch, `transaction_id` is `$_REQUEST['payment_id'] ?? ''` — chosen by the returning browser — and it is written to `tix_transaction_id` even though `get_credit_payment( $request )` on line 145 already produced the payment Instamojo confirms. A buyer appending `&payment_id=anything` stores a bogus transaction id (until a later webhook overwrites it), and that is what reconciliation and refunds key off. Capture the credit object and use `$credit->payment_id`, as `payment_notify()` does on line 234.

---

Nothing blocking in the test files: the PHP floor is 8.3, so the `ReflectionProperty::setValue`/`ReflectionMethod::invoke` calls without `setAccessible()` are fine, and the new `muplugins_loaded` priority-20 gateway loader does run after `camptix.php` defines `CampTix_Payment_Method` (the parent class is `require`d in `CampTix_Plugin::__construct()`, at include time).

KokkieH (commented):
Tests as described, though Claude identified some issues with the code. I also verified finding 1, i.e. if a user has an apostrophe in their name I get a 403, where I'd get a 200 for the same call just without the apostrophe in their name.

bor0:
Thanks — all five are real, and I verified each against the code rather than taking them on trust. Fixed in 0ae13df7, with regression tests for the three that have behaviour to pin.

**1. Signature over slashed input — agreed, and this one was a regression on my part.** I had `wp_unslash()` plus a test for it on an earlier iteration of this branch, and lost both when I rebuilt the change from `production`. Confirmed the concrete failure before fixing:

```
buyer_name as posted : D\'Souza
verifies with $_POST as-is       : NO  <-- signature rejected
verifies with wp_unslash($_POST) : YES
```

Your read of the impact is right, and worse than the old behaviour: previously the mismatch fell through to `PENDING` and an organizer could at least see it; as written it was a 403 and a `draft` that cleanup would eventually cancel. It now signs over `wp_unslash( $_POST )`, with `test_notify_verifies_the_signature_over_unescaped_values` driving a slashed payload through.

**2. Failed fetch vs real mismatch — agreed, and the retry point is the part I had missed.** I knew the two outcomes were conflated and judged it acceptable; I had not thought about 4xx telling Instamojo *not* to retry, which turns a transient outage into a permanently unsettled paid order. `get_payment_request()` now returns `WP_Error` when Instamojo was not reached (a 404 stays a definitive `false`), and the webhook answers 503 so it is retried. `payment_request_owns_token()` also rejects a `WP_Error` explicitly rather than relying on it being falsy — it is not.

**3 & 4. Unguarded `payment_request_id` — agreed on both.** I fixed the sibling `mac`/`payment_id` notices and missed this one. Both paths now read it with `??`, and neither calls the API with an empty id, so an abandoned return no longer makes a blocking outbound request for `.../payment-requests//`.

**5. `transaction_id` from a browser-supplied value — agreed.** The confirmed-return branch already had the credit object in hand and was ignoring it. It now records `$credit->payment_id`, matching what `payment_notify()` does. `test_return_records_the_confirmed_payment_id` sends `&payment_id=MOJO_supplied_by_the_browser` and asserts the stored id is Instamojo's.

Each fix is pinned by reverting it and confirming the suite goes red — 1, 2 and 5 each fail exactly one test when undone. Suite is now 14 tests (from 10); full run 784 passing; `phpcs-branch.php` clean.

Thanks also for checking the bootstrap loader ordering and the PHP floor — that saved me re-deriving both.

PR #1929, opened 2026-08-24
Speaker Feedback: attribute submissions to the current session

mcliwanow:
## Why

The speaker feedback create endpoint took the comment author from the request.

`REST_Feedback_Controller` overrides `create_item_permissions_check()`, and core authorises the `author` parameter only inside the method it replaces. `WP_REST_Comments_Controller::prepare_item_for_database()` then copies the display name, email address, and URL of whatever user `author` names, with no check of its own. `create_item()` reset `content`, `date`, and `parent`, but never `author`.

So a submission naming a user was stored as that user: their `user_id`, their display name, and their email address, on a row organisers read in the feedback moderation UI. Reproduced on a local camp site, where one request naming another account stored that account's `user_id` and address, neither of which the request supplied.

The author has always been a client-side value. `views/form-feedback.php` renders it as a hidden field filled from `get_current_user_id()`, and `assets/js/script.js` forwarded whatever was in it.

## What changed

- The author comes from the session. Logged in, it is the current user, and any `author_name` or `author_email` in the request is dropped. Logged out, there is no author, and the name and email the form collects are used.
- `author_url` is dropped either way. The form collects no URL, and `get_comment_author_link()` renders a stored one as a link on the speaker's feedback page.
- `author_ip` and `status` are dropped. Core's permissions check rejects both for anyone without `moderate_comments`, and this controller replaces that check. Neither reached the database before, since `prepare_item_for_database()` re-guards one and `handle_status_param()` is never called for the other, so this only stops that from depending on core internals.
- The form no longer sends an author it does not decide.
- Resolved the `// TODO` on the permissions check. Submitting without an account is a supported flow, so there is deliberately no capability check there.

### Smaller things

- Return type spacing in the test file, `() : void` to `(): void`. They were the only phpcs violations left in it.

## Other approaches

- Deferring to `parent::create_item_permissions_check()` looked like the obvious option, but core requires `rest_allow_anonymous_comments` (which defaults to false) and open comments, so it would reject the account-free submissions this form supports.
- Dropping `author` outright is not enough on its own, since core never defaults it to the current user and `create_item()` needs a name and email, so logged-in submissions would start failing.
- Adding a nonce was considered and left out: `wp-api-fetch` is already a script dependency, so the REST nonce is sent on every submission.
- `wp_new_comment()` was considered for its `wp_filter_comment` pass, but this controller already calls `wp_allow_comment()` and the length checks itself and always holds the comment, so it would duplicate that work and change how the status is decided.

### Setting up a session that takes feedback

Feedback opens when a session starts and closes two weeks later. Organisers bypass that gate entirely, so the attribution checks below want a logged out visitor rather than an admin.

1. **Put a published session inside that window.**
   ```bash
   wp post meta update <id> _wcpt_session_type session --url=<camp>
   wp post meta update <id> _wcpt_session_time $(( $(date +%s) - 86400 )) --url=<camp>
   ```
   The form is at `<session permalink>feedback/`.

### The form, both ways

Feedback is always held, whatever the discussion settings say, and the speaker's view lists approved rows only. So nothing appears on the page after a submission. Step 6 is how to read the row back.

2. **Logged out.** Name and Email render. Fill them in with a rating and the first question and submit. Expect "Feedback submitted.", and a row carrying what you typed with `user_id` 0.

3. **Logged in.** No Name or Email fields, just "You are leaving feedback as …". Submit and the row is attributed to your account.

4. **The author is no longer the client's to choose.** Still logged in, edit the hidden `sft-author` input to another user's id in devtools, then submit. The row is still yours. On `production` the same edit files the feedback under the other account, carrying their display name and email address.

5. **Without an account there is no identity to borrow.** A submission that names an `author` but supplies no name or email is refused with `rest_feedback_author_data_required`. On `production` it is accepted and stored under that account.

6. **Read the rows back**, and delete them between runs, since an identical resubmission returns 409 from the duplicate check.
   ```bash
   wp comment list --post_id=<id> --type=wc-speaker-feedback --fields=comment_ID,user_id,comment_author,comment_author_email,comment_author_url --url=<camp>
   ```

### Automated

`phpcs` clean on both PHP files, and `phpcs-branch.php` against `production` reports nothing. Six new tests in `Test_SpeakerFeedback_REST_Feedback_Controller` take the plugin suite from 58 tests and 143 assertions to 64 and 179. The full run is 774 tests with no failures and 6 skipped, the CampTix revenue reports that need `intl`. One of the new tests goes through `rest_do_request()` instead of calling the controller, so the permission callback and core's parameter sanitising are exercised too.

Mutation checked rather than assumed. Removing the session block fails all five attribution tests. Removing the `author_url` line fails both the anonymous and the logged-in case, and it failed only the logged-in one until that gap was covered.

`test_create_item_no_user` needed `wp_set_current_user( 0 )` added. Its name promised no user while `setUp()` logged one in, so what it actually asserted was that the endpoint ignores the session, which is the behaviour being removed here. The case it used to cover is now `test_create_item_logged_in_without_author_params`.

Compatibility: the only client is this plugin's own JS, shipped from this repo, and no stored data is read or written differently. Duplicate detection is unchanged, since `duplicate_check()` branches on `user_id`, which is set exactly when `author` is, so logged in yes and anonymous no, which matches what the form produced before. I have no production counts for how many rows this touches, so that rests on these cases rather than on numbers.

One reduction worth naming: `spam_check()` and Akismet now always see an empty `comment_author_url`.


KokkieH (approved):
On my local site the feedback gets held for moderation (even after disabling comment moderation ¯\_(ツ)_/¯), so I couldn't verify `a held row carrying what you typed, with user_id 0.` But otherwise this tests as described--verified via CLI that the correct user_id is given to anon comments.

Claude spotted a couple issues with the tests following these changes. Would be good to address, but doesn't require a second review.

> 1. tests/test-class-rest-feedback-controller.php:177 — test_create_item_user_id no longer covers what its name and `@covers` advertise. create_item() now unconditionally overwrites/unsets author, so the 'author' => subscriber param is discarded; the test only passes because setUp() logs in that same subscriber, making it a duplicate of the new test_create_item_logged_in_without_author_params. The PR stripped the dead author param from test_create_item_duplicate and test_create_item_not_duplicate — this one was missed.
> 2. tests/test-class-rest-feedback-controller.php:414 — test_create_item_no_meta (and test_create_item_permissions_check_is_valid at :505) still pass an ignored 'author' => subscriber. That leaves test_create_item_no_meta quietly dependent on the ambient session: add a wp_set_current_user( 0 ) for hygiene — matching the pattern this PR just introduced in the three anonymous tests — and it fails on rest_feedback_author_data_required instead of rest_feedback_meta_data_required.

mcliwanow:
Thanks, both fixed in 1f2a4746.

1. `test_create_item_user_id` now sets its own session and asserts the stored `user_id` and email instead of just a count. I kept the `author` param deliberately rather than stripping it like the two duplicate tests: session pages are cached, so a visitor can still be running the previous `script.js`, which sent one. That gives the test a job the newer ones don't cover. Checked by mutation, it is the test that fails if `author` is unset for everyone rather than derived from the session.

2. Good catch on the ambient session, though `wp_set_current_user( 0 )` would have moved the assertion rather than tightened it: the author check runs before the meta check, so it would start failing on `rest_feedback_author_data_required` and stop covering the missing meta at all. Set the logged-in session explicitly instead and dropped the discarded param. Also dropped it from `test_create_item_permissions_check_is_valid`, which never reads it.

Suite is 64 tests, 181 assertions.

On the moderation queue, that is not your local site. `add_feedback()` sets `comment_approved` to 0 unconditionally unless the submission is flagged as spam, so feedback is always held no matter what the discussion settings say, and the speaker's view lists approved rows only. Nothing was going to show up on the page. CLI was the right call, and I have added a line to the test instructions saying so, since they implied you would see something.

PR #1932, opened 2026-08-24
WordCamp Reports: align Meetup Status export handler with the other reports

KokkieH:
## Summary

The Meetup Status report's export handler and admin-page render were missing the capability/nonce checks that the other report classes in `wordcamp-reports/classes/report/` already apply consistently. This brings that one handler in line with its siblings:

- `Meetup_Status::export_to_file()` now verifies the `run-report` nonce and the `view_wordcamp_reports` capability before proceeding, matching e.g. `Meetup_Details`, `Meetup_Groups`, etc.
- `Meetup_Status::render_admin_page()` now checks the capability up front, so it fails closed as a whole rather than gating only the data-building step.

Both checks are the same ones already used elsewhere in the plugin — no new behavior for users who have report access.

### Tests

`wordcamp-reports` had no PHPUnit suite, so this also adds one:

- A data-provider test covering **every** report class's export handler, asserting each requires report access — so a newly added report can't accidentally ship without the same check.
- Targeted tests for the Meetup Status handler and admin-page render (capability-present renders the screen; capability-absent and nonce/capability mismatches produce nothing).

Wired into `phpunit.xml.dist` (new `WordCamp Reports` suite) and `phpunit-bootstrap.php`, following the same pattern as the other plugin suites.

## Test plan

- [x] `phpunit --testsuite "WordCamp Reports"` — 18/18 passing
- [x] Full regression run — no new failures (pre-existing Groups-blocks failures need built JS assets, unrelated to this change)
- [x] phpcs clean on all changed/new files
- [x] Manual browser pass (below)

### Manual test steps

As a user with access to the Reports screen (on central):

1. Open **Dashboard → Reports → Meetup Status**. The report screen renders normally (date range, status dropdown, both buttons).
2. Enter a date range and click **Show results** — results render as before.
3. Click **Export resulting Meetups as a CSV** and submit — a CSV downloads as before.

No change is expected in any of these flows for a user who has report access.

https://claude.ai/code/session_01HxMxx2cg9CVabUvdBAbkR4

### DIFF
--- a/phpunit-bootstrap.php
+++ b/phpunit-bootstrap.php
@@ -109,6 +109,7 @@ require_once WP_PLUGIN_DIR . '/wcpt/tests/bootstrap.php';
 require_once( WP_PLUGIN_DIR . '/wordcamp-remote-css/tests/bootstrap.php' );
 require_once WP_PLUGIN_DIR . '/wordcamp-speaker-feedback/tests/bootstrap.php';
 require_once WP_PLUGIN_DIR . '/wordcamp-payments-network/tests/bootstrap.php';
+require_once WP_PLUGIN_DIR . '/wordcamp-reports/tests/bootstrap.php';
 require_once SUT_WPMU_PLUGIN_DIR . '/tests/bootstrap.php';
 require_once SUT_WPMU_PLUGIN_DIR . '/wporg-groups-frontend/tests/bootstrap.php';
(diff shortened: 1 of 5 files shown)


bor0 (approved):
Confirm fix for unauth users and also confirm no regression for auth users

PR #1934, opened 2026-08-25
Groups: consolidate event forms

av3nger:
The event form lived in two places, the "Create event" modal on the group page (event-modal-app.js) and the Events tab of the settings dialog (events-tab.js). The copies had drifted: different strings, different media-library filtering, different error handling, slightly different payloads.

That drift was discussed in #1909 and was the reason why same input could behave differently depending on which button you clicked.

This PR merges both forms into one EventForm component in src/components/event-form/. It owns the field state, the event-form-data fetch and the request payload.

Two additional bugs turned up while testing and are fixed here too:
- The description editor's "+" block inserter drew over the modal header when you scrolled. The header was transparent and sat below the popover's z-index.
- The venue editor's map was unstyled on the front end. Leaflet's CSS was imported from the block editor entry, so it compiled into a stylesheet nothing enqueued, and the tiles flowed unpositioned across the form. Loading the CSS exposed a second problem, a 404 on the marker icon, because Leaflet guesses its image folder from a URL the build hashes. Both fixed. The map stays in its box and the pin renders.

Props vanyukov

### Screenshots

Description editor "+" before:
before-modal-inserter

and after:
after-modal-inserter

Venue editor map, before:
before-modal-venue-map

and after:
after-modal-venue-map

Mostly regression testing.

1. Build the plugin: `npm run build --workspace=public_html/wp-content/mu-plugins/wporg-groups-frontend`.
2. As an editor (Organiser) on a group site, click "+ Create event". Fill every field: title, description, featured image, date, time, duration, a weekly recurrence, venue, online link, a registration question. Wait for "Draft saved at…", then Create. The event page and the notification email should show the right date and venue.
3. Open Settings, Events, "+ Create event" and repeat with a monthly recurrence and a speaker. The speaker should be saved. Clone an event from the list and save the copy. The copy opens in edit mode ("Save changes", no date minimum, recurrence not re-sent).
4. Edit an existing event from the settings list. Change every field, clear the venue, remove the image, switch online off, relabel the question, swap the speaker. Save. No notification email should go out.
5. As an author (Event Organiser), only "+ Create event" is visible, no Settings. Create an event, then edit it from the event page. The edit modal has no draft picker, no autosave label, and the recurrence controls are disabled.
6. In either venue editor, choose "+ Add a new venue", type an address and pick a suggestion. The map renders inside its box with a blue pin, and Cancel is clickable.
7. Scroll the modal with the description editor focused. The "+" inserter stays under the header.


av3nger:
Superseded by https://github.com/WordPress/wordcamp.org/pull/1935

PR #1935, opened 2026-08-25
Groups: consolidate event forms

av3nger:
While investigating and regression-testing #1909, I found that the event form was implemented separately in the group-page modal and the Events tab of Settings. The two copies had drifted in their strings, media handling, error handling and payload construction.

This PR moves the common field state, data loading and payload construction into one EventForm component. Each wrapper keeps its own workflow and navigation.

The existing differences between the two locations remain. The modal owns draft selection and autosave. Settings owns event listing, cloning and Speakers. The incomplete Settings event and draft list is tracked separately in #1977.

Props vanyukov

Mostly regression testing.

1. Build the plugin: `npm run build --workspace=public_html/wp-content/mu-plugins/wporg-groups-frontend`.
2. As an editor (Organiser) on a group site, click "+ Create event". Fill every field: title, description, featured image, date, time, duration, a weekly recurrence, venue, online link, a registration question. Wait for "Draft saved at…", then Create. The event page and the notification email should show the right date and venue.
3. Open Settings, Events, "+ Create event" and repeat with a monthly recurrence and a speaker. The speaker should be saved. Clone an event from the list and save the copy. The copy opens in edit mode with "Save changes" and no date minimum. Its recurrence remains editable until the event is published.
4. Edit an existing event from the settings list. Change every field, clear the venue, remove the image, switch online off, relabel the question, swap the speaker. Save. No notification email should go out.
5. As an author (Event Organiser), only "+ Create event" is visible, no Settings. Create an event, then edit it from the event page. The edit modal has no draft picker, no autosave label, and the recurrence controls are disabled.


public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/event-form.js
```diff
@@ -0,0 +1,380 @@
...
+ * @param {string}   props.className         Class of the `<form>` element.
+ * @param {Function} props.onSubmitPayload   Receives the payload, returns a promise; a rejection's message is shown as the save error.
+ * @param {Function} props.onCancel
+ * @param {Function} props.onOpenVenueEditor Receives the venue id (0 = new venue).
+ * @param {Function} props.onChange          Optional; called when a field value changes.
+ * @param {Function} props.onDirty           Optional; called on any edit, including the description.
+ * @param {Function} props.onLoad            Optional; called once event data has been loaded into the form.
+ * @param {*}        props.header            Rendered above the fields.
+ * @param {*}        props.footerStart       Rendered at the start of the actions row.
+ * @param {*}        props.venueEditor       When set, rendered in place of the fields; the field state is kept.
```
gedex: What keeps the description here? The `venueEditor` return at :250 unmounts DescriptionEditor, so edits die on the venue round-trip. Would hiding the form instead of replacing it work?

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/event-form.js
```diff
@@ -0,0 +1,380 @@
...
+		// Blank-labelled rows are just an empty slot the organizer
+		// added and never filled in; the server drops them too.
+		rsvp_questions: ( form.rsvp_questions || [] ).filter(
+			( q ) => q.label.trim() !== ''
+		),
+	} );
+
+	useImperativeHandle( ref, () => ( {
+		getPayload: buildPayload,
+		loadEvent: ( id ) => {
```
gedex, line 199: Why drop `loadFormData`'s canceller? Two loads in flight both write state, so a slow draft response can land last. Could it be kept in a ref?

gedex (commented):
The shared `buildPayload` sends the same keys as the old events-tab version; the modal drops only `new_venue_name`/`new_venue_address`, which VenueField never let into `venue_select` anyway. `useImperativeHandle` has no dep array, so autosave's `getPayload`/`isSaving`/`isLoading` read the current render.

Worth a look: 1) description edits lost on the settings-tab venue round-trip, and 2) `loadEvent` dropping its canceller.

Coverage: `buildPayload` holds the logic that drifted (edit-mode recurrence, blank RSVP labels, clearing `online_event_link`) and nothing tests it. The e2e covers one create path per surface. Would a jest test over an exported builder be worth it?

With the modal payload change, nothing sends `new_venue_name`/`new_venue_address` any more. They're declared in `event_args_schema()` (`inc/rest.php:643`) and inherited by `draft_args_schema()`, so `POST /wporg-groups/v1/event`, `/event/{id}`, `/draft`, `/draft/{id}` and `/draft/{id}/publish` still accept them while the reads at :903 and :1011 and the `resolve_venue_id()` branch at :1163 are unreachable. That namespace is ours and unreleased, so delete them in a follow-up?

Also: event-form.js is JSX, its siblings use `h()`, and the docblock explaining the old choice is gone. Intentional?

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/event-modal-app.js
```diff
@@ -534,36 +106,24 @@ const MINIMUM_EVENT_DATE = window.wporgGroupsEventModal?.minimumEventDate || '';
...
 			}, AUTOSAVE_INTERVAL_MS );
 			return () => clearInterval( interval );
-		// eslint-disable-next-line react-hooks/exhaustive-deps
-		}, [ isEdit, draftId, saving, loading, form, recurrence ] );
-
-		const updateField = ( field, value ) => {
-			setForm( ( prev ) => ( { ...prev, [ field ]: value } ) );
-			markDirty();
-		};
+		}, [ isEdit, draftId, changeCount ] );
```
gedex, line 109: How does autosave fire with `changeCount` in the deps? Every field keystroke bumps it and clears the interval. Would a ref for the countdown avoid that?

tests/e2e/event-form.spec.js
```diff
@@ -0,0 +1,70 @@
...
+		const modalTitle = `Event form modal E2E ${ Date.now() }`;
+		await page.goto( '' );
+		await page.getByRole( 'button', { name: '+ Create event' } ).click();
+		const modal = page.locator( '.wporg-groups-event-modal' );
+		await fillAndSubmit( modal, modalTitle );
+		await expectPublishedWithDatetimes( page, modalTitle );
+
+		const tabTitle = `Event form settings E2E ${ Date.now() }`;
+		await page.goto( '' );
+		await page.getByRole( 'button', { name: /Settings/ } ).click();
```
gedex: What made the `/Settings/` regex the pick? recurring-events-frontend.spec.js:15 uses `[data-wporg-settings-open]`.

tests/e2e/event-form.spec.js
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	test( 'publishes with the entered datetimes from both the modal and the settings tab', async ( { page } ) => {
+		test.slow(); // Two cold boots of the inline block editor.
+
+		await login( page, 'organiser2', 'password' );
+
+		const modalTitle = `Event form modal E2E ${ Date.now() }`;
+		await page.goto( '' );
+		await page.getByRole( 'button', { name: '+ Create event' } ).click();
```
gedex: Should `+ Create event` be `exact: true` here and at :65? recurring-events-frontend.spec.js:17 sets it.

tests/e2e/event-form.spec.js
```diff
@@ -0,0 +1,70 @@
...
+	 * the saved datetimes.
+	 *
+	 * @param {import('@playwright/test').Page} page
+	 * @param {string}                          title
+	 */
+	async function expectPublishedWithDatetimes( page, title ) {
+		await page.waitForURL( /\/event\//, { timeout: 30000 } );
+		await expect( page.getByRole( 'heading', { name: title } ) ).toBeVisible();
+		await expect( page.locator( 'body' ) ).toContainText( 'Wednesday, March 4' );
+		await expect( page.locator( 'body' ) ).toContainText( '12:00 PM to 1:00 PM' );
```
gedex: Which setting does this time string depend on? Only spec asserting a rendered time.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/featured-image-picker.js
```diff
@@ -0,0 +1,85 @@
...
+		{ className: `${ classPrefix }__field` },
+		h(
+			'label',
+			{ className: 'components-base-control__label' },
+			__( 'Featured image', 'wporg-groups-frontend' )
+		),
+		h(
+			'div',
+			{ className: `${ classPrefix }__featured` },
+			imageId
```
gedex: Why key the preview on `imageId`? Settings tab used `imageUrl`; an id with no URL renders ``. Would `imageId && imageUrl` cover both?

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/featured-image-picker.js
```diff
@@ -0,0 +1,85 @@
...
+				? h(
+					'div',
+					{ className: `${ classPrefix }__featured-preview` },
+					h( 'img', { src: imageUrl, alt: '' } ),
+					h(
+						'div',
+						{ className: `${ classPrefix }__featured-actions` },
+						h(
+							Button,
+							{ variant: 'secondary', onClick: openMediaFrame },
```
gedex: What happened to `isSmall` on Replace and Remove (:73)? Those buttons grow in the settings tab.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/venue-field.js
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * WordPress dependencies.
+ */
+import { createElement as h } from '@wordpress/element';
+import { Button, SelectControl } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+
+export default function VenueField( { venues, venueId, onSelect, onOpenVenueEditor, classPrefix } ) {
+	const options = [
+		{ label: __( '— Select a venue —', 'wporg-groups-frontend' ), value: '' },
```
gedex: Was the modal's label the deliberate pick? This is also the clear-venue option when editing, where the settings tab's "— No venue —" read truer.

gedex:
Continuing from a draft doesn't work for me. Steps to reproduce:

1. Create event from the sidebar

    <kbd>image</kbd>

2. Select existing draft

    <kbd>image</kbd>

3. Click create event and you'll get following error:
 
    <kbd>image</kbd>

Might not be caused by the PR, but seems relevant to PR's test instruction.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/event-modal-app.js
av3nger: It fires the same way it did before the refactor. The interval only saves when the form is dirty, and after a save it clears dirty, so in practice it's "5 seconds after you stop typing".

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/event-form.js
av3nger: Fixed and added regression coverage for the round trip

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/event-form.js
av3nger: Fixed

tests/e2e/event-form.spec.js
av3nger: Agreed. I added `exact: true` to both `+ Create event` lookups.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/featured-image-picker.js
av3nger: Fixed

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/featured-image-picker.js
av3nger: Restored

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/venue-field.js
av3nger: Fixed

tests/e2e/event-form.spec.js
av3nger: No particular reason. Switched to `[data-wporg-settings-open]`

tests/e2e/event-form.spec.js
av3nger: Changed the test to check `gatherpress_datetime_start` and `gatherpress_datetime_end` via the REST endpoint instead

av3nger:
Thanks again for the detailed review. I've worked through the comments and pushed the follow-ups.

> Also: event-form.js is JSX, its siblings use h(), and the docblock explaining the old choice is gone. Intentional?

Yes, intentional. The shared form has a large nested render tree, and JSX makes that section easier to follow. The existing build already compiles it.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/events-tab.js
```diff
@@ -9,405 +9,124 @@
...
+				classPrefix: 'wporg-event-form',
+				className: 'wporg-event-form',
+				onSubmitPayload: submitPayload,
+				onCancel,
+				onOpenVenueEditor: ( id ) => setVenueEditorId( id ),
+				header: h( 'div', { className: 'wporg-event-form__header' },
+					h( Button, { variant: 'tertiary', onClick: onCancel, icon: 'arrow-left-alt2' }, __( 'Back to events', 'wporg-groups-frontend' ) ),
+				),
+			},
+				h( 'div', { className: 'wporg-event-form__field' },
```
gedex, line 110: Is Speakers meant to stay settings-only? It goes in through `children` rather than the shared state, so the two surfaces still offer different fields. Fine as a deliberate split, just want to confirm it's intended rather than left over.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/events-tab.js
```diff
@@ -424,7 +143,7 @@ export default function EventsTab( { eventId: initialEventId, onClose } ) {
 	useEffect( () => {
 		if ( editingId !== null ) return;
 		setLoading( true );
-		apiFetch( { path: '/wp/v2/gatherpress_events?per_page=100&_fields=id,title,meta,status&orderby=date&order=desc' } )
+		apiFetch( { path: '/wp/v2/gatherpress_events?per_page=100&status=publish,draft&_fields=id,title,meta,status&orderby=date&order=desc' } )
```
gedex: Why a different source from the modal's picker? That one calls `/{NS}/drafts`, which returns every draft you can edit, capped at 50 and ordered by modified. This pulls publish and draft together at `per_page=100` ordered by post date, then filters client-side, so drafts fall off once the combined count passes 100. On the same group I get 4 drafts in the modal and 2 here, with the two oldest missing. Both paths check `edit_post`, so it isn't caps.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/build-payload.js
```diff
@@ -0,0 +1,36 @@
...
+		title: form.title,
+		description,
+		date: form.date,
+		time_start: form.time_start,
+		time_end: form.time_end,
+		venue_id: parseInt( form.venue_select, 10 ) || 0,
+		is_online: form.is_online,
+		online_event_link: form.is_online ? form.online_event_link : '',
+		featured_image_id: featuredImageId,
+		// Recurrence is locked once an event is published; drafts stay editable.
```
gedex: Does this comment still hold? The gate is `locked` now, not edit mode, so editing a draft does resend recurrence.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
```diff
@@ -92,6 +92,13 @@ public function test_yearly_leap_day_skips_non_leap_years(): void {
 		);
 	}
 
+	/** A post with no recurrence meta reads back a rule that passes the REST enum. */
+	public function test_rule_from_bare_post_uses_valid_weekday(): void {
```
gedex, line 96: Worth asserting a stored value too? This covers the empty case, but `strtoupper` is the part that changed behaviour, so `'we'` to `'WE'` is the interesting one.

gedex (commented):
On my test the modal picker shows 4 drafts where the settings Drafts section shows 2. They read from different endpoints with different limits and ordering, so the settings one can drop drafts as the event count grows. Details inline.

I also noticed the two surfaces still offer different fields: no Speakers when you create from the group page, Speakers when you create from Settings. Predates this PR and sits outside the shared state by design, so asking inline rather than blocking.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/event-modal-app.js
gedex: Confirmed, production had `form` in the same deps, so the interval was already restarting on every keystroke.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/events-tab.js
gedex: | modal draft picker (4 drafts) | settings drafts section (2 drafts) |
| -- | -- |
| <kbd>image</kbd> | <kbd>image</kbd> |

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/events-tab.js
gedex: | modal picker (no speakers) | settings draft section (has speakers) |
| -- | -- |
| <kbd>image</kbd> | <kbd>image</kbd> |

av3nger: Yes, for this PR. I kept the existing behavior rather than changing which fields each entry point exposes. The shared form owns the common fields, while Settings continues to add Speakers through children.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/components/events-tab.js
av3nger: I have removed that query change from this PR and added the endpoint, ordering, and limit mismatch to #1977.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/build-payload.js
av3nger: Updated it to match the current behavior.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
av3nger: Added a stored lowercase case that checks we is normalized to WE.

av3nger:
@gedex , thanks for the followup review. I have removed the partial Settings draft query from this PR and added the full endpoint, ordering, and limit mismatch to #1977. I also corrected the recurrence comment, added coverage for normalizing a stored lowercase weekday.

gedex (approved):
LGTM. Good call splitting the drafts query into #1977.

Copilot's point still stands though: edit mode and the settings-only Speakers field have no e2e coverage. Not blocking, happy to file it separately.

PR #1938, opened 2026-08-25
Groups: keep the block inserter below the event modal header

### LINKED ISSUE #1937: Groups: description block inserter overlaps the event modal header
### Describe the bug

The Description editor's `+` block inserter can render over the event modal header when the modal is scrolled.

### To reproduce

1. Log in as an Organiser on a group site.
2. Click **+ Create event**.
3. Focus the Description editor and scroll the modal.
4. See the block inserter overlap the modal header.

### Expected behavior

The modal header should remain above the form content, and the block inserter should stay within the scrollable area.

### Screenshots / Screencasts

before-modal-inserter

### WordCamp

Not specific to one site. Reproduced on a local group site.

 - Device: MacBook
 - OS: latest MacOS
 - Browser: Safari/Chrome
 - Version: latest

av3nger:
Updates the event modal header's background and stacking order so Description editor controls no longer render above it while scrolling.

Fixes #1937

Props vanyukov

### Screenshots

Before:
before-modal-inserter

After:
after-modal-inserter

1. Build the Groups frontend:
   `npm run build --workspace=public_html/wp-content/mu-plugins/wporg-groups-frontend`
2. As an Organiser, open **+ Create event**, focus the Description editor, and scroll the modal.
3. Confirm the `+` inserter stays below the header and the header controls remain visible.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/component-accents.css
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/component-accents.css
@@ -21,8 +21,11 @@
 	--wp-components-color-accent-inverted: var(--wp--preset--color--white, #fff);
 }
 
-/*
- * Fix form elements overlapping form header.
+/**
+ * The block editor's inserter popover (`.block-editor-block-popover`,
+ * position: absolute, z-index: 31) draws over the modal header when the
+ * content scrolls, because the upstream header is z-index: 10 with no
+ * background. Opaque and one above the popover keeps the header on top.
  */
 .wporg-groups-modal-accent .components-modal__header {
 	background: var(--wp--preset--color--white, #fff);
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/component-accents.css
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/component-accents.css
@@ -20,3 +20,11 @@
 	--wp-components-color-accent-darker-20: var(--wp--preset--color--dark-blueberry, #1d35b4);
 	--wp-components-color-accent-inverted: var(--wp--preset--color--white, #fff);
 }
+
+/*
+ * Fix form elements overlapping form header.
+ */
+.wporg-groups-modal-accent .components-modal__header {
+	background: var(--wp--preset--color--white, #fff);
+	z-index: 32;
+}
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/style.scss
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/group-settings/style.scss
@@ -21,7 +21,8 @@
 // === Full-screen modal ===
 
 .wporg-group-settings-modal {
-	// Accent tokens come from `components/component-accents.css`.
+	// Accent tokens and the header stacking fix come from
+	// `components/component-accents.css`.
 
 	.components-modal__content {
 		padding: 0;


public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/component-accents.css
```diff
@@ -20,3 +20,11 @@
 	--wp-components-color-accent-darker-20: var(--wp--preset--color--dark-blueberry, #1d35b4);
 	--wp-components-color-accent-inverted: var(--wp--preset--color--white, #fff);
 }
+
+/*
+ * Fix form elements overlapping form header.
+ */
```
gedex: Which elements does "form elements" mean here, and which header is "form header"? Worth a rewording that also describes the state rather than the edit, since "Fix..." won't read well once this is just how the file looks. Something shaped like:

```css
/*
 * `<element>` renders above the modal header while the content scrolls, because
 * the header ships transparent and low in the stacking order. This keeps the
 * title and close button usable.
 */
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/component-accents.css
```diff
@@ -20,3 +20,11 @@
...
+.wporg-groups-modal-accent .components-modal__header {
```
gedex, line 30: Is this file the right home for a stacking fix? Its header comment covers only the four accent tokens, and `group-settings/style.scss:25` points back at it with `// Accent tokens come from components/component-accents.css.`, so both read as stale now.

The shared location makes sense, since this is the only selector that reaches all four modals. Would a separate `src/components/modal-header.css` imported from both entry points work for you? Otherwise, happy with just widening this file's header comment to mention shared modal chrome.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/component-accents.css
```diff
@@ -20,3 +20,11 @@
...
+	background: var(--wp--preset--color--white, #fff);
```
gedex, line 31: Did you get a chance to test the `+` on the very first block? `.components-modal__content` has `margin-top: 72px`, so the header band sits above the scroll viewport, and the popovers are position-fixed. Raising the header to 32 with an opaque background means an inserter landing in that top band is now hidden behind it rather than drawn over it. Might be a non-issue, I just couldn't check without a browser.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/component-accents.css
```diff
@@ -20,3 +20,11 @@
...
+	z-index: 32;
```
gedex, line 32: Where does `32` come from? Can we document that somewhere, so the next person touching this doesn't have to reverse-engineer it?

gedex (commented):
The shared selector lands on all four modals carrying `wporg-groups-modal-accent`, though only the event modal has a block editor in it. Is that broader reach deliberate, or does only the event modal need it?

Otherwise the mechanism checks out. Upstream `.components-modal__header` is `position: absolute; z-index: 10` with no background, so content scrolls through it, and both declarations are needed. `lint:css` and `build` pass, the rule reaches both the `event-manage` and `group-settings` bundles, and `build/` is gitignored, so source-only is right.

Left a few questions inline.

gedex:
Tested on wide and narrow width.

image

gedex (approved):

av3nger:
@akeda, it's actually two forms, not one - the settings and "create event" modals. The class is the one selector that reaches both modals. I've also adjusted the comments to explain the reasoning.

gedex:
Current changes look good — thanks for addressing all the feedback!

PR #1940, opened 2026-08-25
Groups: load Leaflet styles and marker assets on the front end

### LINKED ISSUE #1939: Groups: venue map is unstyled in frontend event forms
### Describe the bug

The venue map in the frontend event forms renders without Leaflet's layout styles. Map tiles overflow their container, and the location marker is missing.

### To reproduce

1. Log in as an Organiser on a group site.
2. Open **+ Create event** or go to **Settings → Events → + Create event**.
3. Choose **+ Add a new venue**, enter an address, and select a suggestion.
4. See the unstyled map and missing marker.

### Expected behavior

The map should remain inside its container and display a location marker for the selected address.

### Screenshots / Screencasts

before-modal-venue-map

### WordCamp

Not specific to one site. Reproduced on a local group site.

 - Device: MacBook
 - OS: MacOS 26.6.2
 - Browser: Safari
 - Version: 26.6.2

av3nger:
Loads the Leaflet stylesheet from the frontend entry points used by both event-form surfaces and explicitly configures the bundled marker images.

This keeps the map inside its container and prevents the marker assets from returning 404 responses.

Fixes #1939

Props vanyukov

### Screenshots

Before:
before-modal-venue-map

After:
after-modal-venue-map

1. Build the Groups frontend:
   `npm run build --workspace=public_html/wp-content/mu-plugins/wporg-groups-frontend`
2. In both **+ Create event** and **Settings → Events**, add a new venue and select an address suggestion.
3. Confirm the map stays inside its container, the marker renders, and no marker-image requests return 404 responses.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/venue-editor.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/venue-editor.js
@@ -24,11 +24,43 @@ import {
 } from '@wordpress/components';
 import apiFetch from '@wordpress/api-fetch';
 import { __, _x } from '@wordpress/i18n';
+import 'leaflet/dist/leaflet.css';
 import markerIcon from 'leaflet/dist/images/marker-icon.png';
 import markerIcon2x from 'leaflet/dist/images/marker-icon-2x.png';
 import markerShadow from 'leaflet/dist/images/marker-shadow.png';
 
 const PHOTON_API = 'https://photon.komoot.io/api/';
+
+let leafletPromise = null;
+
+/**
+ * Loads Leaflet once and points its default icon at the bundled images.
+ * Leaflet guesses its icon folder from the stylesheet's background-image
+ * URL, which the build has hashed.
+ */
+function loadLeaflet() {
+	if ( ! leafletPromise ) {
+		leafletPromise = import( 'leaflet' )
+			.then( ( module ) => {
+				const L = module.default || module;
+				delete L.Icon.Default.prototype._getIconUrl;
+
+				L.Icon.Default.mergeOptions( {
+					iconUrl: markerIcon,
+					iconRetinaUrl: markerIcon2x,
+					shadowUrl: markerShadow,
+				} );
+
+				return L;
+			} )
+			.catch( ( error ) => {
+				leafletPromise = null;
+				throw error;
+			} );
+	}
+
+	return leafletPromise;
+}
 
 export default function VenueEditor( { venueId, onSave, onCancel, inline, hideHeader } ) {
 	const [ loading, setLoading ] = useState( !! venueId );
@@ -132,21 +164,10 @@ export default function VenueEditor( { venueId, onSave, onCancel, inline, hideHe
 				return;
 			}
 
-			import( 'leaflet' ).then( ( module ) => {
+			loadLeaflet().then( ( L ) => {
 				if ( ! mapRef.current ) {
 					return;
 				}
-
-				const L = module.default || module;
-
-				// Leaflet guesses its icon folder from the stylesheet's
-				// background-image URL, which the build has hashed.
-				delete L.Icon.Default.prototype._getIconUrl;
-				L.Icon.Default.mergeOptions( {
-					iconUrl: markerIcon,
-					iconRetinaUrl: markerIcon2x,
-					shadowUrl: markerShadow,
-				} );
 
 				if ( mapInstanceRef.current ) {
 					mapInstanceRef.current.setView( [ lat, lng ], 15 );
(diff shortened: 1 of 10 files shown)


public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/venue-editor.js
```diff
@@ -24,6 +24,9 @@ import {
 } from '@wordpress/components';
 import apiFetch from '@wordpress/api-fetch';
 import { __, _x } from '@wordpress/i18n';
+import markerIcon from 'leaflet/dist/images/marker-icon.png';
```
gedex, line 28: Was jest coverage a consideration? The wp preset maps only `\.(scss|css)$`, so any suite importing venue-editor.js (or event-modal-app / venues-tab) now dies with "Jest encountered an unexpected token". Nothing currently does, so not a blocker. Worth mapping images in `jest.config.js` while you're here?

```js
'\\.(png|jpe?g|gif|webp|svg)$': require.resolve(
	'@wordpress/jest-preset-default/scripts/style-mock.js'
),
```

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/view.js
```diff
@@ -8,4 +8,5 @@
  * @package WordCamp\Groups\Frontend
  */
 
+import 'leaflet/dist/leaflet.css';
```
gedex: Any reason to import leaflet.css here rather than in venue-editor.js, next to the marker images, and drop it from both view.js files? venue-editor is a static import of both entries, so the build produces the same view.css either way. If a third block imports VenueEditor later, it would then get the styles for free instead of regressing to #1939.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/venue-editor.js
```diff
@@ -136,6 +139,15 @@ export default function VenueEditor( { venueId, onSave, onCancel, inline, hideHe
 
 				const L = module.default || module;
 
+				// Leaflet guesses its icon folder from the stylesheet's
+				// background-image URL, which the build has hashed.
+				delete L.Icon.Default.prototype._getIconUrl;
+				L.Icon.Default.mergeOptions( {
```
gedex: Meant to re-run per geocode? Deps are `[latitude, longitude]`, so this fires on every address selection. Idempotent, so nothing breaks. Worth a module-level `configured` guard, or is inline with the import the point?

gedex (commented):
Fix looks right. Good diagnosis: `detectIconPath()` strips a trailing `marker-icon.png`, which the hashed filename defeats, so `delete _getIconUrl` plus `mergeOptions` is the right workaround.

Verified locally:

- Build emits `view.css` for both blocks (10,375 B, leaflet only) plus `-rtl`, so `viewStyle: file:./view.css` resolves.
- `marker-icon-2x` and `marker-shadow` land in `build/blocks/images/`. Auto publicPath resolves to `build/blocks/` (`s.p = e + "../"`) and css-loader writes `url(../images/...)`, so both URL forms are correct.
- `WP_Block::render()` enqueues `view_style_handles` right next to `view_script_handles`, so `view.css` loads under the same conditions as the working `view.js`.
- Dropping leaflet.css from `index.js` is safe, and it explains the bug. Both blocks are `edit: () => null`, and `index.css` isn't in `block.json`, so leaflet was never enqueued anywhere.
- 28 jest tests pass, no new eslint errors.

Two optional questions inline. One tradeoff worth naming: 10 KB of leaflet CSS now loads on every page rendering either block, though the map sits behind a modal and the leaflet JS stays code-split. Unavoidable while venue-editor is a static import. Any lazy option ruled out?

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/venue-editor.js
av3nger: Added

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/view.js
av3nger: Agreed and moved into venue-editor.js, removed from both view.js.

av3nger:
@gedex , thanks for the thorough review. On the build - you are correct on the size. However, it's not just "every page rendering either block". The potential savings are ~5 kb for organiser only. I don't think it's worth it today. But if the map ever moves to a public place (the GatherPress map is different), we can always revisit.

public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-manage/venue-editor.js
av3nger: Moved to a small helper that loads Leaflet once and sets up the marker at the same time.

gedex:
Tested on desktop and mobile. Works as expected. Thanks for the fix!

gedex (approved):

PR #1941, opened 2026-08-25
WordCamp application: attribute the application to the submitting user

KokkieH:
## Summary

`WordCamp_Application::create_post()` set the application's `post_author` from the submitted **WordPress.org Username** field (resolved via `wcorg_get_user_by_canonical_names()`). The application form is only reachable by a logged-in user — `Event_Application::submit_application()` requires `is_user_logged_in()` — so the account that submits the application is the natural owner.

This change attributes the new WordCamp application to the current (submitting) user and keeps the WordPress.org username field as application data for the Community Team to review, rather than using it to set the owner. The initial status-change entry is recorded against the same user. This matches how the Meetup converter attributes its applications and makes ownership predictable.

## Changes

- `post_author` is the current user (falling back to the `wordcamp` account only if there is somehow no current user).
- The WordPress.org username stays as the `WordPress.org Username` meta.
- `_status_change['user_id']` is the current user.

## Testing

1. Check out this branch on a local wordcamp.org multisite (Central site).
2. Log in as the admin account.
3. Open the organizer application form at `https://central.wordcamp.test/wordcamp-organizer-application/` and submit it. In the **WordPress.org Username** field, enter a *different* account's username than the one you're logged in as, e.g. `contributorrole`.
4. Find the application created on Central and confirm the author is the account you submitted with — not the one you typed in the username field:
   ```
   wp --url=https://central.wordcamp.test post list --post_type=wordcamp \
     --fields=ID,post_title,post_author --orderby=ID --order=desc --posts_per_page=1
   ```
   `post_author` should be your logged-in user's ID.
5. Confirm the typed username is still retained as data on the application:
   ```
   wp --url=https://central.wordcamp.test post meta get <application_id> "WordPress.org Username"
   ```
   It should echo back what you entered in step 3.
6. Confirm the initial status entry is attributed to the submitting user:
   ```
   wp --url=https://central.wordcamp.test post meta get <application_id> _status_change --format=json
   ```
   `user_id` should be your logged-in user's ID.

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-events/class-events-application.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-events/class-events-application.php
@@ -150,15 +150,16 @@ class Events_Application extends WordCamp_Application {
 	public function create_post( $data ) {
 		$wordcamp_user_id = get_user_by( 'email', 'support@wordcamp.org' )->ID;
 
-		// Create the post.
-		$user     = wcorg_get_user_by_canonical_names( $data['q_wporg_username'] );
-		$statuses = \WordCamp_Loader::get_post_statuses();
+		// `submit_application()` only reaches this method for a logged-in submitter, so the
+		// application is owned by that account. The username field is stored as meta below.
+		$author_id = get_current_user_id();
+		$statuses  = \WordCamp_Loader::get_post_statuses();
 
 		$post = array(
 			'post_type'   => self::get_event_type(),
 			'post_title'  => esc_html( $data['q_event_location'] ),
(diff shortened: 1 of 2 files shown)


mcliwanow (approved):
Code looks good and I verified it locally

One more thing here might be worth addressing: `wcpt-events/class-events-application.php:161` has the same pattern (author resolved from the typed username), and it goes through the same login-gated `submit_application()`. I'd apply the same change there, in this PR or a follow-up.

nit: I'd cut the new comment down to ~2 lines. Long comments add mental burden.

PR #1942, opened 2026-08-25
WordCamps: decide cancelled camp visibility per camp, not per status

mcliwanow:
## Why

`wcpt-cancelled` is registered `public`, so a cancelled camp resolves for anyone, over REST and at its Central permalink. The flag is there for one consumer: Official WordPress Events reads cancelled camps so it can drop them from the events widget, and that only concerns camps that reached the schedule.

Most never did. `menu_order` holds the date a camp was added to the official schedule (`WordCamp_Loader::set_scheduled_date()`), and it is 0 for 369 of the 424 cancelled camps on Central. Those were cancelled while still applications, so Central never listed them: its public surfaces only ever covered `wcpt-scheduled` and `wcpt-closed`. #1931 moved every other application status to `protected`, and cancelled was held back because the `public` flag was carrying this REST behaviour.

| | count |
|---|---|
| cancelled camps | 424 |
| ever added to the schedule | 55 |
| cancelled while still an application | 369 |

## What changed

`public` is per status and this rule is per post, so the registration stays as it was and the query decides. `WordCamp_Loader::hide_unscheduled_cancellations()` filters `posts_where`, which keeps a collection's items, its `X-WP-Total` and its page count describing the same set. Filtering per row instead returns short and empty pages in the middle of a paginated set, and a client that stops on a short page truncates silently.

Single-item REST reads fetch the post rather than running a query, so they cannot inherit that filter. `WordCamp_Loader::can_read_unscheduled_cancellation()` holds the rule, `check_read_permission()` calls it, and the SQL clause mirrors it. The permalink and the API have to describe the same set, so writing the rule twice is how they drift.

Who keeps a camp: its author, and whoever is named as its mentor. The author rather than `edit_post`, because `add_organizer_to_central()` runs on `wcpt_approved_for_pre_planning`, so an application cancelled before it was accepted never added its author to Central and that author usually holds no role there. The mentor name resolves through `wcorg_get_user_by_canonical_names()`, which is how `map_subrole_caps()` decides who holds `edit_post` on a camp and how the list table in #1943 picks what it shows. A name that is one account's login and another's nicename belongs to the login holder, and only they hold anything over the camp. SQL cannot resolve a name, so `mentor_names_for()` states which stored names resolve back to this user and the clause selects those, rather than the rule widening to match a clause that matched both names.

Wranglers, cron and WP-CLI see everything, so `wp post list` still reports the real set, and a personal data export sees the subject's own record whoever is processing it.

wp-admin is exempt too, on `can_read_unscheduled_cancellation()` as well as on the clause, because every exemption one takes the other has to take: where the clause is the more permissive of the two it selects a row the rule then refuses, and the response comes back short under a total that was never reduced. The exemption names the capability rather than the entry point. Opening a WordCamp screen needs `edit_wordcamps`, while `is_admin()` on its own is true on `admin-post.php`, which defines `WP_ADMIN` and fires `admin_post_nopriv_{$action}` before any capability check. Nothing registers such a handler today, so this is a shape rather than a read, but the capability admits whoever can open the list table through any of those doors and nobody else.

`was_ever_scheduled()` keeps the `menu_order` convention next to `set_scheduled_date()`, which writes it, and the SQL says `menu_order <= 0` so the two agree on values nothing writes yet.

## What the rule is, and what it costs

`menu_order > 0` means the camp reached the official schedule, which is not the same as "was public". `URL` meta is set at `wcpt-needs-site`, before scheduling, so 152 of the 369 got their own camp site and were then cancelled without ever being scheduled. Those 152 stay hidden: their page on Central stops resolving for logged out visitors and REST returns 401. Their own camp sites are untouched, and `get_best_wordcamp_url()` already sends people there. Central never listed them either way, so this is the intended reading, and it rules out using a site URL as the signal since that would bring all 152 back.

The other cost is age. `menu_order` is only written from 2016-06-23, so a camp cancelled before that reads as never scheduled whatever actually happened. That is 34 records, 21 of which had a site and were probably public. Left alone here: a read path should not learn a date. A one-time backfill of `menu_order` on those 21 would recover them without teaching it one, which is a better answer than either and belongs in its own change.

`sanitize_post_statuses()` still lets anyone query `?status=wcpt-cancelled`. Which of them come back is what changed.

`check_read_permission()` keeps testing the public statuses plus `wcpt-cancelled`, rather than `get_publicly_viewable_post_statuses()`, which also carries the six pre-planning statuses. Those resolve at their permalink and stay refused over REST, as on `production`. Aligning the two would publish the meta `register_rest_public_fields()` exposes, `Organizer Name`, `WordPress.org Username`, `Number of Anticipated Attendees` and the venue fields among it, which `single-wordcamp.php` does not render. That is worth doing on its own terms, with the `@todo` above `sanitize_post_statuses()`, not as a side effect of this. It leaves that `@todo`'s own behaviour in place: anyone may still query `?status=wcpt-pre-planning`, and the collection answers 200 with an empty page under a total counting the rows it then refuses. Measured identically on `production`, so this restores it rather than introducing it, and it belongs with the same decision.

## Testing

27 tests in `tests/wcpt-wordcamp/test-cancelled-camp-visibility.php`. The one that matters is `test_the_clause_and_the_rule_admit_the_same_set`: the clause decides in SQL and the rule decides in PHP, about the same camps, and every defect found in review here has been one input they answered differently, each on a dimension nobody had enumerated yet. It compares them across ten viewers, eight camps and three request contexts, and mutating each of those defects back in fails it naming the viewer, the camp and the context. The rest cover both kinds of cancelled camp on both surfaces, an author with no capabilities on Central, mentors by login, by nicename, in a second meta row and by a name another account holds as a login, a wrangler, an unattributed camp, a negative `menu_order`, an administrator without the subrole in wp-admin, the exemption stopping at admin-ajax, an export query, the pre-planning statuses staying refused, and the collection agreeing with its count in each case. Green in CI on 8.3, 8.4 and 8.5, `phpcs` included. Locally the suite also fails `Test_QuickBooks_OAuth_State` and the `Test_Map_Meta_Cap` and `Test_Groups_Blocks` cases, all of which fail on `production` too and pass in CI.

Measured on a local Central, anonymous:

| | production | branch |
|---|---|---|
| cancelled after scheduling, REST / permalink | 200 / resolves | 200 / resolves |
| cancelled as application, REST / permalink | 200 / resolves | 401 / does not resolve |
| scheduled, REST / permalink | 200 / resolves | 200 / resolves |

Driven over HTTP on the branch against an application-stage cancellation, REST with a nonce from `admin-ajax.php?action=rest-nonce`, permalink as a plain GET. Every row agrees across the two surfaces, which is the point of the change:

| viewer | its author's | one they mentor | someone else's |
|---|---|---|---|
| logged out | 401 / empty | 401 / empty | 401 / empty |
| the author, no role on Central | 200 / resolves | 403 / empty | 403 / empty |
| the mentor, no role on Central | 403 / empty | 200 / resolves | 403 / empty |
| a wrangler | 200 / resolves | 200 / resolves | 200 / resolves |

In the list table, where the public surfaces do not apply: an administrator without the subrole sees an application-stage cancellation, as they do every other application status, and a mentor whose name is stored as `MentorJane` reads their camp on both public surfaces and its collection comes back with `items` and `X-WP-Total` both 1, where it was 0 and 1.

A mentor asking for `?status=wcpt-cancelled&per_page=2` gets 2 items, `X-WP-Total` 2 and 1 page, containing the camp they mentor and the scheduled-then-cancelled one and nothing else. `/wordcamps/`, `/schedule/` and the feed list none of the hidden camps anonymously.

With 3 readable of 6 cancelled camps, `per_page=2` returns 2 items on page 1 and 1 on page 2, total 3, pages 2. `/wordcamps/`, its feed, `/schedule/`, `/about/`, `calendar.ics` and site search are unchanged and list none of the hidden camps. A wrangler sees all 6 in the collection and in the admin list; the applicant sees the 4 they authored.

## How to review this locally

Create two cancelled camps, one with a scheduled date and one without, both authored by a contributor:

```
wp post create --post_type=wordcamp --post_status=wcpt-cancelled --post_name=t-sched --menu_order=1560293422 --post_author=<id>
wp post create --post_type=wordcamp --post_status=wcpt-cancelled --post_name=t-app --post_author=<id>
```

Anonymously the first answers 200 at `/wordcamps/t-sched/` and over REST; the second does not resolve and returns 401. (Locally an unresolvable camp permalink redirects to the Central home page rather than 404ing, which is existing behaviour and not something this changes.) Logged in as its author, or as a Wrangler, the second comes back. `wp post list --post_type=wordcamp --post_status=wcpt-cancelled` still counts every one of them.

```
docker compose -f docker-compose.phpunit.yml run --rm phpunit_wp phpunit -c phpunit.xml.dist \
  --filter Test_Cancelled_Camp_Visibility
```


bor0:
Reviewed the branch, ran the suite, and probed the before/after behaviour in the phpunit container.

The core change is sound and does what the description claims. `Test_Cancelled_Camp_Visibility` + `Test_Event_Status_Visibility` come to 22 tests, 155 assertions, green on the branch. Two things in the resulting behaviour deserve discussion before merge.

### Verified before/after (anonymous, same fixtures)

| | production | branch |
|---|---|---|
| cancelled-after-scheduling — REST | 200 | 200 |
| cancelled-after-scheduling — **front-end permalink** | **resolves** | **404** |
| cancelled-as-application — REST | 200 | 401 |
| cancelled-as-application — front-end permalink | resolves | 404 |
| scheduled — REST / front end | 200 / resolves | 200 / resolves |

### 1. The 55 lose their Central page, not just the 369

This is the one I'd push on. The premise here is that a camp which reached the official schedule and was then cancelled *is* a public record. That premise is honoured over REST via `check_read_permission()`, but `public` / `protected` is a per-**status** flag, so dropping `wcpt-cancelled` from `get_publicly_viewable_post_statuses()` 404s the Central permalink for all 424 — including the 55 the description argues are legitimately public. Same for Central site search: with `public => false` they fall out of the default `WP_Query` status set that non-singular queries build.

`single-wordcamp.php` renders date, venue, address and website link — a real public page for a camp that was announced and then cancelled, and one that external sites and search results may link to. The compatibility check in the description only covers *internal* renderers (`get_map_markers()`, the schedule and past-camps templates), which is why this doesn't surface there.

Either outcome is defensible, but the asymmetry (REST 200, permalink 404 for the same post) should be a stated decision — it isn't in "Deliberately unchanged" right now. If the 55 should keep their page, that needs a `pre_get_posts` / `posts_results` filter rather than the status flag, which is worth deciding now rather than later.

### 2. Collection pagination returns short and empty pages

The description flags the `X-WP-Total` mismatch but not this consequence. With 10 cancelled camps of which 1 reached the schedule:

```
branch:      per_page=5 -> page 1: 1 item,  page 2: 0 items, X-WP-Total 10, X-WP-TotalPages 2
production:  per_page=5 -> page 1: 5 items, page 2: 5 items, X-WP-Total 10, X-WP-TotalPages 2
```

A fully empty page in the middle of a paginated set is a different failure mode than an inflated total. A client that stops on a short page — a common pattern — silently truncates. Since the known consumer (Official WordPress Events) lives in meta and isn't checkable from this repo, I'd either confirm how it paginates or filter the query itself: a `posts_where` adding `AND menu_order > 0` when `status` is exactly `wcpt-cancelled` and the caller can't `edit_posts` keeps the total, the page count and the per-item filter all agreeing.

### 3. Smaller things

- **Stale comment.** `sanitize_post_statuses()` still carries "Camps that are scheduled and then cancelled should still be available (though not included by default)" above `$public_statuses[] = 'wcpt-cancelled';`. That is now decided in `check_read_permission()`; the comment should point there, or it reads as the live rule.
- **Inconsistent strictness.** `check_read_permission()` gained a strict `in_array()`; the sibling in `sanitize_post_statuses()` (line 66) stayed loose. Not a bug, but this PR touches exactly that and leaves the other behind.
- **Parent check is skipped for the readable branch.** `is_cancelled_after_scheduling()` returns `true` without going through `WP_REST_Posts_Controller::check_read_permission()`, so `check_is_post_type_allowed()` no longer runs for those posts. Inert today, but it is the first path in this controller that returns `true` with no core check behind it — worth a line in the docblock.
- **Reader is far from the writer.** `is_cancelled_after_scheduling()` is the only place that interprets `menu_order` as "was scheduled", and it lives in the REST controller while `WordCamp_Loader::set_scheduled_date()` writes it. A `WordCamp_Loader::was_ever_scheduled( $post )` would keep the convention in one class and make it reusable — relevant if the permalink in point 1 ends up needing it too.
- **Test gaps.** The refactor rewrote the public-status branch, but nothing asserts that a `wcpt-scheduled` camp still returns 200 over REST. And the front-end behaviour for a scheduled-then-cancelled camp is untested — that is the case in point 1, so it should be pinned whichever way it is decided.

### Drive-by, unrelated to this PR

`WordCamp_Admin::notify_new_wordcamp_in_slack()` (`wcpt-wordcamp/wordcamp-admin.php:1059`) throws a `TypeError` on PHP 8.4 when a camp transitions to `wcpt-scheduled` with an empty `Start Date (YYYY-mm-dd)` meta — `gmdate( 'F j, Y', '' )`. It made writing a scheduled-camp fixture awkward here, and likely will for anyone else adding tests in this area. Probably deserves its own issue.

mcliwanow:
> The 55 lose their Central page, not just the 369

😬  Looking into that and into the pagination.

> WordCamp_Admin::notify_new_wordcamp_in_slack() (wcpt-wordcamp/wordcamp-admin.php:1059) throws a TypeError on PHP 8.4 when a camp transitions to wcpt-scheduled with an empty Start Date (YYYY-mm-dd) meta — gmdate( 'F j, Y', '' )

I'm working on adjacent area in another PR. WIll fix it there.

mcliwanow:
@bor0, I'd appreciate another round of review as the approach changed a bit here. Especially due to addressing feedback on point 1.

The status registration goes back to what it was and the query decides instead. `WordCamp_Loader::hide_unscheduled_cancellations()` filters `posts_where`, which fixes point 2 at the source: items, `X-WP-Total` and page count agree now. With 3 readable of 6 cancelled camps, `per_page=2` gives 2 then 1, total 3, pages 2. Anonymously the permalink and REST agree on both kinds, and a scheduled camp is unchanged. The applicant keeps their own record, and wranglers, cron and WP-CLI see everything, so `wp post list` still reports all 424 rather than 55. (That one matters because it's how we measured the split.)

Smaller things all in: `was_ever_scheduled()` moved to `WordCamp_Loader` next to `set_scheduled_date()`, the stale comment points at the new decision point, its `in_array()` is strict, `check_read_permission()` says why it repeats the rule, and the test gaps are covered.

Your review also surfaced a live one: `get_current_user_id()` is 0 when logged out and an unattributed post has `post_author` 0, so the clause keeping an applicant their own record also read as "or nobody owns it". Central has exactly one such camp (3037886, cancelled, 2014) and without the fix it would have stayed in the anonymous REST collection. No permalink exposure, it has no slug.

Two calls I made rather than changed, both in the description now.

-  `menu_order > 0` means the camp reached the official schedule, which is not the same as "was public". `URL` meta is set at `wcpt-needs-site`, so 152 of the 369 got a camp site and were then cancelled without ever being scheduled. Those 152 stop resolving on Central for logged out visitors and REST returns 401. Their own camp sites are untouched. That looks right to me since Central never listed them anyway, and it rules out using a site URL as the signal.
- The pre-2016 blind spot is also bigger than I said: 34 records, 21 of them with sites. Recovering those means hardcoding 2016-06-23 into a read path, so I left it alone.

Tell me if you read either of those differently.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/class-wp-rest-wordcamps-controller.php
```diff
@@ -94,13 +93,14 @@ public function sanitize_post_statuses( $statuses, $request, $parameter ) {
...
 	 */
 	public function check_read_permission( $post ) {
-		$public_statuses = WordCamp_Loader::get_public_post_statuses();
-
-		// Camps that are scheduled and then cancelled should still be available (though not included by default).
-		$public_statuses[] = 'wcpt-cancelled';
+		// A single-item read fetches the post rather than running a query, so it does not
+		// inherit `WordCamp_Loader::hide_unscheduled_cancellations()`.
+		if ( 'wcpt-cancelled' === $post->post_status && ! WordCamp_Loader::was_ever_scheduled( $post ) ) {
+			return current_user_can( 'edit_post', $post->ID );
```
bor0: The description says the applicant keeps their own record, and `hide_unscheduled_cancellations()` implements that as `post_author = current_user_id`. This line implements the same rule as `current_user_can( 'edit_post' )`, and the two are not the same set.

`edit_post` maps to `edit_wordcamps`, which `register_post_capabilities()` grants to contributor and up. But `WordCamp_Admin::add_organizer_to_central()` (`wcpt-wordcamp/wordcamp-admin.php:1000`) only adds the lead organizer to Central as a contributor *when the application is accepted*. A camp cancelled while it was still an application never reached that point, so its author typically holds no role on Central at all — which is exactly the population this PR is about.

Measured, a cancelled application authored by a subscriber:

| | production | branch |
|---|---|---|
| permalink | resolves | resolves |
| REST single | 200 | 403 |

So the applicant keeps the permalink and loses the REST read, and the two surfaces disagree in the one case the change is meant to serve. `test_the_author_can_read_their_own_cancelled_application()` passes because it creates the author as a contributor; making that user a subscriber reproduces it.

Since the argument for `posts_where` is that the permalink and the API have to describe the same set, it seems worth expressing the rule once — say a `WordCamp_Loader::can_read_unscheduled_cancellation( $post )` that both this method and the SQL clause are built from — rather than author-match on one side and a capability on the other.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/class-wp-rest-wordcamps-controller.php
```diff
@@ -94,13 +93,14 @@ public function sanitize_post_statuses( $statuses, $request, $parameter ) {
...
+		}
 
 		// If post status is not listed as public, it cannot be read.
-		if ( ! in_array( $post->post_status, $public_statuses ) ) {
+		if ( ! in_array( $post->post_status, WordCamp_Loader::get_publicly_viewable_post_statuses(), true ) ) {
```
bor0: This swaps `get_public_post_statuses() + wcpt-cancelled` for `get_publicly_viewable_post_statuses()`, which also brings in all six pre-planning statuses. Anonymous single-item REST reads change accordingly — measured 401 on `production`, 200 on the branch, for `wcpt-pre-planning`, `wcpt-needs-contract` and `wcpt-needs-budget-re`.

I dumped the payload to see what that covers: title, content, excerpt, link, author, no meta, since nothing is registered for REST here. That is the same material the permalink already renders for those statuses after #1931, so I think the alignment is correct and probably even resolves the `sanitize_post_statuses()` `@todo` a few lines up, where `wcpt-pre-planning` has been queryable but filtered back out of the response.

The thing I would not want is for it to land unremarked. Given how carefully the rest of the description accounts for what changes and what it costs, this one deserves a paragraph and a test of its own — or the narrow list back, if the pre-planning half is meant to be a separate decision.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
```diff
@@ -256,6 +254,57 @@ public static function get_publicly_viewable_post_statuses() {
...
+		}
+
+		if ( current_user_can( 'wordcamp_wrangle_wordcamps' ) ) {
+			return $where;
+		}
+
+		// An applicant keeps their own record. `-1` rather than the logged-out `0`, which is
+		// a real `post_author` value and would expose every unattributed camp.
+		return $where . $wpdb->prepare(
+			" AND NOT ( {$wpdb->posts}.post_status = %s AND {$wpdb->posts}.menu_order = 0 AND {$wpdb->posts}.post_author <> %d )",
```
bor0: `was_ever_scheduled()` just below reads `menu_order > 0`, so a negative `menu_order` counts as never scheduled there, while this clause tests `menu_order = 0` and counts it as scheduled. Measured on a cancelled camp with `menu_order = -1`: the permalink resolves, REST returns 401.

Nothing writes a negative value today, so this is a latent inconsistency rather than a live one. But it is the same shape as the `check_read_permission()` comment — one rule written twice, already drifted on the first edge case — which is the argument for deriving both from a single predicate. Failing that, `menu_order <= 0` here matches `was_ever_scheduled()`.

bor0 (commented):
Reviewed by fetching the branch, reading the surrounding code, and running probe tests against both the branch and `production` in the phpunit container. Confirming the description's numbers first: full suite on the branch is **862 tests, 1955 assertions, 0 failures, 6 skipped**, matching what you reported.

The core call is right — `posts_where` rather than per-row filtering, so a collection's items, its `X-WP-Total` and its page count describe the same set. The reasoning in the description is unusually complete, which is why the two findings below stand out: they contradict the design the description argues for, rather than being alternatives to it.

Holding off on approving for the two inline comments on `check_read_permission()` and the SQL clause. The third inline is smaller but I think it wants a decision rather than a silent change.

Non-blocking, fine as follow-ups:

**Mentors get the mirror image of the author problem.** `map_subrole_caps()` (`wcorg-subroles.php:158-180`) grants a mentor `edit_post` on their mentee camp, so a mentor reads a cancelled application over REST but gets nothing at its permalink. Measured for a mentor on an application-stage cancellation: REST 200, permalink 0 posts, against 200 / 1 on `production`. Same root cause as the first inline comment — one rule, two implementations — so it likely falls out of whatever fix that gets.

**The filter is narrower than it reads.** `post_type => 'any'` skips the guard at the top of `hide_unscheduled_cancellations()`; I confirmed a hidden camp comes back that way. And `get_posts()` defaults to `suppress_filters => true`, so `posts_where` never runs — which is how both `get_wordcamps()` and `get_wordcamp_post()` call it. Nothing is affected today, since every public surface on Central pins `post_status` to scheduled/closed/pre-planning, but the docblock currently reads like a general boundary and it is really "queries that name the post type and don't suppress filters". Worth a sentence so the next caller knows.

**The privacy exporter.** `WordCamp\WCPT\Privacy\_get_wordcamp_posts()` (`wcpt-wordcamp/privacy.php:284`) runs `post_type=wordcamp, post_status=any` while handling a personal-data export request, over admin-ajax as whoever is processing it. Super admins clear `wordcamp_wrangle_wordcamps`, but a plain Central administrator without the subrole does not, and their export would quietly omit cancelled applications. The erasers are commented out, so this is export-only for now. Probably just needs confirming who runs those in practice.

A couple of things I checked that turned out to be non-issues, recorded so nobody re-derives them: WP 6.1+ query caching hashes the final SQL in `generate_cache_key()`, so the per-user `WHERE` clause cannot land in a cache entry shared with another user; `?s=` on Central falls through to `post_type = 'post'`, so site search never reaches camps; and `current_user_can()` inside `posts_where` carries no recursion risk, since the wrangler cap maps without touching a post query.

mcliwanow:
All three were the same thing and you named it: one rule, two implementations. `can_read_unscheduled_cancellation()` holds it now, `check_read_permission()` calls it, and the SQL mirrors it.

Reproduced exactly. `test_the_author_can_read_their_own_cancelled_application()` was passing because I made the fixture a contributor, so it encoded the assumption instead of testing it. Subscriber now, and it asserts the user lacks `edit_wordcamps` before anything else.

Mentors keep their mentee camps on both surfaces rather than losing REST, since that is the set the list table in #1943 gives them. Costs a `postmeta` subquery.

`menu_order <= 0` now. Pre-planning stays, with a paragraph in the description and a test per status, and you are right that it settles the `@todo`.

Both follow-ups are in rather than deferred, exporter included. The description has the measured table, both surfaces, every viewer.

871 tests, 0 failures, the new ones verified failing first.

bor0 (commented):
The redesign is the right one and the single-rule refactor landed: `can_read_unscheduled_cancellation()` reads as the definition, `check_read_permission()` calls it, and the docblocks now say why the SQL exists next to it. Full suite on `123d213`: **871 tests, 1979 assertions, 0 failures, 6 skipped**, matching what you reported. The author fix is real — a subscriber-authored cancellation now reads 200 on both surfaces.

Two things I'd still like resolved, and one stale reference. Both of the first two are the same shape as last time: the rule is expressed once in PHP and once in SQL, and the two disagree on inputs neither of us checked.

### 1. Mentor matching drifts between the clause and `is_mentored_by()`

`is_mentored_by()` (`wordcamp-loader.php:312`) compares with `in_array( ..., true )`, which is byte-exact. The SQL at line 372 compares with `meta_value IN ( %s, %s )`, which runs under the column collation — case-insensitive and trailing-space-insensitive. So the two match different sets.

Measured, mentor meta `MentorJane` against a user whose `user_login` is `mentorjane`:

```
query rows=1   can_read_unscheduled_cancellation()=false   REST single=403
```

Same for a stored trailing space (`mentorbob `), and for two `Mentor WordPress.org User Name` rows on one camp, where `IN` matches either and `get_post_meta( ..., true )` reads only the first.

The consequence is the failure mode this whole design exists to prevent. `WP_REST_Posts_Controller::get_items()` re-checks `check_read_permission()` per row after the query, so when SQL is the more permissive of the two the row is fetched and then dropped:

```
GET /wp/v2/wordcamps?status=wcpt-cancelled  ->  items=0, X-WP-Total=1
```

A short page with an inflated total, from a second implementation of the rule — which is exactly what moving to `posts_where` was meant to rule out.

This isn't hypothetical input, either. `Mentor WordPress.org User Name` is a free-text admin field, and every other consumer of it resolves the name through the database rather than in PHP: `map_subrole_caps()` (`wcorg-subroles.php:176`) and `wordcamp-new-site.php:675` both call `wcorg_get_user_by_canonical_names()`, which is `get_user_by( 'login' )` then `get_user_by( 'slug' )` — case-insensitive both times. So a wrangler who types `MentorJane` gets a mentor who holds `edit_post` on the camp and passes the SQL clause, and is refused by `is_mentored_by()` alone. It is the odd one out among three implementations, not the SQL.

Resolving name to user and comparing IDs would fix the drift and retire the collision caveat in the docblock at the same time:

```php
$mentor = wcorg_get_user_by_canonical_names( get_post_meta( $post->ID, 'Mentor WordPress.org User Name', true ) );

return $mentor && $mentor->ID === $user->ID;
```

That matches what `map_subrole_caps()` already decides, and it matches what the SQL already selects.

### 2. A plain Central administrator loses cancelled applications from the admin list table

`hide_unscheduled_cancellations()` filters every wordcamp query, `is_admin()` included, and the exemption is `wordcamp_wrangle_wordcamps`. Nobody below a wrangler holds that, so the WordCamp list table on Central changes for everyone else. Default `edit.php` query, plain administrator, no subrole:

```
production:  pre-planning=yes  needs-site=yes  cancelled=yes  scheduled=yes
branch:      pre-planning=yes  needs-site=yes  cancelled=NO   scheduled=yes
```

That is the one surface where cancelled applications now behave differently from every other application status. #1931 made those `protected`, which is what keeps them off the front end while leaving them in the list table for anyone with `edit_wordcamps`; this clause takes cancelled out of the list table too. So a Central admin looking for a camp that was cancelled as an application won't find it, and nothing tells them the list is filtered.

You already identified this population for the exporter — the comment on `privacy.php:291` is about precisely the administrator who doesn't hold the subrole. The same argument seems to apply here: wp-admin isn't a public surface, and access to it is already gated by `edit_wordcamps`. `is_admin() && ! wp_doing_ajax()` would exempt the list table without opening admin-ajax, and would subsume the export escape hatch rather than needing it alongside. If instead the intent is that only wranglers see these anywhere, that belongs in the description as a stated cost, alongside the 152 and the 34 — it's a bigger behaviour change than either.

### 3. Stale reference

`is_mentored_by()`'s docblock says the comparison is "what `WordCamp_Admin::get_authored_or_mentored_wordcamps()` does". That method doesn't exist on this branch or on `production` — it's #1943. Worth either landing after it or naming something that exists today. (`filter_by_mentor()` at `wordcamp-admin.php:1735`, the nearest thing, matches on `user_login` only, so it isn't the same rule.)

### On the two you flagged

**The 152 with sites.** Agreed, and I'd keep it. `URL` is set at `wcpt-needs-site` and Central's public surfaces only ever covered scheduled and closed, so a site existing never made the Central page a listed one. Using site URL as the signal would also make the rule mean something different from what `menu_order` records, which is the thing you can actually reason about.

**The 34 pre-2016.** I read this one slightly differently, but not enough to block. Those 21 with sites probably were announced, and their Central permalinks may well be linked from elsewhere, which is the argument the description makes for the 55. The alternative to hardcoding a date in the read path is a one-time backfill: set `menu_order` on cancelled camps created before 2016-06-23 that have `URL` meta, so the data carries the fact and `was_ever_scheduled()` stays a pure read of the convention. 21 rows, and the read path never learns the date. Fine as a follow-up either way — just noting the option, since "hardcode 2016-06-23 into a read path" isn't the only way to recover them.

### Verified, recorded so nobody re-derives it

- `wcpt_include_unscheduled_cancellations` is not in `query_vars()`, so `WP::parse_request()` can't set it from a URL. Your test pins this.
- The per-user `WHERE` clause does not leak across users through the WP 6.1+ query cache. Same args, three users, one request: author 1 row, another subscriber 0, logged out 0.
- `add_organizer_to_central()` — confirmed the warning it fixes is live on `production` (`Undefined array key "wcpt_wordpress_org_username"` fires from the test suite today), so that's a genuine fix rather than only a test convenience.

mcliwanow:
All three are in.

Mentor matching resolves through `wcorg_get_user_by_canonical_names()` now, so PHP agrees with the collation. `MentorJane` for `mentorjane`: 200 on both surfaces, and that collection is items=1 against `X-WP-Total` 1, where you measured 0 and 1.

wp-admin is exempt, but it does not subsume the export argument. Exports run inside `wp_ajax_wp_privacy_export_personal_data()`, so `wp_doing_ajax()` is true and the exemption skips them. There is a test that it stops at admin-ajax.

Stale reference gone, and the description takes your backfill point for the 21.

One for the record: this suite defines `WP_ADMIN`, so `is_admin()` is true unless a screen says otherwise, which is why `set_up()` now pins the front end.

Rebased onto current production.

bor0 (commented):
Re-reviewed at `edca06c7` by fetching the branch and running the suite and probe queries against both it and `production` in the phpunit container.

**All three items from last round are resolved.** Verified each:

1. **Mentor drift.** `is_mentored_by()` resolves through `wcorg_get_user_by_canonical_names()` and compares IDs (`wordcamp-loader.php:317-322`), matching `map_subrole_caps()` and `wordcamp-new-site.php`. `MentorJane` for `mentorjane` reads 200 on both surfaces and its collection is `items=1` against `X-WP-Total` 1, where I measured 0 and 1. Worth recording next to it: `Event_Admin::standardize_usernames()` (`class-event-admin.php:673`) already canonicalises to `user_login` on save, so this only ever bit legacy or imported rows. The fix is still the right one, and it retires the collation caveat.
2. **The admin list table.** `is_admin() && ! wp_doing_ajax()` is in, pinned by two tests including the admin-ajax boundary.
3. **The stale `get_authored_or_mentored_wordcamps()` reference** is gone.

Suite on the branch: **920 tests, 2168 assertions, 0 failures, 6 skipped**, against **897** on the `production` the branch merged (+23). `phpcs-branch.php` clean.

---

### 1. The "one deliberate widening" is bigger than the description says

The description says the widened statuses return "title, content, excerpt, link and author, **with no meta, since nothing is registered for REST here**." That isn't the case. `WordCamp_Loader::register_rest_public_fields()` (`wordcamp-loader.php:604`) registers the twenty keys from `get_public_meta_keys()` plus `WordCamp_Admin::get_venue_address_meta_keys()` on the `wordcamp` post type.

Measured anonymously, same fixture on each side:

| | production | branch |
|---|---|---|
| `GET /wp/v2/wordcamps/<id>`, all six pre-planning statuses | 401 | 200 |
| `?status=wcpt-pre-planning` collection | items=0, total=1 | items=1, total=1 |

The 200 carries `Organizer Name`, `WordPress.org Username`, `Number of Anticipated Attendees`, `Physical Address`, `Venue Name`, `Maximum Capacity`, `Available Rooms`, `Exhibition Space Available` and the `_venue_*` / `_host_*` address parts. Those are populated at that stage rather than hypothetical: `class-wordcamp-application.php:218-229` and `jetpack-form-to-wcpt-application.php:143-150` write `Organizer Name`, `Number of Anticipated Attendees` and `WordPress.org Username` on every application.

The "same material the permalink already renders" argument holds for the content, and `single-wordcamp.php:45-72` does render venue name and physical address. It doesn't render the organiser's name, their w.org username, anticipated attendees, capacity, rooms or exhibition space. Those go from 401 to readable, and listable, without credentials.

This isn't needed for the change either. The cancelled rule is fully handled by the branch above it, and `get_publicly_viewable_post_statuses()` differs from the old list by exactly the six pre-planning statuses. So it's either: keep `check_read_permission()` on `get_public_post_statuses() + 'wcpt-cancelled'` and leave the `@todo` for its own change, or keep the widening with its cost stated accurately alongside the 152 and the 34. I'd take the first, since it keeps this PR to its subject — but it's a call rather than a defect: the code does what it says, the description undersells what it opens.

### 2. The wp-admin exemption reintroduces the drift the design is built to rule out

Non-blocking, but the same shape as last round.

`hide_unscheduled_cancellations()` now carries three exemptions — `is_admin() && ! wp_doing_ajax()`, the export query var, and cron/WP-CLI. `can_read_unscheduled_cancellation()` carries only cron/WP-CLI. Inside wp-admin the SQL is the more permissive of the two, and `WP_REST_Posts_Controller::get_items()` re-checks the rule per row after the query:

```
REST collection dispatched with is_admin() true:  items=0, X-WP-Total=1
```

That's the short page under an unreduced total that moving to `posts_where` existed to prevent, now in the fix for last round's list-table finding.

Reachability is thin today: it needs an internal `rest_do_request()` from an admin screen, by someone who is neither the author, the mentor nor a wrangler — and `edit_others_posts` already maps to `wordcamp_wrangle_wordcamps`, so in practice nobody can open that editor. So it's a latent trap rather than a live bug. But the next in-admin REST consumer inherits it silently, and the docblock says this method *is* the rule. Giving `can_read_unscheduled_cancellation()` the same `is_admin() && ! wp_doing_ajax()` exemption restores that property.

### 3. Duplicate mentor meta rows, same shape

Minor. `meta_value IN ( … )` matches any row for the key; `get_post_meta( …, true )` reads only the first. A camp with two `Mentor WordPress.org User Name` rows where the second names the current user gives `php=false sql=true`, and `items=0` against `X-WP-Total` 1 again. Not reachable through the admin, which uses `update_post_meta`, so imports and WP-CLI only. Noting it because it's the same class as 2, and both close together if the mentor lookup and the subquery are stated once.

---

### Verified, recorded so nobody re-derives it

- `get_publicly_viewable_post_statuses()`'s **body** is unchanged from `production` — only its docblock moved. So the pre-planning statuses are already registered `public` and their permalinks already resolve; the widening in 1 is REST-only, as the description says.
- The SQL parenthesises correctly: `AND NOT ( status = cancelled AND menu_order <= 0 AND NOT ( author OR mentor ) )`, with `$readable` prepared and the outer string not re-prepared over it.
- `add_organizer_to_central()`'s guard is a genuine fix, not only a test convenience — the `Undefined array key` warning still fires from the suite on `production` today.
- The branch is two commits behind `production` (`20c11d47`, `27ebc02a`). GitHub still reports it CLEAN, so this is only worth a merge before landing.

Only 1 needs a decision before merge.

bor0:
Addendum to the above, on point 3. I said duplicate meta rows were the only way the two mentor implementations disagree, and that it was import-only. That was too narrow — there are two more reachable inputs, and one of them is the bad direction.

The two sides normalise differently. `is_mentored_by()` asks "which user does this string resolve to", via `sanitize_user()` inside `get_user_by( 'login' )` and then `sanitize_title()` inside `get_user_by( 'slug' )`. The SQL asks "does this string equal *this* user's `user_login` or `user_nicename`" under the column collation, which pads trailing spaces and nothing else.

Measured on the branch, application-stage cancellation, viewer logged in:

| stored mentor value | `is_mentored_by()` | SQL | permalink | single REST | collection |
|---|---|---|---|---|---|
| `"mentorsue "` (trailing space) | true | true | resolves | 200 | items=1, total=1 |
| `" mentorjane"` (leading space) | true | **false** | **empty** | **200** | items=0, total=0 |
| `"mentorbob\n"` (trailing newline) | true | **false** | **empty** | **200** | items=0, total=0 |

Trailing space is the control and agrees. Leading whitespace and a newline don't: `sanitize_user()` trims both, the collation trims neither. The mentor gets 200 from the API and an empty permalink — the two surfaces describing different sets, which is the property the design exists to hold.

The other direction is rarer but it's the failure mode that motivated moving to `posts_where`. If one user's `user_login` is another user's `user_nicename` and the meta holds that string, the SQL selects the row for the second while `is_mentored_by()` resolves it to the first:

```
viewer = the user holding it as nicename:  php=false sql=true  permalink=resolves  single=403  items=0  X-WP-Total=1
```

WordPress keeps `user_login` and `user_nicename` unique within their own columns but not against each other, so this needs a nicename edited away from its login and a second account taking that string as its nicename. Exotic, but nothing prevents it, and `wp_insert_user()` won't be the thing that stops it.

Both close the same way as the duplicate-row case: normalise once and compare one thing. Resolving the stored name to a user id and comparing ids on both sides — the SQL arm selecting on a resolved id rather than on the raw string — makes all four rows above agree by construction, and it's the same shape as the `post_author` arm right next to it.

(Independent second pass over the diff agreed with point 1 and turned up nothing else; it counts ~36 registered keys via `get_public_meta_keys()` plus `get_venue_address_meta_keys()`, and notes they carry no `context` restriction, so they're in every readable response rather than only `edit`.)

mcliwanow:
Thanks, both were right, and the description was wrong about the meta.

`check_read_permission()` is back on the public statuses plus cancelled, so none of what `register_rest_public_fields()` exposes gets published. The rule now takes the same exemptions as the clause. The mentor name resolves canonically again, matching `map_subrole_caps()` and the list table in #1943, and `mentor_names_for()` narrows the clause to the names that resolve back to the user rather than the rule widening to match it.

`test_the_clause_and_the_rule_admit_the_same_set` compares the two across viewers, camps and request contexts. Mutating each defect review has found back in fails it.

Production merged. Another look when you have time?

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
```diff
@@ -256,6 +254,218 @@ public static function get_publicly_viewable_post_statuses() {
...
+	public static function can_read_unscheduled_cancellation( $post ) {
+		// No user to hold the capability, and `wp post list` should report the real set.
+		if ( wp_doing_cron() || ( defined( 'WP_CLI' ) && WP_CLI ) ) {
+			return true;
+		}
+
+		// Every exemption the clause below takes has to be taken here too. Where the clause is
+		// the more permissive of the two, it selects a row this then refuses, and the response
+		// comes back short under a total that was never reduced.
+		if ( is_admin() && ! wp_doing_ajax() ) {
```
bor0: Non-blocking, and there's no live path to it today, but the exemption is wider than the justification for it.

The comment on the clause (line 407) reasons that "wp-admin is not a public surface and reaching a WordCamp screen already needs `edit_wordcamps`", and admin-ajax is carved out because any logged-in user can reach it. `admin-post.php` defines `WP_ADMIN` too, so `is_admin()` is true and `wp_doing_ajax()` is false there, and it's reachable logged out — it dispatches `admin_post_nopriv_{$action}` before any capability check. So the exemption currently reads as: anyone, on a request that isn't a WordCamp screen and needs no capability at all.

It's inert as things stand: I grepped the network and nothing registers an `admin_post_nopriv_*` handler, so there's no request that both takes this branch and renders a camp. It's the shape I'd want closed rather than an actual read.

What would close it without another special case is to say the capability the comment is already appealing to:

```php
if ( is_admin() && current_user_can( 'edit_wordcamps' ) ) {
	return true;
}
```

`register_post_capabilities()` grants `edit_wordcamps` to contributor and up on Central, so this admits exactly the people who can open the list table — which is what the exemption is for — and refuses a subscriber or a logged-out request whatever wp-admin entry point they arrived through. admin-ajax then needs no mention: a contributor reaching it sees what their list table already shows, and anyone below it sees nothing.

The same substitution works on the clause at line 407. The privacy exporter keeps its own opt-in either way, and `test_the_wp_admin_exemption_does_not_reach_admin_ajax()` would want re-pointing at a subscriber, which is the case it's actually about.

Happy for this to be a follow-up if you'd rather not touch the matrix again in this PR.

bor0 (approved):
All three of my earlier points are addressed, and addressed at the level they were about rather than patched at the call site:

- The author/`edit_post` split is gone: `can_read_unscheduled_cancellation()` is now the single rule, `check_read_permission()` calls it and the clause mirrors it, and `test_the_author_can_read_their_own_cancelled_application()` asserts `! current_user_can( 'edit_wordcamps' )` first so it can't quietly pass on a capability again.
- `check_read_permission()` keeps the narrow status list, with the reasoning for not aligning it with `get_publicly_viewable_post_statuses()` written down and the `@todo` left where it was. That's the right call — the meta `register_rest_public_fields()` exposes is not what `single-wordcamp.php` renders, and that decision deserves its own change.
- `menu_order <= 0` now agrees with `was_ever_scheduled()`, and `'negative order'` is a row in the matrix.

`test_the_clause_and_the_rule_admit_the_same_set` is the part worth keeping. Every defect this PR has collected has been the SQL and the PHP answering the same question differently, and a 10 × 8 × 3 matrix comparing `WP_Query` against `check_read_permission()` — the controller method, not the predicate on its own — is a test that will keep catching the next one. The mentor handling is right too: resolving through `wcorg_get_user_by_canonical_names()` and narrowing the clause with `mentor_names_for()`, rather than widening the rule to match the clause, keeps the shared-name case on the side that `map_subrole_caps()` already puts it.

I checked the surfaces the clause doesn't reach. `get_wordcamps()` is `get_posts()`, so `suppress_filters` skips it, but every caller either names an explicit status set that excludes cancelled (`city-landing-pages.php`, the map markers, `get_upcoming_wordcamps_query()`, `get_map_stats()`) or runs from a site where the loader was never instantiated (`wporg-flagship-landing`, which already asks for cancelled deliberately). Nothing there regresses.

One non-blocking note inline about how wide the wp-admin exemption is. Approving — it isn't a live path today.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
mcliwanow: Took it: both sites now check `is_admin() && current_user_can( 'edit_wordcamps' )`, and the test is re-pointed at a subscriber and a request with no user.

PR #1943, opened 2026-08-25
WCPT, WC Post Types: scope the event admin screens, and register the status rules on every write path

mcliwanow:
## Why

Three places in the event admin where the code and the access rules around it disagree.

- **The WordCamp list table is not scoped to the viewer.** `wp_edit_posts_query()` asks for `perm => 'readable'`, which `WP_Query` applies to the literal `private` status. This workflow lives in custom statuses, so the query returns the full pipeline whatever the viewer's relationship to it. The status links above it come from `wp_count_posts( $post_type, 'readable' )` and count the same set.
- **The `wcb_*` list tables print meta the REST API treats as private.** `inc/rest-api.php` registers `_wcb_speaker_email` and `_wcb_sponsor_amount` with `context => edit` and a `meta_auth_callback()`; the Gravatar Email and sponsor Amount columns print them unconditionally. The Sessions screen also dereferences speakers with `post_status => any`, so it lists `private` speaker records that the Speakers screen keeps to their author.
- **The status rules only run in wp-admin.** `enforce_post_status()` is registered in the `WordCamp_Admin` constructor, which `WCPT_Loader::core_admin()` only builds when `is_admin() || DOING_CRON`, so REST writes never reach it. Core's `handle_status_param()` requires `publish_posts` for `publish`, `future` and `private`, and settles the other registered statuses elsewhere.

## What changed

- A `pre_get_posts` filter limits the list table to the applications the viewer authored or mentors, the set `map_subrole_caps()` already lets them edit, and a `wp_count_posts` filter recounts the status links over that same set. Wranglers and Central administrators are unaffected on both: `edit_wordcamps` is granted to editors and administrators as well, so keying only on the wrangler capability would have narrowed the screen for people who administer the site. Central editors are scoped, since there is no capability between the two to key on. Core narrows the default screen to the viewer's own rows whenever they have authored one of these and lack the type's `edit_others_posts`, which is true of exempt viewers as well, so that is cleared before the exemption rather than after it. The count filter only rewrites this post type, so anything else counting in the same request passes through.
- Mentorship is decided by resolving the stored name through `wcorg_get_user_by_canonical_names()`, which is how `map_subrole_caps()` decides it. The meta compare is a prefilter: on its own it also matches a camp whose mentor name is somebody else's login that happens to be this viewer's nicename, and the list would then offer a row that `edit_post` refuses.
- The filter also clears the author `WP_Posts_List_Table` adds to the default screen for anyone who has written one of these and cannot edit others'. Intersected with the scoped set it hid the camps they only mentor while the status links still counted them, so the screen promised more rows than it showed. It clears the request value as well as the query var, because `get_views()` and `is_base_request()` read the first and would otherwise mark Mine active over a table showing more than that. Core only adds it when the request named no author, so choosing Mine is untouched.
- The Gravatar Email and sponsor Amount columns check `edit_post` on the row, as the avatar columns already do via `edit_post_link()`. The Sessions screen skips a `private` speaker the viewer cannot read, and keeps listing drafts.
- `organizer-reminder` registers an explicit `capabilities` map keyed to the `manage_options` its own menu page declares. It had core's generic capabilities, so `post-new.php`, which does not consult the admin menu, opened without it.
- `notify_new_wordcamp_in_slack()` passed the raw start-date meta to `gmdate()`, which TypeErrors on an empty one under PHP 8.4 and would otherwise announce a 1970 camp. Unreachable from the admin path, which requires the field, so it only bites tests today.
- `enforce_post_status()` moves to an always-registered `WordCamp_Status_Guard`, and now also covers writers with no user. Cron is exempt, and WP-CLI is exempt only when the process has no user, so `wp --user=…` is held to the capability. Inserts are left to the caller; `create_posts` is what decides who may create one.
- `set_scheduled_date()` moves to priority 20, after both filters that can revert a status. It reads the submitted status and stamps `menu_order` once, never revising it, so a write that either the capability clamp or `require_complete_meta_to_publish_wordcamp()` rejected used to leave behind a scheduled date the camp never had.

## Deliberately unchanged

`edit_wordcamps` is still granted to `contributor` and `author`: narrowing it needs a `remove_cap()` on Central and would cut the mentor branch in `wcorg-subroles.php`. The `wcb_*` types still have no `capabilities` map. Their screens are meant to be reachable at contributor level, because that is how organizers and volunteers work on their own camp site, so the correction belongs in the columns rather than the capability model. `organizer-reminder` is the opposite case: its menu page already required `manage_options` and only the post type disagreed. `require_complete_meta_to_publish_wordcamp()` and `trigger_schedule_actions()` stay in `WordCamp_Admin`, so a REST status change still fires no Slack notification, and `WCPT_Vetting_Abilities` logs its own.

One edge left in the list filter: `post_status => 'any'` excludes `trash`, so the Trash view is empty for a restricted author, who cannot trash anything anyway. A stored mentor name that no longer resolves to an account shows the camp to nobody, which is what `edit_post` decides for it too.

## Testing

43 new tests: 19 list table, 11 status guard, 8 columns, 5 reminder capabilities. Full suite green on the branch with the same 6 pre-existing skips as `production`. `phpcs` clean on the changed lines.

`test-wcor-mailer.php` needed fixing: three cases moved a fixture to `wcpt-scheduled` while logged out, which the old `is_user_logged_in()` gate allowed. Once the write stopped landing they asserted nothing, so the fixture holds the wrangler capability now and 21 assertions come back. `wc-post-types` called `site_supports_block_templates()`, imported from an mu-plugin, before checking whether the post was one of its own types, so every `the_content` render reached for it. The operands swap, which is also why no test bootstrap has to load that mu-plugin.

As a Contributor on a local install, before and after:

| | before | after |
|---|---|---|
| `edit.php?post_type=wordcamp` | every application | their own |
| status links | every application | their own |
| REST `{"status":"wcpt-scheduled"}` on their own camp | stored | unchanged |
| Gravatar Email and sponsor Amount columns | the values | empty |
| private speaker on a session row | listed | absent |
| a camp they mentor but did not write, default screen | every application | listed, with All marked active and its count matching the rows |

## How to review this locally

`.docker/wp-config.php` gives user 6 (`contributorrole`) the `wordcamp_wrangler` subrole.

1. **Set up.** Add two contributors, `floworg` and `flowmentor`. Create one camp authored by `floworg`, and a second authored by somebody else with `Mentor WordPress.org User Name` set to `flowmentor`.

2. **Check the list table** at `/wp-admin/edit.php?post_type=wordcamp`. `floworg` sees their own camps, with the status links counted over just those. `flowmentor` sees only the camps they mentor. `contributorrole` sees everything, with the global counts and the "Mine (Mentoring)" view. Clicking a scoped link returns the camps it promised. Edit screens are unchanged: 200 on your own or mentored camp, 403 otherwise.

3. **Change a status in the metabox.** As `contributorrole` it sticks. As `floworg`, pick "WordCamp Scheduled" on their own camp: the save redirects, the status does not move, and `menu_order` stays `0`. Same over REST, with a nonce from `admin-ajax.php?action=rest-nonce`. `wp post update <id> --post_status=wcpt-closed` still works, which is what the exemption is for.

4. **Check the private columns on a camp site.** Give `floworg` `contributor`, then add a speaker with `_wcb_speaker_email` and a sponsor with `_wcb_sponsor_amount`. Both columns are empty for `floworg` and populated for `admin`.

5. **Check the reminder screens on Central.** `/wp-admin/post-new.php?post_type=organizer-reminder` is 403 for `floworg` and 200 for `admin`, and the reminder list still shows every reminder to `admin`.

6. **Run the tests.**

```
docker compose -f docker-compose.phpunit.yml run --rm phpunit_wp phpunit -c phpunit.xml.dist \
  --filter 'Test_WordCamp_List_Table_Access|Test_WordCamp_Status_Guard|Test_Private_Column_Meta|Test_WCOR_Reminder_Capabilities'
```

Quick edit and bulk edit both write through `wp_insert_post()`, so the clamp covers them. As the applicant neither moves a status, including on a camp they do not own, and both apply normally as a wrangler. Paged past page one with `per_page` at 2, the list returns 2 then 1, total 3, and never the camps outside the scope.

Not driven by hand: the Campus Connect abilities over MCP, and cron rather than a `wp_doing_cron` filter.


public_html/wp-content/plugins/wc-post-types/wc-post-types.php
```diff
@@ -820,7 +820,7 @@ public function add_avatar_to_speaker_posts( $content ) {
 		global $post;
 		$enabled_site_ids = apply_filters( 'wcpt_speaker_post_avatar_enabled_site_ids', array( 364 ) );    // 2014.sf
 
-		if ( site_supports_block_templates() || ! $this->is_single_cpt_post( 'wcb_speaker') ) {
+		if ( ! $this->is_single_cpt_post( 'wcb_speaker' ) || site_supports_block_templates() ) {
```
mcliwanow, line 823: `$this->is_single_cpt_post( 'wcb_speaker' )` is cheaper than `site_supports_block_templates()`

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
```diff
@@ -20,12 +20,16 @@ class WordCamp_Loader extends Event_Loader {
...
+		add_action( 'wp_insert_post_data',             array( $this, 'set_scheduled_date' ), 20 );
 		add_filter( 'wordcamp_rewrite_rules',          array( $this, 'wordcamp_rewrite_rules'            ) );
 		add_filter( 'query_vars',                      array( $this, 'query_vars'                        ) );
 		add_filter( 'rest_wordcamp_collection_params', array( $this, 'set_rest_post_status_default'      ) );
 		add_action( 'rest_api_init',                   array( $this, 'register_rest_public_fields'       ) );
 		add_action( 'init',                            array( $this, 'register_post_capabilities' ) );
+
+		// Not from `includes()`: that is hooked to `plugins_loaded`, and the Jetpack form
+		// bridge requires this loader from a form handler long after that fires.
+		WordCamp_Status_Guard::init();
```
KokkieH, line 35: `WordCamp_Loader`'s constructor now depends on `WordCamp_Status_Guard`, but this file doesn't load that class — it's only required in `WCPT_Loader::includes()`. Several places require `wordcamp-loader.php` directly without going through the loader (`wordcamp-site-cloner.php`, `sponsor-invoice.php`, `wporg-flagship-landing/functions.php`, `4-helpers-wcpt.php`, `structured-data.php`, and the organizer-reminders test bootstrap). Today those only use constants/statics, so nothing breaks, but any future `new WordCamp_Loader()` from one of those contexts would fatal on a missing class. A `require_once __DIR__ . '/class-wordcamp-status-guard.php';` at the top of this file would make it self-contained.

The comment is also slightly off: `includes()` runs synchronously in `WCPT_Loader::__construct()` — only `core_admin()` is deferred to `plugins_loaded`.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/class-wordcamp-status-guard.php
```diff
@@ -0,0 +1,81 @@
...
+
+	/**
+	 * Enforce a valid post status for WordCamps.
+	 *
+	 * @param array $post_data
+	 * @param array $post_data_raw
+	 * @return array
+	 */
+	public static function enforce_post_status( $post_data, $post_data_raw ) {
+		if ( WCPT_POST_TYPE_ID != $post_data['post_type'] || empty( $post_data_raw['ID'] ) ) {
```
KokkieH: This early return means the status of a brand-new post is never checked by the guard. That's fine today because `create_posts` maps to `wordcamp_wrangle_wordcamps`, so only wranglers and trusted code paths can insert these — but the guard silently depends on that capability mapping staying as strict as it is. Worth either a short comment recording the dependency, or clamping inserts to `WCPT_DEFAULT_STATUS` for writers without the capability so the class holds on its own.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/class-wordcamp-status-guard.php
```diff
@@ -0,0 +1,81 @@
...
+
+		$post = get_post( $post_data_raw['ID'] );
+		if ( ! $post ) {
+			return $post_data;
+		}
+
+		if ( ! empty( $post_data['post_status'] ) ) {
+			// Only Wranglers change statuses. Cron and WP-CLI have no user to hold the
+			// capability, and `close_wordcamps_after_event()` writes `wcpt-closed` from cron.
+			$system_context = wp_doing_cron() || ( defined( 'WP_CLI' ) && WP_CLI );
```
KokkieH: One corner where this is looser than the old `is_user_logged_in()` check: under WP-CLI with `--user=<low-privilege-user>`, the old code clamped the status (logged in, no capability held), while this exempts the whole process regardless of the user set. Probably an acceptable trade-off for an ops-only surface, but flagging it since the description presents the exemption purely as covering user-less writes.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
```diff
@@ -1743,6 +1690,158 @@ public function filter_mentoring_view( $query ) {
...
+		 *
+		 * `wp_edit_posts_query()` asks for `perm => 'readable'` when a status is filtered, and
+		 * `readable` restricts results to their author only for the literal `private` status.
+		 * This workflow is expressed in custom statuses, so without this the screen lists
+		 * every application to anybody who can open it.
+		 *
+		 * @param WP_Query $query The query to filter.
+		 */
+		public function limit_list_to_editable_wordcamps( $query ) {
+			if ( ! $this->is_wordcamp_list_query( $query ) || current_user_can( self::get_edit_capability() ) ) {
```
KokkieH: This keys full-list visibility purely on `wordcamp_wrangle_wordcamps`, but `register_post_capabilities()` grants `edit_wordcamps` to editors and administrators too. A Central editor or administrator without a subrole previously had a read-only view of the whole pipeline; after this change they'll see only what they authored or mentor. Super admins and the `wordcamp_wrangler`/`mentor_manager` subroles are unaffected. It's coherent with the existing edit model, but it is a behaviour change for a persona the description doesn't mention — worth confirming nobody relies on browsing the list with a plain role before this ships.

KokkieH (commented):
Reviewed the full diff and ran the manual testing steps from the description against a local environment. The overall architecture looks right to me: extracting the status rules into an always-registered `WordCamp_Status_Guard` is the correct structural move, and the `wp_insert_post_data` ordering (guard at 9 → meta requirements at 11 → `set_scheduled_date` at 20) cleanly fixes the stale `menu_order` stamp. I also confirmed the REST read side is already handled by `check_read_permission()` in the custom controller, so scoping the admin list table doesn't leave an equivalent gap on the API listing.

Manual verification — every check matched the "after" column of the table in the description:

- List table: a contributor sees only their own applications with matching status-link counts; a mentor sees only mentored camps; a wrangler sees everything with global counts and the Mine (Mentoring) view. A user who authored one camp and mentors another sees both on the default screen, with **All** marked active and its count matching the rows; an explicit `author=` Mine view still narrows to what they authored. Edit screens: 200 on own/mentored, 403 otherwise.
- Status changes: the author's classic-editor save redirects normally but the status doesn't move and `menu_order` stays 0; same over REST with a fresh `rest-nonce` (200, unchanged). A wrangler's change sticks through both paths. `wp post update --post_status=…` with no user works, as intended for the CLI exemption.
- Camp-site columns: Gravatar Email and sponsor Amount render empty for a contributor and populated for an admin, while the rows themselves stay visible. A `private` speaker attached to a session is absent from the contributor's Sessions screen and present on the admin's.
- Reminders: `post-new.php?post_type=organizer-reminder` is 403 for a contributor and 200 for an admin; the reminder list still renders in full for the admin.

Two minor notes that didn't seem worth inline threads:

- A REST client whose status change is reverted gets a 200 with the rest of the update applied — there's no signal the status was refused. That's inherent to doing this in `wp_insert_post_data`, so fine as is; if an explicit error is ever wanted, `rest_pre_insert_wordcamp` would be the place.
- The `$_REQUEST['author']` vs `$_GET['author']` distinction works because `WP_Posts_List_Table::__construct()` writes only `$_GET` and PHP never refreshes `$_REQUEST`. It's coupled to a core internal, but it's well-commented and the tests would fail loudly if that ever changed.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
mcliwanow: Both fixed in 32df4813. `wordcamp-loader.php` requires the guard itself now, so the constructor holds wherever the file is required from.

You are right about the comment too, and I checked: `includes()` runs synchronously in `WCPT_Loader::__construct()`, only `core_admin()` is deferred to `plugins_loaded`. The comment said the opposite. Rewritten to say what it actually depends on, which is that callers reach this file directly.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/class-wordcamp-status-guard.php
mcliwanow: Took the comment rather than the clamp, in 32df4813, and it is worth saying why since I tried the clamp first.

Clamping inserts to `WCPT_DEFAULT_STATUS` for writers without the capability broke two existing tests that create fixtures at a chosen status through `self::factory()->post->create()`. That made me think it is wrong in principle rather than just inconvenient: `wp_insert_post()` is the trusted server-side API and a caller reaching it has already decided, so clamping there means the guard overrules trusted code. The guard's business is submitted changes, which carry a user's intent.

So the dependency is recorded instead: the comment now says `create_posts` is what keeps an application from starting at an arbitrary status, that the mapping is load-bearing, and that it should stay as strict as it is.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/class-wordcamp-status-guard.php
mcliwanow: Tightened in 32df4813 rather than accepted, since you are right that the description presented it as covering user-less writes and it did not.

It is now `wp_doing_cron() || ( defined( 'WP_CLI' ) && WP_CLI && ! is_user_logged_in() )`, so `wp --user=<low-privilege-user>` is held to the capability exactly as the `is_user_logged_in()` check it replaced did. The uses that need the exemption run without `--user`, so nothing legitimate changes: cron writing `wcpt-closed` from `close_wordcamps_after_event()`, and `wp post update` from the command line.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
mcliwanow: Half addressed in 32df4813, and I would rather flag the other half than let it look settled.

Administrators now keep the whole list: the filter exempts `manage_options` alongside the wrangler capability. Editors do not, so the persona you named is still split down the middle.

Having thought about it since, `manage_options` is the wrong discriminator anyway. Keying the exemption on the core `edit_others_posts` primitive covers editors and administrators and excludes contributors and authors, which is exactly the population that previously had the read-only view, and it is the capability core's own list table uses to decide who sees everyone's rows. Happy to switch it if you agree.

The other half of your comment is the part nobody has done: confirming that nobody actually browses this screen with a plain editor role. I have no way to check that from here, so it needs someone who knows who uses the Central pipeline screens. Worth settling before this ships either way, since it is a behaviour change for a persona the description still does not mention.

mcliwanow:
Thanks for going through this properly, and for running the manual steps rather than taking the table on trust. The independent check that `check_read_permission()` already covers the REST read side was the one I most wanted someone else to make.

All four inline comments are answered above and the code is in 32df4813. Three are fixed. The fourth is half done: administrators keep the whole list now, editors do not, and the question you raised about whether anyone actually browses that screen with a plain role is still open. I have suggested `edit_others_posts` as a better discriminator than the `manage_options` I used, in that thread.

Two smaller things from the same commit that you did not raise but are worth knowing about: comments and the description no longer describe the pre-fix behaviour, which was a fair objection you made outside the PR, and the test counts are now stated as what the branch adds rather than as totals, since production moved four times while this was open.

Ready for another look whenever suits.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
```diff
@@ -13,19 +13,25 @@
  * @subpackage Loader
  * @since WordCamp Post Type (0.1)
  */
+require_once __DIR__ . '/class-wordcamp-status-guard.php';
```
KokkieH: The `require_once` landed between the `WordCamp_Loader` docblock and the class, so the docblock now documents the require statement and the class is left without one. Moving the require above the docblock (or up with the `define`s) keeps the class documented.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
```diff
@@ -1743,6 +1690,163 @@ public function filter_mentoring_view( $query ) {
...
+		 * @param WP_Query $query The query to filter.
+		 */
+		public function limit_list_to_editable_wordcamps( $query ) {
+			if ( ! $this->is_wordcamp_list_query( $query ) ) {
+				return;
+			}
+
+			// Wranglers curate the pipeline, and a Central administrator already administers
+			// everything on this site, so both keep the whole list.
+			if ( current_user_can( self::get_edit_capability() ) || current_user_can( 'manage_options' ) ) {
```
KokkieH, line 1735: This new exemption branch is the one behaviour 32df4813 changed in this method, and it has no test — the list-table suite covers wranglers, contributors, and mentors, but not an exempt viewer without the wrangler capability. Worth adding whichever capability wins the discriminator discussion.

A test would also surface one corner the commit message overstates: an exempt viewer without `edit_others_wordcamps` who has *authored* a camp returns early here, before the author-clearing below, so core's default-to-Mine injection still applies — their default screen opens on Mine with only their own camps rather than the whole list. That's core-standard, self-consistent UI (All is one click away), but "keeps the whole list" only holds for exempt viewers who haven't authored one.

KokkieH (commented):
Second pass over 32df4813 — thanks, all the substantive points landed well. I re-verified the CLI change locally: `wp post update --user=<contributor>` is now clamped while the user-less write still goes through, and the insert-path comment recording the `create_posts` dependency reads as the right call. Two small leftovers inline; the capability-discriminator question is answered on the existing thread.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
KokkieH: Agreed — switch it to the core `edit_others_posts` primitive. That stands on its own merits, independent of the usage question: it's the capability core's own list table keys on for who may see rows they don't own, it covers editors and administrators, and it excludes contributors and authors, which is the scoping this PR is for. `manage_options` is an admin-only carve-out that happens to work for one role; there's no case where it's the better answer.

One correction to the framing, though: `edit_others_posts` holders are not "exactly the population that previously had the read-only view". Before this PR the screen only needed `edit_wordcamps`, which contributors and authors hold too, so they saw the full list as well. The honest rationale for the comment is core's convention, not restoring the prior population.

Please also add a list-table test for the exemption branch with whichever capability lands — it's the one behaviour the fix commit changed and it's currently uncovered (details in the new inline comment).

Whether any staff workflow actually depends on a plain editor browsing this screen only affects the stakes, not the choice — we'll settle that on our side, but it shouldn't hold up the swap.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-loader.php
mcliwanow: Fixed in 281a679f, the require sits above the docblock now.

Worth admitting the pattern: this is the second time in this PR that something I inserted landed on the wrong side of a doc comment. The first was `limited_post_in()`, where removing a scratch test took its docblock with it, and phpcs caught that one because the function was left with none at all. This one phpcs could not catch, because the class still had *a* preceding comment, just the wrong one.

public_html/wp-content/plugins/wcpt/wcpt-wordcamp/wordcamp-admin.php
mcliwanow: You found a real defect here, not just a missing test, and the commit message was wrong rather than merely overstated. Both fixed in 281a679f.

The exemption returned before the author-clearing ran, so an exempt viewer who had authored one of these met core's narrowing and opened on their own rows. The clearing has moved above the exemption, where it covers both cases: a scoped viewer keeps the camps they only mentor, and an exempt viewer keeps the whole list. Switching the discriminator would not have fixed this, incidentally, because core tests the type's `edit_others_wordcamps` rather than the core primitive, so an editor would have met the same narrowing.

The test is a data provider over having authored one and having authored none. Only the first fails without the fix; the second passes either way, which is exactly why the claim survived my own testing. Description corrected too, and it now says plainly that Central editors are scoped, since there is no capability between `edit_wordcamps` and `manage_options` to key on.

KokkieH: > Description corrected too, and it now says plainly that Central editors are scoped, since there is no capability between edit_wordcamps and manage_options to key on.

So just confirming, you decided to not swap to `edit_others_posts` as discussed above, so Central editors without a wrangler role will only see camps they author or mentor, unlike Central admins who can see all camps?

Fine if it's an intentional choice. Just making sure you're not conflating the two uses of `edit_others_posts`, i.e. the post type's capability slot vs the generic primitive.

Approving to not block this further if this is intentional.

KokkieH (approved):

PR #1944, opened 2026-08-25
Applications: keep submitted form values as text in posts

KokkieH:
## Summary

The handlers that turn a form submission into a `post_title` treat `sanitize_text_field()` as enough to
reduce a submitted value to text. It isn't, because that sanitiser and `wp_kses()` disagree about what
opens a tag:

- `strip_tags()`, which `wp_strip_all_tags()` and therefore `sanitize_text_field()` are built on, only
  treats `<` as a tag opener when a letter, `/`, `!` or `?` follows it. `strip_tags( 'a < b > c' )`
  returns the string unchanged.
- `wp_kses()`'s tag grammar also accepts whitespace there (`'%^<\s*(/\s*)?([a-zA-Z0-9-]+)([^>]*)>?$%'`).

So a value like `< code >` passes the sanitiser untouched, and the `wp_filter_kses()` that core registers
on `title_save_pre` for users without `unfiltered_html` then rebuilds it into a real element. The handler
meant to store text and stored markup instead.

The Forms to Drafts titles get there by a shorter route: Jetpack hands those values over already through
`wp_kses_post()` (`Contact_Form_Plugin::strip_tags()`, applied at `class-contact-form.php:3055`), so
plain `<strong>` reaches `post_title` today without needing the spaced-tag form at all.

Either way the fix is the same — ask for text explicitly — so this adds one shared helper and applies it
wherever a submitted value becomes a title.

## Changes

- **New** `wcorg_sanitize_plain_text( $value )` in `mu-plugins/3-helpers-misc.php`. It runs
  `wp_check_invalid_utf8()` and `wp_strip_all_tags()`, then encodes any `<` that survived, so nothing is
  left for kses to read back as a tag. Idempotent, and `esc_html()` / `esc_attr()` won't double-encode the
  result (`_wp_specialchars()` is called with `$double_encode = false`).
- Applied to the titles built by the WordCamp application (`class-wordcamp-application.php`), the Campus
  Connect bridge (`events/jetpack-form-to-wcpt-application.php`), and the four Forms to Drafts handlers
  (sponsor, volunteer, speaker, session).
- `class-meetup-application.php` and `class-events-application.php` already `esc_html()` their titles and
  are left alone.

Two things the helper deliberately does **not** do:

- It does not call `sanitize_text_field()`, which would also delete every `%[a-f0-9]{2}` sequence and so
  quietly turn `https://example.org/My%20Notes.pdf` into `https://example.org/MyNotes.pdf`.
- It does not use `wp_kses( $value, array() )`, which deletes whole `<...>` spans rather than neutralising
  them, so `Hall < 100 > seats` comes back as `Hall  seats`, and rewrites every `&`.

### Post bodies are intentionally untouched

An earlier revision of this PR also ran the three Forms to Drafts bodies (`Company Description`,
`Your Bio`, `Topic Description`) through the helper. That is reverted, because it cost something and
bought nothing.

Jetpack has already run those values through `wp_kses_post()` before the handler sees them, and
`content_save_pre` runs them through it again on the way into the database. Measured on the installed
Jetpack:

| submitted | after `Contact_Form_Plugin::strip_tags()` |
|---|---|
| `< code >" data-x="< /code >` | `<code>" data-x="</code>` — already normalised |
| `< script >alert(1)< /script >` | `alert(1)` — dangerous tag already gone |
| `<strong>bold</strong>` | `<strong>bold</strong>` — formatting kept, deliberately |

So there is no sanitiser disagreement left to settle on the body path: the spaced form has already been
rebuilt into ordinary markup, constrained to the `wp_kses_post()` allow-list. Running the helper over it
only stripped formatting that submitters can use today and broke percent-encoded URLs. A test now pins
both behaviours.

### It also recovers text that is lost today

Worth stating, because it argues for the change on its own: a title containing a `<...>` span currently
loses that span and everything in it, because kses deletes it on save. Measured through both call-site
paths -- `sanitize_text_field()` then `title_save_pre` for the application handlers, Jetpack's
`wp_kses_post()` then `title_save_pre` for Forms to Drafts:

| submitted | stored today | stored on this branch |
|---|---|---|
| `Hall < 100 > seats` | `Hall  seats` | `Hall &lt; 100 &gt; seats` |
| `Wine < Beer` | `Wine &lt; Beer` | `Wine &lt; Beer` |
| `Rated <A best` | `Rated &lt;A best` | `Rated &lt;A best` |

Only the first row changes. The other two are already preserved today by `wp_pre_kses_less_than()`, which
encodes a `<` run that contains no `>` -- and the helper runs it for the same reason, so it does not
regress them.

### Note on stored values

Encoding `<` means an affected title is stored as `&lt;`, which renders correctly everywhere it goes
through `the_title` or `esc_*`. Two places consume these titles without decoding, so a title containing a
literal `<` would show `&lt;` there: the Application Tracker's JSON payload (`wcpt-event/tracker.php:88`)
and the new-site `blogname` (`wordcamp-new-site.php:479`). A third goes out as text: the organizer
reminder mails substitute `post_title` into `[wordcamp_name]` (`wcor-mailer.php:395`), and that plugin
never sets a `text/html` content type, so `wp_mail()` sends them plain. Both already show raw entities for any title
containing `&`, since kses stores that as `&amp;`, so this isn't a new class of problem — flagging it
because it's the one visible side effect. A `<` in a camp or venue name is rare enough that I'd rather not
add decoding on those paths speculatively.

This only changes values on the way in. Any existing rows are untouched.

## Testing

- `mu-plugins/tests/test-3-helpers-misc.php` — the helper, including entity-encoded input, NUL bytes,
  percent-encoded URLs, arrays, non-scalars and idempotence.
- `wcpt/tests/wcpt-wordcamp/test-wordcamp-application.php` — the WordCamp application handler end to end.
- `wordcamp-forms-to-drafts/tests/test-draft-content.php` — the four Forms to Drafts titles, plus a case
  asserting bodies keep their formatting and percent-encoding.

The handler tests each fail if the helper is removed from the call site they cover, so the suite notices a
future refactor that drops one. `mu-plugins/tests/bootstrap.php` now requires `3-helpers-misc.php`
directly, rather than relying on an unrelated plugin's bootstrap to have loaded it.

```
docker compose -f docker-compose.phpunit.yml run --rm phpunit_wp \
  phpunit --testsuite "WordCamp MU Plugins","WordCamp Post Type","WordCamp Forms to Drafts"
```

- [x] Full suite: 0 failures (6 pre-existing skips — missing `intl`, one GatherPress data provider)
- [x] `phpcs` reports no new violations; the new files are clean

### Manual test steps

Restart the container first so opcache can't serve the old bytecode:

```
git checkout add/plain-text-form-values
```

Both `wp eval` steps below start with `kses_init_filters()`. WP-CLI removes the kses filters that a normal
web request has, and those filters are the second half of what this PR is about, so without that call the
snippets don't exercise the path a real form submission takes. Both clean up after themselves.

1. **A location containing angle brackets is stored as text.** This runs the real application handler:

   ```
   docker exec wordcamptest-wordcamp.test-1 wp --url=central.wordcamp.test --allow-root eval '
   kses_init_filters();
   require_once WP_PLUGIN_DIR . "/wcpt/wcpt-loader.php";
   $a = new WordPress_Community\Applications\WordCamp_Application();
   $d = $a->get_default_application_values();
   $d["q_1079103_wordcamp_location"] = "Portland< code >\" data-x=\"< /code >Oregon";
   $id = $a->create_post( $d );
   echo "title: " . get_post( $id )->post_title . "\n";
   wp_delete_post( $id, true );
   '
   ```

   Expected on this branch — the submitted text, with the angle brackets encoded:

   ```
   title: WordCamp Portland&lt; code &gt;" data-x="&lt; /code &gt;Oregon
   ```

   On `production` the same command gives `WordCamp Portland<code>" data-x="</code>Oregon` — an element
   where the submitted text should be.

2. **It renders back as what was typed.**

   ```
   docker exec wordcamptest-wordcamp.test-1 wp --url=central.wordcamp.test --allow-root eval '
   kses_init_filters();
   $v = "Portland< code >\" data-x=\"< /code >Oregon";
   $id = wp_insert_post( array( "post_type" => "wordcamp", "post_status" => "wcpt-needs-vetting",
     "post_title" => "WordCamp " . wcorg_sanitize_plain_text( $v ) ) );
   echo "rendered: " . apply_filters( "the_title", get_post( $id )->post_title, $id ) . "\n";
   wp_delete_post( $id, true );
   '
   ```

   Expected on this branch:

   ```
   rendered: WordCamp Portland&lt; code &gt;&#8221; data-x=&#8221;&lt; /code &gt;Oregon
   ```

   The `&lt;` / `&gt;` entities render as `<` and `>`, and `wptexturize()` curls the straight quotes to
   `&#8221;`, so a browser shows:

   ```
   WordCamp Portland< code >” data-x=”< /code >Oregon
   ```

   The curly quotes are the point of comparison here: `wptexturize()` skips the contents of a `code`
   element, so on `production` the same title keeps its straight quotes. Getting them curled means no
   element was formed and the value is being treated as ordinary title text, like any other title.

3. **Ordinary submissions are unaffected.** Submit a WordCamp application at
   `https://central.wordcamp.test/become-an-organizer/` with a normal location (`Portland, Oregon`), and
   confirm the tracker title on `/wp-admin/edit.php?post_type=wordcamp` is `WordCamp Portland, Oregon`,
   exactly as before. A value containing an `&` (`Smith & Sons Hall`) should also be unchanged from
   `production`.

4. **Forms to Drafts still works.** On a camp site, submit the Call for Speakers and Call for Sponsors
   forms and confirm the drafted speaker, session and sponsor posts have the submitted titles and bodies,
   with paragraphs intact in the bodies.

5. **Forms to Drafts, without depending on a form.** Step 4 only reaches the handlers if the form's post
   carries the right `wcfd-key` meta and its fields are labelled `Company Name` / `Company Description` /
   `Website`, which is easy to miss on a local site. This calls the handlers directly instead, so it works
   on any camp site. Substitute your own camp URL, and note this one needs `sh -c` because of the heredoc:

   ```
   docker exec wordcamptest-wordcamp.test-1 sh -c 'cat > /tmp/fd.php <<"PHP"
   <?php
   kses_init_filters();
   $wcfd = $GLOBALS["wordcamp_forms_to_drafts"];
   $text = "Portland< code >\" data-x=\"< /code >Oregon";
   $body = "First paragraph with <strong>bold</strong> and https://example.org/My%20Notes.pdf\n\nSecond paragraph.";

   $form_id = wp_insert_post( array( "post_type" => "post", "post_title" => "tmp form", "post_status" => "draft" ) );
   update_post_meta( $form_id, "wcfd-key", "call-for-sponsors" );
   $sub_id = wp_insert_post( array( "post_type" => "feedback", "post_status" => "publish", "post_parent" => $form_id ) );

   $wcfd->call_for_sponsors( $sub_id, array(
       "1_Company Name"        => $text,
       "2_Company Description" => $body,
       "3_Website"             => "https://example.org",
   ), array() );

   $s = get_posts( array( "post_type" => "wcb_sponsor", "post_status" => "draft", "numberposts" => 1 ) );
   echo "sponsor title:   " . $s[0]->post_title . "\n";
   echo "sponsor content: " . str_replace( "\n", " / ", trim( $s[0]->post_content ) ) . "\n";
   echo "sponsor website: " . get_post_meta( $s[0]->ID, "_wcpt_sponsor_website", true ) . "\n";

   $m = new ReflectionMethod( $wcfd, "create_draft_speaker" );
   $m->setAccessible( true );
   $speaker_id = $m->invoke( $wcfd, array( "Name" => $text, "Your Bio" => $body ) );
   echo "speaker title:   " . get_post( $speaker_id )->post_title . "\n";
   echo "speaker content: " . str_replace( "\n", " / ", trim( get_post( $speaker_id )->post_content ) ) . "\n";

   foreach ( array( $form_id, $sub_id, $s[0]->ID, $speaker_id ) as $id ) {
       wp_delete_post( $id, true );
   }
   PHP
   wp --url=shinynew.wordcamp.test/2023 --allow-root eval-file /tmp/fd.php; rm /tmp/fd.php'
   ```

   Expected on this branch — **titles** have their angle brackets encoded, while the **bodies** keep the
   `<strong>`, the percent-encoded URL and their paragraph split:

   ```
   sponsor title:   Portland&lt; code &gt;" data-x="&lt; /code &gt;Oregon
   sponsor content: First paragraph with <strong>bold</strong> and https://example.org/My%20Notes.pdf /  / Second paragraph.
   sponsor website: https://example.org
   speaker title:   Portland&lt; code &gt;" data-x="&lt; /code &gt;Oregon
   speaker content: <!-- wp:paragraph --> / <p>First paragraph with <strong>bold</strong> and https://example.org/My%20Notes.pdf</p> / <!-- /wp:paragraph --> / <!-- wp:paragraph --> / <p>Second paragraph.</p> / <!-- /wp:paragraph --> / ...
   ```

   The bodies are the point of this step as much as the titles: they should be byte-identical to
   `production`. On `production` the titles instead store `<code>` elements. The script cleans up
   everything it creates.

https://claude.ai/code/session_0113Xx49sx6VqVgEZgymZoM5

### DIFF
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -426,7 +426,10 @@ function wcorg_sanitize_plain_text( $value ) {
 		return '';
 	}
 
+	// `strip_tags()` deletes from an unterminated `<` to the end of the string, so encode that
+	// case first -- `Rated <A best` should keep its text, not become `Rated`.
 	$value = wp_check_invalid_utf8( (string) $value );
+	$value = wp_pre_kses_less_than( $value );
 	$value = wp_strip_all_tags( $value, true );
 
 	return str_replace( '<', '&lt;', $value );
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -397,3 +397,42 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
 		true
 	);
 }
+
+/**
+ * Reduce a submitted value to text that stays text once WordPress saves it.
+ *
+ * `sanitize_text_field()` and `wp_kses()` disagree about what opens a tag: `strip_tags()`, which the
+ * former is built on, needs a letter, `/`, `!` or `?` after the `<`, while kses also accepts whitespace.
+ * A value like `< code >` therefore passes the sanitiser untouched, and the `wp_filter_kses()` that core
+ * puts on `title_save_pre` and `content_save_pre` then rebuilds it into a real element -- so a handler
+ * that meant to store text ends up storing markup. Encoding the `<` that survives settles the
+ * disagreement, and leaves the character visible to readers.
+ *
+ * `wp_kses( $value, array() )` is not a substitute: it drops whole `<...>` spans, turning
+ * `Hall < 100 > seats` into `Hall  seats`, and rewrites every `&`.
+ *
+ * The result is encoded for HTML. Decode it before using it in a plain-text medium such as an email
+ * body or a CSV column.
+ *
+ * @param string|array $value            The submitted value. Arrays are handled recursively; keys are
+ *                                       left alone.
+ * @param bool         $keep_line_breaks Whether to keep newlines, as `sanitize_textarea_field()` does.
+ *
+ * @return string|array A string, or an array of strings when `$value` is an array.
+ */
+function wcorg_sanitize_plain_text( $value, $keep_line_breaks = false ) {
+	if ( is_array( $value ) ) {
+		return array_map(
+			function ( $item ) use ( $keep_line_breaks ) {
+				return wcorg_sanitize_plain_text( $item, $keep_line_breaks );
+			},
+			$value
+		);
+	}
+
+	$value = $keep_line_breaks
+		? sanitize_textarea_field( (string) $value )
+		: sanitize_text_field( (string) $value );
+
+	return str_replace( '<', '&lt;', $value );
+}
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -410,7 +410,8 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
  * percent-encoded URLs; `wp_kses( $value, array() )` drops whole `<...>` spans, turning
  * `Hall < 100 > seats` into `Hall  seats`.
  *
- * The result is HTML-encoded. Decode it for a plain-text medium such as an email body or a CSV column.
+ * The result is HTML-encoded. Decode it for a plain-text medium -- the organizer reminder mails
+ * (`wcor-mailer.php`) are the case already in the tree.
  *
  * @param mixed $value Arrays are handled recursively, with keys left alone. Anything neither array nor
  *                     scalar becomes `''`, as `sanitize_text_field()` does.
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -401,24 +401,19 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
 /**
  * Reduce a submitted value to text that stays text once WordPress saves it.
  *
- * `sanitize_text_field()` and `wp_kses()` disagree about what opens a tag: `strip_tags()`, which the
- * former is built on, needs a letter, `/`, `!` or `?` after the `<`, while kses also accepts whitespace.
- * A value like `< code >` therefore passes the sanitiser untouched, and the `wp_filter_kses()` that core
- * puts on `title_save_pre` then rebuilds it into a real element -- so a handler that meant to store text
- * ends up storing markup. Encoding the `<` that survives settles the disagreement, and leaves the
- * character visible to readers.
+ * `strip_tags()` wants a letter, `/`, `!` or `?` after a `<` before it counts as a tag, but `wp_kses()`
+ * also accepts whitespace. So `< code >` survives `sanitize_text_field()`, and the `wp_filter_kses()`
+ * core puts on `title_save_pre` then rebuilds it into a real element. Encoding the surviving `<` settles
+ * that, and still reads as `<` to a human.
  *
- * This deliberately does not call `sanitize_text_field()`, which would also delete every `%[a-f0-9]{2}`
- * sequence and so quietly break a percent-encoded URL. `wp_kses( $value, array() )` is not usable either:
- * it drops whole `<...>` spans rather than neutralising them, turning `Hall < 100 > seats` into
- * `Hall  seats`, and rewrites every `&`.
+ * Two near-misses to avoid: `sanitize_text_field()` also deletes every `%[a-f0-9]{2}`, which breaks
+ * percent-encoded URLs; `wp_kses( $value, array() )` drops whole `<...>` spans, turning
+ * `Hall < 100 > seats` into `Hall  seats`.
  *
- * The result is encoded for HTML. Decode it before using it in a plain-text medium such as an email
- * body or a CSV column.
+ * The result is HTML-encoded. Decode it for a plain-text medium such as an email body or a CSV column.
  *
- * @param mixed $value The submitted value. Arrays are handled recursively; array keys are left alone,
- *                     and anything that is neither an array nor a scalar becomes an empty string, as
- *                     `sanitize_text_field()` does.
+ * @param mixed $value Arrays are handled recursively, with keys left alone. Anything neither array nor
+ *                     scalar becomes `''`, as `sanitize_text_field()` does.
  *
  * @return string|array A string, or an array of strings when `$value` is an array.
  */
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -404,35 +404,35 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
  * `sanitize_text_field()` and `wp_kses()` disagree about what opens a tag: `strip_tags()`, which the
  * former is built on, needs a letter, `/`, `!` or `?` after the `<`, while kses also accepts whitespace.
  * A value like `< code >` therefore passes the sanitiser untouched, and the `wp_filter_kses()` that core
- * puts on `title_save_pre` and `content_save_pre` then rebuilds it into a real element -- so a handler
- * that meant to store text ends up storing markup. Encoding the `<` that survives settles the
- * disagreement, and leaves the character visible to readers.
+ * puts on `title_save_pre` then rebuilds it into a real element -- so a handler that meant to store text
+ * ends up storing markup. Encoding the `<` that survives settles the disagreement, and leaves the
+ * character visible to readers.
  *
- * `wp_kses( $value, array() )` is not a substitute: it drops whole `<...>` spans, turning
- * `Hall < 100 > seats` into `Hall  seats`, and rewrites every `&`.
+ * This deliberately does not call `sanitize_text_field()`, which would also delete every `%[a-f0-9]{2}`
+ * sequence and so quietly break a percent-encoded URL. `wp_kses( $value, array() )` is not usable either:
+ * it drops whole `<...>` spans rather than neutralising them, turning `Hall < 100 > seats` into
+ * `Hall  seats`, and rewrites every `&`.
  *
  * The result is encoded for HTML. Decode it before using it in a plain-text medium such as an email
  * body or a CSV column.
  *
- * @param string|array $value            The submitted value. Arrays are handled recursively; keys are
- *                                       left alone.
- * @param bool         $keep_line_breaks Whether to keep newlines, as `sanitize_textarea_field()` does.
+ * @param mixed $value The submitted value. Arrays are handled recursively; array keys are left alone,
+ *                     and anything that is neither an array nor a scalar becomes an empty string, as
+ *                     `sanitize_text_field()` does.
  *
  * @return string|array A string, or an array of strings when `$value` is an array.
  */
-function wcorg_sanitize_plain_text( $value, $keep_line_breaks = false ) {
+function wcorg_sanitize_plain_text( $value ) {
 	if ( is_array( $value ) ) {
-		return array_map(
-			function ( $item ) use ( $keep_line_breaks ) {
-				return wcorg_sanitize_plain_text( $item, $keep_line_breaks );
-			},
-			$value
-		);
+		return array_map( 'wcorg_sanitize_plain_text', $value );
 	}
 
-	$value = $keep_line_breaks
-		? sanitize_textarea_field( (string) $value )
-		: sanitize_text_field( (string) $value );
+	if ( ! is_scalar( $value ) ) {
+		return '';
+	}
+
+	$value = wp_check_invalid_utf8( (string) $value );
+	$value = wp_strip_all_tags( $value, true );
 
 	return str_replace( '<', '&lt;', $value );
 }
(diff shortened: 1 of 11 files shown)


bor0:
I read the diff and verified the claims against the code in the tree. The root-cause analysis is correct, and the supporting details hold up:

- `_sanitize_text_fields()` only escapes `<` via `wp_pre_kses_less_than()` when there is no matching `>`, so `< code >` genuinely survives to kses.
- The `esc_html()` / `esc_attr()` non-double-encoding claim holds — `_wp_specialchars()` defaults `$double_encode = false` and passes it to `htmlspecialchars()`.
- All four `post_title` / `post_content` sites in `wordcamp-forms-to-drafts.php` are covered — nothing missed.
- The "meetup and events already `esc_html()`" claim checks out (`class-events-application.php:160`, `class-meetup-application.php:219`).

Five things I would raise.

### 1. The body fields now silently lose percent-encoded sequences

`Company Description`, `Your Bio` and `Topic Description` previously reached `post_content` with no `sanitize_*` pass at all — only `escape_shortcodes()`. They now go through `sanitize_textarea_field()`, which unconditionally deletes every `%[a-f0-9]{2}` match. Running the exact loop from core's `_sanitize_text_fields()`:

```
in : See https://en.wikipedia.org/wiki/Caf%C3%A9 for more.
out: See https://en.wikipedia.org/wiki/Caf for more.

in : Read https://example.org/My%20Talk%20Notes.pdf
out: Read https://example.org/MyTalkNotes.pdf
```

A pasted percent-encoded URL in a speaker bio is not exotic, and the loss is silent and unrecoverable. If the goal is only to neutralise `<`, the helper could skip the percent-strip:

```php
$value = wp_check_invalid_utf8( (string) $value );
$value = wp_strip_all_tags( $value, ! $keep_line_breaks );
return str_replace( '<', '&lt;', $value );
```

### 2. Markup that previously survived in bodies is now stripped, and the description says the opposite about Jetpack

Jetpack does not run `sanitize_textarea_field()` on these values. `Contact_Form_Plugin::strip_tags()` (`jetpack-forms/src/contact-form/class-contact-form-plugin.php:183`) runs `wp_kses_post()` over every entry in `$all_values`. So `<strong>`, `<em>` and `<a href>` typed into a sponsor description or a speaker bio reached `post_content` intact today, and `wp_strip_all_tags()` now removes them.

That may well be the intent, but it is a bigger behavioural change than "encode any `<` that survived" and is not in the description. Worth stating explicitly so a reviewer can sign off on it.

### 3. The new tests only pass by accident of another plugin's bootstrap

`mu-plugins/tests/bootstrap.php` does not load `3-helpers-misc.php`. The only place that file is required in the harness is `wordcamp-speaker-feedback/tests/bootstrap.php:15`. Because `phpunit-bootstrap.php` loads every suite's bootstrap into one process, the function ends up globally defined and the new tests pass — but `test-3-helpers-misc.php`, the new wcpt test, and the existing `Test_Draft_Content` suite now all depend on a line in an unrelated plugin's bootstrap. Removing it there would fatal three suites.

Adding the require to `mu-plugins/tests/bootstrap.php`, where the file under test lives, would make that dependency explicit.

### 4. No test covers the four forms-to-drafts call sites

The PR argues the wcpt test exists "so the suite notices if a future refactor drops the helper from a call site", but that guard only exists for `WordCamp_Application`. There is already a `Test_Draft_Content` suite (`wordcamp-forms-to-drafts/tests/test-draft-content.php`) doing exactly this shape of assertion for `escape_shortcodes()`, with helpers (`make_submission()`, `assertShortcodeNeutralized()`) that drop straight in. Four short cases there would close the gap — manual step 5 in the description exists precisely because that coverage is missing.

### 5. The added comment in the Campus Connect bridge is inaccurate

```php
// Jetpack only puts these through `sanitize_textarea_field()`, which is not enough on its own
```

`$campus` / `$city` / `$country` come from `find_first_field_matching_label()` in that same file, which applies `sanitize_text_field()` (line 205) — not Jetpack, and not the textarea variant. The `sanitize_textarea_field()` call at line 79 is on `$application_data`, which never feeds the title. Suggested wording: "`find_first_field_matching_label()` applies `sanitize_text_field()`, which is not enough on its own…".

### Nit

`(string) $value` in the helper throws an `Error` on an object without `__toString()`, where the `sanitize_*_field()` it wraps returns `''`. An `if ( ! is_scalar( $value ) ) { return ''; }` before the cast would keep the wrapper no less safe than the thing it wraps.

---

Points 1 and 3 are the ones I would want addressed before merge. Point 2 needs a decision rather than a code change.

KokkieH:
Thanks — all five hold up, and I verified each against the tree rather than taking them on trust. Addressed in b59045a9, as a new commit rather than an amend so the change from the review stays visible.

### 1. Percent-encoded sequences — fixed

Reproduced exactly. The helper no longer goes through `sanitize_text_field()` at all:

```php
$value = wp_check_invalid_utf8( (string) $value );
$value = wp_strip_all_tags( $value, true );

return str_replace( '<', '&lt;', $value );
```

Essentially your suggestion. There's a test case pinning `See https://example.org/My%20Notes.pdf` now.

### 2. Markup in bodies — reverted, and you were more right than the point claims

This one changed my mind about the scope of the PR. Following the value further back: `Contact_Form_Plugin::strip_tags()` is applied to `$all_values` at `class-contact-form.php:3055`, and `wp_kses_post()` **already normalises the spaced form**. Measured on the installed Jetpack:

| submitted | after `Contact_Form_Plugin::strip_tags()` |
|---|---|
| `< code >" data-x="< /code >` | `<code>" data-x="</code>` |
| `< script >alert(1)< /script >` | `alert(1)` |
| `<strong>bold</strong>` | `<strong>bold</strong>` |

So on the body path there was no sanitiser disagreement left to settle — the spaced form has already been rebuilt into ordinary markup, constrained to the `wp_kses_post()` allow-list, and `content_save_pre` applies it again on save. Turning `< code >` into `<code>` there produces an element that was already permitted, so the pass bought no integrity and cost submitters their formatting plus any percent-encoded URL.

All three bodies are reverted to `escape_shortcodes()` only. A test now pins that `<strong>` and `My%20Notes.pdf` both survive into `post_content`.

That also corrects the PR description, which claimed Jetpack ran `sanitize_textarea_field()` over these. It doesn't, and the description has been rewritten.

The same trace changes *why* the Forms to Drafts titles need fixing: since Jetpack hands them over already kses'd, plain `<strong>` reaches `post_title` today without the spaced form being involved at all. The fix is the same, but the description no longer implies those four sites share the application handlers' mechanism.

### 3. Bootstrap — fixed

`mu-plugins/tests/bootstrap.php` now requires `3-helpers-misc.php` directly.

One thing worth recording: I removed the `wordcamp-speaker-feedback` require to confirm the dependency, and that suite fatals on its own with `Call to undefined function WordCamp\SpeakerFeedback\wcorg_skip_feature()`. So that line is load-bearing for speaker-feedback itself, not an accidental prop — but the mu-plugins suite no longer leans on it either way.

### 4. Call-site coverage — added

Four cases in `Test_Draft_Content` using the existing `make_submission()` helper, covering the sponsor, volunteer, speaker and session titles, plus the body-preservation case from point 2. I checked they fail for the right reason rather than only that they pass: removing the helper from the sponsor and session titles produces 2 failures.

### 5. Campus Connect comment — fixed

Reworded to `find_first_field_matching_label()` applies `sanitize_text_field()`, which is not enough on its own…`.

### Nit — taken

Added `if ( ! is_scalar( $value ) ) { return ''; }`. Confirmed first that `sanitize_text_field( new stdClass() )` returns `''` rather than throwing, so the wrapper now matches what it wraps. There's a test case for it.

Also dropped `$keep_line_breaks`, which lost its only callers when the bodies were reverted.

---

Full suite is green at 873 tests / 1983 assertions, 0 failures (the 6 skips are the pre-existing `intl` and GatherPress data-provider ones). `phpcs` reports nothing new; the remaining hits on the touched files are all on pre-existing lines.

KokkieH:
@bor0 addressed your comments, but then ran another local agent review which raised more points, but mostly related to test coverage and docblock accuracy.

bor0 (approved):
Re-reviewed, all five points and the nit are addressed. Nice catch on `3476b899` — my suggested snippet dropped `wp_pre_kses_less_than()`, and `strip_tags( 'Rated <A best' )` returns `'Rated '`.

Probed the helper against `sanitize_text_field()` through `wp_filter_kses()`: identical except the two intended changes, idempotent throughout. 305 tests green, `phpcs-changed` clean.

Two non-blocking nits: `[wordcamp_name]` (`wcor-mailer.php:395`) belongs in the "stored values" list since those mails are `text/plain`; and `MARKUP_INPUT` is the pre-Jetpack spaced form, so the production-shaped case (a plain `<strong>`) isn't covered.

LGTM.

PR #1945, opened 2026-08-25
Blocks: refresh the Latest Posts list without ServerSideRender

mcliwanow:
## Why

A `core/latest-posts` block with live update on renders its list server-side, then loses it a second later. `render-frontend.js` mounts the component into the container, `ServerSideRender` throws, and the crash boundary replaces the posts with "There was an error trying to render this content". Two live examples:

- [leipzig.wordcamp.org/2025/day-of-event/](https://leipzig.wordcamp.org/2025/day-of-event/) serves 5 items, shows 0
- [events.wordpress.org/campusconnect/2026/kolkata/day-of-event/](https://events.wordpress.org/campusconnect/2026/kolkata/day-of-event/) serves 2, shows 0

34 sites on the network have a page carrying the block. Three separate things are wrong, and the first one hides the other two.

**`ServerSideRender` is an editor component.** The [handbook](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-server-side-render/) describes it as rendering a preview of dynamic blocks in the editor, and calls it a legacy fallback not appropriate for new development. The blocker is in the code rather than the guidance: since 2.0 it calls `__experimentalSanitizeBlockAttributes()`, which looks the block up with `getBlockType()` and throws when it is missing from the client-side registry. Nothing registers core blocks outside the editor, so it throws before it fetches. 1.9.0 (April 2020) has no such call and 2.0.0 (May 2021) does, so this stopped working when core picked that up, around WP 5.8. The npm pin in `package.json` never mattered, since `@wordpress/*` imports are externalised to core's bundled scripts.

**The renderer route rejects attributes the render path still honours.** `postLayout` and `columns` moved to the layout support and are gone from the block's schema, but `render_block_core_latest_posts()` still reads them. The two WordPress copies in this repo bracket the change: the phpunit suite runs 7.0.4, which declares both, and the site runs 7.2-alpha, which declares neither. A block saved with them draws its grid and then fails every poll. `slovenia.wordcamp.org/2026` is one of them:

```
attributes[postLayout]=grid   400  "postLayout is not a valid property of Object."
the same request without it   200
```

**The nonce expires and nothing recovers it.** `apiFetch` attaches `X-WP-Nonce`. For a logged-out visitor that nonce dies after 12 to 24 hours, and `rest_cookie_check_errors()` refuses an invalid one before the request is routed, so the safelist never runs. api-fetch then tries `admin-ajax.php?action=rest-nonce`, which core registers for logged-in users only. Measured against `us.wordcamp.org/2019`:

```
no nonce      200
stale nonce   403  rest_cookie_invalid_nonce
rest-nonce    400
```

A page served from a cache older than a day is born in that state, which is the hallway-screen case the feature exists for.

## What changed

- `front-end.js` drops `ServerSideRender` and requests the renderer route itself, so nothing looks the block up in a registry it was never added to. The block needs a URL and a fetch, not an editor preview component.
- The response is written straight into the container, so a refreshed list has the same shape the server rendered. That is the markup any theme or custom CSS was written against, and it leaves nothing interposed for the layout rules to land on: the block gap and the grid both address direct children.
- Nothing is fetched until the interval fires, so the list on screen is the server's until there is something newer to show. A failed or refused request leaves it alone.
- The request goes out through `window.fetch`, against a root-relative endpoint `add_script_data()` localises from `rest_url()`. No nonce, so there is nothing to expire, and no absolute host, so it cannot turn cross-origin.
- `render()` keeps unregistered attributes out of `data-attributes`. They only feed container classes and the container on the page keeps its own, so a block saved with `postLayout` keeps its grid and starts polling again.
- Attributes that cannot be read, or are not there at all, stop the refresh rather than starting one. Asking with no attributes would render the block's defaults over a correctly configured list.
- A failure is reported once through `console.warn`, so a list that stops updating can be diagnosed instead of going quiet.

- The list is only rewritten when the markup actually changed. Rebuilding it drops focus and any selection inside it and restarts the images, and most polls return what is already on screen.
- Polling stops when the container leaves the page, when the endpoint is missing, and after three refused requests. A refusal repeats by construction, so retrying it forever is pointless; a dropped connection is not, so those keep retrying and a screen left running unattended survives a bad afternoon.
- The safelisted response carries `Cache-Control: public, s-maxage=60`. Everyone on a page polls the same URL, and a minute is short against the five between polls, so a shared cache can absorb most of the traffic without adding much to how stale the list gets. It is the only response on that route marked cacheable: an authorised render, and anything the guards refused, are not.

Dropping the component takes React with it, and with no strings left to translate the script no longer sets translations either. A page carrying the block went from loading `react-jsx-runtime`, `wp-element`, `wp-i18n`, `wp-hooks` and `wp-server-side-render` to loading `wp-url` alone, with the built file at 947 bytes.

## Test plan

- [x] `--group=blocks` 29/29, and a full `phpunit` run clean apart from the pre-existing skips
- [x] `lint:js` clean, `wp-scripts test-unit-js` 23/23
- [x] `phpcs` clean on the changed lines (the one error it reports, in `check_version_support()`, is on `production` too and is not touched here)
- [x] Browser pass on a local multisite camp, steps below

Two new tests, plus a fix to the suite's environment. The suite loads `controller.php` directly, after `init` has fired, so the hook's own registration never ran and the block's schema was missing `liveUpdateEnabled`. `set_up` now adds it and `tear_down` removes it, which is what lets a test send the attribute the front end actually sends.

`test_unregistered_attributes_are_not_polled` uses a name core will never declare rather than `postLayout`, because the phpunit WordPress still registers `postLayout` and the site does not. `test_polled_attributes_are_accepted_by_the_route` closes the loop the change exists for: it takes the `data-attributes` payload the filter wrote and sends it back through `rest_do_request()`, which fails without the filtering.

### Manual test steps

Two pages on a local camp site, one with plain attributes and one carrying the legacy pair.

```bash
docker compose exec -T wordcamp.test wp --url=seattle.wordcamp.test/2023/ eval '
$a = wp_insert_post( array( "post_type"=>"page", "post_status"=>"publish", "post_title"=>"Modern",
 "post_content"=>"<!-- wp:latest-posts {\"displayPostDate\":true,\"liveUpdateEnabled\":true} /-->" ) );
$b = wp_insert_post( array( "post_type"=>"page", "post_status"=>"publish", "post_title"=>"Legacy",
 "post_content"=>"<!-- wp:latest-posts {\"postsToShow\":3,\"postLayout\":\"grid\",\"columns\":3,\"liveUpdateEnabled\":true} /-->" ) );
foreach ( array( "news one", "news two" ) as $t ) {
  wp_insert_post( array( "post_type"=>"post", "post_status"=>"publish", "post_title"=>$t, "post_content"=>"body" ) );
}
echo get_permalink( $a ) . "\n" . get_permalink( $b ) . "\n";'
```

Load each page logged out and watch the list. On `production` the posts are replaced by the error string within a second and the console carries `Block type 'core/latest-posts' is not registered`. Here the list stays, the console is quiet, and one request to the renderer route goes out on load.

Driven with Playwright on both pages:

|  | `production` | this branch |
| --- | --- | --- |
| list items after mount | 0 | 4 and 3 |
| console | `Block type 'core/latest-posts' is not registered.` | none |
| `X-WP-Nonce` sent | yes | none |
| renderer request on load | none | 200 |

The legacy page keeps `is-grid columns-3` on its container while its `data-attributes` payload comes back as `{"postsToShow":3,"liveUpdateEnabled":true}`.

For the interval itself, install a fake clock, publish a post with the page open, and move five minutes on. The new post arrives at the top of the list and both requests return 200.

#### Before

image

image

#### After

image

image

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
@@ -122,40 +122,6 @@ function safelist_block_renderer( $response, $handler, $request ) {
 add_filter( 'rest_request_after_callbacks', __NAMESPACE__ . '\safelist_block_renderer', 10, 3 );
 
 /**
- * Whether the markup is one `<ul>` container spanning the whole string.
- *
- * The tag swap below renames the first `<ul` and the last `</ul>`, so both have to
- * belong to the same element. A sibling list would otherwise leave
- * `<div>…</ul><ul>…</div>`.
- *
- * The nesting count reads the raw string. That is safe because `esc_attr()` escapes
- * `<`, so `<ul` and `</ul` cannot appear inside an attribute value. If that stops
- * being true the count goes wrong and the rewrite silently stops attaching.
- *
- * @param string $markup Rendered block markup.
- * @return bool
- */
-function is_swappable_container( $markup ) {
-	if ( ! preg_match( '#\A\s*<ul\b#', $markup ) || ! preg_match( '#</ul>\s*\z#', $markup ) ) {
-		return false;
-	}
-
-	preg_match_all( '#<(/?)ul\b#', $markup, $matches );
-	$depth = 0;
-
-	foreach ( $matches[1] as $index => $slash ) {
-		$depth += '' === $slash ? 1 : -1;
-
-		// The container closed, so it has to be the last `</ul>` in the markup.
-		if ( 0 === $depth ) {
-			return count( $matches[1] ) - 1 === $index;
-		}
-	}
-
-	return false;
-}
-
-/**
  * Drop the attributes the renderer route will not accept.
  *
  * The route validates `attributes` against the registered schema and rejects anything
@@ -177,7 +143,16 @@ function pollable_attributes( $attrs ) {
 		return $attrs;
 	}
 
-	return array_intersect_key( $attrs, $block_type->attributes );
+	$registered = array_intersect_key( $attrs, $block_type->attributes );
+
+	// A null serialises into the query string as an empty value, which the route
+	// refuses for anything the schema does not type as a string.
+	return array_filter(
+		$registered,
+		static function ( $value ) {
+			return null !== $value;
+		}
+	);
 }
 
 /**
@@ -199,7 +174,7 @@ function render( $block_content, $block ) {
 	$enabled = isset( $block['attrs']['liveUpdateEnabled'] ) && $block['attrs']['liveUpdateEnabled'];
 	// Order by date, desc is the default, so these properties are not set.
 	$order_date_desc = ! isset( $block['attrs']['orderBy'] ) && ! isset( $block['attrs']['order'] );
-	if ( $enabled && $order_date_desc && is_swappable_container( $block_content ) ) {
+	if ( $enabled && $order_date_desc ) {
 		/*
 		 * Rewrite the container through the HTML API, not by string content:
 		 * `str_replace()` also matched its needle inside attribute values (the
@@ -215,15 +190,6 @@ function render( $block_content, $block ) {
 			$processor->set_attribute( 'data-attributes', rawurlencode( wp_json_encode( pollable_attributes( $block['attrs'] ) ) ) );
 
 			$block_content = $processor->get_updated_html();
-
-			/*
-			 * `render-frontend.js` renders the polled markup into this element, and that
-			 * markup has been through this filter too, so it arrives as a `<div>`. A
-			 * `<ul>` cannot hold it. Anchored to the ends of the string, where an
-			 * attribute value cannot match.
-			 */
-			$block_content = preg_replace( '#\A(\s*)<ul\b#', '$1<div', $block_content, 1 );
-			$block_content = preg_replace( '#</ul>(\s*)\z#', '</div>$1', $block_content, 1 );
 		}
 	}
 
--- a/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
@@ -3,6 +3,8 @@ namespace WordCamp\Blocks\Hooks\Latest_Posts;
 
 defined( 'WPINC' ) || die();
 
+const POLL_CACHE_SECONDS = 60;
+
 /**
  * Check editor versions.
  *
@@ -117,7 +119,18 @@ function safelist_block_renderer( $response, $handler, $request ) {
 		return $response;
 	}
 
-	return call_user_func( $handler['callback'], $request );
+	$response = call_user_func( $handler['callback'], $request );
+
+	/*
+	 * Everyone on a page polls the same URL, so let a cache answer most of them. A
+	 * minute is short against the five the block waits between polls, and core only
+	 * sends no-cache headers for a logged-in request, which this never is.
+	 */
+	if ( $response instanceof \WP_REST_Response ) {
+		$response->header( 'Cache-Control', 'max-age=' . POLL_CACHE_SECONDS );
+	}
+
+	return $response;
 }
 add_filter( 'rest_request_after_callbacks', __NAMESPACE__ . '\safelist_block_renderer', 10, 3 );
 
--- a/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
@@ -57,8 +57,6 @@ function init() {
 		'before'
 	);
 
-	wp_set_script_translations( 'wordcamp-live-posts', 'wordcamporg' );
-
 	$block_type = \WP_Block_Type_Registry::get_instance()->get_registered( 'core/latest-posts' );
 	if ( $block_type ) {
 		unregister_block_type( $block_type->name );
@@ -158,6 +156,31 @@ function is_swappable_container( $markup ) {
 }
 
 /**
+ * Drop the attributes the renderer route will not accept.
+ *
+ * The route validates `attributes` against the registered schema and rejects anything
+ * outside it, while the render path still honours older keys such as `postLayout`.
+ * Those only feed container classes, and the container on the page keeps its own, so
+ * dropping them here costs nothing and keeps the request valid.
+ *
+ * This filters keys, not values. A registered key with a stale value can still be
+ * refused, `categories` saved as a string being the one that used to be: it survives
+ * because core migrates it on `render_block_data`, before this filter runs.
+ *
+ * @param array $attrs Saved block attributes.
+ * @return array
+ */
+function pollable_attributes( $attrs ) {
+	$block_type = \WP_Block_Type_Registry::get_instance()->get_registered( 'core/latest-posts' );
+
+	if ( ! $block_type || empty( $block_type->attributes ) ) {
+		return $attrs;
+	}
+
+	return array_intersect_key( $attrs, $block_type->attributes );
+}
+
+/**
  * Filter the content of the latest posts block.
  *
  * @param string $block_content The block content about to be appended.
@@ -189,7 +212,7 @@ function render( $block_content, $block ) {
 		if ( $processor->next_tag() && 'UL' === $processor->get_tag() ) {
 			$processor->add_class( 'has-live-update' );
 			$processor->add_class( 'is-loading' );
-			$processor->set_attribute( 'data-attributes', rawurlencode( wp_json_encode( $block['attrs'] ) ) );
+			$processor->set_attribute( 'data-attributes', rawurlencode( wp_json_encode( pollable_attributes( $block['attrs'] ) ) ) );
 
 			$block_content = $processor->get_updated_html();
 
@@ -208,7 +231,6 @@ function render( $block_content, $block ) {
 }
 add_filter( 'render_block', __NAMESPACE__ . '\render', 10, 2 );
 
-
 /**
  * Add data to be used by the JS scripts in the block editor.
  *
@@ -218,7 +240,10 @@ add_filter( 'render_block', __NAMESPACE__ . '\render', 10, 2 );
  */
 function add_script_data( array $data ) {
 	if ( check_version_support() ) {
-		$data['latest-posts'] = array();
+		$data['latest-posts'] = array(
+			// Root-relative, so the request stays same-origin on a non-canonical host.
+			'renderer' => wp_make_link_relative( rest_url( 'wp/v2/block-renderer/core/latest-posts' ) ),
+		);
 	}
 
 	return $data;
--- a/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
@@ -122,12 +122,16 @@ function safelist_block_renderer( $response, $handler, $request ) {
 	$response = call_user_func( $handler['callback'], $request );
 
 	/*
-	 * Everyone on a page polls the same URL, so let a cache answer most of them. A
-	 * minute is short against the five the block waits between polls, and core only
-	 * sends no-cache headers for a logged-in request, which this never is.
+	 * Everyone on a page polls the same URL, so let a shared cache answer most of
+	 * them. A minute is short against the five the block waits between polls, and core
+	 * only sends no-cache headers for a logged-in request, which this never is.
+	 *
+	 * `s-maxage` because all of the value is in a shared cache: a browser's own copy
+	 * expires long before that browser polls again. And every guard above has to have
+	 * run first, since this is the one response here that is safe to hold.
 	 */
 	if ( $response instanceof \WP_REST_Response ) {
-		$response->header( 'Cache-Control', 'max-age=' . POLL_CACHE_SECONDS );
+		$response->header( 'Cache-Control', 'public, s-maxage=' . POLL_CACHE_SECONDS );
 	}
 
 	return $response;
@@ -197,7 +201,16 @@ function render( $block_content, $block ) {
 		 */
 		$processor = new \WP_HTML_Tag_Processor( $block_content );
 
-		if ( $processor->next_tag() && 'UL' === $processor->get_tag() ) {
+		// By what it is, not by what comes first: the markup can hold a list belonging
+		// to something else, and a container of ours can sit inside a group.
+		$found = $processor->next_tag(
+			array(
+				'tag_name'   => 'UL',
+				'class_name' => 'wp-block-latest-posts',
+			)
+		);
+
+		if ( $found ) {
 			$processor->add_class( 'has-live-update' );
 			$processor->add_class( 'is-loading' );
 			$processor->set_attribute( 'data-attributes', rawurlencode( wp_json_encode( pollable_attributes( $block['attrs'] ) ) ) );
--- a/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
+++ b/public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
@@ -1,61 +1,159 @@
 /**
  * WordPress dependencies
  */
-import { __ } from '@wordpress/i18n';
-import { Component } from '@wordpress/element';
-import ServerSideRender from '@wordpress/server-side-render';
+import { addQueryArgs } from '@wordpress/url';
+
+const SELECTOR = '.wp-block-latest-posts.has-live-update';
+const REFRESH_INTERVAL = 5 * 60 * 1000; // 5 minutes in milliseconds.
 
 /**
- * Internal dependencies
+ * The endpoint that renders this block, localised by the controller.
+ *
+ * @return {string} The endpoint, or an empty string when the data is missing.
  */
-import renderFrontend from '../../utils/render-frontend';
+function rendererUrl() {
+	const data = window.WordCampBlocks || {};
 
-class LivePosts extends Component {
-	constructor( props ) {
-		super( props );
-		this.renderInterval = setInterval(
-			() => {
-				// `forceUpdate` is a React internal that triggers a render cycle.
-				this.forceUpdate();
-			},
-			5 * 60 * 1000 // 5 minutes in milliseconds.
-		);
+	return ( data[ 'latest-posts' ] || {} ).renderer || '';
+}
+
+/**
+ * Read the block's saved attributes off the container.
+ *
+ * @param {Element} element Container element.
+ * @return {Object|null} The attributes, or null when they cannot be read.
+ */
+function readAttributes( element ) {
+	const encoded = element.dataset.attributes;
+
+	if ( ! encoded ) {
+		return null;
 	}
 
-	componentWillUnmount() {
-		clearInterval( this.renderInterval );
-	}
+	try {
+		const parsed = JSON.parse( decodeURIComponent( encoded ) );
 
-	render() {
-		const { attributes } = this.props;
-		// Remove the helper data-attribute.
-		delete attributes.attributes;
-
-		// Note: the `LoadingResponsePlaceholder` is intentionally an anonymous function to trigger re-render when
-		// `forceUpdate` happens.
-		return (
-			<ServerSideRender
-				block="core/latest-posts"
-				attributes={ attributes }
-				LoadingResponsePlaceholder={ () => (
-					<p>{ __( 'Loading', 'wordcamporg' ) }</p>
-				) }
-			/>
-		);
+		// Anything but a plain object would ask the route for the block's defaults.
+		return parsed && 'object' === typeof parsed && ! Array.isArray( parsed )
+			? parsed
+			: null;
+	} catch {
+		return null;
 	}
 }
 
-const getAttributesFromData = ( element ) => {
-	let parsedAttributes = {};
-	const { attributes } = element.dataset;
-	if ( attributes ) {
-		parsedAttributes = JSON.parse( decodeURIComponent( attributes ) );
-	}
-	return { attributes: parsedAttributes };
-};
+/**
+ * Take the list items out of a rendered response.
+ *
+ * The response has been through the `render_block` filter too, so it arrives wearing a
+ * container of its own. The container already on the page is the one carrying the
+ * block's classes, so only its contents are replaced.
+ *
+ * @param {string} html Rendered block markup.
+ * @return {string} The container's contents, or the markup unchanged if it has none.
+ */
+function unwrapContainer( html ) {
+	const template = document.createElement( 'template' );
+	template.innerHTML = html.trim();
 
-renderFrontend(
-	'.wp-block-latest-posts.has-live-update',
-	LivePosts,
-	getAttributesFromData
-);
+	const container = template.content.querySelector(
+		'.wp-block-latest-posts'
+	);
+
+	return container ? container.innerHTML : html;
+}
+
+/**
+ * Keep one container's list up to date.
+ *
+ * The markup is written straight into the container, so a refreshed list has the same
+ * shape as the one the server rendered and any styling written against that markup
+ * still applies. A failed request leaves the page alone.
+ *
+ * @param {Element} element Container element.
+ */
+function start( element ) {
+	const attributes = readAttributes( element );
+
+	element.classList.remove( 'is-loading' );
+
+	/*
+	 * Without the saved attributes a refresh would ask for the block's default
+	 * settings, replacing a correct list with a wrong one. Leave the page's own.
+	 */
+	if ( ! attributes ) {
+		return;
+	}
+
+	let warned = false;
+	let lastContent = null;
+
+	/**
+	 * Report the first failure, so a list that stops updating can be diagnosed.
+	 *
+	 * @param {string} reason What went wrong.
+	 */
+	function warn( reason ) {
+		if ( ! warned ) {
+			warned = true;
+			// eslint-disable-next-line no-console
+			console.warn( 'Latest Posts live update failed:', reason );
+		}
+	}
+
+	function refresh() {
+		// The container can be taken off the page by something else on the site. Stop
+		// rather than keep polling into a node nobody can see.
+		if ( ! element.isConnected ) {
+			clearInterval( timer );
+			return;
+		}
+
+		const endpoint = rendererUrl();
+
+		if ( ! endpoint ) {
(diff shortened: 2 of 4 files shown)


bor0 (commented):
Reviewed this locally: pulled the branch, ran the suite, built it, and probed the endpoint against a local multisite.

### Claims I verified

| Claim | Result |
| --- | --- |
| `--group=blocks` 28/28 | ✅ `OK (28 tests, 50 assertions)` |
| `lint:js` clean | ✅ |
| Build drops React, `wp-url` only | ✅ `production` emits `react-jsx-runtime, wp-element, wp-i18n, wp-server-side-render`; this branch emits `wp-url` alone, 1.0 KiB |
| Anonymous 200 from the renderer route | ✅ curl with the exact query shape `addQueryArgs` produces |
| Root-relative endpoint | ✅ resolves to `/2023/wp-json/wp/v2/block-renderer/core/latest-posts` on a subdirectory site |
| `pollable_attributes()` matches what the route validates | ✅ `WP_Block_Type::get_attributes()` returns `$this->attributes` verbatim, and the route's validator uses `additionalProperties => false` |
| `context=edit` is required | ✅ the item schema marks `rendered` as `'context' => array( 'edit' )`, so `view` would strip it |

One premise I could not reproduce: my local WP is 7.1-beta2, which still declares `postLayout` and `columns`, so the 400 does not happen here. The filtering is correct and harmless on both versions, and the new test wisely uses a name core will never register so it holds either way — but the production-only half of the diagnosis rests on your measurement rather than anything I could confirm.

Nothing below is a correctness blocker, and this is a clear improvement over a block that currently crashes on 34 sites. Line comments carry the specifics; two more that do not anchor to a line:

### The renderer response carries no cache headers

Confirmed locally — the anonymous response sends no `Cache-Control`, no `Expires`, no `Last-Modified`. This is *new* traffic: today the request never fires, because SSR throws before it gets there. Every visitor with the block open past five minutes now generates an uncacheable dynamic REST hit, across 34 sites. Probably fine at this volume, but worth answering deliberately rather than by omission — a short `max-age` sent from the safelist would let the edge absorb the polling, and the hallway-screen case is exactly the one that benefits.

### No JS tests for the new module

The package has jest wired up already (`source/**/tests/index.test.js`). `readAttributes()` and `unwrapContainer()` are pure functions that encode precisely the contract this PR exists to establish: a non-object payload must not poll, and the response's own container must be stripped before the markup is written. Two small tests would hold that in place — the PHP side has the loop covered end to end, and the JS side has nothing.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
```diff
@@ -189,7 +212,7 @@ function render( $block_content, $block ) {
 		if ( $processor->next_tag() && 'UL' === $processor->get_tag() ) {
 			$processor->add_class( 'has-live-update' );
 			$processor->add_class( 'is-loading' );
-			$processor->set_attribute( 'data-attributes', rawurlencode( wp_json_encode( $block['attrs'] ) ) );
+			$processor->set_attribute( 'data-attributes', rawurlencode( wp_json_encode( pollable_attributes( $block['attrs'] ) ) ) );
```
bor0, line 216: The `<ul>` → `<div>` swap just below (lines 218-226) is vestigial after this change, and the comment justifying it now describes code this PR deletes.

It reads *"`render-frontend.js` renders the polled markup into this element, and that markup has been through this filter too, so it arrives as a `<div>`. A `<ul>` cannot hold it."* That was true of ServerSideRender. It is not true here: `unwrapContainer()` strips the response's own container and writes only the `<li>` children, so a `<ul>` holds them fine.

What ships today, rendered locally:

```html
<div data-attributes="…" class="wp-block-latest-posts__list wp-block-latest-posts has-live-update is-loading"><li>…</li>
```

`<li>` as a direct child of a `<div>` is invalid HTML, drops the list semantics assistive tech reads, and breaks any theme or custom CSS written against `ul > li`. The PR's own rationale — *"a refreshed list has the same shape the server rendered"* — argues for dropping the swap, since the container itself is still being reshaped.

Removing it is the better fix. If you would rather keep this PR tight, the comment has to change regardless, and the swap is worth a follow-up.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
```diff
@@ -1,61 +1,159 @@
...
+		// The container can be taken off the page by something else on the site. Stop
+		// rather than keep polling into a node nobody can see.
+		if ( ! element.isConnected ) {
+			clearInterval( timer );
+			return;
+		}
+
+		const endpoint = rendererUrl();
+
+		if ( ! endpoint ) {
```
bor0, line 120: This branch warns once and then keeps running forever, re-reading a constant every five minutes with no fetch and no exit. `clearInterval( timer )` belongs here — if the endpoint is missing on the first poll it will be missing on every one after.

It is a real state, not a hypothetical. `WordCampBlocks` is inlined into the page HTML, so a page cached before the deploy carries the old payload with no `renderer` key, while the versioned script URL still serves the new file. Every such page lands in exactly this branch until the cache turns over.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
```diff
@@ -1,61 +1,159 @@
...
+				 * what is already on screen, so only write when something changed.
+				 */
+				if ( content === lastContent ) {
+					return;
+				}
+
+				lastContent = content;
+				element.innerHTML = content;
+			} )
+			.catch( ( error ) => warn( error.message ) );
```
bor0: Same shape as the missing-endpoint branch: the first failure is reported and the polling continues indefinitely.

The companion cached-page case is a request that keeps being refused — HTML cached before the deploy carries a `data-attributes` payload written by the old, unfiltered `render()`, so it still has `postLayout` and draws a 400 on every poll until the cache expires.

A consecutive-failure counter that clears the interval after two or three would bound both. Nothing user-visible goes wrong either way — the server-rendered list stays on screen, which is already the point — but a failure that repeats on a timer forever is worth closing while you are here.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
```diff
@@ -1,61 +1,159 @@
...
+			return;
+		}
+
+		/*
+		 * Deliberately not `apiFetch`: it attaches an `X-WP-Nonce`, which expires for a
+		 * logged-out visitor, and an expired nonce is refused before the request is
+		 * routed. Sending none leaves the request anonymous, which is what this needs.
+		 */
+		window
+			.fetch( addQueryArgs( endpoint, { context: 'edit', attributes } ) )
```
bor0, line 134: A `null` attribute value serialises to an empty string and the route refuses it. Checked against the real `addQueryArgs`:

```
addQueryArgs( '/x/', { attributes: { foo: null } } )
→ /x/?attributes%5Bfoo%5D=
```

For any non-string type in the schema that fails validation, so the block renders once and never updates — and per the comment on the `.catch`, silently and forever.

Narrow, since the editor usually drops an attribute rather than nulling it. But `pollable_attributes()` is already the place where attributes the route would refuse get dropped, so nulls belong in the same pass rather than in a second one later.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
```diff
@@ -1,61 +1,159 @@
...
+	/*
+	 * Without the saved attributes a refresh would ask for the block's default
+	 * settings, replacing a correct list with a wrong one. Leave the page's own.
+	 */
+	if ( ! attributes ) {
+		return;
+	}
+
+	let warned = false;
+	let lastContent = null;
```
bor0, line 95: Minor: seeded to `null`, so the first poll always rewrites the list even when nothing changed. The comment below it — focus, selection, restarted images — describes a cost that is avoided from the second poll on, not from the first.

Seeding from `element.innerHTML` would not work reliably (whitespace differs between the page's copy and the response), so this may just be the honest ceiling. Worth softening the comment if so.

public_html/wp-content/mu-plugins/tests/test-blocks-latest-posts.php
```diff
@@ -253,6 +294,59 @@ public function test_container_still_gets_live_update_markup() {
...
+		$output = $this->render_container(
+			'<ul class="wp-block-latest-posts is-grid"><li>a</li></ul>',
+			array(
+				'postsToShow'         => 3,
+				'wcorgRetiredSetting' => 'grid',
+				'liveUpdateEnabled'   => true,
+			)
+		);
+
+		preg_match( '/data-attributes="([^"]*)"/', $output, $matches );
```
bor0: `$matches[1]` is read without asserting the match first, here and at line 342. If a regression stops `render()` writing `data-attributes` at all — the thing both tests are built on — this fails as `Undefined array key 1` rather than saying what broke.

`$this->assertSame( 1, preg_match( … ), 'the container carries no data-attributes.' );` keeps the failure legible.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
mcliwanow: Right on both counts, and I had missed that the comment describes code this PR deletes. `render-frontend.js` is not imported any more, and `unwrapContainer()` strips the response's container and writes only the items, which a `<ul>` holds.

Dropped the swap. That takes `is_swappable_container()` with it: it existed to prove the first `<ul` and the last `</ul>` were the same element, which is meaningless once nothing is renamed. 45 lines out, 11 in.

It also closes something I filed against #1917 and never fixed. A sibling list, or anything either side of the container, used to fail that check and opt out of live update silently. Those get marked now, which `test_markup_around_the_container_still_gets_marked` pins.

Checked in a browser: the container stays a `UL`, its children are `LI` only, refreshed content lands, and `ul.wp-block-latest-posts > li` matches again.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
mcliwanow: Fixed, `clearInterval` there now. The cached-page reasoning is the part I had not considered: the inline data sits in the HTML while the script URL is versioned, so old HTML plus new script lands in exactly that branch and stays there.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
mcliwanow: Added a consecutive-failure budget that stops after three.

Worth recording that the first version of this was wrong: I reset the counter where the content is written, and since most polls return unchanged content and return early, it never reset. Two blips twenty minutes apart plus a third would have retired a healthy poller. It resets on any answered request now.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
mcliwanow: Good catch, and you are right that `pollable_attributes()` is the place. Nulls are filtered there now, in the same pass as the unregistered keys.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
mcliwanow: Correct, and it is the honest ceiling. The page's copy has been through the parser and is not byte-identical to the response, so it cannot seed the comparison. Softened the comment to say the saving starts from the second poll.

public_html/wp-content/mu-plugins/tests/test-blocks-latest-posts.php
mcliwanow: Done, both places.

mcliwanow:
Thanks for pulling the branch and running it rather than reading it. Everything is in, in `a1a081b4` and `ec6915f0`.

**Cache headers.** Went with `max-age=60` on the safelisted response. Verified it reaches the wire: an anonymous poll returns `cache-control: max-age=60`, and a request that is not safelisted still 401s without one. Core only sends no-cache headers when `is_user_logged_in()`, and this request carries no nonce, so it is anonymous and the header stands. A minute is short against the five between polls, so it barely moves staleness while still collapsing a crowd on one page onto a single origin hit.

**JS tests.** Added `source/hooks/latest-posts/tests/index.test.js`, eight tests over `readAttributes()` and `unwrapContainer()`, both now exported. It covers the contract you named: a payload that is not a plain object returns null so it cannot poll, and the response's own container is stripped before anything is written. Suite is 3 files, 23 tests.

**On 7.1-beta2.** Fair flag, and nothing to fix. Your WP still declares `postLayout`, so the 400 cannot happen there; it is the newer schema that drops it. The version boundary is in the description, and the new test uses a name core will never register so it holds either side of the change.

Removing the swap turned out to be the most valuable comment of the set. It deleted more than it added and closed an edge case left over from #1917.

bor0 (commented):
Re-reviewed at ec6915f0. All six are fixed, and I checked each one rather than taking the replies for it.

| Round 1 | Fix | Verified |
| --- | --- | --- |
| `<ul>` → `<div>` swap, and its comment describing deleted code | swap and `is_swappable_container()` both gone | container renders as `<ul …>` with `<li>` children, and the response's own `<ul>` unwraps to its items |
| missing endpoint polls forever | `clearInterval` | ✅ |
| repeated failure polls forever | budget of 3, reset on any answered request | ✅ the reset is on the answered path, not the written one |
| a `null` attribute value | filtered in `pollable_attributes()` | ✅ |
| `lastContent` seeded to `null` | comment softened | ✅ |
| `$matches[1]` read unasserted | `assertSame( 1, preg_match( … ) )`, both places | ✅ |

Re-ran on the branch:

| | |
| --- | --- |
| `phpunit --group=blocks` | `OK (28 tests, 56 assertions)` |
| `wp-scripts test-unit-js` | 23/23 across 3 suites |
| `lint-js source/hooks/latest-posts` | clean |
| `wp-scripts build` | clean, `live-posts` deps are `array( 'wp-url' )`, 1131 bytes |
| renderer route, anonymous | `200`, `cache-control: max-age=60` |
| `render()` against a real page | `<ul data-attributes="…" class="… wp-block-latest-posts has-live-update is-loading">` |

Four things below. The first is a bug I can reproduce; the rest are the two new commits' loose ends.

Nit while you are in the file: `check_version_support()`'s docblock still opens with *"we need the `wp.serverSideRender` component available"*. That component is the one this PR stops using.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
```diff
@@ -189,26 +200,16 @@ function render( $block_content, $block ) {
 		if ( $processor->next_tag() && 'UL' === $processor->get_tag() ) {
```
bor0: `next_tag()` takes the **first** tag in `$block_content`, and nothing checks that it belongs to this block. Dropping `is_swappable_container()` was right, but that function did carry one thing nothing here replaces: it refused markup where the block's list was not the whole string.

Run against a preceding sibling list, on this branch:

```php
render( '<ul class="other"><li>b</li></ul><ul class="wp-block-latest-posts"><li>a</li></ul>', … )
```
```html
<ul data-attributes="%7B%22liveUpdateEnabled%22%3Atrue%7D" class="other has-live-update is-loading"><li>b</li></ul><ul class="wp-block-latest-posts"><li>a</li></ul>
```

The stranger's list takes the marking. It does not match `SELECTOR`, so `start()` never runs on it and `is-loading` stays on it for the life of the page — while the block's own list, unmarked, never polls at all.

`test_markup_around_the_container_still_gets_marked` reads as though it covers this. Its `siblings` case is `<ul class="wp-block-latest-posts">…</ul><ul class="other">…</ul>`, which passes because the block's list happens to come first; reversing the two fails it. The docblock's *"whatever sits beside it is irrelevant"* is the claim, and it is only half true.

Addressing the element by what it is says what you mean and fixes it:

```php
if ( $processor->next_tag( array( 'tag_name' => 'UL', 'class_name' => 'wp-block-latest-posts' ) ) ) {
```

Checked against four shapes:

| markup | today | with `class_name` |
| --- | --- | --- |
| `<ul class="…__list wp-block-latest-posts …">` | marked | marked |
| `<ul class="other">` first | wrong list marked | block's list marked |
| `<div class="wp-block-group"><ul class="wp-block-latest-posts">` | not marked | marked |
| `<ul class="wp-block-latest-posts__list">` alone | marked | not marked |

Row three comes free: a wrapped container opts out silently today. Row four is the token match doing its job — `__list` is not `wp-block-latest-posts`.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
```diff
@@ -119,42 +119,53 @@ function safelist_block_renderer( $response, $handler, $request ) {
...
-	return call_user_func( $handler['callback'], $request );
+	$response = call_user_func( $handler['callback'], $request );
+
+	/*
+	 * Everyone on a page polls the same URL, so let a cache answer most of them. A
+	 * minute is short against the five the block waits between polls, and core only
+	 * sends no-cache headers for a logged-in request, which this never is.
+	 */
+	if ( $response instanceof \WP_REST_Response ) {
+		$response->header( 'Cache-Control', 'max-age=' . POLL_CACHE_SECONDS );
```
bor0: This is new since the last round, and it lands in the one branch of this function with no test behind it.

The header itself holds. On the wire, anonymous:

```
cache-control: max-age=60
```

and core defaults `rest_send_nocache_headers` to `is_user_logged_in()`, so nothing competes for the request this is aimed at.

What is not pinned is the **placement**. The header is only correct because every early return above it has already run — the `name` guard, `post_id`, `postsToShow`, and "leave any other objection standing". Move it up one guard and this starts telling caches to hold a Coming Soon 403 for a minute, with the suite still green. `test_other_errors_are_not_discarded` is right there and one assertion away:

```php
$this->assertArrayNotHasKey( 'Cache-Control', $response->get_headers() );
```

plus the positive one on `test_anonymous_can_render_latest_posts`.

Two things around it:

The PR description does not mention this. It changes the response headers of a public endpoint, which is what whoever deploys it would rather read in the body than find in a commit. The test-plan counts have drifted too — `test-unit-js` is 23/23 now, not 15/15.

And the question the change turns on: does the edge in front of wordcamp.org cache `/wp-json/` GETs at all? A browser cache cannot use this — 60s against a five-minute poll never hits, and a reload starts a fresh poller that fetches five minutes later — so every bit of the value is in a shared cache. If it does cache, `s-maxage` and `public` would say so rather than leave it to be inferred.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
```diff
@@ -1,61 +1,183 @@
...
+				warn( error.message );
+
+				/*
+				 * A refusal usually repeats: a page cached before the deploy sends
+				 * attributes the route no longer accepts, and will until it expires.
+				 * Give the network a couple of chances, then stop asking.
+				 */
+				failures++;
+
+				if ( failures >= MAX_FAILURES ) {
```
bor0, line 193: The budget bounds the case I raised. It also bounds one I did not: three consecutive transport failures retire the poller for the life of the page.

The failure I described repeats by construction — a page cached with `postLayout` draws a 400 on every poll until the cache turns over, and no amount of retrying will fix it. A flaky network is the opposite. Fifteen minutes of it and a hallway screen meant to run unattended for three days stops updating, with nothing to bring it back short of someone walking over to reload it. That screen is the case the feature exists for.

The two are distinguishable where the error is raised:

```js
if ( ! response.ok ) {
	const error = new Error( `HTTP ${ response.status }` );
	error.permanent = response.status >= 400 && response.status < 500;
	throw error;
}
```

Then count only `error.permanent` toward the budget. A refused request still retires after three; a dropped connection keeps trying every five minutes, which costs nothing.

Either way, `warn()` has latched long before the interval is cleared, so the retirement itself is silent — the one warning on screen is the first failure, not the decision to stop. A second `console.warn` where `clearInterval` runs would say which of the two happened.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
```diff
@@ -1,61 +1,183 @@
...
+ */
+export function unwrapContainer( html ) {
+	const template = document.createElement( 'template' );
+	template.innerHTML = html.trim();
+
+	const container = template.content.querySelector(
+		'.wp-block-latest-posts'
+	);
+
+	return container ? container.innerHTML : html;
```
bor0: Narrow, and it follows from the swap coming out: this fallback was written when the target was a `<div>`.

`start()` writes the return value into `element`, which is now the `<ul>` core rendered, and a `<ul>` can only legally hold `<li>`. On the branch where no container is found, returning `html` unchanged puts whatever the response was straight inside a list — the shape removing the swap exists to stop.

It can only fire if the response stops carrying `.wp-block-latest-posts`, so it is the fallback rather than the path. But there is no rendering of this block where writing unrecognised markup beats leaving the server's list where it is. Returning `null` and skipping the write in the caller keeps the guarantee the rest of the change rests on, and `returns the markup unchanged when there is no container to strip` becomes `leaves the list alone`.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
mcliwanow: This one is mine and you are right. Removing `is_swappable_container()` dropped a guarantee I did not notice it was carrying: that the block's list was the whole string. Fixed with the `class_name` match.

The part worth recording is that my test could not see it. It asserted the output contained `has-live-update` somewhere, which holds when the wrong element carries it, and the `siblings` case put the block's list first. I added your reversed case and it still passed. Only tightening the assertion to require the class on the block's own list made it fail, and then the fix made it pass. Both orders are covered now.

Row three of your table came free as you said: a container inside a group is found now, where it used to be skipped silently.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/controller.php
mcliwanow: Both fair. Header is `public, s-maxage=60` now, since as you say the value is entirely in a shared cache and a browser's copy expires long before it polls again. The description covers it.

On the assertions: the one on `test_other_errors_are_not_discarded` does not actually bite. A lockdown arrives as a `WP_Error`, so it never reaches the `instanceof WP_REST_Response` check and cannot gain a header however the code is arranged. I added it as documentation but pinned the case that does bite instead: an authorised editor's render must not be marked cacheable, since that response returns before the header and marking it would hand a logged-in render to everyone behind the cache. Verified it fails if the header moves above the guards.

Counts in the test plan are corrected. On whether the edge caches `/wp-json/` GETs, I do not know and cannot see it from here. The header states the intent; if it turns out nothing honours it, it costs nothing.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
mcliwanow: Good catch, and the distinction is the right one. Only a 4xx counts against the budget now; a dropped connection retries indefinitely. The unattended screen is the case the feature exists for and retiring it after a bad afternoon would have been worse than the problem the budget solves.

Added the second `console.warn` where `clearInterval` runs, for the reason you give: `warn()` latched on the first failure, so the decision to stop was silent.

public_html/wp-content/mu-plugins/blocks/source/hooks/latest-posts/front-end.js
mcliwanow: Agreed, and it follows from the swap coming out exactly as you say. `unwrapContainer()` returns `null` now and the caller skips the write, so unrecognised markup never goes into the list. The JS test is renamed to match what it now asserts.

mcliwanow:
@bor0 , are you up for one more review pass? 🙏

bor0 (approved):
Round 3 at `d7e6f6b9`. Both new commits check out and I probed the parts the tests can't self-certify.

The `class_name` match against a real core render, rather than a hand-copied fixture:

```
<ul data-attributes="…" class="wp-block-latest-posts__list has-dates wp-block-latest-posts has-live-update is-loading">
<ul data-attributes="…" class="wp-block-latest-posts__list is-grid columns-3 wp-block-latest-posts has-live-update is-loading">
```

Marking lands on the block's own list in both, `__list` doesn't false-match the token, and a block without the toggle or with `orderBy` set comes back untouched.

The one thing I wanted to be sure of before the safelist re-runs the callback: that it isn't re-running it on unvalidated input, since the permission error arrives before the route's own validation in some orderings. It doesn't — anonymous, on this branch:

| request | result |
| --- | --- |
| unregistered attribute | 400 `rest_invalid_param`, no cache header |
| `postsToShow=abc` | 400 `rest_invalid_param`, no cache header |
| `attributes` as a string | 400 `rest_invalid_param`, no cache header |
| `post_id=abc` | 400 `rest_invalid_param`, no cache header |
| `postsToShow=101` | 401 `block_cannot_read`, no cache header |
| `postsToShow=100` | 200, `public, s-maxage=60` |

And the loop the change exists for, driven through a real query string rather than a PHP array — booleans as `true`/`false` and everything else as strings, the way `buildQueryString()` sends it — across plain, `categories`, excerpt, author and featured-image attribute sets: 200 and a rendered list on all five.

| | |
| --- | --- |
| `phpunit --group=blocks` | `OK (30 tests, 65 assertions)` |
| `wp-scripts test-unit-js` | 23/23, 3 suites |
| `lint-js source/hooks/latest-posts` | clean |
| `wp-scripts build` | clean, `live-posts` deps `array( 'wp-url' )` |

Also confirmed `postLayout` and `columns` only ever reach `$classes[]` in core's render, so dropping them from the poll can't change the list — the wrapper the classes live on is the one being kept.

Approving. Two things not worth another round: `check_version_support()`'s docblock still opens with *"we need the `wp.serverSideRender` component available"*, which is the component this PR removes; and the test plan says `--group=blocks` 29/29 and 947 bytes, where I get 30/30 and 1289. Grab both if you touch the branch again.

mcliwanow:
The block works fine now. Examples on:
- https://events.wordpress.org/campusconnect/2026/kolkata/day-of-event/
- https://leipzig.wordcamp.org/2025/day-of-event/

Issue #1946, opened 2026-08-25
Reports: Sponsor ROI reporting for multi-event sponsor agreements (tracking)


wputasic:
## Problem

Global sponsors (Automattic, Google, PayPal, …) sponsor many events through a single Multi-Event Sponsorship agreement, but we have no way to show them what that investment reaches. The data exists — it's just never joined:

- **Spend** lives in Sponsor Invoices (`wcb_sponsor_invoice` posts, QBO-backed) on each camp site.
- **Reach** lives in CampTix (`tix_attendee` posts; `tix_attended` check-in meta where organizers use the Attendance tools).
- **The agreement** (`_mes_id` on each local sponsor post) is the key that links a sponsor's presence across camps — and the unit sponsors actually think in.

Today, answering "what did sponsor X get across all camps last year" means manual per-site spelunking. At WCEU scale it took hand-built spreadsheet exports to demonstrate this was even possible.

A second, structural problem the report has to surface honestly: **attendance is often not measured** — `tix_attended` sits empty at many events (WCEU included, several years running), and some locales can't use CampTix at all. A report that silently treats "not measured" as "zero attended" would produce wrong ratios.

## Design decisions

- **Rollup unit = the MES agreement** (`_mes_id`), with per-camp drill-down. Both counts shown: registered and checked-in.
- **Aggregates only, internal audience.** No per-attendee rows anywhere; spend fields are private-context only. Summaries get *sent* to sponsors — no sponsor self-service.
- **Paid invoices only** count as spend by default (sent/approved optionally includable, clearly labeled).
- **Unmeasured attendance is flagged, never faked**: camps with no check-in data add to registered but not attended, and the ratio carries an explicit "attendance not measured at N of M events" caveat.

## Phases

- [ ] **Phase 1 — the report:** new `Sponsor_ROI` report in `wordcamp-reports` (spend × reach per agreement, CSV export). PR to follow.
- [ ] **Phase 1b — pilot:** run it for one global sponsor's portfolio (Automattic) and reconcile against the manually-built WCEU numbers.
- [ ] **Fast-follow — engagement via Jetpack Stats API:** page views / outbound clicks per sponsor are already collected by Jetpack; automate the pull instead of manual exports.
- [ ] **Later — click-through precision:** first-party redirect + impression counting (Jetpack click tracking is destination-attributed today).
- [ ] **Later — demographics with consent:** standardized questions + explicit sharing consent and minimum cell sizes before anything leaves the internal audience.

[...]

PR #1947, opened 2026-08-25
Reports: add a Sponsor ROI report

wputasic:
Part of #1946 (Sponsor ROI tracking issue — Phase 1).

## What this adds

A new **Sponsor ROI** report in `wordcamp-reports`: spend × reach per sponsor, rolled up across camps by Multi-Event Sponsorship agreement.

For each camp in the date range, the report finds every sponsor carrying an `_mes_id`, joins:

- **Spend** — sum of `_wcbsi_amount` on that sponsor's `wcb_sponsor_invoice` posts. **Paid invoices only** (`wcbsi_paid`) by default; an `include_sent_invoices` option also counts `wcbsi_approved`, for forecasting. Multi-currency amounts are normalized to USD with `Currency_XRT_Client` at the event-date rate (base currency short-circuited, `unknown_currency` tolerated as 0).
- **Reach** — registered = published `tix_attendee` count; checked in = attendees with `tix_attended` meta.
- **Tier** — `wcb_sponsor_level` term.

…then rolls rows up by `_mes_id` into per-agreement portfolios with totals and cost-per-registered / cost-per-attended.

## Semantics worth reviewing

- **"Attendance not measured" is a first-class state.** A camp with zero `tix_attended` flags counts toward registered but never toward attended, increments `totals.unmeasured_camps`, and renders as "not measured" (not `0`). The ratio carries an explicit caveat. This matters because check-in coverage is patchy network-wide, and pretending otherwise produces confidently wrong numbers.
- **Spend is private-context only.** `spend_usd`, `has_invoice`, `site_id`, `sponsor_id` sit in `$private_data_fields`, so they're absent from any public-context run. Aggregates only — no per-attendee data anywhere, including the CSV.

## Implementation notes

- **`WP_Query` fails open on unregistered statuses** — statuses missing from `$wp_post_statuses` are silently dropped from `post_status`, so the invoice query can return *every* status (caught live: $104,999 reported for a $5,000 sponsor). The query result is re-checked in PHP against the counted-status list; a regression test unsets the registered statuses and proves the filter holds.
- **Cache keys include the report parameters.** The base class keys the cache on the report slug alone, which would serve one date-range/MES-filter's cached data to a different query for a week. `get_cache_key()` / `get_cache_expiration()` are overridden accordingly.
- **Exporter follows the current authorization convention** (as in `class-meetup-status.php`): `export_to_file()` early-returns unless `$_GET['report']` matches its own slug (all exporters share `admin_init` and the `run-report` nonce), then requires nonce + capability; `render_admin_page()` fails closed on the capability. The existing `test-report-export-authorization.php` dataProvider picks the new class up automatically.

## Tests

21 tests (unit + subsite integration) in the existing `WordCamp Reports` suite — spend join and status filtering, currency normalization, reach and the unmeasured-camp semantics, `_mes_id` rollup, field privacy, cache keys, export authorization. Two harness notes:

- Attendee-creating tests use camptix's `CampTix_Root_Blog_Fixture`, per #1879.
- Invoice fixtures suspend wordcamp-payments' `set_invoice_status` (`wp_insert_post_data`) around the insert — the same idiom sponsor-invoice.php itself uses for programmatic status writes — since that hook otherwise forces incomplete test invoices back to `draft`.

Suite result: 40 tests / 97 assertions green (the 21 new plus the existing reports tests, which now also exercise this class), no new warnings, no DB-error noise.

## Verified / not verified

- Verified locally (docker sandbox): registration, the full data path against seeded multisite data (portfolio rollup, paid-only filtering with an approved-invoice decoy excluded, unmeasured-camp rendering), and the fail-closed admin page.
- **The Export CSV button is not verifiable locally** — `WordPressdotorg\MU_Plugins\Utilities\Export_CSV` only exists in production. Its authorization path is covered by the guard tests; the button itself gets its first real click in production.

## Post-merge follow-ups (not in this PR)

- Add the report to the central `/reports/` landing page (content edit).
- Pilot run for one agreement's portfolio (Automattic) and reconciliation against the manually-built WCEU exports.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-sponsor-roi.php
+++ b/public_html/wp-content/plugins/wordcamp-reports/classes/report/class-sponsor-roi.php
@@ -1,0 +1,749 @@
+<?php
+/**
+ * @package WordCamp\Reports
+ */
+
+namespace WordCamp\Reports\Report;
+defined( 'WPINC' ) || die();
+
+use DateTime;
+use Exception;
+use WordCamp\Reports\Utility\Date_Range;
+use WordCamp\Utilities\Currency_XRT_Client;
+use WordPressdotorg\MU_Plugins\Utilities\Export_CSV;
+use const WordCamp\Reports\CAPABILITY;
+use function WordCamp\Reports\get_views_dir_path;
+use function WordCamp\Reports\Validation\validate_date_range;
+
+/**
+ * Class Sponsor_ROI
+ *
+ * Joins sponsor spend (Sponsor Invoices) with event reach (registered + checked-in attendee
+ * counts), rolled up across each sponsor's portfolio (their Multi-Event Sponsor agreement).
+ * Aggregates only — no new tracking.
+ *
+ * @package WordCamp\Reports\Report
+ */
+class Sponsor_ROI extends Base {
+	/**
+	 * Invoice statuses that count toward spend, default mode: paid only.
+	 *
+	 * @var array
+	 */
+	const COUNTED_STATUSES_PAID = array( 'wcbsi_paid' );
+
+	/**
+	 * Invoice statuses that count toward spend when including sent invoices.
+	 *
+	 * @var array
+	 */
+	const COUNTED_STATUSES_PAID_SENT = array( 'wcbsi_paid', 'wcbsi_approved' );
+
+	/**
+	 * The currency that all spend amounts are normalized to.
+	 *
+	 * @var string
+	 */
+	const BASE_CURRENCY = 'USD';
+
+	/**
+	 * Sponsor country meta key.
+	 *
+	 * @var string
+	 */
+	const COUNTRY_META = '_wcpt_sponsor_country';
+
+	/**
+	 * Report name.
+	 *
+	 * @var string
+	 */
+	public static $name = 'Sponsor ROI';
+
+	/**
+	 * Report slug.
+	 *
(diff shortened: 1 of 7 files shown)


gedex (commented):
One blocker, inline. A single failed exchange-rate lookup turns into a clean-looking report with $0 spend for that sponsor.

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-sponsor-roi.php
```diff
@@ -0,0 +1,749 @@
...
+				? gmdate( 'Y-m-d', (int) $info['Start Date (YYYY-mm-dd)'] )
+				: '';
+
+			$camp_rows[] = $this->get_camp_sponsors( $valid->site_id, $valid->post_id, $info['Name'], $event_date );
+		}
+
+		$data = $camp_rows ? array_merge( ...$camp_rows ) : array();
+
+		$data = $this->filter_data_fields( $data );
+		$this->maybe_cache_data( $data );
```
gedex: When a currency conversion fails, the rows still get cached with that spend as 0, so later runs show $0 and no error for up to a week. Skip the cache when there's an error, or convert to USD at render time like the other spend reports do.

gedex (changes requested):
Marking this as changes requested for the caching issue above. Happy to take another look once it's in.

PR #1948, opened 2026-08-26
Post Types: consolidate the participant username sanitizer

KokkieH:
## What

Participant records (speaker, organizer, volunteer) name the WordPress.org account they represent, stored in `_wcpt_user_name` and resolved to `_wcpt_user_id`. This makes explicit, and consistent across every write path, who may set that account:

- A user may link a record to **their own** account.
- Linking a record to a **different** account is reserved for users who can edit other authors' posts (organizers/admins).

The rule lives in one shared helper, `wcorg_get_linkable_user_login()`, applied at each write boundary:

- **REST** — a `rest_pre_insert_wcb_{speaker,organizer,volunteer}` guard rejects a disallowed link with an error, instead of letting the meta sanitizer quietly drop it.
- **Forms to Drafts** — Call for Speakers and Call for Volunteers route the submitted username through the same helper; a logged-in submitter links their own account (or another, if entitled), and an anonymous submission links no account.
- **Username validation route** — validates a name as *linkable by the current user*, so the block editor's check matches what the save path allows.

The meta `sanitize_callback` goes back to resolving/existence only — a sanitize callback runs on every write path and can only rewrite a value, never reject one, so it can't hold an authorization rule. Two byte-for-byte duplicated closures (`sanitize` and `prepare`) are also pulled into shared named functions.

## Why

- Keeps the record-ownership check and the account-entitlement check as two separate questions, applied the same way whether a record is created via the block editor, the REST API, or a form submission.
- Removes duplicated closures in favour of shared, documented functions.

## Testing

Automated:

```
docker compose -f docker-compose.phpunit.yml run --rm phpunit_wp \
  phpunit --testsuite "WordCamp MU Plugins","WordCamp Post Type","WordCamp Post Types","WordCamp Forms to Drafts"
```

Manual — use a genuine Contributor (only the contributor role) and an Editor/Admin. A fresh contributor on the Seattle test site:

```
wp user create speakertester speakertester@example.test \
  --role=contributor --user_pass=password \
  --url=https://seattle.wordcamp.test/2023/
```

(In the Docker dev environment, prefix with `docker compose exec wordcamp.test`.)

### Block editor / REST

1. As the **Contributor**, create a Speaker draft and set its WordPress.org Username to **your own** username. Save.
   - Expected: saved, and the record resolves to your account.
2. As the **same Contributor**, set the WordPress.org Username to **another existing** username. Save.
   - Expected: refused with an error (nothing linked).
3. As an **Editor/Admin**, set the WordPress.org Username to **another existing** username. Save.
   - Expected: saved, resolves to that account.
4. Re-save an already-linked record as a lower-capability user who can still edit it.
   - Expected: the existing link is preserved (not wiped).

Repeat 1–3 for the Organizer and Volunteer post types; behaviour is identical.

### Forms to Drafts

5. Submit **Call for Volunteers** while logged out, entering any existing username.
   - Expected: a volunteer draft is created with **no** linked account. (The submitted name is still on the Jetpack feedback entry for an organizer to review.)
6. Submit **Call for Volunteers** (or **Call for Speakers**) while logged in as the **Contributor**, entering **another** existing username.
   - Expected: the draft is linked to **you**, not the named account.
7. Same submission as an **Editor/Admin** naming another account.
   - Expected: the draft is linked to the named account.

### Note on "existing" usernames

"Existing" means present in the WordCamp network user table (resolved locally, not fetched from WordPress.org). A user who exists on the network but isn't a member of this site still resolves; a WordPress.org account never provisioned here does not.

## Notes

- Only changes who may set the link. Existing stored values are untouched, and an already-linked record keeps its link when re-saved.
- Anonymous Call for Volunteers submissions now store an empty `_wcpt_user_name` by design (the same state an unknown name already produced), so an organizer sets the account during review.
- `phpcs` (branch scan) clean; the touched suites pass.

https://claude.ai/code/session_01G1iP5g1aPd87xudF8SbixF


public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -372,6 +354,37 @@ function register_volunteer_post_meta() {
...
+ * it is resolved to a canonical `user_login`. A record may only be linked to an
+ * account other than the current user's by someone who can edit other authors'
+ * posts; everyone else may link their own account. Unknown or disallowed values
+ * are stored as an empty string.
+ *
+ * @param string $value The submitted username.
+ *
+ * @return string The canonical `user_login`, or an empty string.
+ */
+function sanitize_user_name_meta( $value ) {
```
mcliwanow, line 367: This runs on the Call for Volunteers submission and breaks it. `call_for_volunteers()` in wordcamp-forms-to-drafts writes `_wcpt_user_name` as the submitting visitor, and that form doesn't require login (only Call for Speakers does). So `get_current_user_id()` is 0, it never matches the resolved account, the visitor has no `edit_others_posts`, and the username stores '' for every submission.

I replayed it locally on a wcb_volunteer draft: as user 0 a valid username stored '' and `_wcpt_user_id` never got derived, as an admin the same value stored fine. Before this change the login was stored regardless of who submitted.

Then `_wcpt_user_id` stays empty (no participation badge), and wordcamp-reports uses `_wcpt_user_name` as the volunteer dedupe key, so the blanks turn into per-post unique values and repeat volunteers stop deduping.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -372,6 +354,37 @@ function register_volunteer_post_meta() {
...
+
+	if ( ! $wporg_user ) {
+		return '';
+	}
+
+	// Linking an account other than the current user's is limited to those who can
+	// edit other authors' posts. When there is no current user (e.g. an anonymous
+	// submission), `get_current_user_id()` is 0 and only a self-link would match,
+	// so a named third-party account is not stored.
+	if ( get_current_user_id() !== $wporg_user->ID && ! current_user_can( 'edit_others_posts' ) ) {
```
mcliwanow: This depends on the current user, so re-writing a value that's already stored can destroy it. An organizer links a record to user X. Later someone who can edit that post but doesn't have `edit_others_posts` (`meta_auth_callback` only needs `edit_post`) saves it again with the same value, and this returns '' because their ID isn't X's. Core sanitizes before the same-value check, so '' differs from the stored `x_login` and the write goes through. `update_wcorg_user_id` then resolves '' to nobody and deletes `_wcpt_user_id`.

Replayed locally: admin links user 5, then a user-0 resave of the same value wiped both `_wcpt_user_name` and `_wcpt_user_id`. So a no-op edit by a lower-cap user silently drops an organizer's link, and on a published post the participation-notifier badge goes with it.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -372,6 +354,37 @@ function register_volunteer_post_meta() {
...
+ * @return string The canonical `user_login`, or an empty string.
+ */
+function sanitize_user_name_meta( $value ) {
+	$wporg_user = wcorg_get_user_by_canonical_names( $value );
+
+	if ( ! $wporg_user ) {
+		return '';
+	}
+
+	// Linking an account other than the current user's is limited to those who can
```
mcliwanow: nit: this restates the docblock right above. I'd keep the policy in the docblock and leave only the non-obvious part here, that user 0 can't match any account so anonymous submissions never link.

mcliwanow (commented):
Left the detail inline. Short version: pulling the three closures into one function is good, but the capability check added alongside it has a problem I think blocks merge.

The check sits in a `register_post_meta` sanitize callback, so it runs on every write path (REST edits, admin saves, and plain `update_post_meta` calls from other plugins). I replayed the writes locally and two things break, both flagged inline: the Call for Volunteers form stores '' for every submitted username, and re-saving an already-linked record from a lower-capability (or no-user) context wipes the stored value and deletes the derived `_wcpt_user_id`. The second one contradicts the "existing stored values are untouched" note in the description.

A sanitize callback can only rewrite the value, it can't reject the write, so a denied edit returns 200 and drops the data silently. If the goal is to stop non-organizers attaching someone else's account, I'd move the check to a layer that can reject: a `rest_pre_insert_wcb_*` filter returning `WP_Error` (it sees both the value and the post ID), or the `update_post_metadata` short-circuit that gatherpress-recurring-events already uses. That also leaves the trusted form writers alone.

Smaller things:

- The check only covers one of the two write paths. `create_draft_speaker()` in wordcamp-forms-to-drafts writes `_wcpt_user_id` straight from the submitted username field with no equivalent check, and that's the meta the prepare_callback, the session author, and the avatars actually read. If both aren't covered the guarantee doesn't really hold.
- The `/wc-post-types/v1/validation` route still checks existence only, so UsernameControl marks a name valid right before the save drops it, with nothing surfaced to the user. Whatever rule the sanitizer ends up applying, the endpoint should apply the same one.
- `edit_others_posts` is hardcoded, while the neighboring `meta_auth_callback` goes through the `edit_post` meta cap. Same result today (all three types are `capability_type => 'post'`), but they'd drift if any of them gets custom caps later. The subtype comes in as the 4th filter arg if you want `get_post_type_object( $subtype )->cap->edit_others_posts`.
- Nothing tests the new behavior. The three cases in your testing steps are unit-testable with `wp_set_current_user` + `update_post_meta`, and would have caught the volunteer regression.
- The three `_wcpt_user_name` prepare_callback closures are still duplicated (lines 90, 252, 310). Same extraction you did for sanitize would finish it.
- The Call for Volunteers stub still shows the WordPress.org Username field as required and tells organizers it feeds the automation. If the form keeps clearing anonymous submissions, that copy needs a look (or the form should require login, like Call for Speakers).

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
KokkieH: Fixed in 837e5d21. Enforcement moved out of the sanitizer into the form handlers, which now resolve the linked account explicitly: a logged-in submitter links their own account, and an anonymous Call for Volunteers submission links none. A valid username is no longer blanked, so the participation badge and `wordcamp-reports` dedupe are unaffected. Covered by `test-participant-link.php`.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
KokkieH: Fixed in 4edd9778. The sanitizer no longer makes any capability decision, so a lower-cap or no-user re-save of the same value can no longer blank it. Entitlement is now enforced at `rest_pre_insert`, which rejects the write with an error rather than silently rewriting it.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
KokkieH: Gone — that comment was removed along with the capability check when it moved to the REST guard.

KokkieH:
Thanks for the thorough review — all addressed across four commits, with enforcement moved to the write boundaries so it can reject rather than silently drop:

- **REST** guard returns `WP_Error` (`4edd9778`).
- **Forms** — the `_wcpt_user_id` write in `create_draft_speaker()` and the Call for Volunteers path both go through the shared `wcorg_get_linkable_user_login()` helper now, so the free-text username field is covered too (`837e5d21`).
- **Validation route** applies the same rule, so it no longer green-lights a name the save path will refuse (`4cd4ceb1`).
- **Cap** flows through the shared helper rather than a hardcoded string; **tests** added for the REST guard, the form handlers, and the validation route; the three `prepare_callback` closures are deduped (`5659ff23`).

One item left out on purpose: the Call for Volunteers form copy / login-requirement question is a product decision rather than part of this change, so I'd suggest a separate follow-up.

mcliwanow (commented):
Re-tested on 5659ff23. The rework fixes the real damage from the first revision: a re-save now preserves the stored link instead of blanking it, `_wcpt_user_id` is derived from the resolved username rather than raw input, and the check can actually refuse a write instead of quietly rewriting it. I ran the suites (325 tests, OK) and the branch phpcs scan (clean), and walked the manual steps for all three post types. Steps 1-3 behave as described.

Two things I'd fix before merge, both small. Left them inline.

Worth answering, probably not blocking: the rule now lives only in the REST guard, while the meta itself is still writable by anything holding `edit_post`. Are we confident REST is the only way in here? Happy to go through what I looked at somewhere other than this thread.

Smaller stuff:

- The validation route changed meaning but `UsernameControl` still renders `rest_invalid_param` as "Invalid username", so a real account now reads as a nonexistent one for anyone without `edit_others_posts`. They'll retype it thinking they typo'd. Either pass the reason through or reword the notice to cover both.
- The tests call `guard_user_name_meta()` directly and never dispatch, so nothing covers the three `add_filter()` lines. I typo'd one to `rest_pre_insert_wcb_volunteers` locally and the suite still reported OK (52 tests), while a contributor could then link someone else's account to a volunteer record. A `has_filter()` assertion, or one dispatched request per type, would catch it. The wiring is correct today, I checked all three.
- `create_draft_speaker()` still reads `_wcpt_user_id` out of the raw form array. Fine because its one caller resolves the value first, but nothing in the method says that's required, and it's the one writer of that key no guard covers.

On anonymous Call for Volunteers: reading the updated description, this is a deliberate call now, so treating it as a question rather than a finding. Are organizers happy re-keying the username from the feedback entry for every anonymous submission? That's most of them, since the form isn't login-gated or pre-filled. Worth noting the reply on my earlier comment says that case is fixed. The behaviour is the same as the revision I commented on (still stores an empty `_wcpt_user_name`), so I want to be sure we're not merging on the assumption it changed.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
```diff
@@ -372,6 +330,89 @@ function register_volunteer_post_meta() {
...
+ *
+ * @param object          $prepared_post The post about to be inserted or updated.
+ * @param WP_REST_Request $request       The REST request.
+ *
+ * @return object|WP_Error The unchanged post, or an error if the link is not allowed.
+ */
+function guard_user_name_meta( $prepared_post, $request ) {
+	$meta = $request['meta'] ?? null;
+
+	if ( ! is_array( $meta ) || ! array_key_exists( '_wcpt_user_name', $meta ) ) {
```
mcliwanow, line 392: The guard keys on `_wcpt_user_name` being present in the payload, not on the link actually changing, so re-submitting the value that's already stored is refused. That takes the whole request with it, since `update_item()` returns before it writes any meta.

Dispatched against this branch, contributor editing a speaker they can edit, with the unchanged username in the same `meta` object:

```
contributor re-saves SAME value       http=403 wcpt_cannot_link_user  name='linktarget' uid='97'
contributor edits email + echoes name http=403 wcpt_cannot_link_user  email_saved=''
```

The link surviving is the fix working, that part's good. But the email edit is gone, and on a record linked to someone else this user can't save anything through a client that round-trips meta.

`get_post_meta( $post_id, '_wcpt_user_name', true )` is available here via `$request['id']`, so guarding only an actual change would cover it.

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
```diff
@@ -360,6 +360,29 @@ protected function get_user_id_from_username( $username ) {
...
+	protected function resolve_linked_username( $submitted_username ) {
+		$current_user = wp_get_current_user();
+
+		if ( ! $current_user->exists() ) {
+			return '';
+		}
+
+		$linkable = wcorg_get_linkable_user_login( $submitted_username, $current_user->ID );
+
+		return '' !== $linkable ? $linkable : $current_user->user_login;
```
mcliwanow: When the submitter names an account they can't link, this hands back their own login, and `get_speaker_from_user_id()` then finds their existing speaker post and skips draft creation at line 486.

Ran two Call for Speakers submissions on this branch as one contributor, naming two different colleagues:

```
newest speaker draft before=Alice Smith after=Alice Smith -> no new draft
session draft: Bob talk    speakers_meta='Alice Smith'
session draft: Alice talk  speakers_meta='Alice Smith'
```

Bob's talk is credited to Alice, and Bob's submitted name, bio and email only exist on the Jetpack feedback entry. Both proposals collapse onto whichever speaker post the submitter already has.

Refusing the link is right. Rewriting the submitted identity to the submitter's is a separate decision, and it's silent. Leaving the draft unlinked, the way the volunteer path does, would keep the submitted data for an organizer to sort out.

KokkieH:
> Are organizers happy re-keying the username from the feedback entry for every anonymous submission?

Based on checking a sampling of upcoming WordCamps, many appear to be using their own Jetpack forms, or even Google forms, rather than the shortcode-generated form, so it would appear the common practice is to create volunteer posts manually anyway. 

Even with this form not all fields are transferred to the post, so organisers would need to cross reference with the feedback entry regardless.

public_html/wp-content/plugins/wc-post-types/inc/rest-api.php
KokkieH: Fixed in 3130df64. The guard now compares the submitted account against the one already stored (via `$request['id']`) and only gates an actual change, so re-saving a record with its unchanged username — and the unrelated edits riding along — goes through. Changing the link to a disallowed account is still refused. Added tests for the re-save, the change, and a `has_filter()` assertion per post type so a renamed hook is caught.

public_html/wp-content/plugins/wordcamp-forms-to-drafts/wordcamp-forms-to-drafts.php
KokkieH: Fixed in 595f9d93. Naming an account the submitter can't link now leaves the draft unlinked rather than rewriting it to their own login, so the submitted identity isn't silently changed and proposals no longer collapse onto the submitter's existing speaker post (only a linked account reuses an existing one now). Blank username still defaults to the submitter. The submitted name remains on the feedback entry for an organizer to resolve.

KokkieH:
Both inline findings fixed (replied on the threads). The smaller items:

- **Validation route** — reverted to existence-only (`3a8415ee`). Enforcing entitlement there made a real account render as "Invalid username" for non-organizers, and the save-time guard already refuses a disallowed link with its own visible error, so the route only needs to report existence.
- **Filter wiring** — added a `has_filter()` assertion for all three `rest_pre_insert_wcb_*` hooks, so the renamed-filter typo you hit would now fail the suite.
- **`create_draft_speaker()`** — added a comment at the `_wcpt_user_id` write noting the caller resolves the username first, so the assumption is documented at the write site.

On "is REST the only way in": the two HTTP-reachable writers — the REST endpoint and the two form handlers — are both gated now. The meta stays writable by server-side `update_post_meta()` (e.g. an importer or another plugin), which is trusted code rather than an untrusted request; the sanitizer can't gate that, and gating it there is what broke the trusted paths last round. Happy to talk through what you looked at if you'd like — a wp-admin metabox for these keys doesn't exist, so REST + forms is the reachable surface.

mcliwanow (approved):
Approving.

Both blockers are fixed. I re-tested them on 3a8415ee rather than taking the commits at their word:

- re-saving an unchanged link returns 200 now, and an unrelated edit riding along in the same `meta` payload survives. That was the one that made the record unsaveable for anyone without `edit_others_posts`.
- naming an account you can't link leaves the draft unlinked instead of quietly re-pointing it at the submitter, and two submissions from one contributor produce two drafts with the sessions credited to the right speaker.

Also ran the suites (349 tests, OK), the branch phpcs scan (clean), and the manual steps across all three post types. The `has_filter()` assertion does what it says: I re-applied the same `rest_pre_insert_wcb_volunteers` typo and the suite fails on it now instead of passing green.

Followed up on the "is REST the only way in" question separately. Nothing there blocks this, it behaves the same before and after your change.

Two non-blocking follow-ups whenever you get to them:

- reverting the validation route took its four tests with it, so that route has no coverage now. Valid name and unknown name would put the floor back.
- `tests/stubs.php` was only needed by the dispatch-based test that got deleted, and the real `site_supports_block_templates()` has no `function_exists` guard of its own, so it's a redeclare waiting to happen. Probably just delete it.

KokkieH:
Thanks, Michal. Fixed the test coverage issues in 122c465

PR #1949, opened 2026-08-26
QuickBooks: verify the OAuth `state` on the token exchange

bor0:
## Summary

The QuickBooks OAuth `exchange` leg had nothing tying a callback to the authorize request that started it. The SDK emits a `state`, but never stores one, so there was nothing to compare against when Intuit sends the user back.

- `Client::get_authorize_url()` now generates a random 32-character `state`, stores it for the user who started the process, and hands it to the SDK to build the URL sent to Intuit.
- `Client::maybe_exchange_code_for_token()` requires a `hash_equals()` match before it will exchange the code, and surfaces an error through the existing error view otherwise.
- The stored value expires after 15 minutes, and is discarded once it's served its purpose: a token came back, or a connection already exists. Verification itself doesn't touch it, so a stray hit on the callback endpoint can't throw away an authorization that's still in flight, and a callback that fails upstream can be retried.

## How to review this PR

Read `includes/client.php` top to bottom; the new helpers sit next to the existing token-storage helpers and follow the same `generate_*_key()` naming.

Two decisions worth calling out:

- **User meta, not a transient.** The redirect URI is hardcoded to the WordCamp network, but the token option is deliberately shared with Events (`get_network_option( WORDCAMP_NETWORK_ID, ... )`). A process that starts on Events finishes on WordCamp, and transients are per-site — user meta is global across the install, so it survives the round trip.
- **`state`, not a nonce.** The `authorize` and `revoke` legs verify nonces because they're our own form posts. `exchange` arrives from Intuit at a redirect URI registered with them, so it can't carry a nonce of its own — `state` is the equivalent for that leg.
- The existing `current_user_can( OAUTH_CAP )` gate in `admin.php` and the "refuse if a valid token already exists" check are unchanged; this only adds the missing link between the two legs.

**Deploy note:** any authorize started before this ships won't have a stored value, so that one callback will be turned away. Clicking Connect again resolves it.

## Test plan

- [x] phpunit — full suite 867/867 passing (6 pre-existing skips: `intl` extension, GatherPress data provider)
- [x] phpunit — new `quickbooks` group, 15/15 passing
- [x] phpcs clean on both changed files
- [x] Manual browser pass as a network admin (steps below)

New coverage in `tests/test-quickbooks-oauth-state.php`: storage, match, mismatched/empty/non-string input on both sides, expiry, wrong user, logged out. Plus the seam itself, all without leaving the local stack — the authorize URL carries the stored value, and the exchange refuses a bad `state`, ignores an unusable `code`, writes no token in either case, and keeps a pending value for a retry.

### Manual test steps

All as a network admin on the local stack, at `/wp-admin/network/settings.php?page=quickbooks`.

1. **Page renders.** Load the settings page. It should show "Not connected" with an enabled **Connect** button, no errors.

2. **Callback with no `state` is refused.** Load this directly, substituting your local host:

   ```
   /wp-admin/admin-post.php?action=wordcamp-qbo-oauth&cmd=exchange&code=ANYTHING&realmId=9999999999
   ```

   Expected: *"Your request could not be validated. Please start the connection process again."* On `production` this instead reaches Intuit and comes back with `{"error":"invalid_client"}` — that difference is the point of the change.

3. **Authorize sends a `state`, and we remember it.** Click **Connect**. You'll land on Intuit's sign-in page — read the `state` query param out of that URL, then check it against what got stored:

   ```bash
   wp --url=https://central.wordcamp.test user meta get admin "wordcamp-qbo_oauth_state"
   ```

   Expected: the `state` in the array matches the one in the Intuit URL, with `expires` ~15 minutes out. No need to actually sign in to Intuit.

4. **Wrong `state` is refused.** With that process still pending, load the step-2 URL again but append `&state=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`. Expected: the same "could not be validated" message.

5. **Matching `state` gets through.** Seed a known value and use it:

   ```bash
   wp --url=https://central.wordcamp.test eval '
     update_user_meta( get_user_by( "login", "admin" )->ID, "wordcamp-qbo_oauth_state",
       array( "state" => "KNOWNGOODSTATEVALUE1234567890abc", "expires" => time() + 900 ) );'
   ```

   Then load the step-2 URL with `&state=KNOWNGOODSTATEVALUE1234567890abc`. Expected: `Exchange Authorization Code for Access Token failed. Body: [{"error":"invalid_client"}]` — an upstream error, meaning the request got past the check and reached Intuit. Reloading gives the same upstream error rather than "could not be validated": the value survives a failed exchange so the callback can be retried, and is only discarded once a token comes back.

6. **Nothing was stored along the way.** After all of the above:

   ```bash
   wp --url=https://central.wordcamp.test eval 'var_dump( get_network_option( 1, "wordcamp-qbo_oauth", "NOT SET" ) );'
   ```

   Expected: `NOT SET` — no token data written by any of the attempts.

A full round trip against a real QBO company wasn't exercised — the local stack has placeholder client credentials, so step 5 stops at Intuit's client check. Worth someone with sandbox credentials completing an actual Connect → authorize → callback cycle before this comes out of draft.

Note the error page in steps 2/4 renders unstyled — `admin-post.php` has no admin chrome. That matches the existing nonce-failure path on the `authorize` and `revoke` legs.

### DIFF
--- a/public_html/wp-content/mu-plugins/quickbooks/includes/client.php
+++ b/public_html/wp-content/mu-plugins/quickbooks/includes/client.php
@@ -215,6 +215,66 @@ class Client {
 	}
 
 	/**
+	 * The user meta key for storing the OAuth `state` value.
+	 *
+	 * User meta is used rather than an option or a transient because it's shared across every site and network
+	 * in the install, while the redirect URI always points at the WordCamp network. That way an authorization
+	 * that starts on the Events network can still be matched up when QBO sends the user back here.
+	 *
+	 * @return string
+	 */
+	protected static function generate_oauth_state_key() {
+		return PLUGIN_PREFIX . '_oauth_state';
+	}
+
+	/**
+	 * Generate a `state` value for an authorization request, and remember it for the callback.
+	 *
+	 * @return string
+	 */
+	protected static function create_oauth_state() {
+		$state = wp_generate_password( 32, false );
+
+		update_user_meta(
+			get_current_user_id(),
+			self::generate_oauth_state_key(),
+			array(
+				'state'   => $state,
+				'expires' => time() + ( 15 * MINUTE_IN_SECONDS ),
+			)
+		);
+
+		return $state;
+	}
+
+	/**
+	 * Check that a callback's `state` matches the authorization request that this user started.
+	 *
+	 * The stored value is good for one callback only, so it's discarded either way.
+	 *
+	 * @param string $state
+	 *
+	 * @return bool
+	 */
+	protected static function verify_oauth_state( $state ) {
+		$user_id = get_current_user_id();
+		$key     = self::generate_oauth_state_key();
+		$stored  = get_user_meta( $user_id, $key, true );
+
+		delete_user_meta( $user_id, $key );
+
+		if ( ! is_string( $state ) || ! $state || ! is_array( $stored ) ) {
+			return false;
+		}
+
+		if ( empty( $stored['state'] ) || empty( $stored['expires'] ) || time() > $stored['expires'] ) {
+			return false;
+		}
+
+		return hash_equals( $stored['state'], $state );
+	}
+
+	/**
 	 * Shortcut for importing an Exception into the client's WP_Error instance.
 	 *
 	 * @param Exception $exception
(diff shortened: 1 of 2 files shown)


mcliwanow:
Nice, this closes a real gap: the exchange leg had nothing tying the callback back to the authorize request. A few things before it ships, most of them around one theme, the stored state gets thrown away too eagerly.

**State is consumed before it's validated** (`verify_oauth_state`). `delete_user_meta()` runs at the top, before the `is_string`/`is_array` guards, so any stray hit on the exchange endpoint mid-flow (a reloaded callback tab, an old one from history) wipes the pending state and the genuine Intuit callback then fails with "start the connection process again". It only needs `code` and `realmId` present, no valid state. I'd consume the row only when there's a real stored value to compare against (or only on a `hash_equals` match). `test_state_can_only_be_used_once` still holds that way.

**Same root, two more:**

- The delete happens before `exchangeAuthorizationCodeForToken()`, so a transient Intuit failure can no longer be retried by reloading the callback. The reload now dead-ends at `invalid_state`.
- The `has_valid_token()` early return skips `verify_oauth_state()`, so the state isn't discarded on that path. The docblock says "discarded either way", which isn't the case here, and it leaves an orphaned meta row.

**Two smaller correctness bits:**

- `hash_equals( $stored['state'], $state )` checks the incoming `$state` is a string but not the stored side. `empty()` lets a non-empty array through, and `hash_equals(array, string)` throws a TypeError that isn't caught (the call sits outside the try). One `is_string( $stored['state'] )` in the guard fixes it. Only reachable if something else writes that meta key, so low, but it's one line.
- `code` and `realmId` still come through raw while `state` now gets `sanitize_text_field`. `?code[]=x` is a truthy array that reaches the SDK and can fatal the same way. Since you're already on these lines, worth sanitizing them too.

**Tests:** the helpers are well covered, but the seam this PR adds isn't: that `maybe_exchange_code_for_token()` refuses a bad state and writes no token, and that `get_authorize_url()` actually carries the stored value. Both run locally with dummy creds, no HTTP. Without them a reverted gate stays green.

Not blocking, one design note: the SDK takes a caller-supplied `state` via `DataService::Configure()` (`$settings['state']`), so you could hand it your value instead of rewriting the finished URL with `add_query_arg`. It keeps the helper's own state in sync, and avoids the fact that `add_query_arg` doesn't url-encode the value (fine today only because `wp_generate_password` is alphanumeric). If you go that way, mint it lazily in `get_authorize_url()`, not in the config array, or you'll write user meta on every Client construction.

bor0:
Thanks — all of it addressed in 5f66d39, and you were right that the three big ones are the same bug.

**State consumed before it's validated.** `verify_oauth_state()` no longer touches the stored row; it just reads and compares. Consumption moved to `maybe_exchange_code_for_token()`, which is the only place that knows whether the value actually did its job.

**Retry after a transient failure.** The row is now discarded *after* `exchangeAuthorizationCodeForToken()` comes back with a token, so a callback that dies upstream can be reloaded until the 15 minutes are up. Confirmed on the local stack: two consecutive callbacks with a matching `state` both reach Intuit and come back `invalid_client`, with the value still stored after each.

**`has_valid_token()` early return.** Split out of the guard clause. Missing `code`/`realmId` still returns without touching anything, but a callback that arrives once a connection exists discards whatever that user had pending, so there's no orphaned row. Docblocks updated to say when the value is discarded rather than "either way".

**Non-string stored value.** `is_string( $stored['state'] )` added to the guard, so `hash_equals()` never sees an array.

**Raw `code`/`realmId`.** Both go through `sanitize_text_field()` now. `?code[]=x` sanitizes to an empty string and returns at the same place a missing `code` does, before the SDK sees anything.

**Tests.** The seam is covered now, all locally with no HTTP: the authorize URL carries the stored value; the exchange refuses a mismatched `state` (and writes no token); an unusable `code` is dropped before the SDK; a failed exchange keeps the value for a retry; and an already-connected callback discards it. Checked both ways — stubbing the gate out fails `test_exchange_is_refused_without_a_matching_state`, and putting the delete back at the top of `verify_oauth_state()` fails three. 15/15 in the group, 867/867 for the full suite, phpcs clean.

**SDK-supplied `state`.** Took it. `get_authorize_url()` mints the value and passes it through `DataService::Configure()`, so `add_query_arg` is gone and `http_build_query` does the encoding. Minted in `get_authorize_url()` rather than the config array, as you suggested, so only a real authorization request writes user meta. Against the local config the URL comes back with an encoded `state` matching what's stored.

The PR description's step 5 changed accordingly — reloading a failed callback now returns the same upstream error instead of "could not be validated".

mcliwanow (approved):
Looks right to me. Moving the delete out of `verify_oauth_state()` into the exchange is the change I wanted, and it takes the retry case and the early-return mismatch with it.

Ran the suite locally rather than trusting CI: 15 tests, 26 assertions, green on 8.4 multisite. `DataService.php:507` does read `$settings['state']`, so the config path is the real one.

Worth saying out loud: the value isn't strictly single use any more, it lives until an exchange succeeds or the 15 minutes run out. Your docblocks say as much, so I read it as deliberate, and it doesn't weaken anything I can see.

Before merge I'd still want the sandbox round trip you called out in the description. Every test stops short of Intuit, so nothing covers a real callback actually connecting. If you've done it since taking this out of draft, ship it.

Small stuff for a follow-up if you think it's worth anything: the unchecked `update_user_meta()` return, a second Connect click still overwriting the first pending value, and nothing pinning that `delete_oauth_state()` runs after a successful exchange.

PR #1953, opened 2026-08-26
PayPal: settle an order only from the transaction that paid for it

mcliwanow:
## Why

Line references in this section are against `production`, since that is where the behaviour being described lives.

`payment_notify()` settles whichever order the `tix_payment_token` query parameter names. It confirms the IPN is genuine (`payment-paypal.php:274`), fetches the transaction with a credentialed `GetTransactionDetails` call (`:296`), then hands the request's token to `payment_result()` (`:317`). The transaction it authenticated and the order it settles are two separate values, and nothing checks they have anything to do with each other.

`payment_return()` has the same shape: `tix_payment_token` and PayPal's `token` arrive as independent request values (`:502-504`), tied together only by comparing the amount (`:550`), which two orders of the same price both satisfy. The `@todo maybe check tix_paypal_token for security` above it is about that.

Driven against a stubbed PayPal on the dev site, a transaction that was made for
one order settles a different one: the order named in the request is published and
has that transaction recorded against it, while the order the payment was actually
for stays in draft. The same gap lets a transaction from one camp settle an order
on another, and lets a later payment overwrite the transaction an organizer would
refund.

The cross-site part matters because many WordCamps have no merchant account of their own: `mu-plugins/camptix-tweaks/camptix-tweaks.php:173-188` provisions one shared set of live credentials, and a read-only audit of the network puts **910 of the 1,186 sites with PayPal enabled** on them, so "paid to this merchant" spans sites for most of the network. There are **192,833 PayPal orders** across 920 sites.

The overwrite costs money rather than tickets: `payment_result()` rewrites `tix_transaction_id` on every call (`camptix.php:6752`) and `send_refund_request()` refunds whatever is recorded there (`payment-paypal.php:790`), so the last writer decides who gets refunded.

## What changed

All in `public_html/wp-content/plugins/camptix/addons/payment-paypal.php`.

- `fill_payload_with_order()` sends `PAYMENTREQUEST_0_CUSTOM` as `<blog_id>:<payment_token>`, so PayPal echoes back which order a transaction was made for. Set there rather than at the two call sites, so it covers both `SetExpressCheckout` and `DoExpressCheckoutPayment` and the `camptix_paypal_payload` filter cannot drop it.
- An order that already names a transaction accepts only that transaction.
- An order that names none accepts a transaction that echoes back a reference matching it, is of a type CampTix could have started, was taken through this gateway, and covers the order total in the order's currency. Covers rather than equals, since some sites charge a fee on top.
- An order with no positive total is never settled. Free orders skip `payment_checkout()` entirely (`camptix.php:6409`) but still carry a token and a `tix_order` of zero, which any amount would cover. **175,064** attendees network wide are that shape.
- `payment_return()` refuses a PayPal checkout whose echoed reference belongs to another order. One carrying no reference is charged on the amount, as before.
- 503 when PayPal cannot be reached or will not describe the transaction, 403 when PayPal will not confirm the message as its own, 200 for anything that will never settle. All of these answered 200 before, including the retryable ones.
- An IPN carrying no transaction id is rejected instead of asking PayPal about the literal string `None`.
- Fixed a `sprintf()` call on the transaction-fetch failure line that swallowed the log's `$attendee_id` and `$txn_details` arguments.

### Why it is shaped this way

**The order's own record comes first, not stored checkout meta.** `payment_return()` already writes `tix_transaction_id` in the request that charges the card, so every order settled before this ships is covered without a migration, and refunds arriving months later still land. No deploy window, no legacy branch. Requiring it to be *that same* transaction is also what stops a later payment displacing the one the buyer made.

**The amount and type back up the reference, which is forgeable.** `CUSTOM` is a payer-settable HTML variable on a Payments Standard payment, so a matching reference alone proves nothing. [PayPal documents](https://developer.paypal.com/api/nvp-soap/get-transaction-details-nvp) `TRANSACTIONTYPE` as `cart` or `express-checkout` for the checkouts this gateway starts, and that is what separates one of ours from a payment somebody sent the merchant themselves. Both checks sit on the first-settlement arm only, so a real payment — matched on its recorded transaction — never reaches them. The type is compared with separators stripped, since PayPal spells that family differently between its IPN variables and its NVP responses.

**Status codes are chosen around the shared account.** PayPal resends anything it does not get a 200 for, for up to four days. With one account serving many camps, a refusal it can never satisfy is a liability for all of them, so those answer 200 and log the reason.

**Two things deliberately absent.** No explicit merchant check: the API call runs on this site's credentials, so a transaction belonging to anyone else comes back `ACK=Failure` and never reaches the binding. No network-wide transaction-id ledger: the reference carries the blog id and payment token, so a transaction can only settle the one order on the one site, and an order that already names one will not take a second.

## Testing

```bash
docker compose -f docker-compose.phpunit.yml run --rm phpunit_wp phpunit
BASE_REF=production php .github/bin/phpcs-branch.php
```

- [x] phpunit: `CampTix` goes from 86 tests to 118, all passing, same 5 pre-existing `intl` skips. Full run 882 passing, 6 skipped
- [x] phpcs clean on changed lines; the file carries the same 15 sniff types as it does on `production`, and fewer violations
- [x] Mutation-tested: all 22 checks below turn the suite red when removed
- [x] Driven end-to-end over HTTP against a stubbed PayPal, on this branch and on `production`'s copy of the file: 15 cases through `payment_notify()` and 5 through `payment_return()`
- [x] Read-only production audit across all 1,650 live sites, for the figures quoted above and to size what the change affects

<details>
<summary><b>Mutation table</b> — each edit applied on its own, failures in <code>phpunit --filter PayPal</code></summary>

| Edit | Failures |
| --- | --- |
| Drop the blog id from `get_order_reference()` | 9 |
| `transaction_was_made_for_order()` → `return true` | 5 |
| Fall through to the reference when a transaction is already recorded | 4 |
| `transaction_covers_order()` → `return true` | 3 |
| Drop the return path's reference check | 3 |
| Drop the `CUSTOM` echo comparison | 2 |
| Compare the transaction type without stripping separators | 2 |
| Answer 2xx instead of 5xx when PayPal is unreachable | 2 |
| Accept any transaction for an order that already names one | 1 |
| Drop the transaction-type gate | 1 |
| Drop the payment-method check | 1 |
| Allow a zero or missing order total | 1 |
| Drop the currency comparison | 1 |
| Amount check `covers` → `equals` | 1 |
| Drop the amount comparison | 1 |
| Accept an IPN PayPal does not verify | 1 |
| Drop the payment-status gate | 1 |
| Answer 2xx instead of 5xx when the transaction cannot be fetched | 1 |
| Drop the unknown-order check | 1 |
| Read only the prefixed spelling of the return path's reference | 1 |
| Return path refuses a checkout carrying no reference | 1 |
| Stop sending `PAYMENTREQUEST_0_CUSTOM` at checkout | 1 |

</details>

### Compatibility

The three legitimate payments settle identically before and after, and the refund is recorded identically. Nothing legitimate reaches the refusal branches:

- The charge is created by `DoExpressCheckoutPayment` inside `payment_return()`, which records `tix_transaction_id` in the same request, so a genuine IPN matches on the recorded transaction even for an order created before this ships.
- `payment_notify_back_compat()`, serving the CampTix 1.1 notify URL, finds its attendee by transaction id before delegating, so it matches by construction.
- Attendees migrated by the 1.1 upgrade (`camptix.php:1528-1542`) have no `tix_order` at all, and are matched on their recorded transaction without one. There are **17,625** of them, every one on the oldest 200 sites and none created since.
- `payment_return()`'s new check is skipped when PayPal echoes no reference, so a checkout started before this ships and finished after it is charged as before. PayPal expires those tokens within hours, so the window closes on its own.

All four are covered by tests.

`payment_notify()` now resolves the order before anything reaches `payment_result()`, and the two use different post-status filters: `get_order()` passes `'any'`, `payment_result()` an explicit list. `'any'` is the superset, because CampTix registers `cancel`, `failed`, `timeout` and `refund` with `exclude_from_search = false`, so nothing that could have reached `payment_result()` is turned away earlier. Worth stating because the cancel-then-webhook behaviour in `payment_result()` rests on that and it is not visible in this diff.

Two populations were checked directly rather than reasoned about. **About 1,190 published PayPal orders carry no transaction id**; sampling and then a full count show none of them holds a stored gateway response, so no PayPal transaction ever existed for them and no notification can arrive. They are imported attendees (981) or orders an organizer published by hand (194). Separately, **193 trashed orders do name a transaction**, and a notification for one finds no order because `get_order()` uses `post_status => 'any'`, which excludes trash. That is unchanged by this PR: the old `payment_result()` query excluded trash too.

At rest the network has **no draft and two pending** PayPal orders without a transaction, so the first-settlement path this PR adds checks to is exercised by almost nothing outside a live checkout.

One behaviour worth naming: a site that changes its configured currency will no longer settle *first* payments for orders created before the change, since `tix_order` records no currency of its own. Those orders are matched on the recorded transaction once they have one, so it only affects an order in flight across the change.

---

### DIFF
--- a/public_html/wp-content/plugins/camptix/addons/payment-paypal.php
+++ b/public_html/wp-content/plugins/camptix/addons/payment-paypal.php
@@ -260,32 +260,62 @@ class CampTix_Payment_Method_PayPal extends CampTix_Payment_Method {
 	 * payload in $_POST. Verify the payload and use $this->payment_result
 	 * to signal a transaction result back to CampTix.
 	 *
-	 * @return mixed Null if returning early, or an integer matching one of the CampTix_Plugin::PAYMENT_STATUS_{status} constants
+	 * The token in the request says which order to settle, but the sender chooses it
+	 * and it is not private: payment_checkout() puts it in the CANCELURL. So the
+	 * transaction has to be shown to belong to the order that was named.
+	 *
+	 * @return mixed A CampTix_Plugin::PAYMENT_STATUS_{status} constant. Anything not
+	 *               settled ends in ipn_die() instead of returning.
 	 */
 	function payment_notify() {
 		/** @var $camptix CampTix_Plugin */
 		global $camptix;
 
-		$payment_token = isset( $_REQUEST['tix_payment_token'] ) ? trim( $_REQUEST['tix_payment_token'] ) : '';
+		$payment_token = isset( $_REQUEST['tix_payment_token'] ) ? sanitize_text_field( wp_unslash( $_REQUEST['tix_payment_token'] ) ) : '';
 
 		// Verify the IPN came from PayPal.
-		$payload = stripslashes_deep( $_POST );
+		$payload  = stripslashes_deep( $_POST );
 		$response = $this->verify_ipn( $payload );
-		if ( '200' != wp_remote_retrieve_response_code( $response ) || 'VERIFIED' != wp_remote_retrieve_body( $response ) ) {
+
+		// Unreachable, so whether this IPN is genuine is unknown. Ask for a resend.
+		if ( is_wp_error( $response ) || 200 !== (int) wp_remote_retrieve_response_code( $response ) ) {
+			$camptix->log( 'Could not reach PayPal to verify an IPN.', 0, compact( 'response' ) );
+
+			$this->ipn_die( 503 );
+		}
+
+		if ( 'VERIFIED' !== trim( wp_remote_retrieve_body( $response ) ) ) {
 			$camptix->log( 'Could not verify PayPal IPN.', 0, null );
-			return;
+
+			$this->ipn_die( 403 );
 		}
 
 		// Grab the txn id (or the parent id in case of refunds, cancels, etc)
-		$txn_id = ! empty( $payload['txn_id'] ) ? $payload['txn_id'] : 'None';
+		$txn_id = ! empty( $payload['txn_id'] ) ? sanitize_text_field( $payload['txn_id'] ) : '';
 		if ( ! empty( $payload['parent_txn_id'] ) ) {
-			$txn_id = $payload['parent_txn_id'];
+			$txn_id = sanitize_text_field( $payload['parent_txn_id'] );
+		}
+
+		// There is no transaction to look up, and sending the message again would not produce one.
+		if ( ! $txn_id ) {
+			$camptix->log( 'Received IPN with no transaction id.', 0, $payload );
+
+			$this->ipn_die( 200 );
 		}
 
 		// Make sure we have a status
 		if ( empty( $payload['payment_status'] ) ) {
 			$camptix->log( sprintf( 'Received IPN with no payment status %s', $txn_id ), 0, $payload );
-			return;
+
+			$this->ipn_die( 200 );
+		}
+
+		$order = $this->get_order( $payment_token );
+
+		if ( ! $order ) {
+			$camptix->log( sprintf( 'Received IPN for %s naming an order that does not exist.', $txn_id ), 0, compact( 'payment_token' ) );
+
+			$this->ipn_die( 200 );
 		}
 
 		// Fetch latest transaction details to avoid race conditions.
@@ -294,12 +324,38 @@ class CampTix_Payment_Method_PayPal extends CampTix_Payment_Method {
 			'TRANSACTIONID' => $txn_id,
 		);
 		$txn_details         = wp_parse_args( wp_remote_retrieve_body( $this->request( $txn_details_payload ) ) );
-		if ( ! isset( $txn_details['ACK'] ) || 'Success' != $txn_details['ACK'] ) {
-			$camptix->log( sprintf( 'Fetching transaction after IPN failed %s.', $txn_id, 0, $txn_details ) );
-			return;
+		if ( ! isset( $txn_details['ACK'] ) || 'Success' !== $txn_details['ACK'] ) {
+			// Nothing about the transaction is known, so ask for the IPN again rather than reporting it handled.
+			$camptix->log( sprintf( 'Fetching transaction after IPN failed %s.', $txn_id ), $order['attendee_id'], $txn_details );
+
+			$this->ipn_die( 503 );
 		}
 
 		$camptix->log( sprintf( 'Payment details for %s via IPN', $txn_id ), null, $txn_details );
+
+		$recorded_txn_id = (string) get_post_meta( $order['attendee_id'], 'tix_transaction_id', true );
+
+		if ( '' !== $recorded_txn_id ) {
+			/*
+			 * The order names its own transaction, so it is this order's by our own record.
+			 * Repeat IPNs, refunds and reversals arrive this way, including for orders that
+			 * predate the reference below. Requiring that same transaction also stops a
+			 * later payment displacing the one an organizer would refund.
+			 */
+			if ( ! hash_equals( $recorded_txn_id, $txn_id ) ) {
+				$this->ignore_ipn( sprintf( 'this order was paid by %1$s, not %2$s', $recorded_txn_id, $txn_id ), $order, compact( 'payment_token', 'txn_details' ) );
+			}
+		} else {
+			// Nothing recorded, so the echoed reference ties them together, backed by the amount.
+			if ( ! $this->transaction_was_made_for_order( $order, $txn_details ) ) {
+				$this->ignore_ipn( sprintf( 'transaction %s was not made for this order', $txn_id ), $order, compact( 'payment_token', 'txn_details' ) );
+			}
+
+			if ( ! $this->transaction_covers_order( $order, $txn_details ) ) {
+				$this->ignore_ipn( sprintf( 'transaction %s does not cover this order', $txn_id ), $order, compact( 'payment_token', 'txn_details' ) );
+			}
+		}
+
 		$payment_status = $txn_details['PAYMENTSTATUS'];
 
 		$payment_data = array(
@@ -315,6 +371,123 @@ class CampTix_Payment_Method_PayPal extends CampTix_Payment_Method {
 		 * payment result twice. In fact, it's typical for payment methods with IPN support.
 		 */
 		return $camptix->payment_result( $payment_token, $this->get_status_from_string( $payment_status ), $payment_data );
+	}
+
+	/**
+	 * End an IPN request without recording a payment result.
+	 *
+	 * PayPal resends anything it does not get a 200 for, for up to four days, and one
+	 * account serves many camps, so a retry it can never satisfy is everyone's problem:
+	 *
+	 * - 503 when the outcome could not be determined, such as PayPal being unreachable.
+	 * - 200 when the message was understood and will never settle this order.
+	 * - 403 when PayPal will not confirm the message as its own. Resent like any
+	 *   non-200, which is wanted: a mangled but genuine IPN can pass on a resend.
+	 *
+	 * @param int $status_code
+	 *
+	 * @return void
+	 */
+	protected function ipn_die( $status_code ) {
+		wp_die(
+			esc_html__( 'This payment notification could not be processed.', 'wordcamporg' ),
+			esc_html__( 'PayPal notification', 'wordcamporg' ),
+			array( 'response' => absint( $status_code ) )
+		);
+	}
+
+	/**
+	 * Log why an IPN will never settle the order it names, and stop.
+	 *
+	 * @param string $reason
+	 * @param array  $order
+	 * @param array  $context
+	 *
+	 * @return void
+	 */
+	protected function ignore_ipn( $reason, $order, $context ) {
+		/** @var $camptix CampTix_Plugin */
+		global $camptix;
+
+		$camptix->log( 'Ignoring IPN: ' . $reason . '.', $order['attendee_id'], $context );
+
+		$this->ipn_die( 200 );
+	}
+
+	/**
+	 * Whether PayPal says this transaction was made for the given CampTix order.
+	 *
+	 * The evidence is the reference fill_payload_with_order() sends as
+	 * PAYMENTREQUEST_0_CUSTOM and GetTransactionDetails echoes back, compared against
+	 * the order's stored token rather than the request's: get_order() matches with a
+	 * MySQL CHAR comparison, which ignores case and trailing spaces, hash_equals()
+	 * neither. First settlement only, per payment_notify().
+	 *
+	 * @param array $order
+	 * @param array $txn_details
+	 *
+	 * @return bool
+	 */
+	protected function transaction_was_made_for_order( $order, $txn_details ) {
+		// An order taken through another gateway has no PayPal transaction to be settled by.
+		if ( get_post_meta( $order['attendee_id'], 'tix_payment_method', true ) !== $this->id ) {
+			return false;
+		}
+
+		/*
+		 * Only a checkout CampTix started can settle an order with no transaction yet;
+		 * a payment sent any other way had its CUSTOM chosen by the payer. Separators
+		 * are stripped because PayPal spells this differently across its APIs.
+		 */
+		$transaction_type = isset( $txn_details['TRANSACTIONTYPE'] ) ? (string) $txn_details['TRANSACTIONTYPE'] : '';
+		$transaction_type = strtolower( preg_replace( '/[^a-zA-Z]/', '', $transaction_type ) );
+
+		if ( ! in_array( $transaction_type, array( 'cart', 'expresscheckout' ), true ) ) {
+			return false;
+		}
+
+		$echoed_reference = isset( $txn_details['CUSTOM'] ) ? (string) $txn_details['CUSTOM'] : '';
+		$payment_token    = (string) get_post_meta( $order['attendee_id'], 'tix_payment_token', true );
+
+		if ( '' === $echoed_reference || '' === $payment_token ) {
+			return false;
+		}
+
+		return hash_equals( $this->get_order_reference( $payment_token ), $echoed_reference );
+	}
+
+	/**
+	 * Whether a PayPal transaction paid at least what the given order asks for.
+	 *
+	 * The reference above is echoed faithfully but is not proof on its own, since a
+	 * payer outside Express Checkout sets `custom` themselves. The amount is what they
+	 * cannot set. "Covers" not "equals", because some sites charge a fee on top;
+	 * compared as integers because verify_order() accumulates the total in floats.
+	 *
+	 * @param array $order
+	 * @param array $txn_details
+	 *
+	 * @return bool
+	 */
+	protected function transaction_covers_order( $order, $txn_details ) {
+		// A free order never went to PayPal, so no transaction was ever made for one.
+		$due = isset( $order['total'] ) ? (float) $order['total'] : 0;
+
+		if ( $due <= 0 || ! isset( $txn_details['AMT'] ) ) {
+			return false;
+		}
+
+		$expected_currency = strtoupper( (string) $this->camptix_options['currency'] );
+		$paid_currency     = isset( $txn_details['CURRENCYCODE'] ) ? strtoupper( (string) $txn_details['CURRENCYCODE'] ) : '';
+
+		if ( '' === $paid_currency || $expected_currency !== $paid_currency ) {
+			return false;
+		}
+
+		// NVP amounts document the thousands separator as an optional comma, and "1,234.56" casts to 1.0.
+		$paid = (float) str_replace( ',', '', (string) $txn_details['AMT'] );
+
+		return (int) round( $paid * 100 ) >= (int) round( $due * 100 );
 	}
 
 	/**
@@ -517,10 +690,6 @@ class CampTix_Payment_Method_PayPal extends CampTix_Payment_Method {
 			wp_die( 'could not find order' );
 		}
 
-		/**
-		 * @todo maybe check tix_paypal_token for security.
-		 */
-
 		$payload = array(
 			'METHOD' => 'GetExpressCheckoutDetails',
 			'TOKEN'  => $paypal_token,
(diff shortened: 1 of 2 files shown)


bor0 (commented):
The core of this is right, and it is shaped the way I would want it shaped. Binding on the order's own `tix_transaction_id` first is the part that makes it deployable: every order settled before this ships already has one written by `payment_return()` in the same request that charged the card, so there is no migration, no legacy branch and no window where refunds arriving months later stop landing. And requiring it to be *that* transaction is what stops a later payment displacing the id `send_refund_request()` reaches for, which is the half of this that costs money rather than tickets.

I checked the compatibility claims against the code rather than taking them:

- `tix_payment_method` is written in `CampTix_Plugin::form_attendee_info()` at attendee creation (`camptix.php:6361`), before checkout, so the gateway check in `transaction_was_made_for_order()` can't refuse a first settlement on an order that simply hasn't been charged yet. That was the one thing that would have quietly broken live checkouts.
- `get_order_by_attendee_id()` casts a missing `tix_order` with `(array) ''`, which is `array( '' )` and truthy, so 1.1-migrated attendees still get an `attendee_id` back and still take the recorded-transaction arm. The `bare` fixture covers it.
- A failed `payment_return()` writes `tix_transaction_id` as `null` (`camptix.php:6751`), so it reads back `''` and a later genuine IPN for a charge we lost the response to still settles through the first-settlement arm rather than being refused as a mismatch.
- `payment_notify_back_compat()` resolves the attendee by transaction id and then hands over the stored token, so it matches on the recorded arm by construction.

The test data provider is doing real work — `no_api` pinning that the empty-status gate fires *before* the transaction fetch is the kind of assertion that stops a later refactor from reordering the checks and leaving every scenario still green.

Three things below. The first is the one I would like an answer on before approving, because it lands on the shared merchant account rather than on one camp.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
```diff
@@ -294,12 +324,38 @@ function payment_notify() {
...
 		);
 		$txn_details         = wp_parse_args( wp_remote_retrieve_body( $this->request( $txn_details_payload ) ) );
-		if ( ! isset( $txn_details['ACK'] ) || 'Success' != $txn_details['ACK'] ) {
-			$camptix->log( sprintf( 'Fetching transaction after IPN failed %s.', $txn_id, 0, $txn_details ) );
-			return;
+		if ( ! isset( $txn_details['ACK'] ) || 'Success' !== $txn_details['ACK'] ) {
+			// Nothing about the transaction is known, so ask for the IPN again rather than reporting it handled.
+			$camptix->log( sprintf( 'Fetching transaction after IPN failed %s.', $txn_id ), $order['attendee_id'], $txn_details );
+
+			$this->ipn_die( 503 );
```
bor0, line 335: This treats every `ACK !== 'Success'` as retryable, and I don't think all of them are.

The rule the PR sets out is that a refusal PayPal can never satisfy should answer 200, precisely because one merchant account serves ~910 sites and four days of resends is everyone's problem. A transaction id that this site's credentials will never be able to describe is exactly that kind of refusal — and the description leans on it deliberately: "No explicit merchant check: the API call runs on this site's credentials, so a transaction belonging to anyone else comes back `ACK=Failure` and never reaches the binding." That is the merchant check working as designed, and the reward for it working is 15 resends over four days.

The two cases really are different, though, which is why I don't think "just make it 200" is right either:

- No `ACK` at all — `$this->request()` returned a `WP_Error`, or a body that didn't parse. Nothing is known, 503 and a resend is exactly right, and that arm is already covered by `! isset( $txn_details['ACK'] )`.
- `ACK === 'Failure'` with a credentials error (10002 and friends). Also genuinely retryable: fix the API credentials inside four days and the resend settles the order. 503 is right here too, and it's the case that makes me not want a blanket flip.
- `ACK === 'Failure'` because the id isn't one this account can see (10004 and the invalid-argument family). Permanent. No resend can change it.

So the split I'd suggest is on the error code rather than on `ACK`:

```php
if ( ! isset( $txn_details['ACK'] ) || 'Success' !== $txn_details['ACK'] ) {
	$camptix->log( sprintf( 'Fetching transaction after IPN failed %s.', $txn_id ), $order['attendee_id'], $txn_details );

	$error_code = isset( $txn_details['L_ERRORCODE0'] ) ? (string) $txn_details['L_ERRORCODE0'] : '';

	// PayPal answered and will answer the same way to a resend.
	$this->ipn_die( in_array( $error_code, self::PERMANENT_FETCH_ERRORS, true ) ? 200 : 503 );
}
```

If you'd rather not enumerate PayPal error codes — which is a fair position, they are not stable and the list would rot — then I think the docblock on `ipn_die()` should say that this arm knowingly resends a permanently-unanswerable fetch, and why that was judged the safer default. Right now that docblock reads as though the 200/503 split is exhaustive, and this is the one place it isn't. Either resolution is fine by me; what I don't want is for the exception to be invisible.

Worth saying: `'a transaction PayPal will not describe'` in the provider pins the current behaviour, so whichever way this goes it wants a second row next to it.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
```diff
@@ -545,7 +714,31 @@ function payment_return() {
...
+			$echoed_reference = '';
+
+			foreach ( array( 'PAYMENTREQUEST_0_CUSTOM', 'CUSTOM' ) as $field ) {
+				if ( isset( $checkout_details[ $field ] ) && '' !== $checkout_details[ $field ] ) {
+					$echoed_reference = (string) $checkout_details[ $field ];
+					break;
+				}
+			}
+
+			if ( '' !== $echoed_reference && ! hash_equals( $this->get_order_reference( $payment_token ), $echoed_reference ) ) {
```
bor0: This builds the expected reference from the request's `$payment_token`, but `transaction_was_made_for_order()` deliberately doesn't, and the reason it gives applies here unchanged:

> compared against the order's stored token rather than the request's: `get_order()` matches with a MySQL CHAR comparison, which ignores case and trailing spaces, `hash_equals()` neither

`payment_return()` reads `$payment_token` with `trim()` at line 674 and hands it straight to `get_order()`, which resolves it under the column collation — so a token differing only in case or trailing whitespace still finds the order, and then `get_order_reference()` builds `<blog>:TOK_RET` against an echoed `<blog>:tok_ret` and `hash_equals()` refuses it.

It isn't reachable in practice: the token round-trips through PayPal's `RETURNURL`, which we set ourselves, so nothing mangles it, and the failure is a `wp_die()` before `DoExpressCheckoutPayment` runs — the buyer sees an error and is not charged, rather than being charged and losing the order. So this is consistency rather than a live defect.

Still, the note in `transaction_was_made_for_order()` is what makes it worth doing: it identifies a specific way the two lookups can disagree, and this is the other place they meet.

```php
$stored_token = (string) get_post_meta( $order['attendee_id'], 'tix_payment_token', true );

if ( '' !== $echoed_reference && ! hash_equals( $this->get_order_reference( $stored_token ), $echoed_reference ) ) {
```

The payload at line 716 should use the same value, or the `CUSTOM` sent on `DoExpressCheckoutPayment` disagrees with the one `payment_notify()` will later match against.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
```diff
@@ -294,12 +324,38 @@ function payment_notify() {
...
+			 * predate the reference below. Requiring that same transaction also stops a
+			 * later payment displacing the one an organizer would refund.
+			 */
+			if ( ! hash_equals( $recorded_txn_id, $txn_id ) ) {
+				$this->ignore_ipn( sprintf( 'this order was paid by %1$s, not %2$s', $recorded_txn_id, $txn_id ), $order, compact( 'payment_token', 'txn_details' ) );
+			}
+		} else {
+			// Nothing recorded, so the echoed reference ties them together, backed by the amount.
+			if ( ! $this->transaction_was_made_for_order( $order, $txn_details ) ) {
+				$this->ignore_ipn( sprintf( 'transaction %s was not made for this order', $txn_id ), $order, compact( 'payment_token', 'txn_details' ) );
```
bor0: Minor, and only about the log line.

`transaction_was_made_for_order()` refuses for three distinct reasons — the order was taken through another gateway, the transaction type isn't one this gateway starts, or the reference doesn't match — and all three land here as "transaction %s was not made for this order". The `$txn_details` in the context means someone can work out which, so nothing is lost, but they have to know to look.

The reason it's worth separating is that the three have very different meanings when they show up. A gateway mismatch or a reference mismatch is the check doing its job. An otherwise-good Express Checkout payment refused on an unexpected `TRANSACTIONTYPE` is a person who has been charged and has no ticket, and it is the only refusal in this method that can happen to a legitimate payment — the type list is our reading of what PayPal returns, and if PayPal ever returns a fourth spelling for a checkout we started, this is where it surfaces. That case only bites when the buyer never makes it back through `payment_return()`, so the IPN is the sole settlement path, which is also when nobody is watching.

Returning a reason string instead of a bool would make the log say which gate closed:

```php
$refusal = $this->reason_transaction_is_not_for_order( $order, $txn_details );

if ( $refusal ) {
	$this->ignore_ipn( sprintf( 'transaction %1$s was not made for this order: %2$s', $txn_id, $refusal ), $order, compact( 'payment_token', 'txn_details' ) );
}
```

Happy for this to be a follow-up, or to be told the context array is enough.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
mcliwanow: You're right. The description leans on `ACK=Failure` as the merchant check and then answers 503 to it, so the check working costs four days of resends on an account 910 sites share. Triggerable too, since `notify_url` is payer-settable.

I didn't want to enumerate codes I can't verify, having already guessed `expresscheckout` when PayPal's value is `express-checkout`. But the reference makes it smaller than expected: 10004 is the only error documented for `GetTransactionDetails`, and 10001 is documented against it too. 10002 and 10007 are not listed for this call.

```php
$refused_argument = isset( $txn_details['L_ERRORCODE0'] ) && '10004' === (string) $txn_details['L_ERRORCODE0'];

$this->ipn_die( $refused_argument ? 200 : 503 );
```

It doesn't need the list to be complete: a resend carries the same `TRANSACTIONID`, so whatever produced "invalid argument" produces it again. Default stays 503, so an unlisted code keeps today's behaviour, and only a wrongly added one could drop a payment. Your credentials case keeps its resend.

Documented, not observed. I have not seen a real 10004 from a foreign id, and PayPal's error page renders badly enough that I trust the operation mapping less than I'd like. The resend argument does not depend on it.

Provider row added, plus mutations both ways (10004 forced retryable, and every failure forced permanent) so neither direction can drift.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
mcliwanow: Agreed, fixed in both places. You're right that the payload has to move with the check, or the `CUSTOM` sent on `DoExpressCheckoutPayment` stops matching what `payment_notify()` compares later.

Added a return case where the request names the order as `TOK_RET` against a stored `tok_ret` and expects the charge to go through. It fails without the change.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
mcliwanow: Taking it now rather than as a follow-up, for your reason: the type case is the only refusal there that can reach somebody already charged, and it rests on our reading of what PayPal returns.

Now `reason_transaction_is_not_for_order()`, returning which gate closed: "the order was taken through another gateway", "unexpected transaction type web-accept", "the reference belongs to another order".

Writing the test caught something neither of us called out: it was logging the normalised `webaccept` rather than the `web-accept` PayPal actually sent, which is the string you would check against their docs. Fixed.

The rows assert the reason through `camptix_log_raw`, so a refactor cannot merge them back into one message and leave everything green.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
```diff
@@ -317,6 +382,129 @@ function payment_notify() {
...
+
+		/*
+		 * Only a checkout CampTix started can settle an order with no transaction yet;
+		 * a payment sent any other way had its CUSTOM chosen by the payer. Separators
+		 * are stripped because PayPal spells this differently across its APIs.
+		 */
+		$transaction_type = isset( $txn_details['TRANSACTIONTYPE'] ) ? (string) $txn_details['TRANSACTIONTYPE'] : '';
+		$comparable_type  = strtolower( preg_replace( '/[^a-zA-Z]/', '', $transaction_type ) );
+
+		if ( ! in_array( $comparable_type, array( 'cart', 'expresscheckout' ), true ) ) {
```
bor0, line 458: This is the one new gate that can turn away a payment somebody really made, as the comment above acknowledges — and pulling it out as its own log reason in the follow-up commit was the right call.

Worth noting that the first-settlement arm is warmer than the description suggests, though. `payment_return()` sets `PAYMENTREQUEST_0_NOTIFYURL` on `DoExpressCheckoutPayment` (~L729), so the notification that call triggers can land before that same request has written `tix_transaction_id`. Every live checkout has that window, and it is exactly where this check runs — not just the two pending orders the audit found at rest.

So if PayPal ever answers with a type outside `{cart, expresscheckout}` for a checkout we started, the notification is dropped with a 200 and never sent again, and the order settles only if the buyer completes the return leg.

Would you consider 503 for the unknown-type case specifically? It is the only reason here where a resend could plausibly help, it keeps the window open while somebody reads the log, and the other two reasons stay 200 as they are.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
```diff
@@ -260,32 +260,62 @@ function template_redirect() {
...
+			$camptix->log( 'Could not reach PayPal to verify an IPN.', 0, compact( 'response' ) );
+
+			$this->ipn_die( 503 );
+		}
+
+		if ( 'VERIFIED' !== trim( wp_remote_retrieve_body( $response ) ) ) {
 			$camptix->log( 'Could not verify PayPal IPN.', 0, null );
-			return;
+
+			$this->ipn_die( 403 );
```
bor0, line 290: This cuts slightly against the principle set out in `ipn_die()`'s docblock — that anything which will never settle answers 200, because resends on the shared account are everyone's problem.

A `_notify-validate` failure is deterministic in practice: the resend carries the same payload and earns the same answer. So 403 buys four days of resends for precisely the case it is aimed at, while the far more common source of unconfirmed messages never resends at all.

I am not arguing for a particular answer, only that this reads as a consequence of "non-200 means resend" rather than a decision taken on its own terms. If it is deliberate, a line in the docblock saying why this one is worth resending would keep somebody from quietly flipping it to 200 later.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
```diff
@@ -317,6 +382,129 @@ function payment_notify() {
...
+	 * @param array $order
+	 * @param array $txn_details
+	 *
+	 * @return bool
+	 */
+	protected function transaction_covers_order( $order, $txn_details ) {
+		// A free order never went to PayPal, so no transaction was ever made for one.
+		$due = isset( $order['total'] ) ? (float) $order['total'] : 0;
+
+		if ( $due <= 0 || ! isset( $txn_details['AMT'] ) ) {
```
bor0: Minor: the zero-total case is refused here but logged by the caller as *"does not cover this order"*, which reads oddly for an order that has nothing to cover.

The follow-up commit gave `reason_transaction_is_not_for_order()` distinct reasons for exactly this kind of blurring. The same treatment here would let a free order be told apart from a genuine underpayment in the log, which are quite different things to find.

bor0 (commented):
Reviewed and ran this locally.

**Verified.** Full PHPUnit suite 886 passing / 6 pre-existing skips; `--filter PayPal` 34 tests, 103 assertions; phpcs clean on both changed files. I also drove `payment_notify()` over HTTP against a stubbed PayPal on a dev site, on this branch and on `production`'s copy of the file. The settle/refuse outcomes and the status codes matched what the description claims in every case I tried, and the legitimate paths still settle identically on both — own transaction under both type spellings, refund via `parent_txn_id`, and the 1.1 notify URL.

**One thing worth stating in the compatibility notes.** `payment_notify()` now gates on `get_order()` before anything reaches `payment_result()`, and the two use different post-status filters — `'any'` here, an explicit list there. That turns out fine, because CampTix registers `cancel` / `failed` / `timeout` / `refund` with `exclude_from_search = false`, so `'any'` is a superset. But it means the recent cancel-then-webhook behaviour in `payment_result()` is load-bearing on a detail that isn't visible in this diff, so it's worth a line.

Three notes inline, none blocking. One more minor: `payment_return()` still reads the request token with `trim()`, while `payment_notify()` moved to `sanitize_text_field( wp_unslash( ... ) )`. Not a behaviour difference — the reference is built from the stored token either way — just inconsistent now that they sit next to each other.

Also worth rebasing on `production` before merge; the branch is 22 commits behind.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
mcliwanow: thanks, done

mcliwanow:
> Three notes inline, none blocking. One more minor: payment_return() still reads the request token with trim(), while payment_notify() moved to sanitize_text_field( wp_unslash( ... ) ). Not a behaviour difference — the reference is built from the stored token either way — just inconsistent now that they sit next to each other.

Thanks for the suggestions. Updated. Looking into remaining comments

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
mcliwanow: It's deliberate, and the docblock gives the reason: we want PayPal to send it again, because a genuine IPN can fail validation just by getting damaged on the way. The resend is a fresh copy, not the same bytes, so it can pass where the first attempt didn't.

public_html/wp-content/plugins/camptix/addons/payment-paypal.php
mcliwanow: Good call on the race. I had it as two rare events, and it is a standing condition times a per-checkout failure rate.

I went back through the network log. GetTransactionDetails has failed after an IPN only a handful of times in the last several years, and nearly all of those sit inside one short window where the shared credentials were failing authentication. A few of the same transactions were served later, but months later, so none of it looks like a transaction that was briefly unavailable. The refused-argument code does not appear at all.

So that special case is gone. Every GetTransactionDetails failure now answers 503 and asks for a resend, which would have recovered that window in days rather than months.

On the type check: all transactions on record are `cart`. A payment made through our own checkout cannot reach that reason, including in the race you describe, since it comes back `cart` with our reference and settles. Left at 200.

bor0 (approved):
Re-reviewed at `31bf8641`. Everything I raised last time is answered, and the two follow-up commits are the shape I was hoping for.

**Verified locally.** Full suite on this branch: 1159 tests, 3146 assertions, 1 failure and 6 skips — the failure is `Test_QuickBooks_OAuth_State::test_authorize_url_carries_the_stored_state`, which fails identically on `origin/production`, and the skips are the five `intl` ones plus the GatherPress contract. `--filter PayPal` is 34 tests / 106 assertions green. `BASE_REF=origin/production php .github/bin/phpcs-branch.php` is clean. Branch is now two commits behind `production` rather than twenty-two.

**On dropping the 10004 special case.** You went and looked, which is the right way to settle it — one credentials window, error 10002, thirteen camps, and those transactions came back once it passed. That is exactly the case my original comment said was worth resending, and 200 would have held those orders open for months instead of days. The nice side effect is that `ipn_die()`'s docblock is exhaustive again: there is no longer an arm that knowingly answers the "wrong" code, which was the thing I did not want left invisible. The residual exposure — a payment to the shared merchant with `notify_url` aimed here, failing the fetch on this site's credentials and earning four days of resends — is bounded at fifteen retries for one message and has never appeared on record. Fine by me.

**On the type check staying 200.** Accepting that on your evidence: every transaction on record is `cart`, and a checkout we started comes back with our reference and settles, race or no race. It now has its own log reason, so if PayPal ever does return a fourth spelling it is findable rather than buried in `$txn_details`.

`reason_transaction_does_not_cover_order()` reads well — "the order has nothing to pay" against "paid 0.01 of 750" is the distinction I was after, and pinning both through `camptix_log_raw` keeps them from collapsing back together. The `payment_return()` request reads now match `payment_notify()`'s, and the reference on both sides is built from the stored token, so the `CUSTOM` sent on `DoExpressCheckoutPayment` and the one compared on the IPN cannot drift apart.

Nothing further from me. Worth pulling `production` in before merge, but that is the only thing left.

PR #1955, opened 2026-08-26
Budget: give the payment post types an explicit capability model

KokkieH:
## Summary

The three payment post types — Reimbursement Requests, Vendor Payments and Sponsor Invoices — are placed under the **Budget** menu with `show_in_menu`, but they had no capability model of their own, so they inherited the generic `post` capabilities. That left the capability a screen enforces dependent on the menu placement rather than on the post type itself.

This gives each of them an explicit capability model aligned with the Budget menu, so the list, creation and deletion paths, and the queries behind them, all enforce the same access.

## Changes

- Map each payment post type's list, creation and deletion capabilities to `WordCamp_Budgets::VIEWER_CAP` (the same capability the Budget menu uses), so the screens and the menu agree.
- Keep `map_meta_cap` enabled — supplying a `capabilities` array otherwise disables the mapping core turns on by default — so per-record edit/delete checks keep routing through the existing `modify_capabilities()` logic.
- Add a `pre_get_posts` guard as a second line of defence, so the admin list query for these post types only returns rows to Budget viewers.
- Gate the sponsor **Invoice Sponsor** metabox on the invoice creation capability, so its contents follow the same access as the invoice screens.
- Add PHPUnit coverage for the capability map, `map_meta_cap`, and the per-role edit/delete outcome.

## Testing

Automated: `phpunit --testsuite "WordCamp Budgets Dashboard"` (adds `test-post-type-capabilities.php`).

Manual — confirms the Budget screens enforce the same capability consistently across roles. Run on a local environment (`<camp-url>` = a camp site, e.g. `https://seattle.wordcamp.test/2023/`).

1. Make sure some payment records exist. As an admin, open **Budget → Vendor Payments / Reimbursements / Sponsor Invoices** and add one of each (or use existing records).
2. Create a user of each role on the camp site:
   ```
   wp user create budgeteditor  editor@example.test  --role=editor      --url=<camp-url>
   wp user create budgetauthor  author@example.test  --role=author      --url=<camp-url>
   wp user create budgetcontrib contrib@example.test --role=contributor --url=<camp-url>
   ```
3. As **admin** or **editor**: confirm the **Budget** menu appears, all three list screens load and show records, and you can create and save a record.
4. As an **author**: confirm the Budget menu appears and the screens load (own records shown by default; `&all_posts=1` shows all).
5. As a **contributor**: confirm the Budget menu does not appear, and opening a screen directly (`/wp-admin/edit.php?post_type=wcb_reimbursement`) returns a permissions error instead of the list.

https://claude.ai/code/session_0129jm2mweoSuZPkqazWaXd6


bor0 (commented):
Reviewed at `92cdd7ca` by fetching the branch, reading the surrounding capability logic, and measuring the resulting capability map against `production` in the phpunit container. Suite on the branch: **897 tests, 2124 assertions, 0 failures, 6 skipped** — unchanged from the `production` it merged, which is the point below. `phpcs-branch.php` clean.

The change is correctly scoped, and `map_meta_cap => true` is the load-bearing part rather than a formality. Measured, registering the same args three ways:

```
bare (no capabilities, no map_meta_cap)   map_meta_cap = true
capabilities supplied, map_meta_cap unset map_meta_cap = false
map_meta_cap => false                     map_meta_cap = false
```

So the comment is right: supplying `capabilities` really does switch off the mapping core otherwise turns on, and without the explicit `true` every `edit_post` check would map to the primitive `'edit_post'`, which no role holds. Good catch, and it's the kind of thing that would have been found in production rather than review.

The resulting delta is exactly what the summary claims and nothing else:

| cap key | production | branch |
|---|---|---|
| `edit_posts` | `edit_posts` | `publish_posts` |
| `create_posts` | `edit_posts` | `publish_posts` |
| `delete_posts` | `delete_posts` | `publish_posts` |
| `edit_others_posts`, `edit_published_posts`, `delete_others_posts`, `publish_posts`, `read_private_posts`, and the three meta caps | — | unchanged |

And by role, `current_user_can( 'edit_post', <own draft> )` / the list screen:

| role | own draft | list screen |
|---|---|---|
| administrator, editor, author | true → true | true → true |
| **contributor** | **true → false** | **true → false** |
| subscriber | false → false | false → false |

Someone else's record is unchanged for every role. So contributors are the entire behavioural delta, which is the stated intent.

---

### 1. No test coverage

This is my one real ask. It's an access-control change across three post types and two plugins, and the suite count is identical before and after, so nothing in CI would notice if any of it regressed.

The `map_meta_cap => true` line is the case that worries me most. It's a one-line deletion away from mapping `edit_post` to a primitive nobody holds, the comment above it is the only thing protecting it, and the failure mode is "nobody below network admin can edit any payment record" — total, silent, and not visible in any existing test. The three tables above are more or less the test: assert the cap map for each of the three types, assert the contributor delta, and assert `get_post_type_object( … )->map_meta_cap` is still true.

### 2. What is `restrict_payment_queries()` protecting?

Not a defect, but the docblock doesn't let the next reader work it out. It says "The list screen and the post type it queries are resolved from separate request parameters" — in `edit.php` both come from `$_GET['post_type']`, and core checks `$post_type_object->cap->edit_posts` there before it builds the list table. With `edit_posts` now mapped to `VIEWER_CAP`, that check already turns away exactly the users this guard filters, before any query runs. So as far as I can find it never fires on the screen it describes.

Worth keeping as belt-and-braces, but either name the query it's actually for or say plainly that it's a second line of defence. Two things to note while it's there: `post_type => 'any'` skips it, since `array_intersect( array( 'any' ), PAYMENT_POST_TYPES )` is empty; and the literal-slug list has to stay in step with the three `POST_TYPE` constants by hand (they match today — I checked). The reason given for the literals is sound: `wcp_payment_request` and `wcb_reimbursement` aren't registered on cron, where only `sponsor-invoice.php` loads.

### 3. Existing records owned by contributors

Worth confirming before merge rather than after. Any payment record whose author holds only `contributor` on that camp site becomes invisible and uneditable to the person who filed it, and the `draft_post` change in `modify_capabilities()` stops them saving further drafts of one. The testing steps check that a contributor loses the menu, which is the intent, but not what happens to a request they already own and have in flight.

Contributor is an ordinary role for organizers to hand to a volunteer on a camp site, so this isn't hypothetical by construction. A count across the network of `wcb_reimbursement` / `wcp_payment_request` / `wcb_sponsor_invoice` rows whose author holds only contributor on that site would settle whether anyone is actually affected, and if some are, they need to be re-roled rather than surprised.

### 4. Minor

The summary says "list and creation capabilities"; `POST_TYPE_CAPABILITIES` also maps `delete_posts`, as its own docblock says. Worth aligning the summary.

---

### Verified, recorded so nobody re-derives it

- All three types are `public => false`, so none of this touches a front-end surface, and `has_archive => true` alongside it is pre-existing and inert.
- The metabox's `\WordCamp\Budgets\Sponsor_Invoices\POST_TYPE` reference adds no new cross-plugin coupling — `wc-post-types.php:1362` already resolves the same constant to build `$existing_invoices` — and `wcb_sponsor_invoice` is registered on `init`, well before metaboxes render, so `get_post_type_object()` can't return null there.
- The new `require_once` in `bootstrap.php` is inert on cron: `wordcamp-budgets.php` is a class definition with no file-scope side effects, and it's correctly ordered ahead of `sponsor-invoice.php`.
- `restrict_payment_queries()` is only hooked when `is_admin()`, since the constructor runs inside that block — so the `is_admin()` check inside the method is redundant but harmless, and cron never reaches it.
- `post__in => array( 0 )` does reliably return nothing: WP_Query treats the array as non-empty and emits `ID IN (0)`.

The branch is two commits behind `production` (`20c11d47`, `27ebc02a`), which accounts for 897 against 919 on current `production`.

bor0 (approved):
None of the comments are merge blockers

mcliwanow:
Adding to @bor0's review: one behavioural gap in the metabox change itself. Gating only the "Add New Invoice" button leaves the "Existing Invoices" list above it visible. It's fed by the plain get_posts() in metabox_invoice_sponsor(), so restrict_payment_queries() never touches it, and a contributor who authored a draft/pending sponsor (invoiceable per sponsor-invoice.php:222-226) sees that sponsor's invoice titles even though the list screen now turns them away. Moving the check up to wrap the whole else branch closes it and also avoids the box rendering empty.

Minor: the new wordcamp-budgets.php require in bootstrap.php duplicates the one in the is_admin() block above; worth hoisting one above both so the cron-critical copy doesn't read as deletable.

KokkieH:
> 3. Existing records owned by contributors

Currently the budget menu is visible to contributors on production, but clicking through to any of the payment post type list screens in that menu leads to the generic 403 error page.

So the UI would seem to indicate it shouldn't be possible for contributors to create/manage these post types. However, it is currently possible for them to create, e.g. reimbursement requests, via the admin bar New menu, and to edit these posts via direct link. So checking for existing cases is prudent.

While there are several cases across the network, there are only posts from two contributor users across two sites covering 2025 and 2026, and in both cases the invoices/reimbursements already show as paid, so I think it's safe to close the gap--once a record is past draft status a contributor cannot edit it anyway, so we're only removing read-visibility of completed work, making the experience consistent for contributors across both list and editor views for these post types.

Addressed the other feedback in 8ea3161, most notably, added test coverage, and gated the Invoice Sponsor metabox body for contributors so it's consistent with other locations.

PR #1956, opened 2026-08-26
Importer: don't import post meta that names a file on disk

mcliwanow:
## Why

The WXR importer copies post meta straight out of the export file. A few keys hold a filesystem path rather than data, and a path only means something on the site that produced the export. The importer does not bring the files across, so after an import those values name a location that has nothing to do with the imported post.

The plugin already knows this. `is_valid_meta_key()` skips two of them:

```php
if ( in_array( $key, array( '_wp_attached_file', '_wp_attachment_metadata', '_edit_lock' ), true ) ) {
	return false;
}
```

The comment above it says "skip attachment metadata since we'll regenerate it from scratch", which is accurate for those two: the importer downloads the attachment and rebuilds both from the real file. That list predates WordPress 6.5, which added the font library and with it a third path-bearing key, `_wp_font_face_file`. It was never added, so the importer stores whatever the export file says, and nothing rebuilds it afterwards because font files are not attachments and never get fetched.

The key is protected meta, so the REST `meta` field, the Custom Fields panel, and `wp_ajax_add_meta` all refuse it. The font-face REST controller writes it through `wp_handle_upload()`, which normalizes the name. The importer stores the raw value.

Two deployment details make this worth handling on our side rather than waiting for the plugin:

- `wordpress-importer` is network-activated on wordcamp.org (`wcorg-network-plugin-control.php:60`), so it runs on every camp site.
- We install it from `wordpress-importer.latest-stable.zip` (`composer-dev.json`) rather than pinning a version. The vendored copy is replaced on every install, so it cannot be patched in place. Upstream `master` still ships the same list.

Same shape as #1919, one importer over.

## What changed

- New mu-plugin `importer-tweaks.php` filters `import_post_meta_key` and skips `_wp_attached_file`, `_wp_attachment_metadata`, and `_wp_font_face_file`.
- The two attachment keys are listed again even though the plugin skips them itself. The plugin is unpinned, so this keeps the behavior the same whichever release lands.
- The filter names the keys it skips. A rule based on the `_` prefix would have been shorter, but the importer depends on `_thumbnail_id` and the `_menu_item_*` family, so it would drop featured images and nav menus on every import.

## Testing

13 new tests in `mu-plugins/tests/test-importer-tweaks.php`: each of the three keys is skipped; `_thumbnail_id`, `_wp_page_template`, three `_menu_item_*` keys, `_wp_attachment_image_alt`, and an unprotected key all still import; and non-string values (`false`, `null`, an array) pass through untouched, since another filter's return value has to survive this one.

The keep-list cases are the ones a prefix rule would break, so they are there to fail if someone later simplifies the filter that way.

Tests go through `apply_filters()` rather than calling the function, so hook registration is covered too. Commenting out the `add_filter()` line fails three of them.

Suite on this branch is 217 tests, 336 assertions, no failures. Clean `origin/production` in the same environment is 204 tests, 323 assertions, also no failures. So this adds 13 tests and no failures. `phpcs` is clean on both new files.

Compatibility, checked because it is the part that would fail quietly:

- Nothing in `wp-content` reads or writes `_wp_font_face_file`. The only match outside the new files is Gutenberg's own bundled handler.
- No WXR fixture in the repo contains a `wp_font_face` item.
- `wc-fonts` is Typekit and Google Fonts, unrelated to the core font library.
- An imported font face keeps its post, its title, and its parent. It loses a value that pointed at a file the import never brought across.

## How to review this locally

Check out the branch and restart the container so opcache cannot serve the old bytecode:

```
git checkout claude/importer-meta
```

**1. Write a small WXR** with a font face carrying one path-bearing key and two ordinary ones:

```
docker exec -i wordcamptest-wordcamp.test-1 sh -c 'cat > /tmp/font-face.xml' <<'XML'
<?xml version="1.0" encoding="UTF-8" ?>
<rss version="2.0" xmlns:wp="http://wordpress.org/export/1.2/" xmlns:content="http://purl.org/rss/1.0/modules/content/" xmlns:excerpt="http://wordpress.org/export/1.2/excerpt/">
<channel><title>t</title><link>https://example.test</link>
<wp:wxr_version>1.2</wp:wxr_version><wp:base_site_url>https://example.test</wp:base_site_url><wp:base_blog_url>https://example.test</wp:base_blog_url>
<item><title>ReviewFamily</title><content:encoded><![CDATA[]]></content:encoded><excerpt:encoded><![CDATA[]]></excerpt:encoded>
<wp:post_id>9920</wp:post_id><wp:status><![CDATA[publish]]></wp:status><wp:post_parent>0</wp:post_parent><wp:post_type><![CDATA[wp_font_family]]></wp:post_type></item>
<item><title>ReviewFace</title><content:encoded><![CDATA[]]></content:encoded><excerpt:encoded><![CDATA[]]></excerpt:encoded>
<wp:post_id>9921</wp:post_id><wp:status><![CDATA[publish]]></wp:status><wp:post_parent>9920</wp:post_parent><wp:post_type><![CDATA[wp_font_face]]></wp:post_type>
<wp:postmeta><wp:meta_key><![CDATA[_wp_font_face_file]]></wp:meta_key><wp:meta_value><![CDATA[review-font.ttf]]></wp:meta_value></wp:postmeta>
<wp:postmeta><wp:meta_key><![CDATA[_thumbnail_id]]></wp:meta_key><wp:meta_value><![CDATA[4242]]></wp:meta_value></wp:postmeta>
<wp:postmeta><wp:meta_key><![CDATA[_menu_item_type]]></wp:meta_key><wp:meta_value><![CDATA[custom]]></wp:meta_value></wp:postmeta>
</item></channel></rss>
XML
```

**2. Import it.** The importer logs each key it stores. `_wp_font_face_file` should not be among them:

```
docker exec wordcamptest-wordcamp.test-1 wp --url=https://seattle.wordcamp.test/2023/ --allow-root \
  import /tmp/font-face.xml --authors=skip
```

```
-- Imported post as post_id #9921
-- Added post_meta _thumbnail_id
-- Added post_meta _menu_item_type
```

On `production` the same run adds a third line, `-- Added post_meta _wp_font_face_file`.

**3. Confirm what was stored.** The font face keeps its ordinary meta and has no file key:

```
docker exec wordcamptest-wordcamp.test-1 wp --url=https://seattle.wordcamp.test/2023/ --allow-root \
  post meta list 9921 --format=csv
```

```
post_id,meta_key,meta_value
9921,_thumbnail_id,4242
9921,_menu_item_type,custom
```

**4. Confirm the post itself imported.** Nothing about the post is dropped, only the one meta key:

```
docker exec wordcamptest-wordcamp.test-1 wp --url=https://seattle.wordcamp.test/2023/ --allow-root \
  post list --post_type=wp_font_face --fields=ID,post_title,post_status,post_parent --format=csv
```

**5. Remove the setup:**

```
docker exec wordcamptest-wordcamp.test-1 wp --url=https://seattle.wordcamp.test/2023/ --allow-root \
  post delete 9921 9920 --force
docker exec wordcamptest-wordcamp.test-1 rm -f /tmp/font-face.xml
```

### Automated

```
docker compose -f docker-compose.phpunit.yml exec phpunit_wp phpunit --group importer-tweaks
```

### DIFF
--- a/public_html/wp-content/mu-plugins/importer-tweaks.php
+++ b/public_html/wp-content/mu-plugins/importer-tweaks.php
@@ -29,10 +29,6 @@ add_filter( 'import_post_meta_key', __NAMESPACE__ . '\skip_file_path_meta' );
  * @return mixed The meta key, or false to skip it. Anything that isn't a string is left alone.
  */
 function skip_file_path_meta( $key ) {
-	if ( ! is_string( $key ) ) {
-		return $key;
-	}
-
 	$file_path_meta = array(
 		'_wp_attached_file',
 		'_wp_attachment_metadata',
--- a/public_html/wp-content/mu-plugins/importer-tweaks.php
+++ b/public_html/wp-content/mu-plugins/importer-tweaks.php
@@ -29,15 +29,25 @@ add_filter( 'import_post_meta_key', __NAMESPACE__ . '\skip_file_path_meta' );
  * @return mixed The meta key, or false to skip it. Anything that isn't a string is left alone.
  */
 function skip_file_path_meta( $key ) {
-	$file_path_meta = array(
-		'_wp_attached_file',
-		'_wp_attachment_metadata',
-		'_wp_font_face_file',
-	);
-
-	if ( in_array( $key, $file_path_meta, true ) ) {
+	if ( in_array( $key, file_path_meta_keys(), true ) ) {
 		return false;
 	}
 
 	return $key;
+}
+
+/**
+ * The post meta keys that store a filesystem path.
+ *
+ * Shared with the Jetpack importer, which writes meta on its own REST routes rather than through
+ * `import_post_meta_key`, so both importers drop the same keys.
+ *
+ * @return string[]
+ */
+function file_path_meta_keys() {
+	return array(
+		'_wp_attached_file',
+		'_wp_attachment_metadata',
+		'_wp_font_face_file',
+	);
 }
--- a/public_html/wp-content/mu-plugins/importer-tweaks.php
+++ b/public_html/wp-content/mu-plugins/importer-tweaks.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Tweaks for the WordPress Importer.
+ *
+ * @package WordCamp\Importer_Tweaks
+ */
+
+namespace WordCamp\Importer_Tweaks;
+
+defined( 'WPINC' ) || die();
+
+add_filter( 'import_post_meta_key', __NAMESPACE__ . '\skip_file_path_meta' );
+
+/**
+ * Skip imported post meta that stores a filesystem path.
+ *
+ * A WXR file carries meta verbatim from the site that produced it, but the importer does not bring
+ * the referenced files across, so a path value lands pointing at something that has nothing to do
+ * with the imported post. The importer downloads attachments and rebuilds the two attachment keys
+ * from the real file. Nothing rebuilds `_wp_font_face_file`, so an imported font face loses its file
+ * binding rather than keeping a path that means nothing here.
+ *
+ * The importer's own `is_valid_meta_key()` skips the two attachment keys already. It has not picked
+ * up `_wp_font_face_file`, which the font library added in 6.5, and the plugin is installed from the
+ * latest published release rather than pinned here, so all three are listed.
+ *
+ * @param mixed $key The meta key, or whatever a lower-priority filter returned in its place.
+ *
+ * @return mixed The meta key, or false to skip it. Anything that isn't a string is left alone.
+ */
+function skip_file_path_meta( $key ) {
+	if ( ! is_string( $key ) ) {
+		return $key;
+	}
+
+	$file_path_meta = array(
+		'_wp_attached_file',
+		'_wp_attachment_metadata',
+		'_wp_font_face_file',
+	);
+
+	if ( in_array( $key, $file_path_meta, true ) ) {
+		return false;
+	}
+
+	return $key;
+}
--- a/public_html/wp-content/mu-plugins/jetpack-tweaks/import-meta.php
+++ b/public_html/wp-content/mu-plugins/jetpack-tweaks/import-meta.php
@@ -8,19 +8,24 @@
  * classes -- and hand the plain result back to the importer. Normal imports keep working, and the
  * importer only ever receives plain data from the request.
  *
+ * The same routes also drop the meta keys that store a filesystem path, which the WXR importer
+ * drops through `import_post_meta_key`. These routes write meta directly, so they need their own
+ * pass over the same list.
+ *
  * @package WordCamp\Jetpack_Tweaks
  */
 
 namespace WordCamp\Jetpack_Tweaks\Import_Meta;
 
 use WP_REST_Request;
+use function WordCamp\Importer_Tweaks\file_path_meta_keys;
 
 defined( 'WPINC' ) || die();
 
 add_filter( 'rest_request_before_callbacks', __NAMESPACE__ . '\normalize_import_meta', 10, 3 );
 
 /**
- * Replace serialized meta values on the Jetpack import routes with plain, object-free values.
+ * Drop file path meta and replace serialized values on the Jetpack import routes.
  *
  * @param mixed $response The response so far. Returned unchanged.
  * @param array $handler  The matched route handler.
@@ -47,7 +52,15 @@ function normalize_import_meta( $response, $handler, $request ) {
 
 	$changed = false;
 
+	$file_path_meta = file_path_meta_keys();
+
 	foreach ( $metas as $key => $value ) {
+		if ( in_array( $key, $file_path_meta, true ) ) {
+			unset( $metas[ $key ] );
+			$changed = true;
+			continue;
+		}
+
 		$clean = decode_without_objects( $value );
 
 		if ( $clean !== $value ) {
--- a/public_html/wp-content/mu-plugins/tests/bootstrap.php
+++ b/public_html/wp-content/mu-plugins/tests/bootstrap.php
@@ -35,6 +35,7 @@ function manually_load_plugins() {
 	require_once dirname( __DIR__ ) . '/wcorg-subroles.php';
 	require_once dirname( __DIR__ ) . '/wcorg-network-theme-control.php';
 	require_once dirname( __DIR__ ) . '/jetpack-tweaks/import-meta.php';
+	require_once dirname( __DIR__ ) . '/importer-tweaks.php';
 }
 
 tests_add_filter( 'muplugins_loaded', __NAMESPACE__ . '\manually_load_plugins' );
--- a/public_html/wp-content/mu-plugins/tests/test-importer-tweaks.php
+++ b/public_html/wp-content/mu-plugins/tests/test-importer-tweaks.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace WordCamp\Tests;
+
+use WP_UnitTestCase;
+
+defined( 'WPINC' ) || die();
+
+/**
+ * Class Test_Importer_Tweaks
+ *
+ * @group mu-plugins
+ * @group importer-tweaks
+ *
+ * @package WordCamp\Tests
+ */
+class Test_Importer_Tweaks extends WP_UnitTestCase {
+
+	/**
+	 * Run a meta key through the registered `import_post_meta_key` filters.
+	 *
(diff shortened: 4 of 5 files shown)


bor0 (commented):
Reviewed locally: checked out the branch, ran the suite, and exercised the import path in the dev container.

### Verification

| Check | Result |
|---|---|
| `phpunit --group importer-tweaks` | 13 tests, 13 assertions, OK |
| Full suite | 887 tests, 2092 assertions, 6 skipped, 0 failures |
| `phpcs` on both new files | clean |
| Functional, this branch | `_wp_font_face_file` skipped; `_thumbnail_id` and `_menu_item_type` still imported |
| Functional, mu-plugin disabled (i.e. `production` today) | `_wp_font_face_file` stored verbatim |

The code is correct. `WP_Import::process_post_meta()` (`class-wp-import.php:981-995`) does `if ( ! $key ) return;`, so returning `false` is the right contract, and the mu-plugin registers at load time so it runs ahead of the plugin's own `is_valid_meta_key()` — `false` propagates through it unchanged. Root-level mu-plugins auto-load, so no loader change was needed. Nothing else in `wp-content` hooks `import_post_meta_key`, so there is no other regression surface.

### Findings

**1. The other importer isn't covered.** Jetpack's Unified Importer (`jetpack-import/src/endpoints/class-post.php:156-180`, the route #1919 hardened) writes meta with `add_post_meta()` and never applies `import_post_meta_key`. Its only exclusion is keys containing `_oembed`. So `_wp_font_face_file` — and `_wp_attached_file` — still go through untouched on that path. Since #1919 established those routes are in use here, this PR closes one of two doors. Either scope that in, or note it explicitly as a follow-up so it doesn't get lost. This is the one I'd like answered before merge.

**2. The imported value is not inert, so I'd raise the priority.** The description reads as though a stale path just sits there unused. It doesn't — core reads `_wp_font_face_file` back later in the post lifecycle and acts on it, and I confirmed in a local environment that an imported value is used as-is. That makes this more than hygiene, and it also raises the question of whether sites that have already imported content need a cleanup pass, plus an upstream report. I'll follow up through the usual private channel with the specifics rather than here; the current wording of the PR is fine as it stands and I'd keep it that way.

**3. The `is_string()` guard is redundant.** In `importer-tweaks.php`, `in_array( $key, $file_path_meta, true )` already returns `false` for `false`, `null`, and arrays, so the function returns `$key` unchanged without the guard — and three of the 13 tests only exercise that branch. Harmless as a statement of intent; drop it if you'd rather not pin behaviour that can't differ.

**4. Small inconsistency in the rationale.** The "this also covers callers that drive `WP_Import` methods without `import()`" argument would apply equally to `_edit_lock`, which the plugin skips and this filter doesn't. Either the list is "keys that name a file" (matching the function name, in which case the claim about non-`import()` callers overreaches) or it's "the plugin's list plus one". The code is fine; it's the description I'd trim.

**5. Out of scope, but worth knowing.** An imported font face keeps its `font_face_settings` JSON in `post_content`, whose `src` still points at the source site's URL. Dropping the meta is right, but the post is still only half-meaningful afterwards — whether `wp_font_face` / `wp_font_family` items should be imported at all is a separate question.

None of these block merge.

mcliwanow:
Thanks, that's a thorough pass. Answers in order.

**1. Jetpack's Unified Importer.** You're right that it writes meta with a bare `add_post_meta()` and never applies `import_post_meta_key`. I checked whether either key can reach anything through it, and for these two it can't:

- `_wp_font_face_file` needs to land on a `wp_font_face` post to matter. Every endpoint fixes its post type in the constructor (`attachment`, `page`, `wp_block`, `wp_template`, `nav_menu_item` and so on), none of them is `wp_font_face` or `wp_font_family`, and nothing reads a post type from the request.
- `_wp_attached_file` does go through untouched, but core resolves it through `wp_delete_file_from_directory()` on every leg of `wp_delete_attachment_files()`, including `wp_delete_file_from_directory( $file, $uploadpath['basedir'] )` for the main file. That helper does the `realpath()` prefix check, so a value outside the uploads tree is refused.

So it isn't a second door on this particular path. It is still a route that accepts arbitrary protected meta keys on `current_user_can( 'import' )`, which I agree is worth its own look, and I'd rather do that as a follow-up than widen this. Happy to open it now if you'd prefer it tracked before merge.

**2.** Understood, and agreed on keeping the wording as it is. I'll pick it up on the private channel, including the question about already-imported content.

**3.** Dropped. You're right that strict `in_array()` already returns the key unchanged for a non-string, so the guard could not change the result. I kept the three tests, since they pin the contract rather than the guard, and they still pass without it.

**4.** Trimmed from the description. The list is "keys that name a file", and the non-`import()` argument didn't fit that, as you say.

**5.** Agreed, and noted. Whether `wp_font_face` and `wp_font_family` should be importable at all is a better question than the one this PR answers, but a separate one.

mcliwanow:
On the second thought, I'll fold the Jetpack fix too. It's not reachable but it's super easy to patch and will keep us secure

mcliwanow:
Folded it in as a separate commit, so it can be dropped without touching the rest.

Both importers now drop the same keys from one shared list in `importer-tweaks.php`. The Jetpack side reuses the existing `rest_request_before_callbacks` hook from #1919, which already guards the route prefix and rewrites the `meta` param, so it is a few lines inside that pass.

Worth restating that this is consistency rather than a live gap: as above, no endpoint can create a `wp_font_face` post, and core resolves `_wp_attached_file` through `wp_delete_file_from_directory()`.

Two things in there worth a look, since they touch #1919:

- **`_wp_attachment_metadata` is now dropped on those routes too.** That is safe because `WP_REST_Attachments_Controller::create_item()` regenerates it with `wp_generate_attachment_metadata()` from the uploaded file, and Jetpack's endpoint calls `parent::create_item()`. The value the client sends is redundant, and since the trait adds meta after the parent runs, dropping it also avoids a second row landing on top of the generated one.
- **One of your tests moved off that key.** `test_serialized_array_round_trips` used `_wp_attachment_metadata` as its serialized fixture, which the drop list now removes. The test is about the value rather than the key, so it uses `tix_questions` now. Shout if you would rather I left that key alone on this path.

Suite is 222 tests, 343 assertions, no failures, and `phpcs` is clean on all four files.

bor0 (approved):
Re-reviewed at `4d80a2a1`, including the folded-in Jetpack commit. All five points from the last round are settled, and I traced the new commit rather than taking the reasoning for it.

**The Jetpack side is safe, and for a narrower reason than the one in your reply.** Arbitrary meta writing on those routes exists in exactly one place — `Post::process_post_meta()` (`class-post.php:156`). `add_post_meta`/`update_post_meta` appear nowhere else in `jetpack-import/src`, and only `Post` and `Page` (which extends it) call it. Every other endpoint, `Attachment` included, goes through the ordinary REST `meta` field, which won't write an unregistered protected key at all. So `_wp_attachment_metadata` was never writable on the attachment route to begin with — the regeneration argument holds, but the drop is a no-op there rather than a behaviour change, and on `/posts` and `/pages` none of the three keys means anything. Nothing to regress.

Verified live in the dev container rather than only through the unit tests:

```
/jetpack/v4/import/posts       => _thumbnail_id, tix_coupon
/jetpack/v4/import/attachment  => _thumbnail_id, tix_coupon
/wp/v2/posts                   => _wp_font_face_file, _wp_attached_file, _wp_attachment_metadata, _thumbnail_id, tix_coupon
```

and the WXR side after the guard came out:

```
function_exists( 'WordCamp\Importer_Tweaks\file_path_meta_keys' )  bool(true)
import_post_meta_key( '_wp_font_face_file' )                        bool(false)
import_post_meta_key( '_wp_attached_file' )                         bool(false)
import_post_meta_key( '_thumbnail_id' )                             "_thumbnail_id"
```

The cross-file call is fine on load order twice over: root mu-plugins glob alphabetically so `importer-tweaks.php` precedes `load-other-mu-plugins.php`, and the call is deferred to dispatch anyway.

| | |
| --- | --- |
| `phpunit --group importer-tweaks` | `OK (13 tests, 13 assertions)` |
| Full suite | 892 tests, 2099 assertions, 6 skipped, 0 failures |

Moving `test_serialized_array_round_trips` to `tix_questions` is the right call — the test is about the value surviving the decode, and the key was incidental.

Approving. One thing to pick up whenever: the description still describes only the WXR half. `import-meta.php`, the five tests that came with it, and the fact that these routes now drop meta a client sends aren't in "What changed" or "Testing" — and the counts read 13 tests / 217, where the branch is 18 / 892. That is the half a deployer would want in the body rather than in a comment thread.

PR #1957, opened 2026-08-26
Groups: apply design revision to group sites

raicem:
## Summary

Applies the group site design revision based feedback from the design team

- The group home page leads with an About teaser (new `excerpt` mode on `wporg/page-content`), and the hero becomes a split layout: group details left, 16:9 photo on a retintable colour mat right, plus a share row.
- The single event page gets a matching hero with a host row and "Part of {group}" line.
- The breadcrumb moves into the sticky local-navigation bar (w.org convention), retiring the per-template back links; inner-page headers lose their tinted band.
- The front page and events archive widen to 1280px, and the hero type, About prose, spacing, and location pin are brought to the design file's specs.
- Smaller fixes: footer callout buttons become a ruled link list, icon search buttons go grey, sidebar buttons no longer grow on focus, and the event sidebar stays under the sticky header.


bor0 (approved):
Nice, thank you!

Some screenshots

Image
Image
Image
Image
Image
Image
Image
Image

carstingaxion:
Wow, this looks amazing! Thanks for the work @raicem and thank you for the subtle look behind the curtain @bor0 !

PR #1958, opened 2026-08-26
Groups: apply `wcorg_sanitize_plain_text()` to the event REST titles

bor0:
Stacked on **`add/plain-text-form-values`** — this consumes the `wcorg_sanitize_plain_text()` helper added there and does not redefine it. Review/merge that one first.

## Summary

`add/plain-text-form-values` applied that helper to the WordCamp / Campus Connect application handlers and the forms-to-drafts titles. The Groups front-end REST layer has the same shape and wasn't covered: `title` and `new_venue_name` are declared `'sanitize_callback' => 'sanitize_text_field'` (`inc/rest.php:592,647`) and then written straight to `post_title`. As the helper's docblock explains, `sanitize_text_field()` isn't the last word for a value that later passes through `wp_kses()`.

It matters for titles specifically because `core/post-title` emits `get_the_title()` into element content without running `wp_kses_post()` over it — so whatever is stored is what renders, on `single-event.html:16,42` and `event-card.php:28,34`.

This applies the existing helper at the three `post_title` write sites in `wporg-groups-frontend/inc/rest.php`:

| Write site | Reached by |
| --- | --- |
| `persist_event()` (`inc/rest.php:1058`) | `POST /event`, `POST /event/{id}`, `POST /draft/{id}/publish` |
| `save_draft()` (`inc/rest.php:864`) | `POST /draft`, `POST /draft/{id}` (autosave) |
| `resolve_venue_id()` (`inc/rest.php:1176`) | `new_venue_name` on **any** of the above — `save_draft()` calls it too (`inc/rest.php:900`) |

Encoding at the write site puts an obligation on the read sites: anything handing that value to a non-HTML medium has to decode it. `list_drafts()` and the venue list already did; `096513fb` adds the same to the form pre-fill and to the two export read sites, so all five are consistent. See item 1 below.

`b131ff73` then closes the one write path none of this reached — venues (and events) written through core's own REST route, which is what the modal actually uses. See item 2.

No new helper, no change to `3-helpers-misc.php`.

Measured end to end, first on the phpunit stack and then in a real browser against the local groups site (`raw` stored value):

| Typed | Before | After |
| --- | --- | --- |
| `Hall < 100 > seats` | `Hall  seats` | `Hall &lt; 100 &gt; seats` |
| `Meetup< a href="…" >click here< /a >` | `Meetup<a href="…">click here</a>` | `Meetup&lt; a href="…" &gt;click here&lt; /a &gt;` |
| `Coffee & Code — Vol. 2 <3` | `Coffee &amp; Code — Vol. 2 &lt;3` | unchanged |

Both new forms render as the literal text the organizer typed. Re-saving is idempotent — the third row is identical before and after, which is the check that nothing is encoded twice.

## How to review this PR

- **Why at the write site, not in the arg schema.** Keeps the raw value available to other consumers of `$fields`, including anything hanging off the `wporg_groups_frontend_event_args_schema` / `..._event_form_fields` filters. Trade-off: a fourth write site would have to remember to call it.
- **`save_draft()` is included**, and is easy to miss when looking only at the event and venue writes: `draft_args_schema()` inherits `event_args_schema()`, and `save_draft()` writes `post_title` from the same param.
- **Two paths deliberately left alone.** `update_group_info()` writes `blogname`/`blogdescription` through `update_option()`, and `sanitize_option()` already runs `esc_html()` over both. `list_drafts()` (`inc/rest.php:833`) returns `html_entity_decode( $post->post_title )` — decoding is the one operation that could undo the encoding, but the value is only ever used as a React `<option>` label in the draft picker (`event-modal-app.js:685`), i.e. rendered as text. Verified in the browser: the picker option renders the angle brackets as text with zero child elements. Also covered by a test.
- `new_venue_address` is unaffected — it goes to post meta, which isn't run through kses on the way in.

### Reviewer follow-ups

Both surfaced during review and were then confirmed in a real browser. Both are now fixed in this PR.

**1. Read sites now decode. (Fixed — `096513fb`.)** Encoding at the write site means every read site that hands the value to a non-HTML medium has to decode it. Two of them already did (`list_drafts()` at `:833`, the venue list at `:777`); three did not:

| Read site | Medium | Was | Now |
| --- | --- | --- | --- |
| `get_event_form_data()` (`rest.php:715`) | text input | `Hall &lt; 100 &gt; seats` | `Hall < 100 > seats` |
| `collect_export_data()` (`export.php:449`) | CSV / JSON export | `Hall &lt; 100 &gt; seats` | `Hall < 100 > seats` |
| `get_event_venue_names()` (`export.php:790`) | CSV / JSON export | `Hall &lt; 100 &gt; seats` | `Hall < 100 > seats` |

The export was the one worth catching: `collect_export_data()` reads the raw `post_title` rather than `get_the_title()` *precisely because* entity-encoding belongs to HTML output and not to a spreadsheet, so honouring that comment's intent now means decoding. Venue names reach it the same way — via the venue `post_title` and via the `$term->name` fallback, both stored encoded, so the decode sits on the combined value.

Round-tripping the decoded title back through the write path lands on the same bytes; `test_event_form_data_title_is_decoded` pins that, so a decode that widened the input would show up as double encoding.

One thing the decode does not fix, and can't: search matches stored bytes, so searching `Hall < 100` won't find the encoded title. That's inherent to encoding at rest. It doesn't bite here — the only search in this plugin is `Members_Controller`, over user fields, not post titles.

**2. Venue names written through core's route. (Fixed — `b131ff73`.)** Choosing "+ Add a new venue" calls `onOpenVenueEditor( 0 )` (`event-modal-app.js:325`), which POSTs the name to core's `/wp/v2/gatherpress_venues` (`venue-editor.js:209,259`) and never sends `new_venue_name`. So `resolve_venue_id()` — correct as it is — was never on the path the UI actually uses. A venue named `Community Hall< a href="…" >click here< /a >` stored as real markup and the Location block rendered a working link to example.org.

**This is core behaviour, not a GatherPress defect.** Checked rather than assumed — a plain `post` and `page` store exactly the same thing:

```
user can unfiltered_html: no
post               stored: Hall<a href="https://example.org">click</a>
page               stored: Hall<a href="https://example.org">click</a>
gatherpress_venue  stored: Hall<a href="https://example.org">click</a>
```

`wp_filter_kses()` permits `<a href title>` in every `post_title` (it's in `$allowedtags`) and `core/post-title` emits `get_the_title()` without `wp_kses_post()`. GatherPress registers ordinary post types with no custom REST controller and adds nothing to the storage path, so an upstream report would rightly be closed. The divergence here is wordcamp.org policy — that policy *is* `wcorg_sanitize_plain_text()` — so it belongs in this plugin.

**Scope of the risk.** kses does stop the dangerous parts, so this is link injection, not XSS:

| Input | Survives kses |
| --- | --- |
| `javascript:` href | No |
| `onclick` | No |
| `<script>` | No |
| `<a href="https://…">` | **Yes** |

**Why `rest_pre_insert_{$post_type}`.** It is per-post-type, so nothing else on the site changes, *and* it runs before kses. Ordering is the whole point: by `wp_insert_post_data` the value has already been through kses and `Hall < 100 > seats` is already `Hall  seats`, so no later hook can preserve what the helper exists to preserve. Verified by probing the hook order rather than reasoning about it.

Events are covered too: `/wp/v2/gatherpress_events` is registered and reachable on production, and bypasses `inc/rest.php` the same way. The helper is a no-op on its own output, so the two paths can't compound.

One trap worth flagging for review: `inc/post-titles.php` is required *above* the `class_exists()` guard in `bootstrap()`, so `PLAIN_TEXT_TITLE_POST_TYPES` spells the post types as literals. Using `Venue::POST_TYPE` there is a **fatal on every network site without GatherPress** — including the network root. I hit exactly that, caught it against a real site, and `test_post_type_names_match_gatherpress()` now pins the literals to the classes.

## Test plan

All green, run on the local phpunit stack against this branch.

- [x] Full suite as CI runs it (`phpunit -c phpunit.xml.dist`) — **979 tests, 0 failures**, and PHP's error log clean of deprecations/warnings
- [x] `phpunit --testsuite "WordCamp MU Plugins"` — 243 tests, OK
- [x] `phpunit --testsuite "WordCamp Post Type"` — 70 tests, OK
- [x] `phpunit --testsuite "WordCamp Forms to Drafts"` — 15 tests, OK
- [x] Reverting `inc/rest.php` to the base branch fails **all five** write-site tests — including `keeps_angle_brackets_used_as_prose` (`Hall  seats`), so they pin behaviour rather than passing vacuously
- [x] `phpcs` on the changed files — 0 errors; the one warning is the pre-existing `numberposts => 200` on `inc/rest.php:784`, untouched here
- [x] **Manual browser pass — done** (see below)
- No JS changed, so no `npm run build`.

Five tests added to `wporg-groups-frontend/tests/test-rest.php`:

| Test | Covers |
| --- | --- |
| `test_event_title_is_stored_as_text` | `POST /event` |
| `test_event_update_title_is_stored_as_text` | `POST /event/{id}` |
| `test_new_venue_name_is_stored_as_text` | `resolve_venue_id()`, resolved via `get_event_venue_post_id()` |
| `test_draft_title_is_stored_as_text` | `POST /draft`, plus the `list_drafts()` decode |
| `test_event_title_keeps_angle_brackets_used_as_prose` | `Hall < 100 > seats` still readable end to end |

Three more added for the read sites (`096513fb`):

| Test | Covers |
| --- | --- |
| `test_event_form_data_title_is_decoded` | `rest.php:715`, plus the write-path round trip |
| `test_export_event_title_is_decoded` | `export.php:449`, JSON and CSV |
| `test_export_venue_name_is_decoded` | `export.php:790` |

All three fail with `inc/` reverted, each with the reported symptom (`Hall &lt; 100 &gt; seats`).

Seven more for item 2 (`b131ff73`), in `tests/test-post-titles.php`, dispatched against *core's* posts controller because that's the route the filter exists for:

| Test | Covers |
| --- | --- |
| `test_venue_title_via_core_rest_is_stored_as_text` | the hole itself |
| `test_event_title_via_core_rest_is_stored_as_text` | same via `/wp/v2/gatherpress_events` |
| `test_prose_angle_brackets_survive_core_rest` | the destructive case — pins that the filter runs *before* kses |
| `test_update_without_a_title_does_not_blank_it` | partial updates leave `post_title` unset |
| `test_resaving_an_encoded_title_is_idempotent` | no double encoding where both paths apply |
| `test_other_post_types_are_untouched` | the filter is scoped, not global — a plain `post` keeps core behaviour |
| `test_post_type_names_match_gatherpress` | the literals still name GatherPress's post types |

Disabling the filter fails exactly the three behaviour tests, showing the real `<a href="https://example.org">click here</a>`; the scoping and idempotency tests stay green, which is what they should do.

Two test-environment notes, both pre-existing and neither a product change: the suite now loads `theme-templates/bootstrap.php` (core's `rest_after_insert_*` runs `the_content`, and `wc-post-types` calls `site_supports_block_templates()`), and `Test_Groups_Post_Titles::set_up()` re-registers GatherPress's post types and rebuilds the REST route map, because `mu-plugins/tests/test-groups-my-events.php` re-registers `gatherpress_event` as `array( 'public' => true )` and drops `show_in_rest` for the rest of a full run.

The first four go through `dispatch_json_request()` rather than calling the REST callbacks directly, so the registered `sanitize_callback` actually runs first — that's the point of the test, and a `set_param()`-built request would skip it and pass for the wrong reason.

`tests/bootstrap.php` now loads `3-helpers-misc.php`, matching what the base branch did for the `wcpt` and `wordcamp-forms-to-drafts` suites (and what `wordcamp-speaker-feedback` already does). The helper's own unit tests live on the base branch in `mu-plugins/tests/test-3-helpers-misc.php` and aren't duplicated here.

### Reproducing the automated run

```bash
docker exec wordcamporg-phpunit_wp-1 bash -lc 'cd /app && phpunit --testsuite "WordCamp Groups"'

# Confirm the tests actually pin the behaviour: revert the write sites, expect 5 failures.
git checkout origin/add/plain-text-form-values -- public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
docker exec wordcamporg-phpunit_wp-1 bash -lc \
  'cd /app && phpunit --testsuite "WordCamp Groups" --filter "is_stored_as_text|keeps_angle_brackets"'
git checkout HEAD -- public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php

# The base branch's suites must stay green.
for s in "WordCamp MU Plugins" "WordCamp Post Type" "WordCamp Forms to Drafts"; do
  docker exec wordcamporg-phpunit_wp-1 bash -lc "cd /app && phpunit --testsuite \"$s\""
done

public_html/wp-content/mu-plugins/vendor/bin/phpcs \
  public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php \
  public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/bootstrap.php \
  public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/test-rest.php
```

### Manual browser pass — results

Run through the real front-end event modal on `events.wordpress.test/group/sunshine-coast-qld/`, as a user with `unfiltered_html=no` (so kses behaves as it does for an Organizer).

| # | Step | Result |
|---|---|---|
| 1 | Publish an event titled `Meetup< a href="https://example.org" >click here< /a >` | **Pass.** Stored `Meetup&lt; a href=…`. Single-event page and group event card both render it as literal text — 0 anchors, 0 element children in the title. |
| 2 | Publish `Coffee & Code — Vol. 2 <3` (regression) | **Pass.** Renders exactly as typed. Stored value byte-identical to the pre-branch value. |
| 3 | Publish `Hall < 100 > seats` (regression) | **Pass.** Renders exactly as typed. Was `Hall  seats` before this branch. |
| 4 | Reopen the step-3 event, save unchanged (idempotency) | **Pass** on storage and render — unchanged, not double-encoded. Title input now pre-fills as `Hall < 100 > seats` (was `Hall &lt; 100 &gt; seats`); see item 1. |
| 5 | Name a new venue `Community Hall< a href="…" >click here< /a >` | **Not covered by this PR** — the modal routes this through core's `/wp/v2/gatherpress_venues`, and the Location block renders a live anchor. See item 2 above. |
| 6 | Autosave a draft titled `Draft< a href="…" >click here< /a >`, reopen the picker | **Pass.** Stored `Draft&lt; a href=…`. Picker option renders the brackets as text (0 anchors, 0 child elements), confirming the `list_drafts()` decode is safe. |
| 7 | Open a pre-existing event, save unchanged (regression) | **Pass.** Title untouched. |
| 8 | `GET /export` in both formats, after `096513fb` | **Pass.** Zero HTML entities anywhere in either payload; `Hall < 100 > seats` and `Coffee & Code — Vol. 2 <3` both export as typed. |
| 9 | Re-run step 5 against a real site after `b131ff73` | **Pass.** The venue stores `Community Hall&lt; a href=…` and the rendered title contains **0 anchors**. `/wp/v2/gatherpress_events` likewise stores `Hall &lt; 100 &gt; seats` rather than destroying it to `Hall  seats`. |

To exercise `resolve_venue_id()` itself, POST to `/wp-json/wporg-groups/v1/event` directly with a `new_venue_name` — that path is covered by `test_new_venue_name_is_stored_as_text`.

This is an input-side change only — titles already stored with markup aren't cleaned up by it. The `post_title LIKE '%<%'` sweep over `gatherpress_event` and `gatherpress_venue` is still worth doing separately, and now has no new rows arriving to chase.

### DIFF
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -426,7 +426,10 @@ function wcorg_sanitize_plain_text( $value ) {
 		return '';
 	}
 
+	// `strip_tags()` deletes from an unterminated `<` to the end of the string, so encode that
+	// case first -- `Rated <A best` should keep its text, not become `Rated`.
 	$value = wp_check_invalid_utf8( (string) $value );
+	$value = wp_pre_kses_less_than( $value );
 	$value = wp_strip_all_tags( $value, true );
 
 	return str_replace( '<', '&lt;', $value );
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -397,3 +397,42 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
 		true
 	);
 }
+
+/**
+ * Reduce a submitted value to text that stays text once WordPress saves it.
+ *
+ * `sanitize_text_field()` and `wp_kses()` disagree about what opens a tag: `strip_tags()`, which the
+ * former is built on, needs a letter, `/`, `!` or `?` after the `<`, while kses also accepts whitespace.
+ * A value like `< code >` therefore passes the sanitiser untouched, and the `wp_filter_kses()` that core
+ * puts on `title_save_pre` and `content_save_pre` then rebuilds it into a real element -- so a handler
+ * that meant to store text ends up storing markup. Encoding the `<` that survives settles the
+ * disagreement, and leaves the character visible to readers.
+ *
+ * `wp_kses( $value, array() )` is not a substitute: it drops whole `<...>` spans, turning
+ * `Hall < 100 > seats` into `Hall  seats`, and rewrites every `&`.
+ *
+ * The result is encoded for HTML. Decode it before using it in a plain-text medium such as an email
+ * body or a CSV column.
+ *
+ * @param string|array $value            The submitted value. Arrays are handled recursively; keys are
+ *                                       left alone.
+ * @param bool         $keep_line_breaks Whether to keep newlines, as `sanitize_textarea_field()` does.
+ *
+ * @return string|array A string, or an array of strings when `$value` is an array.
+ */
+function wcorg_sanitize_plain_text( $value, $keep_line_breaks = false ) {
+	if ( is_array( $value ) ) {
+		return array_map(
+			function ( $item ) use ( $keep_line_breaks ) {
+				return wcorg_sanitize_plain_text( $item, $keep_line_breaks );
+			},
+			$value
+		);
+	}
+
+	$value = $keep_line_breaks
+		? sanitize_textarea_field( (string) $value )
+		: sanitize_text_field( (string) $value );
+
+	return str_replace( '<', '&lt;', $value );
+}
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -410,7 +410,8 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
  * percent-encoded URLs; `wp_kses( $value, array() )` drops whole `<...>` spans, turning
  * `Hall < 100 > seats` into `Hall  seats`.
  *
- * The result is HTML-encoded. Decode it for a plain-text medium such as an email body or a CSV column.
+ * The result is HTML-encoded. Decode it for a plain-text medium -- the organizer reminder mails
+ * (`wcor-mailer.php`) are the case already in the tree.
  *
  * @param mixed $value Arrays are handled recursively, with keys left alone. Anything neither array nor
  *                     scalar becomes `''`, as `sanitize_text_field()` does.
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -401,24 +401,19 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
 /**
  * Reduce a submitted value to text that stays text once WordPress saves it.
  *
- * `sanitize_text_field()` and `wp_kses()` disagree about what opens a tag: `strip_tags()`, which the
- * former is built on, needs a letter, `/`, `!` or `?` after the `<`, while kses also accepts whitespace.
- * A value like `< code >` therefore passes the sanitiser untouched, and the `wp_filter_kses()` that core
- * puts on `title_save_pre` then rebuilds it into a real element -- so a handler that meant to store text
- * ends up storing markup. Encoding the `<` that survives settles the disagreement, and leaves the
- * character visible to readers.
+ * `strip_tags()` wants a letter, `/`, `!` or `?` after a `<` before it counts as a tag, but `wp_kses()`
+ * also accepts whitespace. So `< code >` survives `sanitize_text_field()`, and the `wp_filter_kses()`
+ * core puts on `title_save_pre` then rebuilds it into a real element. Encoding the surviving `<` settles
+ * that, and still reads as `<` to a human.
  *
- * This deliberately does not call `sanitize_text_field()`, which would also delete every `%[a-f0-9]{2}`
- * sequence and so quietly break a percent-encoded URL. `wp_kses( $value, array() )` is not usable either:
- * it drops whole `<...>` spans rather than neutralising them, turning `Hall < 100 > seats` into
- * `Hall  seats`, and rewrites every `&`.
+ * Two near-misses to avoid: `sanitize_text_field()` also deletes every `%[a-f0-9]{2}`, which breaks
+ * percent-encoded URLs; `wp_kses( $value, array() )` drops whole `<...>` spans, turning
+ * `Hall < 100 > seats` into `Hall  seats`.
  *
- * The result is encoded for HTML. Decode it before using it in a plain-text medium such as an email
- * body or a CSV column.
+ * The result is HTML-encoded. Decode it for a plain-text medium such as an email body or a CSV column.
  *
- * @param mixed $value The submitted value. Arrays are handled recursively; array keys are left alone,
- *                     and anything that is neither an array nor a scalar becomes an empty string, as
- *                     `sanitize_text_field()` does.
+ * @param mixed $value Arrays are handled recursively, with keys left alone. Anything neither array nor
+ *                     scalar becomes `''`, as `sanitize_text_field()` does.
  *
  * @return string|array A string, or an array of strings when `$value` is an array.
  */
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -404,35 +404,35 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
  * `sanitize_text_field()` and `wp_kses()` disagree about what opens a tag: `strip_tags()`, which the
  * former is built on, needs a letter, `/`, `!` or `?` after the `<`, while kses also accepts whitespace.
  * A value like `< code >` therefore passes the sanitiser untouched, and the `wp_filter_kses()` that core
- * puts on `title_save_pre` and `content_save_pre` then rebuilds it into a real element -- so a handler
- * that meant to store text ends up storing markup. Encoding the `<` that survives settles the
- * disagreement, and leaves the character visible to readers.
+ * puts on `title_save_pre` then rebuilds it into a real element -- so a handler that meant to store text
+ * ends up storing markup. Encoding the `<` that survives settles the disagreement, and leaves the
+ * character visible to readers.
  *
- * `wp_kses( $value, array() )` is not a substitute: it drops whole `<...>` spans, turning
- * `Hall < 100 > seats` into `Hall  seats`, and rewrites every `&`.
+ * This deliberately does not call `sanitize_text_field()`, which would also delete every `%[a-f0-9]{2}`
+ * sequence and so quietly break a percent-encoded URL. `wp_kses( $value, array() )` is not usable either:
+ * it drops whole `<...>` spans rather than neutralising them, turning `Hall < 100 > seats` into
+ * `Hall  seats`, and rewrites every `&`.
  *
  * The result is encoded for HTML. Decode it before using it in a plain-text medium such as an email
  * body or a CSV column.
  *
- * @param string|array $value            The submitted value. Arrays are handled recursively; keys are
- *                                       left alone.
- * @param bool         $keep_line_breaks Whether to keep newlines, as `sanitize_textarea_field()` does.
+ * @param mixed $value The submitted value. Arrays are handled recursively; array keys are left alone,
+ *                     and anything that is neither an array nor a scalar becomes an empty string, as
+ *                     `sanitize_text_field()` does.
  *
  * @return string|array A string, or an array of strings when `$value` is an array.
  */
-function wcorg_sanitize_plain_text( $value, $keep_line_breaks = false ) {
+function wcorg_sanitize_plain_text( $value ) {
 	if ( is_array( $value ) ) {
-		return array_map(
-			function ( $item ) use ( $keep_line_breaks ) {
-				return wcorg_sanitize_plain_text( $item, $keep_line_breaks );
-			},
-			$value
-		);
+		return array_map( 'wcorg_sanitize_plain_text', $value );
 	}
 
-	$value = $keep_line_breaks
-		? sanitize_textarea_field( (string) $value )
-		: sanitize_text_field( (string) $value );
+	if ( ! is_scalar( $value ) ) {
+		return '';
+	}
+
+	$value = wp_check_invalid_utf8( (string) $value );
+	$value = wp_strip_all_tags( $value, true );
 
 	return str_replace( '<', '&lt;', $value );
 }
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -183,6 +183,35 @@ function wcorg_get_user_by_canonical_names( $name ) {
 }
 
 /**
+ * Resolve a submitted username to the account a user is allowed to link a participant record to.
+ *
+ * A user may name their own account; naming a different one is reserved for those who can edit other
+ * authors' posts. An unknown name, or one they may not claim, resolves to an empty string.
+ *
+ * @param string   $name    The submitted username (matched against login and nicename).
+ * @param int|null $user_id The user doing the linking. Defaults to the current user.
+ *
+ * @return string The canonical `user_login` the user may link, or an empty string.
+ */
+function wcorg_get_linkable_user_login( $name, $user_id = null ) {
+	if ( is_null( $user_id ) ) {
+		$user_id = get_current_user_id();
+	}
+
+	$wporg_user = wcorg_get_user_by_canonical_names( $name );
+
+	if ( ! $wporg_user ) {
+		return '';
+	}
+
+	if ( (int) $user_id === (int) $wporg_user->ID || user_can( $user_id, 'edit_others_posts' ) ) {
+		return $wporg_user->user_login;
+	}
+
+	return '';
+}
+
+/**
  * Get CLDR country names and codes.
  *
  * @param array $args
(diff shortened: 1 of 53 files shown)


KokkieH:
Ran the full test plan against this branch: all four suites green (Groups is now 298 tests after the production merge), the 5 new tests fail with `inc/rest.php` reverted to the base branch, and the manual browser pass reproduces as documented — including the encoded pre-fill in item 1 and byte-identical re-saves.

One side effect the PR doesn't cover: the RSVP export. `inc/export.php:449` passes the raw `post_title` through precisely because "entity-encoding belongs to HTML output, not to a data export" — but the raw stored value is now the encoded form. Verified against `GET /export?format=json` after publishing the test titles: the export contains `Hall &lt; 100 &gt; seats`, which is what an organizer will see in the spreadsheet. Venue names flow in the same way via `export.php:778`. An `html_entity_decode()` at those two read sites (the `list_drafts()` pattern) would restore the comment's intent — fine as part of the same follow-up as item 1 if you'd rather keep this PR to the write sites.

On item 1: +1 to adding `html_entity_decode()` at `rest.php:715`. Same family, for that follow-up: search matches stored bytes, so searching `Hall < 100` won't find the encoded title. 

On item 2: agreed this doesn't block — the modal writes venues through core's posts controller, so that path is untouched (and unworsened) by this branch, and the `resolve_venue_id()` change still correctly covers the REST-direct param. For the follow-up, a server-side filter scoped to the venue post type would cover the core REST route, wp-admin, and the modal in one place, rather than fixing the JS.

bor0:
Thanks for running the full plan — and for catching the export. Folded the read-side decodes into this PR in `096513fb`, since the export regression is one this branch causes and shipping it separately would mean landing a known wrong spreadsheet in between.

**The export, and the rest of the read sites.** Your framing is the one I used: `collect_export_data()` reads the raw `post_title` *precisely because* encoding belongs to HTML output and not to a data export, so honouring that comment now means decoding. Three sites needed it, joining the two that already did:

| Read site | Medium | Was | Now |
| --- | --- | --- | --- |
| `get_event_form_data()` (`rest.php:715`) | text input | `Hall &lt; 100 &gt; seats` | `Hall < 100 > seats` |
| `collect_export_data()` (`export.php:449`) | CSV / JSON | `Hall &lt; 100 &gt; seats` | `Hall < 100 > seats` |
| `get_event_venue_names()` (`export.php:790`) | CSV / JSON | `Hall &lt; 100 &gt; seats` | `Hall < 100 > seats` |

On the venue one: I put the decode on the combined value rather than on `$venue_post->post_title` alone, because the `$term->name` fallback beside it is encoded too — `pre_term_name` carries `wp_filter_kses`, so a venue reached via the fallback would otherwise still export encoded.

**Verified against your own fixtures**, not fresh ones — events 574/576/578 on `events.wordpress.test`, still in the state your pass left them:

- `GET /export` in both formats: zero HTML entities anywhere in either payload. `Hall < 100 > seats` and `Coffee & Code — Vol. 2 <3` both export as typed. CSV row: `"576","Hall < 100 > seats",…`.
- The modal on event 576 pre-fills `Hall < 100 > seats`. That's `res.fields.title` → `form.title` → `TextControl value`, so no JS change and no build.

Three tests added; all three fail with `inc/` reverted, each with exactly the symptom you reported. Suites: Groups **301 OK**, MU Plugins 243, Post Type 70, Forms to Drafts 15. `phpcs` clean on the four changed files (the `numberposts => 200` warning is the pre-existing one).

**On search** — agreed, and worth being explicit that the decode doesn't help there: search matches stored bytes, so `Hall < 100` won't find it. That's inherent to encoding at rest rather than something a read site can undo. It doesn't bite in this plugin today — the only search here is `Members_Controller`, over user fields, not post titles — so I've noted it in the description rather than chasing it.

bor0:
Item 2 is in as well now — `b131ff73`, taking the server-side filter you suggested rather than touching the JS.

**It isn't a GatherPress bug.** I checked before writing anything, since it would have changed where the fix belongs:

```
user can unfiltered_html: no
post               stored: Hall<a href="https://example.org">click</a>
page               stored: Hall<a href="https://example.org">click</a>
gatherpress_venue  stored: Hall<a href="https://example.org">click</a>
```

A plain `post` does the same thing. `wp_filter_kses()` permits `<a href title>` in every `post_title` and `core/post-title` emits `get_the_title()` without `wp_kses_post()`. GatherPress registers ordinary post types with no custom REST controller and adds nothing to the storage path, so upstream would rightly close it. The divergence is our policy — that policy *is* `wcorg_sanitize_plain_text()` — which is what makes this the right place, as you said.

**Also worth correcting my own framing:** kses strips `javascript:`, `onclick` and `<script>`; only plain `<a href>` survives. So this is link injection, not XSS. Lower severity than "renders a live anchor" made it sound.

**The hook choice turned out to matter more than expected.** `rest_pre_insert_{$post_type}` is scoped per-post-type *and* runs before kses. That second half is load-bearing — I probed the ordering, and by `wp_insert_post_data` the value has already been through kses, so `Hall < 100 > seats` is already `Hall  seats`. Nothing later can preserve what the helper exists to preserve.

I extended it to `gatherpress_event` too: core's `/wp/v2/gatherpress_events` is registered on production and bypasses `inc/rest.php` the same way. The helper is a no-op on its own output, so the two paths can't compound — pinned by a test. Happy to drop the event half if you'd rather keep this strictly to venues.

**One trap worth knowing about, since it nearly shipped.** I first wrote the post-type list as `Venue::POST_TYPE` / `Event::POST_TYPE`. `inc/post-titles.php` is required *above* the `class_exists()` guard in `bootstrap()`, so that's a fatal on every network site that doesn't run GatherPress — including the network root. Every test passed, because GatherPress is always loaded in the suite; only running it against a real site caught it. It's literals now, with `test_post_type_names_match_gatherpress()` pinning them to the classes.

**Verification.** Full suite the way CI runs it: 979 tests, 0 failures, PHP error log clean, phpcs clean, all six CI checks green. Disabling the filter fails exactly the three behaviour tests with the real `<a href="https://example.org">click here</a>` visible, while the scoping and idempotency tests stay green. Against a real site the venue now stores `Community Hall&lt; a href=…` and the rendered title has **0 anchors**.

Two notes that aren't product changes:

- The suite now loads `theme-templates/bootstrap.php` — core's `rest_after_insert_*` runs `the_content`, and `wc-post-types` calls `site_supports_block_templates()`.
- `Test_Groups_Post_Titles::set_up()` re-registers GatherPress's post types and rebuilds the REST route map, because `mu-plugins/tests/test-groups-my-events.php` re-registers `gatherpress_event` as `array( 'public' => true )` and drops `show_in_rest` for everything after it in a full run. I worked around it here rather than touching that suite, but it's a landmine for any future REST test and probably wants its own fix.

The `post_title LIKE '%<%'` sweep is still outstanding, though no new rows can arrive now.

KokkieH:
Took another full pass on `b131ff73`. Everything re-verified; one residual path and one doc ask, neither blocking.

**Verification.** Groups **308 OK**, Post Type 70, Forms to Drafts 23. MU Plugins has 1 failure — `Test_QuickBooks_OAuth_State::test_authorize_url_carries_the_stored_state` — in files byte-identical to `production` (came in with the merge; looks environmental), not this branch. Revert checks hold: the 3 read-site tests fail with `inc/` reverted, and the 3 storage tests in `test-post-titles.php` fail with the filter registration disabled. Browser, as an editor without `unfiltered_html`: the modal pre-fills `Hall < 100 > seats` decoded and an unchanged save lands on identical bytes; naming a venue through the "+ Add a new venue" overlay now stores encoded text and the Location block renders it with zero anchors — the step-5 hole is closed; `GET /export` in both formats carries the typed text with no entities anywhere, and the CSV cell quoting is right.

**The residual path — fine as a follow-up.** `rest_pre_insert_*` only sees REST writes, so the non-REST admin paths (Quick Edit / bulk edit, via `admin-ajax` → `edit_post()` → `wp_update_post()`) still store what kses allows — verified end-to-end: Quick Edit on the events list stored a working `<a href>` title as that same editor. In practice it's events-only, since the venue list table isn't reachable at that role. I don't consider it a blocker: it's pre-existing behaviour, not a regression — this branch strictly shrinks the exposed surface, and every path the product actually steers people down is now covered.

Worth being precise about what the follow-up is, though: it isn't patching a WordPress hole. Core's only title guard is kses, and kses draws the line at XSS, not link injection — `<a href title>` is in `$allowedtags` deliberately, with protocol filtering on the href. Title markup is a formatting feature in core's model; this network's "titles are plain text" rule is a stricter local policy, which is exactly why it lives in this mu-plugin and not upstream. Finishing that policy across all write routes has one post-type-aware choke point: the `wp_insert_post_data` filter, whose third argument (`$unsanitized_postarr`, WP 6.0+) still carries the pre-kses title. That change touches every `wp_insert_post()` for these post types — including programmatic writers like provisioning and archive restore — so it deserves its own test pass rather than being commit four here.

**One ask for this PR:** the `post-titles.php` header says the filter keeps titles as text "whichever route writes them" — the commit message's scoping ("on core's REST route") is the accurate one. Narrowing that sentence to the routes actually covered keeps the next auditor from reading the residual path as closed.

KokkieH (approved):

PR #1959, opened 2026-08-26
Budgets: apply payment file visibility rules on every request

### LINKED ISSUE #1316: Non admin users can't view more than the first page in media library grid

bor0:
## Summary

Attachments on budget requests (reimbursements and vendor payments) are meant to be visible to the organizer who uploaded them, the organizer who filed the request, and network admins. That scoping lives in `wordcamp-payments/includes/privacy.php`, and it wasn't applying consistently:

- The file was only loaded for admin, cron, and ajax requests, so the rules didn't apply on REST or XML-RPC requests at all.
- The rules hung off `the_posts`, which `WP_Query` skips when `suppress_filters` is set — the default for `get_posts()` and therefore `get_children()`.
- The post type was compared strictly against `'attachment'`, so a `post_type => 'any'` query never matched.

This loads the file on every request and moves the exclusion into the query itself, so it no longer depends on each call site opting in.

Fixes #1316 — doing the exclusion in SQL keeps `found_posts` in step with the results, which is what the Media Library grid needs to offer "Load more" to non-admins. Supersedes #1679, which addressed that symptom alone.

## How to review this PR

`includes/privacy.php` is the change; the rest is test scaffolding.

- `unsuppress_attachment_query_filters()` — `pre_get_posts` is the one hook `WP_Query` fires *before* it reads `suppress_filters`, so it's the only place this can be corrected from.
- `exclude_others_payment_files()` — the SQL condition. `fields => ids` and the `found_posts` count both return before `the_posts`, so a results filter can't reach either.
- `hide_others_payment_files()` — kept on as a second pass, for results supplied through `posts_pre_query` (an external search index, say) where the clauses are built but discarded.
- `get_budget_request_post_types()` — the file can't reference the `POST_TYPE` constants any more, since it now loads on requests where those files don't. A test pins the two together so they can't drift.

**Not applied to WP-CLI:** `privacy.php` never reached WP-CLI before this PR, and CLI runs with no user set, so leaving the guard on there would have `wp media regenerate`, `wp post list --post_type=attachment` and any export or cleanup script quietly working from an incomplete set. The hooks aren't registered under `WP_CLI` — see `is_cli_request()`. Cron is deliberately not exempted: it already loaded this file, so there's nothing to restore. Raised in review by @KokkieH.

**Scope of the guard:** it only attaches to queries that can actually return an attachment — `post_type` naming `attachment` or `any`, or an attachment permalink. A query that names no post type is *not* included: `WP_Query` narrows those to `page` or `post` (search included), so they can't surface an attachment, and that's the shape of nearly every front-end query on the site. Measured on a home-page-shaped query against a local subsite: no guard subquery is added. Two tests pin both sides of that boundary.

## Test plan

- [x] phpunit — 949/949 passing, including 15 new tests in a new `WordCamp Payments` suite
- [x] phpcs clean on changed files (3 remaining notices in `privacy.php` are pre-existing)
- [x] Manual browser pass, all three roles
- [x] Manual WP-CLI pass on a local subsite, with and without the exemption

### Manual test steps

Setup on a local subsite (e.g. `seattle.wordcamp.test/2023`):

1. Create two Editors, `organizer-a` and `organizer-b`.
2. As `organizer-a`: Budget → Vendor Payments → create a request and upload a PDF to it.
3. As `organizer-a`: upload a second, unrelated PDF straight to the Media Library.
4. As `organizer-b`: Media → Library (grid). **Expect the unrelated PDF only.**
5. As `organizer-a`: same screen. **Expect both.**
6. As a network admin: same screen. **Expect both.**

Steps 1–6 were run by hand in the grid view. Still worth a reviewer's pass:

7. Repeat 4–6 in list view — it's a different query path from the grid.
8. Confirm `/wp-json/wp/v2/media` returns the same set as the screen does for each of the three, so the API and the UI agree. (Checked for `organizer-b` only.)

One thing the tests pin that isn't obvious from the diff: `WP_Query` keys its result cache partly on the assembled SQL, and the guard puts the current user's ID into that SQL — so two organizers running the same query get different cache keys. That's what stops a permissive result being served to a restricted user under a persistent object cache, and it would never show up in a test run without being asserted.

Covered by the automated tests rather than repeated by hand: `get_posts()`/`get_children()` defaults, `post_type => 'any'`, `fields => ids`, `found_posts` vs. returned rows, the XML-RPC media methods, the uploader-on-someone-else's-request case, and a logged-out front-end search.

#### Screenshots

| Description | Screenshot |
| --- | --- |
| `organizer-b` — other organizer's request file is not listed | 1-organizer-b-other-request-file-hidden |
| `organizer-a` — own request file listed | 2-organizer-a-own-request-file-visible |
| Network admin — everything listed | 3-network-admin-all-files-visible |


public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
```diff
@@ -9,44 +9,215 @@
...
  *
- * @param WP_Post[] $attachments
- * @param WP_Query  $wp_query
+ * `get_posts()` -- and so `get_children()` -- defaults to `suppress_filters => true`, which skips `posts_clauses`
+ * and `the_posts` alike. `pre_get_posts` is the only hook `WP_Query` fires before it reads that var, so it's the
+ * one place this can be corrected from.
+ *
+ * @param WP_Query $wp_query
+ */
+function unsuppress_attachment_query_filters( $wp_query ) {
```
KokkieH, line 69: Now that this loads on every request, WP-CLI is treated as an anonymous visitor: CLI commands run with no user set, so attachment queries from maintenance tooling silently return an incomplete set. Concretely, `wp post list --post_type=attachment` no longer lists payment files, and anything that iterates attachments (`wp media regenerate`, export/cleanup scripts) will skip them without any indication it did. I hit the same shape in #1948 — a rule keyed on the current user unintentionally removing abilities from the CLI user.

Could we exempt system contexts, keeping the guard for actual visitors? Something like:

```php
/**
 * Whether the request is a system context acting on the site itself rather than on behalf of a visitor.
 *
 * CLI and cron run with no user set; treating them as an anonymous visitor would make maintenance
 * commands operate on an incomplete set of attachments without any indication that rows were skipped.
 */
function is_system_request() {
	return ( defined( 'WP_CLI' ) && WP_CLI ) || wp_doing_cron();
}
```

…with an early return in `unsuppress_attachment_query_filters()`, `exclude_others_payment_files()`, and `hide_others_payment_files()`. One thing to weigh: `privacy.php` already loaded on cron before this PR, so cron was partially covered by the old `the_posts` filter — exempting it is a small deliberate loosening, while exempting `WP_CLI` only restores what CLI had before this PR. Fine either way as long as it's a conscious choice; the logged-out front-end case must of course stay guarded (the existing test pins that).

public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
```diff
@@ -57,15 +228,37 @@ function hide_others_payment_files( $attachments, $wp_query ) {
...
+ *
+ * `wp.getMediaItem` looks an attachment up by ID with `get_post()`, so it never runs a `WP_Query` for the query
+ * guards to apply to. Returning an empty struct keeps the method able to answer without describing the file.
+ *
+ * @param array   $media_item_struct
+ * @param WP_Post $media_item
+ *
+ * @return array
+ */
+function redact_others_payment_files( $media_item_struct, $media_item ) {
```
KokkieH, line 278: Same shape as the `wp.getMediaItem` case this handles, one path over: `WP_REST_Attachments_Controller::get_item()` also resolves the attachment with `get_post()`, so no query runs and none of the guards apply. The result is that the collection endpoint and the item-by-ID endpoint currently disagree about the same attachment for the same user — `/wp/v2/media` omits it while `/wp/v2/media/<id>` returns it in full.

Suggest mirroring this redaction for REST so the by-ID path answers the same way the attachment permalink does:

```php
add_filter( 'rest_prepare_attachment', __NAMESPACE__ . '\redact_others_payment_files_rest', 10, 2 );

/**
 * Apply the same rules to REST item-by-ID lookups, which resolve via `get_post()` and so never run
 * a query the other guards could reach.
 *
 * @param WP_REST_Response $response
 * @param WP_Post          $attachment
 *
 * @return WP_REST_Response|WP_Error
 */
function redact_others_payment_files_rest( $response, $attachment ) {
	if ( current_user_can( 'manage_options' ) ) {
		return $response;
	}

	if ( get_hidden_payment_file_ids( array( $attachment ) ) ) {
		return new WP_Error( 'rest_post_invalid_id', __( 'Invalid post ID.' ), array( 'status' => 404 ) );
	}

	return $response;
}
```

Returning the same 404 core uses for an unknown ID keeps the two endpoints, the permalink, and the XML-RPC method all telling the same story. Worth a test alongside the XML-RPC one.

public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
```diff
@@ -9,44 +9,215 @@
 defined( 'WPINC' ) || die();
 
 
-add_filter( 'the_posts',                          __NAMESPACE__ . '\hide_others_payment_files', 10, 2 );
+add_action( 'pre_get_posts',                      __NAMESPACE__ . '\unsuppress_attachment_query_filters', 1 );
```
KokkieH: Minor: at priority 1 this runs before other `pre_get_posts` callbacks, so a suppressed query whose `post_type` is set to `attachment` by a later callback would be missed. Hooking at a very late priority instead would catch that shape. Not aware of a caller that does this today, so fine to leave as is.

KokkieH (commented):
Ran the manual pass (grid and list views for all three roles, plus the REST collection for each) and the suite — everything behaves as described, and the list-view count now agrees with the visible rows. Two changes requested, both about making the rules apply consistently across contexts; neither blocks the approach.

public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
bor0: Good catch — taken.

Went with WP-CLI only rather than CLI + cron. The CLI half restores exactly what it had before this PR; cron already loaded `privacy.php`, so there's nothing to restore there, and exempting it would give up coverage that exists in production today for no reported need. Conscious choice, as you asked.

Made it a registration-time check rather than three early returns, so CLI doesn't pay for the hooks at all:

```php
if ( ! is_cli_request() ) {
	add_action( 'pre_get_posts', __NAMESPACE__ . '\unsuppress_attachment_query_filters', PHP_INT_MAX );
	add_filter( 'posts_clauses', __NAMESPACE__ . '\exclude_others_payment_files', 10, 2 );
	add_filter( 'the_posts',     __NAMESPACE__ . '\hide_others_payment_files' );
}
```

Verified against the local dev subsite, both directions: with the exemption, `wp post list --post_type=attachment` lists the file attached to the budget request; with it removed, that row silently disappears while the unrelated one stays — the shape you described.

Not unit-tested directly, because flipping `WP_CLI` mid-run means defining a constant that then leaks into every later test in the process. Verified by hand instead. The logged-out front-end case is untouched and `test_logged_out_search_hides_payment_files` still pins it.

public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
bor0: Taken — moved to `PHP_INT_MAX`.

It turns out to cost nothing: every priority on `pre_get_posts` runs before `WP_Query` reads `suppress_filters`, so going last only widens what's caught. Added `test_guard_applies_when_a_later_callback_names_the_post_type`, which sets `post_type` from a callback at priority 100; it fails at the old priority 1 and passes now.

public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
bor0: You're right about the mechanism — the item-by-ID path resolves through `get_post()`, so no query guard reaches it, and the redaction you've written does the right thing.

I'd still rather not land it in this PR. On a second read that route isn't the only one with this shape: there are other by-ID paths that resolve an attachment without running a query, and they don't share a single chokepoint, so a filter here would close part of the surface while reading as though it closed all of it. Some of them aren't reachable through per-post capability mapping either, so it's genuinely two layers of work rather than one filter.

None of that changes with this PR — `/wp/v2/media` behaves the same before and after — so I'd like to keep #1959 to the query layer and do the by-ID surface as its own change, where it can be reviewed as a set. @mcliwanow raised the same thing and offered to pick it up separately this week; Michal, your suggestion here is the right starting point for it.

The `xmlrpc_prepare_media_item` filter stays in this PR, but scoped to the media methods it names rather than claiming the by-ID surface generally.

KokkieH (approved):
Thanks for the changes. Not seeing any other issues.

PR #1962, opened 2026-08-27
Blocks: Fix text domain in block patterns registration

### LINKED ISSUE #1927: i18n: several custom text domains appear not to be loaded
## Summary

Custom PHP code across the Groups platform and the blocks mu-plugin uses text domains other than `wordcamporg`, which appears to be the only custom domain with a loaded `.mo` file. If that reading is right, those strings render untranslated regardless of the visitor's locale.

Raising this as an audit finding rather than a bug report, because the right resolution depends on decisions I can't see from the repository.

## What the repository documents

`mu-plugins/wcorg-misc.php`, in `wcorg_load_wordcamp_textdomain()`:

> `wordcamporg` is used by all the custom plugins and themes, so that translators only have to deal with a single GlotPress project, and we only have to install/update a single mofile per locale.

It is loaded on `plugins_loaded` and `change_locale` from `WP_CONTENT_DIR/languages/wordcamporg/wordcamporg-{locale}.mo`.

Searching `public_html` for `load_textdomain`, `load_plugin_textdomain` and `load_muplugin_textdomain`, the only other hit is `campt-indian-payment-gateway`, a bundled third-party plugin with its own `languages/` directory.

## Relevant scan results

Token-aware scan of 358 PHP files under `mu-plugins/` and `themes/`, excluding tests, vendor and node_modules, counting only gettext calls with a literal domain. Showing the custom domains relevant here:

| Domain | Strings | Files |
| --- | --- | --- |
| `wordcamporg` | 706 | 110 |
| `wporg-groups-frontend` | 131 | 20 |
| `groups-site` | 12 | 2 |
| `wordcamp` | 8 | 4 |
| `gpre` | 8 | 2 |

Neither `mu-plugins/wporg-groups-frontend/`, `mu-plugins/gatherpress-recurring-events/` nor `themes/groups-site/` calls a loader.

## Clearest case: `wordcamp` in `mu-plugins/blocks/patterns.php`

That file uses both domains:

- line 31: `_x( 'WordCamp', 'Block pattern category', 'wordcamporg' )`
- lines 63, 78, 98: `__( ..., 'wordcamp' )`
- lines 144, 147: `translate_with_gettext_context( ..., 'wordcamp' )`

Same file, two domains, one of them the documented shared one. This looks like a typo rather than a decision.

## Examples from the other domains

- `__( 'Add New Sponsor', 'wporg-groups-frontend' )`
- `__( 'Be the first to RSVP', 'wporg-groups-frontend' )`
- `esc_html__( 'This occurrence has been cancelled.', 'gpre' )`

## Where I am less certain

`themes/groups-site/style.css` declares `Text Domain: groups-site`, so its 12 strings may be handled by theme translation loading, or by language files deployed to production outside this repository. `themes/campsite-2017` declares `Text Domain: wordcamporg` instead, so both patterns exist in the repo. I can't tell from here which is intended for `groups-site`.

## Questions

1. Is `wordcamporg` still the intended domain for all custom code, or have the Groups plugins and theme deliberately moved to their own domains?
2. If they are intended to keep separate domains, do they need loaders and a GlotPress project, or is that handled during deployment?
3. Is `wordcamp` in `patterns.php` simply a typo for `wordcamporg`?

Happy to open a PR once the direction is clear. Normalizing to `wordcamporg` would be mechanical; adding loaders for the separate domains would be small too.

## Related

- #1898 - the JavaScript side uses the same domains (`wporg-groups-frontend`, `gpre`) and would need `wp_set_script_translations()` rather than a rename, so I've scoped this to PHP.
- #1096, #1440, #1356 - existing i18n issues in adjacent areas.

wprashed:
### Summary

Fixes #1927.

In `mu-plugins/blocks/patterns.php`, strings were registered under the text domain `wordcamp` instead of the canonical `wordcamporg` text domain loaded across the network and used by the meta/wordcamp GlotPress project.

### Changes

- Replaced `wordcamp` with `wordcamporg` in gettext and `translate_with_gettext_context()` calls in `mu-plugins/blocks/patterns.php`.

### Testing Instructions

1. Ensure `patterns.php` passes syntax checks.
2. Verify block patterns in WordCamp sites load and translate titles/descriptions under the `wordcamporg` domain.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/patterns.php
+++ b/public_html/wp-content/mu-plugins/blocks/patterns.php
@@ -60,7 +60,7 @@ function register_patterns() {
 						esc_html(
 							sprintf(
 								/* translators: %s: file name. */
-								__( 'Could not register file "%s" as a block pattern ("Slug" field missing)', 'wordcamp' ),
+								__( 'Could not register file "%s" as a block pattern ("Slug" field missing)', 'wordcamporg' ),
 								$file
 							)
 						),
@@ -75,7 +75,7 @@ function register_patterns() {
 						esc_html(
 							sprintf(
 								/* translators: %1s: file name; %2s: slug value found. */
-								__( 'Could not register file "%1$s" as a block pattern (invalid slug "%2$s")', 'wordcamp' ),
+								__( 'Could not register file "%1$s" as a block pattern (invalid slug "%2$s")', 'wordcamporg' ),
 								$file,
 								$pattern_data['slug']
 							)
@@ -95,7 +95,7 @@ function register_patterns() {
 						esc_html(
 							sprintf(
 								/* translators: %1s: file name; %2s: slug value found. */
-								__( 'Could not register file "%s" as a block pattern ("Title" field missing)', 'wordcamp' ),
+								__( 'Could not register file "%s" as a block pattern ("Title" field missing)', 'wordcamporg' ),
 								$file
 							)
 						),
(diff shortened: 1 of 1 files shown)


bor0:
Thanks for your contribution! 🙌

wprashed:
@bor0 Will I get a badge on my wp.org profile?

bor0:
@wprashed I think you should, but not sure how often the sync happens. @obenland do you know more?

PR #1963, opened 2026-08-27
Reports: add a REST endpoint to the Campus Connect Details report

maciejpilarski:
## What

Adds a REST endpoint to the **Campus Connect Details** report, and requires a capability for report REST routes generally.

## Why

`CampusConnect_Details` is currently reachable only through wp-admin. Its column set is exactly what the Campus Connect programme reports on — statuses across the whole organiser pipeline, `Actual Attendees`, institution, city and country — and none of that is derivable from the public API:

- `WordCamp_Loader::get_public_post_statuses()` is `['wcpt-scheduled', 'wcpt-closed']`, so the ~46 events in pre-planning statuses are never exposed.
- `/wp-json/wp/v2/wordcamps/<id>` returns `meta: []` and omits `Actual Attendees` entirely.

So anyone reporting on Campus Connect has to open the report, tick the boxes, export the CSV and paste it somewhere — by hand, every time. This makes that step automatable.

## The permission_callback part

`register_rest_endpoints()` in `index.php` already had the machinery to expose any report declaring `$rest_base` + `rest_callback()`, but no report used it, so nothing exercised the path.

Registering a route there today would have exposed it **anonymously** — `register_rest_route()` was called without a `permission_callback`, which on WP 5.5+ both triggers a `_doing_it_wrong()` notice and leaves the route publicly readable. Given reports read private post meta and non-public post statuses, that seemed worth fixing in the same change rather than adding the first consumer on top of it.

The default is `current_user_can( CAPABILITY )` (`view_wordcamp_reports`). A report class can override by defining its own `rest_permission_callback()`, so a genuinely public report is still possible — it just has to say so explicitly.

## The endpoint

```
GET /wp-json/wordcamp-reports/v1/campus-connect-details
```

Returns the same rows the admin screen renders, using the same field set that is checked by default there, so the endpoint and the CSV export stay in step. No request parameters; it returns all Campus Connect events.

## Testing

Both files are `php -l` clean, and the change follows the existing `get_data()` / `prepare_data_for_display()` contract that `render_admin_page()` already uses.

**I have not been able to exercise this against a running Central instance** — I don't have a local wordcamp.org environment, so the endpoint's live response has not been verified end to end. Please treat the runtime behaviour as unverified. I am opening this as a draft for that reason, and I am happy to adjust the shape — field set, parameters, whether the capability default belongs here at all — to whatever you would prefer.

## Notes

- No existing behaviour changes: the admin screen, the CSV export and every other report are untouched.
- The `permission_callback` addition is the only change outside the Campus Connect report class, and it affects a code path that currently has no consumers.


public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
```diff
@@ -200,6 +211,81 @@ public function get_event_posts() {
...
+			'Start Date (YYYY-mm-dd)',
+			'End Date (YYYY-mm-dd)',
+			'Status',
+			'Name',
+			'Organizer Name',
+			'Venue Name',
+			'_venue_city',
+			'_venue_country_name',
+			'Number of Anticipated Attendees',
+			'Actual Attendees',
```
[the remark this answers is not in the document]
maciejpilarski: Confirmed and fixed in b07356b. You are right: `meta_keys()` unsets `Series Event` (alongside `Transparency Report Received`) unless the caller has `manage_options`, so `validate_fields_input()` rejected the whole request and a `report_viewer` — who passes the permission check — got a 500 instead of the report. The endpoint now treats its fixed list as an upper bound and intersects it with `get_data_fields_safelist()` before building, so each caller gets exactly the fields they are entitled to. I kept an explicit list rather than defaulting to the full safelist, because the unrestricted set includes organiser e-mail addresses and telephone numbers.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

public_html/wp-content/plugins/wordcamp-reports/index.php
```diff
@@ -343,12 +343,25 @@ function register_rest_endpoints() {
...
+
 			register_rest_route(
 				$namespace,
 				'/' . $class::$rest_base,
 				array(
-					'methods'  => array( 'GET' ),
-					'callback' => array( $class, 'rest_callback' ),
+					'methods'             => array( 'GET' ),
+					'callback'            => array( $class, 'rest_callback' ),
+					'permission_callback' => $permission_callback,
```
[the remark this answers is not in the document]
maciejpilarski: Added in 277afb5 — `tests/test-report-rest-authorization.php`, following the idioms in the existing `test-report-export-authorization.php`. It covers: the route is registered at all (so the other assertions cannot pass as 404s); an anonymous caller gets 401; a logged-in user without the capability gets 403; and a user holding only `view_wordcamp_reports` gets a 200 with array data. That last case is what catches the field-safelist 500 you identified in the other comment — it failed before the fix and passes after. A fifth test asserts directly that `Series Event` is absent from a report-only safelist while `Actual Attendees` is present, so the trap is documented rather than just guarded.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

KokkieH:
## Local environment testing steps as devised by Claude Code:

Note: after any host-side edit to `.docker/wp-config.php`, restart the container — the single-file bind mount goes stale when the file is rewritten (a container restart fixes it).

1. Run the unit tests:
   ```
   docker restart wordcamptest-phpunit_wp-1 && sleep 3
   docker exec -w /app wordcamptest-phpunit_wp-1 ./public_html/wp-content/mu-plugins/vendor/bin/phpunit -c phpunit.xml.dist --testsuite "WordCamp Reports"
   ```
   Expect: `OK (23 tests, 27 assertions)`.

2. Create two test users on central (a report viewer and a control with no capability):
   ```
   docker exec wordcamptest-wordcamp.test-1 wp user create reportviewer reportviewer@example.test --role=subscriber --user_pass=password --porcelain --url=central.wordcamp.test --allow-root
   docker exec wordcamptest-wordcamp.test-1 wp user create nocapuser nocapuser@example.test --role=subscriber --user_pass=password --porcelain --url=central.wordcamp.test --allow-root
   ```
   Note the two user IDs it prints.

3. Grant the subrole: edit `.docker/wp-config.php` and add an entry to the existing `$wcorg_subroles` array, using the reportviewer ID from step 2:
   ```php
   24 => array( 'report_viewer' ),
   ```
   Then restart so the stale mount refreshes: `docker restart wordcamptest-wordcamp.test-1`.

4. Create a Campus Connect event in a non-public status, with a private-meta canary:
   ```
   docker exec wordcamptest-wordcamp.test-1 bash -c '
   ID=$(wp post create --post_type=wordcamp --post_title="CC Test University" --post_status=wcpt-pre-planning --porcelain --url=central.wordcamp.test --allow-root)
   wp post meta set $ID event_subtype campusconnect --url=central.wordcamp.test --allow-root
   wp post meta set $ID "Venue Name" "Test University" --url=central.wordcamp.test --allow-root
   wp post meta set $ID "Location" "Cape Town, South Africa" --url=central.wordcamp.test --allow-root
   wp post meta set $ID "E-mail Address" "canary@example.test" --url=central.wordcamp.test --allow-root
   wp post meta set $ID "Series Event" 1 --url=central.wordcamp.test --allow-root
   echo "Created $ID"'
   ```

5. Create application passwords for the three callers (each prints a password — note them):
   ```
   docker exec wordcamptest-wordcamp.test-1 wp user application-password create reportviewer test --porcelain --url=central.wordcamp.test --allow-root
   docker exec wordcamptest-wordcamp.test-1 wp user application-password create nocapuser test --porcelain --url=central.wordcamp.test --allow-root
   docker exec wordcamptest-wordcamp.test-1 wp user application-password create admin test --porcelain --url=central.wordcamp.test --allow-root
   ```

6. Hit the endpoint as each caller and check the four outcomes:
   ```
   BASE="https://central.wordcamp.test/wp-json/wordcamp-reports/v1/campus-connect-details"
   curl -sk -o /dev/null -w "%{http_code}\n" "$BASE"                          # expect 401
   curl -sk -u "nocapuser:<pass>" -o /dev/null -w "%{http_code}\n" "$BASE"    # expect 403
   curl -sk -u "reportviewer:<pass>" "$BASE" | python3 -m json.tool           # expect 200, rows WITHOUT "Series Event", no email anywhere
   curl -sk -u "admin:<pass>" "$BASE" | python3 -m json.tool                  # expect 200, rows WITH "Series Event": "1"
   ```
   Also confirm the `wcpt-pre-planning` event appears (proving non-public statuses are included, as intended) and that `canary@example.test` appears in no response.

7. Clean up (delete the event ID from step 4, the users, the admin app password, and revert the config):
   ```
   docker exec wordcamptest-wordcamp.test-1 wp post delete <event-id> --force --url=central.wordcamp.test --allow-root
   docker exec wordcamptest-wordcamp.test-1 wp user delete reportviewer nocapuser --network --yes --url=central.wordcamp.test --allow-root
   docker exec wordcamptest-wordcamp.test-1 wp user application-password delete admin --all --url=central.wordcamp.test --allow-root
   git checkout -- .docker/wp-config.php
   ```

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
```diff
@@ -200,6 +211,100 @@ public function get_event_posts() {
...
+	 * @return \WP_REST_Response|\WP_Error
+	 */
+	public static function rest_callback( $request ) {
+		/*
+		 * Ask only for fields this caller can actually be given. The safelist is
+		 * capability-filtered, and `validate_fields_input()` rejects the whole
+		 * request if it names one field the caller cannot have -- so without this
+		 * intersection a user holding exactly `view_wordcamp_reports` passes the
+		 * permission check and then receives a 500 instead of their report.
+		 */
```
KokkieH, line 308: **Priority/blocking.** This comment (and the ones on `get_rest_fields()` / `rest_permission_callback()`) credit the capability-filtered safelist with keeping private fields off the wire — but the safelist never actually filters rows on this path. In `Base_Details::get_data()`:

```php
$this->filter_data_fields( $data );
```

`filter_data_fields()` takes `$data` by value and returns the filtered copy, so the result is discarded. Every other report in the plugin assigns it (`$data = $this->filter_data_fields( $data );` in eleven call sites); `class-base-details.php:152` is the only one that doesn't. As a result `get_data()` returns full rows — organiser e-mail included — and the only thing keeping them out of the response is the `unset` loop in `prepare_data_for_display()` combined with the intersected `fields` option. That held up in my live test (seeded contact meta never appeared), but it's a single layer where the comments promise two.

This is a pre-existing quirk shared by the admin screen and CSV export rather than something this PR introduces, but since this PR's comments lean on the safelist, I'd fix it here: assign the return value in `class-base-details.php`, or reword these comments to describe what actually enforces the field set.

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
```diff
@@ -200,6 +211,100 @@ public function get_event_posts() {
...
+		$probe  = new static( null, null, false, array( 'public' => false ) );
+		$fields = array_values(
+			array_intersect(
+				self::get_rest_fields(),
```
KokkieH: Non-blocking: the rest of this method uses `new static`, so a subclass overriding `get_rest_fields()` would be silently ignored here.

```suggestion
				static::get_rest_fields(),
```

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
```diff
@@ -200,6 +211,100 @@ public function get_event_posts() {
...
+			'Start Date (YYYY-mm-dd)',
+			'End Date (YYYY-mm-dd)',
+			'Status',
+			'Name',
+			'Organizer Name',
+			'Venue Name',
+			'_venue_city',
+			'_venue_country_name',
+			'Number of Anticipated Attendees',
+			'Actual Attendees',
```
KokkieH, line 271: Non-blocking: this field's presence depends on request context in a non-obvious way. `WordCamp_Admin::meta_keys()` unsets `Actual Attendees` whenever `get_post()` returns a post whose status isn't `wcpt-closed` — it survives here only because there's no global `$post` during a REST dispatch. That held in testing, but it's fragile; worth a short comment so a future change that sets up postdata before this runs doesn't silently drop the field.

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
```diff
@@ -200,6 +211,100 @@ public function get_event_posts() {
...
+			'Series Event',
+			'Created',
+			'Tracker URL',
```
KokkieH: Non-blocking: `Tracker URL` comes from `get_edit_post_link()`, which returns `null` for callers who can't edit the post — i.e. exactly the `report_viewer`-only audience this endpoint is for (confirmed live: every row had `"Tracker URL": null`). Either drop it from this list or build the `post.php?post=…&action=edit` URL directly.

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
```diff
@@ -200,6 +211,100 @@ public function get_event_posts() {
...
+
+		if ( ! empty( $messages ) ) {
+			return new WP_Error(
+				'wcr_campus_connect_details_failed',
+				implode( ' ', $messages ),
+				array( 'status' => 500 )
+			);
+		}
+
+		return rest_ensure_response( $report->prepare_data_for_display( $report->get_data() ) );
```
KokkieH: Non-blocking: this returns display-formatted output — translated status labels, renamed columns, JSON keys containing spaces and parentheses. Any consumer that parses `Status` or keys off column names breaks when a label or the locale changes. Since the stated goal is automation, emitting `get_data()` values (with the existing `Base::prepare_rest_response()` wrapper, which `Ticket_Revenue::rest_callback()` already uses) plus a route schema would be the more stable contract. Related nits: the route registers no `args`/schema, and `$request` is unused.

public_html/wp-content/plugins/wordcamp-reports/tests/test-report-rest-authorization.php
```diff
@@ -0,0 +1,159 @@
...
+	 * subrole -- must actually receive the report, not a 500.
+	 */
+	public function test_allows_user_with_only_the_report_capability() {
+		wp_set_current_user( self::$subscriber_id );
+		add_filter( 'user_has_cap', array( $this, 'grant_report_cap' ) );
+
+		$response = $this->dispatch();
+
+		$this->assertSame( 200, $response->get_status() );
+		$this->assertIsArray( $response->get_data() );
```
KokkieH: **Priority/blocking.** This is the test the file's docblock calls load-bearing, but no `campusconnect` fixtures exist, so `get_event_posts()` returns `[]` and this reduces to `assertIsArray( [] )`. It proves the authorization contract (valuable!) but not the field contract: adding `E-mail Address` to `get_rest_fields()` would pass every test in this file.

Suggest creating a fixture post in `wpSetUpBeforeClass()` — `wordcamp` post type, `event_subtype => campusconnect`, a non-public status, plus canary meta like `E-mail Address`/`Telephone` — then asserting the exact key set of a returned row (and that the canaries are absent).

public_html/wp-content/plugins/wordcamp-reports/index.php
```diff
@@ -343,12 +343,25 @@ function register_rest_endpoints() {
...
+			 * Reports read private post meta and non-public post statuses, so a
+			 * route is restricted to users who may view reports unless the
+			 * report class opts into something else with its own
+			 * `rest_permission_callback()`.
+			 */
+			$permission_callback = method_exists( $class, 'rest_permission_callback' )
+				? array( $class, 'rest_permission_callback' )
+				: function () {
+					return current_user_can( CAPABILITY );
+				};
```
KokkieH: Non-blocking: this default is the right architecture (default-deny, explicit opt-out), but it's currently dead code under test — the only registered route defines its own `rest_permission_callback()`, so nothing exercises this closure. A small test registering a dummy report class without the method would cover it, since this is the line protecting every *future* report.

KokkieH (changes requested):
Thanks for this — the authorization model is sound, and the intersection trick to keep a `report_viewer`-only caller from hitting a 500 is a nice catch. I reviewed the full permission/safelist chain and tested the endpoint end to end against a local environment (details below). Requesting changes for two priority items; everything else is non-blocking.

## Priority / blocking

1. **The safelist doesn't actually filter on this code path** — see inline comment on `rest_callback()`. One-line fix in `class-base-details.php`.
2. **The success-path test asserts against an empty dataset** — see inline comment on the test file. Needs a fixture so the field contract is actually exercised.

## Non-blocking (inline where possible)

- `self::` → `static::` in `rest_callback()` (suggestion inline).
- `Tracker URL` is `null` for exactly the audience this endpoint targets (inline).
- `Actual Attendees` inclusion silently depends on there being no global `$post` during REST requests (inline).
- The response is display-formatted output rather than stable data (inline).
- The default `permission_callback` in `index.php` is never exercised by the tests (inline).
- `get_rest_fields()` duplicates the first 15 entries of `get_field_order()` verbatim; the two will drift with nothing catching it. A test asserting the subset relationship (or deriving one from the other) would pin it.
- In the parent `prepare_data_for_display()`, `$all_statuses[ $value ]` assumes every returned status is a `wcpt-*` status. `post_status => 'any'` also matches core statuses (`draft`, `publish`, …), which would trigger an undefined-key warning and pass `null` onward. A `?? $value` guard would close it.
- The 500 error path forwards internal validation messages to the client. They aren't sensitive today, but a generic message plus a logged detail would be more conventional.
- **If anyone later adds caching to this report** (it currently has none, and the class comment saying it's admin-only is now outdated): note that `Base::get_cache_key()` varies only on slug + public/private, while the safelist is per-user capability. A naive short-TTL cache would serve an admin's rows — including `Series Event` — to `report_viewer` callers. Any cache key here must incorporate capability context.

## Testing performed

- `WordCamp Reports` PHPUnit suite: **OK (23 tests, 27 assertions)**.
- Live authorization matrix against `GET /wp-json/wordcamp-reports/v1/campus-connect-details` on a local environment, using application passwords:
  - Anonymous → **401**
  - Authenticated user without the capability → **403**
  - Subscriber with only the `report_viewer` subrole → **200**, rows **without** `Series Event`
  - Super admin → **200**, rows **with** `Series Event`
- Seeded a `campusconnect` event in `wcpt-pre-planning` with `E-mail Address` and `Telephone` meta: the non-public-status event correctly appeared in both 200 responses, and the contact-detail canaries appeared in **neither**.

## Verified as sound

- The capability can't be self-granted: it comes only from the `$wcorg_subroles` config global, and `omit_usermeta_caps()` strips usermeta-stored caps. The plugin (and therefore the route) is active only on central, which is also the only site where the subrole grants the cap.
- The `Series Event` capability-gating claim is accurate (`WordCamp_Admin::meta_keys()` drops it without `manage_options`), and the intersection handles it — verified live in both directions.
- Adding the default `permission_callback` closes no previously-public route (`Ticket_Revenue::$rest_base` is commented out), and it flips the plugin to default-deny for future reports, which is the right architecture.
- phpcs: no new violations.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
maciejpilarski: Fixed in `class-base-details.php` — `$data = $this->filter_data_fields( $data );`. You are right that it was the odd one out; `filter_data_fields()` takes `array $data` by value and returns the filtered copy, so this call site was a no-op while the other eleven assign it.

I took the fix rather than rewording the comments, because after tracing it the comments were describing the design correctly and the code was the thing out of step. It also matters more here than on the admin path: I have moved the row narrowing into `rest_callback()` (see the other thread), and that only counts as a second layer if `get_data()` is genuinely filtering as the first.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

public_html/wp-content/plugins/wordcamp-reports/tests/test-report-rest-authorization.php
maciejpilarski: Fixed. The test file now seeds a `campusconnect` event in `wpSetUpBeforeClass()` — a non-public status where wcpt has registered one, falling back to draft — with `E-mail Address` and `Telephone` canaries, and tears it down after.

The success case now locates that fixture row by ID and asserts its **exact key set** rather than probing for known-bad keys, which is what makes your example fail: adding `E-mail Address` to `get_rest_fields()` now breaks `test_report_viewer_receives_exactly_the_published_fields`. There is also a canary test that scans the whole serialised response for both values, for a report_viewer and an administrator, so a leak through some other column fails too.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
maciejpilarski: Applied — `static::get_rest_fields()`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
maciejpilarski: Documented. `get_rest_fields()` now carries a paragraph saying `Actual Attendees` survives only because there is no global `$post` during a REST dispatch, and that a future change setting up postdata before this runs would silently drop it — flagged as a deliberate dependency rather than an accident.

I left the behaviour as is. Forcing the field would mean reaching around `meta_keys()`, and the field genuinely is capability/context-gated by design.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
maciejpilarski: Dropped from `get_rest_fields()`. Building the URL by hand would hand every caller a link to a screen the `report_viewer` audience cannot open, which seemed worse than omitting the column.

It stays in `get_field_order()`, so the admin screen and CSV export are unchanged. There is now a test asserting it is not published, so it does not drift back in.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

public_html/wp-content/plugins/wordcamp-reports/index.php
maciejpilarski: Extracted the closure into a named function, `WordCamp\Reports\default_rest_permission_callback()`, so it can be called directly, and added `test_default_permission_callback_requires_the_capability` covering anonymous, logged-in-without-cap, and capability-holder.

That seemed better than registering a dummy report class: `get_report_classes()` is a hardcoded list, so injecting one would have meant a seam that exists only for the test.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

public_html/wp-content/plugins/wordcamp-reports/classes/report/class-campusconnect-details.php
maciejpilarski: Agreed and switched — the callback now emits `get_data()` values wrapped in `prepare_rest_response()`, so keys are the raw meta names, `Status` is the `wcpt-*` slug rather than a translated label, and dates are the stored values.

**One thing worth flagging, because taking the suggestion literally would have leaked.** `options['fields']` is consumed in exactly one place — the `unset` loop in `WordCamp_Details::prepare_data_for_display()`. `get_data()` never narrows by it. So dropping `prepare_data_for_display()` and returning `get_data()` widens the response from the 14 published fields to the entire private safelist, `E-mail Address` and `Telephone` included. It is the same coupling you identified in the `filter_data_fields()` thread, seen from the other end.

So the narrowing is now explicit in `rest_callback()`: `array_intersect_key()` against the intersected field list. With the `filter_data_fields()` fix that gives two real layers — safelist, then published list — instead of one layer and a comment claiming two. The canary tests cover it in both directions.

`$request` is now documented as unused rather than silently ignored: the route takes no parameters.

**I did not add a schema, deliberately.** The values are post meta and come back as untyped strings and occasional nulls, so anything I declared would be describing what I want rather than what the endpoint returns, and I have no live instance to check it against. Happy to add one if you would rather have it — I would want to write it from a real response rather than from the field list.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

maciejpilarski:
Thanks — that was a genuinely useful review, and the local-environment matrix especially. All seven inline comments are addressed (replies on each thread), and the four items that were review-body-only are done too:

- **`get_rest_fields()` drifting from `get_field_order()`** — `test_published_fields_are_a_subset_of_the_field_order()` asserts `array_diff()` is empty. I kept them as separate lists rather than deriving one from the other, because the published set is a deliberate editorial choice (it omits `Tracker URL` and the whole private tail) rather than a prefix of the order, and a test pins the relationship without pretending otherwise.
- **`$all_statuses[ $value ]` on a non-wcpt status** — `?? $value` in `WordCamp_Details::prepare_data_for_display()`, with a comment noting `post_status => 'any'` also matches core statuses. This only affects the admin and CSV paths now, since the REST path no longer goes through that method.
- **Internal validation detail in the 500** — now logged via `wp_trigger_error()` and the caller gets a generic "The report could not be generated."
- **Caching** — the class docblock has a standing warning: `Base::get_cache_key()` varies only on slug and public/private, while the safelist is per-user capability, so any cache added here must incorporate capability context or it will serve an administrator's rows to a `report_viewer`. I also corrected the stale "only used in WP Admin" line you spotted.

Two things I did **not** do, both explained on their threads: no route schema (the values are untyped post meta and I have no live instance to write it against — happy to add if you want it), and `Actual Attendees` keeps its context dependency, now documented rather than removed.

Worth calling out one thing your `filter_data_fields()` finding turned up. Taking the "emit `get_data()` values" suggestion literally would have leaked contact details: `options['fields']` is consumed only inside `prepare_data_for_display()`, so bypassing that method widens the response to the whole private safelist. The narrowing is now explicit in `rest_callback()` instead, which is what makes the two layers your comment asked for real.

_🤖 Replied via [Claude Code](https://claude.com/claude-code)_

maciejpilarski:
Heads up on a sixth file appearing in the diff: `wcpt/wcpt-wordcamp/wordcamp-admin.php`.

The new fixture in `test-report-rest-authorization.php` creates a Campus Connect event in `wcpt-pre-planning`, which fires `add_organizer_to_central()`. That method read `$_POST['wcpt_wordpress_org_username']` unguarded, so outside an admin POST the key is absent — a warning, and then `null` into `get_user_by()`, which PHP 8 deprecates. The suite itself was green (1109 tests, 2793 assertions); it was the "unexpected entries in the error log" gate that failed the build.

I fixed the read rather than working around it in the fixture, since a wordcamp can legitimately reach pre-planning without an admin POST — WP-CLI, an importer, a test — and the warning would fire there too. Behaviour is unchanged when the field is present; when it is absent the method now returns early instead of looking up a user by null.

Happy to split it into its own PR if you would rather keep this one to `wordcamp-reports` — it just has to land somewhere before this branch can go green.

_🤖 Reported via [Claude Code](https://claude.com/claude-code)_

KokkieH (approved):
Claude ran through the test steps on my local environment again:

Full matrix passed:

1. Anonymous → 401; subscriber without the cap → 403.
2. report_viewer (subrole via $wcorg_subroles, app-password curl) → 200 with exactly the 13 promised fields — raw keys, raw values (wcpt-pre-planning slug, Y-m-d date), wrapped in {report_name, report_description, data}, no Series Event.
3. Administrator → 200 with 14 fields including Series Event: "1".
4. Canaries (manual-canary@example.test, 555-0100-manual-canary) and the keys E-mail Address, Telephone, Tracker URL absent from both full response bodies.
5. Admin report/CSV path still intact after the base-class fix: prepare_data_for_display() as an administrator returns the selected columns, including private ones like E-mail Address, with display renaming working.

Critical issues have also been addressed, and the expanded test suite still passes, so this looks good to go to me.

maciejpilarski:
Thanks for the approval and for re-running the matrix.

The branch had gone into conflict, and the cause turned out to be a happy one: `755f8d91` fixed `add_organizer_to_central()` on production while this was open, with the same guard-and-early-return shape I had added — and better wording on the comment. So I have reverted my copy and taken production's, which resolves the conflict and drops that sixth file. **This PR is back to `wordcamp-reports` only**, exactly as it was when you reviewed it: no change to any file you approved.

`production` is merged in and the branch is up to date.

_🤖 Updated via [Claude Code](https://claude.com/claude-code)_

KokkieH:
Confirmed the PR is still in the correct state after conflict resolution. This is good to go 🚢

Issue #1964, opened 2026-08-27
i18n: remaining unloaded custom text domains after #1927


ekamran:
#1927 documented custom text domains without a loaded mo file. #1962 fixed the wordcamp domain in the block pattern registration code, and the issue was closed. Two findings remain from that audit, verified against current production after the merge.

1. Three pattern content files still carried the wordcamp domain in their Read More labels: speaker-list-basic.php, speaker-grid-centered.php and speaker-grid-basic.php. Pattern files execute through ob_start() and are included by patterns.php, so these calls run. I am opening a PR that finishes the swap from #1962.

2. The Groups code still uses two custom domains with no loader in this repository: wporg-groups-frontend, 138 gettext calls in 22 files, and groups-site, 19 calls in 2 files. The only custom domain loaded by the platform is wordcamporg, and no wp.org plugin named wporg-groups-frontend exists for just in time loading to find. groups-site is a theme domain, so the same argument is weaker there.

Raising point 2 as an audit question rather than a bug report, same as #1927: I cannot tell from the repository whether the Groups code serves this network, is synced from a repository where those domains are intentional, or whether mo files for these domains are provisioned outside the repository. If the domains are intentional, closing this with a short note would settle the question. If not, the fix is either swapping the domains to wordcamporg or loading the extra domains alongside it.

bor0:
Both points are resolved.

1. The three speaker patterns use `wordcamporg` for their Read More labels now.
2. #2113 moved the `wporg-groups-frontend` and `groups-site` strings onto `wordcamporg` and added `wp_set_script_translations()` for the Groups blocks. The only `groups-site` left is a block pattern category slug, not a text domain.

Verified locally with a test `es_ES` mo and a user whose profile locale is `es_ES`: theme nav and Groups block strings both render translated on the group front end.

PR #1965, opened 2026-08-27
Blocks: Fix text domain in speaker pattern content files

ekamran:
Follows up #1962, which replaced the wordcamp text domain with wordcamporg in the pattern registration code. Three pattern content files carry the same wrong domain in their Read More labels. Pattern files execute through ob_start() and are included by patterns.php, so these calls run at registration and the label can never translate: only the wordcamporg domain has a loaded mo file, and no wp.org plugin named wordcamp exists for just in time loading to find.

Same class of change as #1962, three files, three lines.

Fixes the first point of #1964. See #1927 for the original audit; the remaining Groups domains question stays in #1964.


bor0:
Thanks for the contrib! 🙌

PR #1966, opened 2026-08-28
Budget: resolve the map_meta_cap subject from the check's arguments

KokkieH:
`WordCamp_Budgets::get_map_meta_cap_post()` returns the post that a `map_meta_cap` callback is being asked about. The three budget CPTs use it to decide whether a record past draft needs `manage_network` to edit or delete, so it has to resolve the post the capability check actually names.

It didn't, in two ways:

- It preferred the global `$post` over `$args`, ignoring the argument entirely whenever a global happened to exist. The global is frequently a different post -- during a bulk edit, or while a list table builds the row actions for a post other than the one in the loop -- so the decision could be made about the wrong record, in either direction.
- Its `$args[0]` fallback required `is_int()`. A post ID reaches `current_user_can()` as whatever the caller passed, and Core itself passes strings in places (`wp_ajax_upload_attachment()` forwards `$_REQUEST['post_id']` uncast), so those lookups returned `null` and the callbacks fell through to their early return.

Now the argument comes first and the global is only a fallback for checks that don't name a post, matching how Core's own `map_meta_cap()` resolves its subject. `WP_Post` objects and numeric-string IDs both resolve; anything that names nothing resolvable stays `null`. The "don't create a global `$post` if there isn't one" behaviour is unchanged.

The numeric check mirrors `get_post()`'s own contract rather than being an extra restriction: `get_post()` gates on `is_numeric()` before calling `WP_Post::get_instance()`, and `map_meta_cap()` appends `do_not_allow` when it gets `null` back. So an ID carrying trailing characters names no post here for the same reason it names none in Core.

The second commit keeps that lookup off capability checks the callbacks don't act on. `map_meta_cap` runs for every capability check site-wide, and all three callbacks resolved the post before looking at which capability was requested, so an `$args[0]` that isn't a post ID -- a user or term ID, say -- meant a `get_post()` miss per callback, and `WP_Post::get_instance()` doesn't cache misses. Each callback now returns early unless the requested capability is one of the ones it acts on (`edit_post`/`delete_post`, plus `draft_post` for reimbursements). Behaviour is identical; for the capabilities that still reach the lookup, Core has already primed the cache with its own `get_post( $args[0] )`.

`budget-tool.php`'s `map_meta_cap` filter doesn't use this helper and is unaffected.

### Tests

New `Test_Map_Meta_Cap` in the Budgets Dashboard suite, whose bootstrap already loads the three payment modules. For each budget CPT it asserts the `edit_post`/`delete_post` outcome across all three argument forms -- int ID, numeric-string ID, `WP_Post` -- and with both an unrelated post and a budget record sitting in the global `$post`. Network admins and an author's own draft are covered too, so the mapping is asserted in both directions rather than only as a denial. A further case pins `get_post()`'s contract for an ID with trailing characters, so a change to it upstream surfaces here rather than silently shifting what this helper resolves.

The suite bootstrap now registers the shared `wcb-*` post statuses. `WordCamp_Budgets`' constructor does that at runtime, but the plugin only instantiates it in the admin, so storing a post at `wcb-approved` was emitting `Attempt to read property "label" on null` from `reimbursement-request.php`.

1. Check out this branch and run the suite:
   ```
   ./public_html/wp-content/mu-plugins/vendor/bin/phpunit -c phpunit.xml.dist --testsuite "WordCamp Budgets Dashboard"
   ```
   Expect `OK (40 tests, 143 assertions)`, with no `Attempt to read property "label" on null` warnings in the output.

2. Confirm the new tests are load-bearing by restoring the helper to its `production` version, re-running, and putting it back:
   ```
   git checkout production -- public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
   ./public_html/wp-content/mu-plugins/vendor/bin/phpunit -c phpunit.xml.dist --filter Test_Map_Meta_Cap
   git checkout HEAD -- public_html/wp-content/plugins/wordcamp-payments/includes/wordcamp-budgets.php
   ```
   Expect 13 failures across `Test_Map_Meta_Cap` on the middle command, and a clean `git status` after the third.

3. Run the full suite to confirm nothing else moved:
   ```
   npm run build --workspace=public_html/wp-content/mu-plugins/wporg-groups-frontend
   ./public_html/wp-content/mu-plugins/vendor/bin/phpunit -c phpunit.xml.dist
   ```
   Expect 1049 tests. Six skip without the `intl` extension. Without the build step above, the eight `Test_Groups_Blocks` page-content tests fail -- that baseline is identical on `production` at 0296b646 and is unrelated to this branch.

4. In wp-admin on a camp site, as a network admin, go to **Budget > Reimbursement Requests** (and the Vendor Payments and Sponsor Invoices screens). Confirm the list tables render, rows are linked, and the Edit row action appears.

5. As an organizer who is not a network admin, create a reimbursement request, save it as a draft, and confirm the form fields are editable. Submit it for review, reopen it, and confirm the request is still reachable from the list table and its details are still readable.

6. As a network admin, open a request that has been approved and confirm it can still be edited and saved.

7. Visit a few screens that check capabilities against things that aren't posts -- **Users**, **Comments**, and a taxonomy screen such as **Posts > Tags** -- and confirm they render normally and that row actions appear as before.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments-network/tests/bootstrap.php
+++ b/public_html/wp-content/plugins/wordcamp-payments-network/tests/bootstrap.php
@@ -28,6 +28,10 @@ function manually_load_plugin() {
 	// budget CPTs.
 	$GLOBALS['wcp_payment_request'] = new \WCP_Payment_Request();
 
+	// The shared `wcb-*` statuses are registered by `WordCamp_Budgets`' constructor, which the plugin only
+	// instantiates in the admin. Register them on their own so posts can be stored at those statuses.
+	add_action( 'init', array( '\WordCamp_Budgets', 'register_post_statuses' ) );
+
 	require_once dirname( __DIR__ )  . '/includes/payment-requests-dashboard.php';
 	require_once dirname( __DIR__ )  . '/includes/wordcamp-budgets-dashboard.php';
 }
--- a/public_html/wp-content/plugins/wordcamp-payments-network/tests/test-map-meta-cap.php
+++ b/public_html/wp-content/plugins/wordcamp-payments-network/tests/test-map-meta-cap.php
@@ -83,30 +83,34 @@ class Test_Map_Meta_Cap extends WP_UnitTestCase {
 
 		unset( $GLOBALS['post'] );
 
-		remove_action( 'save_post', 'WordCamp\Budgets\Reimbursement_Requests\save_request', 10 );
-		remove_action( 'save_post', 'WordCamp\Budgets\Sponsor_Invoices\save_invoice', 10 );
-		remove_action( 'save_post', array( $GLOBALS['wcp_payment_request'], 'save_payment' ), 10 );
-
-		remove_filter( 'wp_insert_post_data', 'WordCamp\Budgets\Reimbursement_Requests\set_request_status', 10 );
-		remove_filter( 'wp_insert_post_data', 'WordCamp\Budgets\Sponsor_Invoices\set_invoice_status', 10 );
-		remove_filter( 'wp_insert_post_data', array( $GLOBALS['wcp_payment_request'], 'wp_insert_post_data' ), 10 );
+		foreach ( $this->save_handlers() as $handler ) {
+			remove_filter( $handler[0], $handler[1], 10 );
+		}
 	}
 
 	/**
-	 * Re-attach the handlers detached in `set_up()`.
+	 * `WP_UnitTestCase` restores `$wp_filter`, so the handlers detached in `set_up()` re-attach themselves.
 	 */
 	public function tear_down() {
-		add_action( 'save_post', 'WordCamp\Budgets\Reimbursement_Requests\save_request', 10, 2 );
-		add_action( 'save_post', 'WordCamp\Budgets\Sponsor_Invoices\save_invoice', 10, 2 );
-		add_action( 'save_post', array( $GLOBALS['wcp_payment_request'], 'save_payment' ), 10, 2 );
-
-		add_filter( 'wp_insert_post_data', 'WordCamp\Budgets\Reimbursement_Requests\set_request_status', 10, 2 );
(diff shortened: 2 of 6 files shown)


bor0:
Tested this locally against current `production` (`a88ea211`, so including #1955) — merges cleanly.

**Test results**

- Full suite with this branch merged in: 1063 tests, 0 failures (6 known skips — `intl`, and the GatherPress data-provider skip).
- Budgets Dashboard suite on its own: 54 tests, 0 failures.
- `phpcs` on the branch diff: clean.

I also confirmed the new tests are load-bearing by reverting the helper to its `production` version and re-running — the `Test_Map_Meta_Cap` cases fail as described, and pass again once restored.

**Behaviour check**

Beyond the unit tests, I exercised the four argument/global-`$post` combinations on a local site with the full plugin stack loaded, for both a network admin and an organizer who isn't one, across all three budget CPTs. This branch corrects the outcome in every case where `production` currently gets it wrong, and it does so in both directions — the over-restriction case is fixed alongside the under-restriction ones. Network admin behaviour is identical throughout, and capability checks that aren't about posts (users, comments, terms) are unaffected.

The early-return capability lists in the second commit are exhaustive against each callback's branches (`edit_post`/`delete_post`, plus `draft_post` for reimbursements), so nothing is dropped, and the bootstrap change matches what `WordCamp_Budgets`' constructor registers.

**One note for anyone else reviewing this**

The `is_numeric()` check in `get_map_meta_cap_post()` mirrors a change that landed in WordPress 7.1 — before that, `get_post()` passed its argument to `WP_Post::get_instance()` without the numeric gate. If your local PHPUnit suite is pinned to an older core you'll see three failures in `Test_Map_Meta_Cap` that are purely an artefact of that. Worth checking, because `install-wp-tests.sh` returns early when the core directory already exists, so a local suite never picks up newer versions on its own:

```
grep wp_version .docker/test_suite/wordpress/wp-includes/version.php
```

CI installs `latest` every run, which is why it's green.

LGTM.

bor0 (approved):

PR #1968, opened 2026-08-28
Groups: give the post editor canvas a side gutter (#1967)

### LINKED ISSUE #1967: Group site: post editor content has no side gutter
Editing a page on a group site puts the content flush against the left edge of the editor frame. Same page type on the main events site is fine.

### Steps

1. Open `/group/<slug>/wp-admin/post.php?post=<id>&action=edit`.
2. Look at the left edge of the canvas.

### Cause

`groups-site` sets `contentSize: 1160px` and no root padding in `theme.json`. The canvas is 1160px wide at a 1440px window (1440 minus the 280px settings sidebar), so the content-width cap never binds and the root container is left with zero padding.

The front end never shows this because the gutter comes from the `edge-space` group that wraps `post-content` in `templates/page.html`, and the post editor does not render the template.

The main events site is fine only because its `contentSize` is 680px, narrower than the canvas.

### Screenshots

| Description | Screenshot |
| --- | --- |
| Group site page editor | <kbd>Image</kbd> |
| Main site page editor | <kbd>Image</kbd> |

gedex:
Fixes #1967.

Page content in the group site post editor sits flush against the left edge of the frame.

The theme sets `contentSize: 1160px` and no root padding, and the canvas is 1160px wide at a 1440px window, so the content-width cap never binds. On the front end the gutter comes from the `edge-space` group wrapping `post-content` in the templates, which the post editor does not render.

Fix pads the canvas root container and the title wrapper, scoped to the post editor. The site editor renders the template and already has its gutter.

## Test plan

- [x] phpcs clean
- [x] Manual pass at 1440px: title and first paragraph both start at x=70 instead of x=0
- [x] Site editor unchanged, no double gutter
- [ ] CSS only, no phpunit change

### Screenshots

| Description | Screenshot |
| --- | --- |
| Before | <kbd>Screenshot 2026-08-28 at 16 35 49</kbd> |
| After | <kbd>image</kbd> |

### DIFF
--- a/public_html/wp-content/themes/groups-site/functions.php
+++ b/public_html/wp-content/themes/groups-site/functions.php
@@ -54,17 +54,14 @@ add_action( 'enqueue_block_editor_assets', __NAMESPACE__ . '\enqueue_assets' );
 /**
  * Give the post editor canvas a side gutter. See #1967.
  *
- * The front end gets its gutter from the `edge-space` group wrapping
- * `post-content` in the templates, which the post editor never renders. Not
- * done as `styles.spacing.padding` in `theme.json`, because root padding also
- * lands on `.wp-site-blocks` and would stack with the templates' own.
- *
- * Only the post editor. The site editor draws the template and its gutter.
+ * Not `styles.spacing.padding` in `theme.json`: root padding also hits
+ * `.wp-site-blocks`, stacking with the gutter the templates already set.
  *
  * @param array                    $settings Block editor settings.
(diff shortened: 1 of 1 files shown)


bor0 (approved):
Discussed with Akeda

The theme's group-frontend tweak drops the editor's default gutter, and this is the only hook with editor context, so the site editor isn't double-padded. theme.json root padding would leak to the front end. Groups-site block editor only, so no regression risk.

We can revisit the underlying theme tweak separately.

PR #1969, opened 2026-08-28
Groups: let beta testers switch their own role tier

bor0:
## Summary

The "Call for testing: WordPress community groups on events.WordPress.org" post tells people who want to try the organizer side to comment on the post or ping `#gatherpress-migration` so we can promote them by hand. This makes that self-serve on the testing groups.

- **"Your role in this group" panel** above the member grid on `/members/`, with a card per tier and a one-line description of what each one grants.
- **`POST /wporg-groups/v1/members/me/role`** — new endpoint, deliberately separate from `/members/{id}/role` (the organizer-facing member manager, which keeps refusing self-changes).
- **Allow-listed to testing groups** via `Capabilities\SELF_SERVE_ROLE_GROUPS`, keyed by host so a community group can't inherit it by sharing a slug with a fixture. Currently `events.wordpress.org/group/internal-testing-group/` plus the local/PHPUnit fixture group. A `wporg_groups_frontend_self_serve_roles_enabled` filter is there for sandboxes.

## How to review this PR

Suggested order: `inc/capabilities.php` → `inc/class-members-controller.php` → `src/blocks/group-members/render.php` → `view.js` → `style.css`.

- **Why an allow-list rather than a general feature.** The Organizer tier grants management of everyone's events, member roles, group info and design, and ownership transfer. Self-serve promotion into it only makes sense on a group that exists to be poked at, so the gate is a code constant with no admin toggle — nobody should be able to switch this on for a live group from the UI.
- **Why a separate route.** Relaxing `update_member_role()`'s `cannot_change_own_role` guard would have changed behaviour for organizers managing other members. The new route reuses that controller's `can_demote_organizer()` floor instead, so the last organizer still can't step down and strand the group.
- **Administrators are excluded**, matching `update_member_role()`'s existing refusal to touch them — the group's admin account is provisioned in wp-admin, and a front-end self-demotion could leave nobody able to restore it.
- **`current_user_can_switch_own_role()` is the single authority.** The block calls it to decide whether to render, and the permission callback ends with it as a catch-all, so the button and the request it makes can't drift apart.
- The click reloads the page rather than patching in place, matching what `group-membership` does for join/leave — a role change also affects the sidebar badge, the event-manage controls, and the settings tab, all server-rendered.

## Test plan

- [x] PHPUnit `--group groups` — 375/375 passing, including 18 new
- [x] phpcs — 0 errors across the plugin
- [x] stylelint (`npm run lint:css`) — clean
- [x] Jest — 33/33 passing, including 5 new
- [x] `npm run build`
- [x] Manual browser pass (see below)

New automated coverage: allow-list on and off, filter override in both directions, per-role capability matrix, self-promotion and self-demotion happy paths, last-organizer self-demotion blocked, `administrator` rejected both as an actor and as a target role value, non-member rejected, logged-out rejected, and the JS fetch/reload/error-surfacing paths.

### Manual test steps

Verified on the local stack (`events.wordpress.test/group/sunshine-coast-qld/`, which is on the allow-list):

1. **Logged out, members page** — no panel rendered.
2. **Server-rendered markup per role** — rendered the block as each of `member1` (subscriber), `eventorganiser1` (author), `organiser` (editor), `organiser1` (administrator). The first three get the panel with the correct card marked `aria-pressed="true"` and `is-current`; the administrator gets no panel.
3. **REST round trip against the dev database**, as `member1`: Member → Organizer → Event Organizer → Member, each returning the new role and label, each confirmed against `wp user list`. Then `administrator` as a role value → `invalid_role`, and `organiser1` (site admin) attempting a self-demotion → `cannot_manage_administrator`. Fixture data restored to its starting state afterwards.
4. **Unauthenticated endpoint** — `POST` with no session returns `rest_not_logged_in` (401).
5. **Layout** — panel checked at 360 / 560 / 800px container widths; the card grid reflows 1 → 2 → 3 columns with no horizontal overflow. No console errors on page load.

6. **Logged-in click-through**, as `member1` on the members page: clicked through Member → Organizer → Event Organizer → Member, confirming each transition landed in the database with `wp user get 13 --fields=roles` between clicks. Left back at `subscriber`.

The screenshot below is the real server-rendered markup and real block CSS, injected into a logged-out page for the layout check — the authenticated behaviour it illustrates is covered by step 6.

#### Screenshots

Screenshot 2026-08-28 at 13 48 34

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/capabilities.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/capabilities.php
@@ -111,3 +111,94 @@ function current_user_can_manage_group_settings(): bool {
 		current_user_can( 'edit_others_posts' )
 	);
 }
+
+/**
+ * Group sites where members may change their own role tier, keyed by the
+ * host the group lives on.
+ *
+ * Self-serve role switching is a beta-testing affordance, not a product
+ * feature: promoting yourself to `editor` grants the full "Organizer" tier
+ * (everyone's events, member roles, group info and design, ownership
+ * transfer), so on a real community group it would be a privilege
+ * escalation for any logged-in WordPress.org account that joined. The
+ * allow-list keeps it to the groups that exist to be poked at.
+ *
+ * Keyed by host rather than by slug alone so a real community group that
+ * happens to share a slug with a local fixture can't inherit it. Sites are
+ * added here by code change on purpose — there's no admin toggle, because
+ * nobody should be able to switch this on for a live group from the UI.
+ *
+ * @var array<string, string[]>
+ */
+const SELF_SERVE_ROLE_GROUPS = array(
+	// The beta's public testing group (see the "Call for testing" post).
+	'events.wordpress.org'  => array( 'internal-testing-group' ),
+
+	// Local development and the PHPUnit fixture site.
+	'events.wordpress.test' => array( 'sunshine-coast-qld' ),
+);
+
+/**
+ * The `/group/{slug}/` segment identifying the current group site.
+ *
+ * Returns an empty string off the groups network (or on its `/group/`
+ * root, which is the landing page rather than a group).
+ */
+function get_current_group_slug(): string {
+	$site = get_site();
+
+	if ( ! $site ) {
+		return '';
+	}
+
+	$segments = explode( '/', trim( (string) $site->path, '/' ) );
+	$slug     = (string) end( $segments );
+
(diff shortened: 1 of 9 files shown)


bor0:
@gedex feel free to provide post-merge review, I will address

gedex:
Self-promoted `author` can create a draft event and mail every member, unthrottled and repeatable. The concern is the volume and the wordpress.org sending address.

`author` grants `upload_files`, so a tester can put files on a wordpress.org urls. Something we need to check (and maybe cleanup) post-CFT.

PR #1970, opened 2026-08-28
Content: add a WP-CLI command to find references to lapsed domains

bor0:
## Summary

Published content across the network references a lot of third-party sites — links, embeds, and script sources. Some of those domains eventually lapse, and our content keeps referencing them. We have no way to notice when that happens.

`wp wc-dangling scan` walks published posts and pages, collects the third-party hosts they reference, and reports the ones whose domains no longer resolve, so stale references can be updated or removed.

- Separates a lapsed domain from a subdomain that merely stopped resolving under a domain that's still registered. They need different follow-up, so the report shouldn't conflate them.
- Confirms candidates against RDAP, so a DNS failure mid-scan doesn't get reported as a lapsed registration. Anything inconclusive downgrades rather than asserting something unverified.
- Scans `_oembed_*` postmeta alongside `post_content` — for a classic-editor embed the markup exists only in the cache, so scanning content alone misses the host entirely.
- Report-only — it has no write path at all, so there's no `--dry-run` to add. It exits non-zero when it finds something, so a scheduled run surfaces as a failure rather than succeeding with findings buried in its output.
- Each finding records the site it came from, and `--fields` selects the output columns, so "which sites are affected?" is `--fields=site --format=csv | tail -n +2 | sort -u`.

Logic lives in `mu-plugins/dangling-hosts.php` (hook-free, so it's testable on its own); `wp-cli-commands/dangling-hosts.php` is a thin wrapper.

Not wired into `cron.php` here — the scan's runtime against the real network hasn't been measured, and the command is schedulable as-is. Worth a follow-up with real timings behind it.

- Only reports hosts that could actually have been registered. A href that swept up the comma after it in the prose parses to `example.net,`, and an IP literal reduces to a "domain" of `1.1`; neither is a registration, so reporting either as lapsed asserts something untrue. Both are common enough in old content to bury the findings that matter, and skipping them keeps the DNS and RDAP lookups off them too.

- Only a real answer from the registry overturns what DNS found. A request that fails, is rate limited, or returns an unexpected status decides nothing, so the finding stands and is reported as unconfirmed rather than quietly downgraded. This matters more than it sounds: one TLD's registry closes the connection uncleanly on exactly the "no such registration" responses, so reading a failed request as "registered" would have suppressed every genuine finding under it while every healthy domain kept verifying normally.

## Test plan

- [x] phpunit — 320/321 passing (`WordCamp MU Plugins`). The one failure, `Test_QuickBooks_OAuth_State::test_authorize_url_carries_the_stored_state`, is pre-existing — verified by stashing this branch and re-running it on `production`.
- [x] phpcs clean on all 5 changed files
- [ ] npm run build — n/a, no JS
- [x] Manual CLI pass against the local stack (below)

63 new tests cover reference extraction per kind, host-shape validation, protocol-relative/relative/non-HTTP URLs, first-party filtering, registrable-domain derivation including multi-label suffixes, the three host statuses, `_oembed_` postmeta scanning, draft exclusion, kind filtering, and result ordering. DNS is stubbed through a filter so the suite never depends on the resolver or on a real domain's registration status.

### Sandbox smoke test

Also run against real content on a sandbox, which is what shook out the two fixes above -- neither was reachable from the stubbed suite:

- Reported known-stale references on a site that has some, at the right status, and correctly separated them from a domain that no longer resolves but is still registered.
- Confirmed the registry step downgrades that still-registered domain rather than reporting it as lapsed, and that the command names any domain it could not confirm.
- Verified the scan issues no writes, which is what makes it safe to point at a production database. There is now a test asserting that, so it holds for future changes rather than being re-derived by reading the code.
- Content walk measured at roughly 20-40s for a single site. That is the number the cron decision needs, and it is why this still isn't scheduled here.

### Manual test steps

No browser pass — this is CLI-only with no UI surface. Verified against the local Docker stack instead:

1. Seeded one published post on `central.wordcamp.test` covering all three outcomes: a live host, a nonexistent subdomain of a live domain, and a nonexistent registrable domain placed **only** in `_oembed_*` postmeta.
2. `wp wc-dangling scan --site=5 --skip-verify` — reported the lapsed domain as `dangling` (found via postmeta, confirming that path) and the missing subdomain as `unresolved` against its still-registered parent. Live hosts correctly absent.
3. `--kinds=embed` / `--kinds=script` — filtering works; the latter reports no findings and exits 0.
4. `--include-ok --format=count` — 18 references vs 2, i.e. the full external surface.
5. Exit code 1 with findings, 0 without.
6. `--fields=site --format=csv` — returns just the affected site URLs.
7. `--kinds=bogus`, `--fields=bogus`, and `--site=nope.example.test` — all rejected with a clear message. `--fields` is validated before the scan starts, so it fails in 0.6s rather than after a full network walk.
8. Fixture posts deleted afterward; confirmed gone via `wp post list`.

**The manual pass caught a bug the stubbed unit tests structurally could not.** `host_resolves()` originally issued one combined `dns_get_record( $host, DNS_A | DNS_AAAA | DNS_CNAME )`. That call returns `false` — not the address — when any type in the mask errors out, which is what happens on a host that has an A record but no CNAME. Every host looked unresolvable and the whole report was false positives. Now each type is queried separately, stopping at the first answer, so the common case is still a single lookup. Scan time on the fixture site dropped from 1:15 to 0:08 as a side effect.

## After merge

Nothing here changes the site on its own -- the command reports, and someone has to run it and act on what it says.

1. Run a scan across the network and triage the findings. Each one is a judgement call about a particular post: update the reference, unlink it, or point it at an archived copy.
2. Prefer a host that can reach the registries. Where it can't, findings arrive flagged unconfirmed and resting on DNS alone; `whois` settles those in seconds, and the command tells you which ones need it.
3. Decide on scheduling once there are real timings for a full run. The per-site cost is known; the network-wide cost isn't, and `scan_network()` collects every reference before it checks any host, so memory is the thing to watch on a first full run rather than the lookups.

### DIFF
--- a/public_html/wp-content/mu-plugins/dangling-hosts.php
+++ b/public_html/wp-content/mu-plugins/dangling-hosts.php
@@ -401,11 +401,27 @@ function verify_unregistered( $domain ) {
 		)
 	);
 
+	/*
+	 * A registry that couldn't be reached hasn't said the domain is registered -- it hasn't said anything.
+	 * Reporting that as `false` would read an unreachable network as evidence of registration, and quietly
+	 * downgrade every genuine finding on a host with no outbound HTTPS.
+	 */
 	if ( is_wp_error( $response ) ) {
+		return null;
+	}
+
+	$code = wp_remote_retrieve_response_code( $response );
+
+	if ( 404 === $code ) {
+		return true;
+	}
+
+	if ( 200 === $code ) {
 		return false;
 	}
 
-	return 404 === wp_remote_retrieve_response_code( $response );
+	// Rate limiting, a 5xx, an unexpected redirect. The registry didn't answer the question either way.
+	return null;
 }
 
 /**
@@ -586,13 +602,21 @@ function scan_network( array $args = array() ) {
 
 			$domain = $result['domain'];
 
-			if ( ! isset( $verified[ $domain ] ) ) {
+			// `array_key_exists`, not `isset`, so a cached `null` isn't looked up again on every host.
+			if ( ! array_key_exists( $domain, $verified ) ) {
 				$verified[ $domain ] = verify_unregistered( $domain );
 			}
 
-			if ( ! $verified[ $domain ] ) {
+			/*
+			 * Only a registry that answered can overturn what DNS found. When it couldn't be asked, the DNS
+			 * evidence is all there is, so the finding stands -- flagged as unconfirmed rather than dropped,
+			 * which is what `verified` carries to the caller.
+			 */
+			if ( false === $verified[ $domain ] ) {
 				$checked[ $host ]['status'] = 'unresolved';
 			}
+
+			$checked[ $host ]['verified'] = $verified[ $domain ];
 		}
 	}
 
@@ -605,8 +629,9 @@ function scan_network( array $args = array() ) {
 			continue;
 		}
 
-		$reference['status'] = $result['status'];
-		$reference['domain'] = $result['domain'];
+		$reference['status']   = $result['status'];
+		$reference['domain']   = $result['domain'];
+		$reference['verified'] = $result['verified'] ?? null;
 
 		$results[] = $reference;
 	}
--- a/public_html/wp-content/mu-plugins/dangling-hosts.php
+++ b/public_html/wp-content/mu-plugins/dangling-hosts.php
@@ -3,11 +3,7 @@
  * Find external references in published content whose domains no longer exist.
  *
  * Old WordCamp content links to and embeds a lot of third-party sites, and some of those domains eventually
- * lapse. A lapsed domain can be re-registered by anyone, at which point our content is pointing at whatever the
- * new owner puts there. That matters more for some references than others: a stale `<a href>` is a broken link,
- * but a WordPress oEmbed is a live consumer. Core's `wp-embed.min.js` un-hides the `wp-embedded-content` iframe
- * as soon as the framed document answers the embed handshake, so whatever the domain serves ends up rendered in
- * our page without the visitor doing anything.
+ * lapse. Our content goes on referencing them, and nothing currently notices.
  *
  * These functions only report. Deciding what to do about a reference -- update it, unlink it, point it at an
  * archived copy -- needs a human looking at the post.
@@ -54,11 +50,11 @@ const MULTI_LABEL_PUBLIC_SUFFIXES = array(
 );
 
 /**
- * Reference kinds, ordered by how much a re-registered domain would get to do with one.
+ * Reference kinds, ordered by how much the referenced host contributes to the rendered page.
  *
- * `script` is first because a script source runs in our origin. `embed` is a Core oEmbed iframe, which is
- * sandboxed but gets revealed automatically. `url` is a bare URL on its own line, which `WP_Embed::autoembed()`
- * turns into an `embed` when the post is rendered -- so it carries the same weight as one, just later.
+ * A `script` or an `embed` is loaded and used as part of the page, where a `link` is only followed if somebody
+ * clicks it, so the first two are worth reviewing first. `url` is a bare URL on its own line, which
+ * `WP_Embed::autoembed()` turns into an `embed` when the post is rendered, so it belongs with them.
  */
 const REFERENCE_KINDS = array( 'script', 'embed', 'iframe', 'img', 'link', 'url' );
 
@@ -578,10 +574,10 @@ function scan_network( array $args = array() ) {
 }
 
 /**
- * Sort the worst findings to the top.
+ * Sort the findings most worth reviewing to the top.
  *
- * Status first, then reference kind, since a dangling script source deserves attention before a dangling link
- * on the same domain.
+ * Status first, then reference kind, since a stale script source is more worth looking at than a stale link on
+ * the same domain.
  *
  * @param array $a
  * @param array $b
--- a/public_html/wp-content/mu-plugins/dangling-hosts.php
+++ b/public_html/wp-content/mu-plugins/dangling-hosts.php
@@ -385,11 +385,19 @@ function dns_records( $host, $type ) {
  * that lapsed. RDAP is authoritative, and only ever runs against the handful of candidates that reached
  * 'dangling', so it costs almost nothing.
  *
+ * Expect no answer for a lapsed `.com` in particular. Verisign's RDAP server, which every `.com` resolves to,
+ * closes the connection without a TLS `close_notify` on its 404s, and OpenSSL 3 treats that as an error, so
+ * `wp_remote_get()` discards a response it did in fact receive. Registered domains answer 200 and are
+ * unaffected, which puts the failure squarely on the domains worth reporting. Those are left standing on their
+ * DNS evidence and flagged unconfirmed, for a human to settle with `whois`. Inferring a lapse from a failed
+ * request would assert the very thing that couldn't be checked.
+ *
  * @param string $domain
  *
- * @return bool True only when the registry positively reports no such registration. Anything ambiguous (a
- *              timeout, a rate limit, an unexpected status) returns false, so an inconclusive answer downgrades
- *              the finding rather than asserting something we haven't confirmed.
+ * @return bool|null True when the registry positively reports no such registration, false when it reports one,
+ *                   and null when it didn't answer the question -- an unreachable host, a rate limit, an
+ *                   unexpected status. Null is not evidence in either direction, and only a real answer is
+ *                   allowed to overturn what DNS found.
  */
 function verify_unregistered( $domain ) {
 	$response = wp_remote_get(
--- a/public_html/wp-content/mu-plugins/dangling-hosts.php
+++ b/public_html/wp-content/mu-plugins/dangling-hosts.php
@@ -1,0 +1,611 @@
+<?php
+/*
+ * Find external references in published content whose domains no longer exist.
+ *
+ * Old WordCamp content links to and embeds a lot of third-party sites, and some of those domains eventually
+ * lapse. A lapsed domain can be re-registered by anyone, at which point our content is pointing at whatever the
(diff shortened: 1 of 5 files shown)


bor0:
## Runtime estimate for a full-network scan

Sampled random sites on a sandbox (production database) to size a full run before scheduling one. Script below; registry lookups excluded, since those only run on the few candidates that reach `dangling` and would measure network luck rather than the part that scales with site count.

| sample | elapsed | per site | peak memory |
|-------:|--------:|---------:|------------:|
|     20 |   83.5s |    4.17s |      108 MB |
|     60 |  289.5s |    4.83s |      130 MB |

1,646 sites in scope. `memory_limit` is `-1` on this host.

**Time:** ~2.2 hours, and that's a floor — a random sample misses the largest sites, and cost scales with content. Note per-site already rose from 4.17s to 4.83s between the two samples, which is the tail starting to show.

**Memory:** the raw 108 MB is mostly WordPress itself, not per-site data. The slope is what matters:

```
(130 - 108) MB / 40 sites = 0.55 MB per site
baseline: 108 - (20 x 0.55) ~ 97 MB
full run: 97 + (1646 x 0.55) ~ 1.0 GB
```

**Conclusion:** ~1 GB and a few hours. Runs straight through, no chunking needed, which is what I wanted to confirm before wiring anything to cron.

<details>
<summary>dangling-estimate.php</summary>

```php
<?php
/**
 * Rough estimate of how long a full-network dangling-host scan would take.
 *
 * Usage: wp eval-file dangling-estimate.php [sample-size]
 */

$sample = isset( $args[0] ) ? (int) $args[0] : 10;

$all = get_sites( array(
	'number'   => 0,
	'fields'   => 'ids',
	'archived' => 0,
	'deleted'  => 0,
	'spam'     => 0,
) );

$total = count( $all );

shuffle( $all );
$ids = array_slice( $all, 0, $sample );

$start = microtime( true );
\WordCamp\Dangling_Hosts\scan_network( array( 'blog_ids' => $ids, 'verify' => false ) );
$elapsed = microtime( true ) - $start;

$per  = $elapsed / count( $ids );
$full = $per * $total;

printf( "sites in scope: %d\n", $total );
printf( "sampled:        %d in %.1fs\n", count( $ids ), $elapsed );
printf( "per site:       %.2fs\n", $per );
printf( "full scan:      ~%d min (%.1f hours)\n", round( $full / 60 ), $full / 3600 );
printf( "peak memory:    %.0f MB for %d sites\n", memory_get_peak_usage( true ) / 1048576, count( $ids ) );
```

</details>

public_html/wp-content/mu-plugins/dangling-hosts.php
```diff
@@ -0,0 +1,683 @@
...
+		$sql = "SELECT ID, post_content FROM {$wpdb->posts} WHERE post_status = 'publish' AND post_type IN ( $placeholders )";
+
+		$posts = $wpdb->get_results( $wpdb->prepare( $sql, $post_types ) ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared -- Prepared immediately above.
+
+		foreach ( $posts as $post ) {
+			collect_references( $references, extract_references( $post->post_content ), (int) $post->ID, $blog_id );
+		}
+
+		// Cached oEmbed markup, which usually isn't present in `post_content`.
+		$oembed_meta = $wpdb->get_results(
```
mcliwanow: This query doesn't join posts, so oembed meta on drafts (and post types outside `--post-types`) gets scanned too. Verified locally: put `_oembed_*` meta on a draft and the scan reported it as dangling. A draft nobody can see can then fail a scheduled run. I'd join `wp_posts` and filter status + post type here. The existing draft test only covers `post_content`, so it needs a case for this path too.

public_html/wp-content/mu-plugins/dangling-hosts.php
```diff
@@ -0,0 +1,683 @@
...
+	 * The `wp:embed` block stores its URL as a JSON attribute rather than on its own line, so match that
+	 * shape too.
+	 */
+	if ( preg_match_all( '#^\s*(https?://[^\s<>"\']+)\s*$#im', $content, $bare ) ) {
+		foreach ( $bare[1] as $url ) {
+			add_reference( $references, $url, 'url' );
+		}
+	}
+
+	if ( preg_match_all( '#"url"\s*:\s*"(https?://[^"\\\\]+)"#i', $content, $block_attrs ) ) {
```
mcliwanow: nit: this matches any block with a `url` attribute, not just embeds. A `wp:button` reports twice (link from the anchor, url from the JSON attr), and url sorts above link. Checked locally with a button block. I'd anchor it to the `wp:embed` comment, or note in the comment above that the noise is deliberate.

public_html/wp-content/mu-plugins/wp-cli-commands/dangling-hosts.php
```diff
@@ -0,0 +1,297 @@
...
+		// Validated up front -- a bad column name shouldn't surface only after a full network scan.
+		$fields = $this->parse_fields( WP_CLI\Utils\get_flag_value( $assoc_args, 'fields', '' ) );
+
+		// A progress bar would interleave with the report itself in the machine-readable formats.
+		$show_progress = in_array( $format, array( 'table', 'count' ), true );
+		$notify        = null;
+
+		if ( $show_progress ) {
+			$site_count = $blog_ids ? count( $blog_ids ) : (int) get_sites( array( 'count' => true ) );
+			$notify     = new Bar( 'Scanning sites', $site_count );
```
mcliwanow: nit: raw `Bar` prints into piped output, so `--format=count | ...` captures the bar lines before the number. `WP_CLI\Utils\make_progress_bar()` no-ops when stdout isn't a TTY, which matters for cron. Also cosmetic: the total here counts archived/spam/deleted sites that `scan_network()` skips, so the bar can end short on a network run.

mcliwanow (commented):
Reviewed the code and ran the full test plan locally: phpunit 326 tests with the same single pre-existing QBO failure (re-checked it on production), phpcs clean, and all 8 manual steps against the docker stack. Got your exact 18 vs 2 counts, and the RDAP pass matched the Verisign note (both fake .coms came back unconfirmed, findings kept as dangling with the warning naming them).

One small bug inline (oembed meta on drafts gets scanned), rest are nits. Fine to fix here or as a follow-up.

public_html/wp-content/mu-plugins/wp-cli-commands/dangling-hosts.php
```diff
@@ -0,0 +1,297 @@
...
+	 * ---
+	 * default: post,page
+	 * ---
+	 *
+	 * [--kinds=<kinds>]
+	 * : Comma-separated reference kinds to report: script, embed, iframe, img, link, url.
+	 * Defaults to all of them.
+	 *
+	 * [--fields=<fields>]
+	 * : Comma-separated columns to output: status, kind, host, domain, site, blog_id, post_id, permalink, url.
```
mcliwanow: nit: `verified` is missing from this list (`parse_fields` accepts it).

public_html/wp-content/mu-plugins/dangling-hosts.php
bor0: Good catch, and confirmed — wrote your case as a failing test first and it reproduced immediately. Postmeta carries no status or type of its own, so those rows had to be qualified by the post they hang off.

Joined `wp_posts` and filtered on both status and post type. Two tests added, since the post-type half fails independently of the draft half: `test_scan_site_ignores_cached_oembed_on_unpublished_posts` and `test_scan_site_ignores_cached_oembed_on_other_post_types`. Both failed before the change.

Fixed in dbd0b800.

public_html/wp-content/mu-plugins/dangling-hosts.php
bor0: Agreed, and I went with anchoring rather than documenting it — on reflection the duplicate is the smaller half of the problem. `url` means "a bare URL that autoembed will turn into an embed", i.e. something the page loads on its own. A button is not that, so classifying it as `url` inflated the category the ordering exists to surface first, and pushed it above the `link` it duplicated.

Now scoped to `wp:embed` blocks specifically. Added your button case, plus a multiline-JSON embed block since the tighter pattern has more to get wrong.

Fixed in dbd0b800.

public_html/wp-content/mu-plugins/wp-cli-commands/dangling-hosts.php
bor0: Both fixed, and the piping one was worse than cosmetic — `--format=count | ...` was returning the bar lines along with the number.

Swapped to `WP_CLI\Utils\make_progress_bar()`, and also gated `$show_progress` on `\cli\Shell::isPiped()`. The second part matters because the no-op bar still left the blank separator line that used to divide bar from table, so piped output was `\n1\n` rather than `1\n`. Verified with `od -c`: it is now exactly `1 \n`.

Site count now uses the same `archived`/`deleted`/`spam` filters as `scan_network()`, so the bar reaches its total.

Fixed in dbd0b800.

public_html/wp-content/mu-plugins/wp-cli-commands/dangling-hosts.php
bor0: Fixed in dbd0b800 — `verified` added to the list.

mcliwanow (approved):
Re-ran everything on dbd0b800: phpunit 330 tests with only the pre-existing QBO failure, phpcs clean, and the manual pass again. Draft's oembed meta is no longer reported while the published one still comes back dangling, button block yields a single link, and piped `--format=count` is exactly `2\n` (same `od -c` check as yours).

One thing I checked on the tighter embed regex: legacy `wp:core-embed/*` blocks don't match it anymore, but their URL sits on its own line inside the wrapper, so the bare-URL fallback still catches them. No coverage lost.

LGTM, thanks.

PR #1975, opened 2026-08-31
Budgets: scope payment file edits to the people who can see them

bor0:
## Summary

`privacy.php` decides who may see a reimbursement or vendor payment attachment from the request it hangs off — `get_hidden_payment_file_ids()` keeps it visible to the uploader, to the person who filed the request, and to network admins. That makes `post_parent` the field that grants access to the file.

Core treats `post_parent` as an ordinary field, though: whoever holds `edit_post` on an attachment may set it. An organizer is an Editor on their camp site, so `edit_others_posts` covers another organizer's uploads. The rules and the capabilities disagree.

- **Capability layer.** A `map_meta_cap` callback declines `edit_post` on a payment file the visibility rules already hide. One choke point covers everything that checks capabilities before writing that field — the media REST endpoints, the Media Library's Attach/Detach actions (`wp_media_attach_action()`), XML-RPC — instead of a guard per route. It matches the `map_meta_cap` callbacks the budget post types already use.

  Only `edit_post`: that is the capability every one of those routes checks, so it is the whole of what this needs. Deletion is a separate question and isn't answered here — a co-organizer can delete another organizer's payment file, as they could before this PR.
- **The one caller that checks nothing.** `wp_update_post()` performs no capability check, so `WordCamp_Budgets::attach_existing_files()` needs its own. The Files metabox only offers files that are unattached or already on this request, but it assembles that list in the browser; the same rule now holds where the field it posts is read. The field's JSON is treated as whatever was submitted rather than as a list of IDs, which also clears a PHP 8 warning on non-numeric entries.

Nothing changes for the people the rules already let through: the uploader, the requester, and `manage_options` keep the file and their rights over it.

### Cost, since this runs everywhere

`privacy.php` loads on every request, and `map_meta_cap` runs once per attachment on the Media Library screens — including on sites that have never filed a budget request. So the guard settles the two common cases from the post cache before it reaches the lookup that costs a query: you own the file, or its parent isn't a budget request. Both restate a condition `get_hidden_payment_file_ids()` applies itself, so neither can change an answer.

`test_ordinary_media_reaches_a_verdict_without_a_query` pins that. Removing the pre-filters makes it fail with one extra query per ordinary attachment, which on a twenty-row `upload.php` is twenty.

## How to review this PR

Suggested order: `includes/privacy.php` → `includes/wordcamp-budgets.php` → `tests/test-privacy.php`.

- The callback lives in `privacy.php` rather than alongside the other `map_meta_cap` callbacks because it has to run outside the admin, and `wordcamp-budgets.php` only loads on admin, cron, and Ajax requests.
- For the same reason it can't call `WordCamp_Budgets::get_map_meta_cap_post()` and carries its own small resolver. That one falls back to the global `$post` for screens that leave it implicit; this one doesn't need to, because Core's `map_meta_cap()` denies `edit_post` outright when no post is named. `test_privacy_does_not_reference_admin_only_symbols` is what pins the separation.
- `get_hidden_payment_file_ids()` takes an optional user ID now. `map_meta_cap` is asked about a named user as often as about the current one, and `user_can( $other, ... )` would otherwise have been answered for whoever was logged in.
- The two early returns before the lookup are cost guards, not rule changes — see "Cost" above. `get_post_type( $attachment->post_parent )` also covers the unattached case, since it returns `false` for parent `0`.
- `tests/bootstrap.php` now also loads `wordcamp-budgets.php`. It defines a class and nothing else, so nothing is instantiated.

## Test plan

- [x] phpunit — 1114 tests, 1 pre-existing unrelated failure (`Test_QuickBooks_OAuth_State::test_authorize_url_carries_the_stored_state`, confirmed failing on `production` with these changes stashed)
- [x] `WordCamp Payments` suite — 32/32 passing, 20 of them new
- [x] No PHP warnings or notices in the error log, which is what CI asserts on top of the suite
- [x] Negative control on every new guard: disabled, the matching tests fail; restored, all pass
- [x] phpcs clean — no new violations, and one pre-existing error removed
- [x] npm run build — not applicable, no JS changed

### Manual test steps

Two Editors on `shinynew.wordcamp.test/2023/`, each with a draft reimbursement request. Organizer A's request carried a PDF they uploaded; Organizer B's started empty. Every end state was cross-checked with `wp post list --post_type=attachment --fields=ID,post_author,post_parent` rather than read off the UI.

1. **Organizer B, Media Library** — Organizer A's file is absent, as before this change.
2. **Organizer B, REST** — `POST wp/v2/media/<A's file>` with `{"post": <B's request>}` returns `403 rest_cannot_edit`.
3. **Organizer B, Media Library Attach** — posting `action=attach`, `media[]=<A's file>`, `found_post_id=<B's request>` to `upload.php` with a valid nonce leaves the parent where it was. (This is the route a post-save filter can't reach; `wp_media_attach_action()` writes `post_parent` with direct SQL.)
4. **Organizer B, Files metabox** — setting `#wcb_existing_files_to_attach` to A's file ID and saving the draft: "Post updated", and A's file stays on A's request.
5. **Organizer B, happy path** — added their own unattached PDF through the Add files modal and saved. It attaches, the file shows under Attached files, and "Submit for Review" becomes available.
6. **Organizer A** — still sees their file on their own request, and still holds `edit_post` on it (`wp/v2/media/<id>?context=edit` → 200).
7. **Network admin** — sees both organizers' files in the Media Library and keeps edit rights over them.

Test users, requests, and attachments were deleted afterwards.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/privacy.php
@@ -12,6 +12,7 @@ if ( ! is_cli_request() ) {
 	add_action( 'pre_get_posts',                  __NAMESPACE__ . '\unsuppress_attachment_query_filters', PHP_INT_MAX );
 	add_filter( 'posts_clauses',                  __NAMESPACE__ . '\exclude_others_payment_files', 10, 2 );
 	add_filter( 'the_posts',                      __NAMESPACE__ . '\hide_others_payment_files' );
+	add_filter( 'map_meta_cap',                   __NAMESPACE__ . '\restrict_others_payment_file_caps', 10, 4 );
 }
 
 add_filter( 'xmlrpc_prepare_media_item',          __NAMESPACE__ . '\redact_others_payment_files', 10, 2 );
@@ -233,14 +234,16 @@ function get_attachments_from_results( $posts ) {
 }
 
 /**
- * Which of the given attachments are payment files the current user may not see.
+ * Which of the given attachments are payment files the given user may not see.
  *
  * @param WP_Post[] $attachments
+ * @param int|null  $user_id     Defaults to the current user. Named explicitly by callers that are asked about
+ *                               somebody else, like a `map_meta_cap` callback.
  *
  * @return int[]
  */
-function get_hidden_payment_file_ids( $attachments ) {
-	$user_id           = get_current_user_id();
+function get_hidden_payment_file_ids( $attachments, $user_id = null ) {
+	$user_id           = is_null( $user_id ) ? get_current_user_id() : (int) $user_id;
 	$payment_posts_ids = get_payment_file_parent_ids( $attachments );
 	$hidden            = array();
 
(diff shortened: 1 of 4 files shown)


mcliwanow (approved):
Went through the whole test plan locally, two Editors on a local camp. REST reparent returns `rest_cannot_edit` with the parent unchanged, and the nonce-valid `upload.php` attach succeeds for B's own file but leaves A's where it was, so it's the guard doing the work, not a broken form.

Files metabox behaves the same, and the happy path still works (own unattached file attaches, Submit for Review shows up).

🚢

PR #1979, opened 2026-08-31
Groups: honour page visibility in the page-content block

KokkieH:
The `wporg/page-content` block embeds a page's body into another template by slug. It handles two states for the target page — missing and draft — but treats every published page alike, without consulting the page's visibility
setting. This adds handling for a page set to "Password protected" in the editor.

Both of the block's modes need the check, since each reads the stored content directly rather than going through `get_the_content()`: excerpt mode parses and renders the leading blocks, and full mode passes the content to the
`the_content` filter.

This adds the visibility check. The block now has a third target state alongside the two it already handled:

| Target page | Visitor | Organizer |
| --- | --- | --- |
| Missing | nothing | heading + "Create" button |
| Draft | nothing | heading + "Finish the draft" button |
| Password protected | nothing | heading + edit link + notice |

Keeping the heading and the edit link matches the two existing states, so the front page stays consistent for organizers rather than dropping a section. The notice reuses core's own wording from `get_the_password_form()`, without the part prompting for a password — the block sits inside another page's template, where a password field for a different page would be confusing. A visitor who has entered the page's password sees the content as before.

Requires a build, as the block renders from `build/`:

```bash
npm run build --workspace=public_html/wp-content/mu-plugins/wporg-groups-frontend
```

1. Open the local groups site, https://events.wordpress.test/group/sunshine-coast-qld/
2. Click the button to create an About page, add some content, set Status & visibility →Visibility to "Password protected", give it a password, and publish.
3. Load the group's front page in a logged-out browser or private window. The About section should not appear at all.
WordPress_Sunshine_Coast_–_Community_group_for_WordPress_users_of_all_levels_—_beginners__business_owners__and_developers__Free_monthly_meetups_covering_WordPress_topics__networking__and_skill-building_

4. Load the same front page as an organizer. The "About this group" heading and the "Edit this content" link should both be present, with the notice "This content is password-protected." in place of the page body. Confirm the page's actual text is not shown.
WordPress_Sunshine_Coast_–_Community_group_for_WordPress_users_of_all_levels_—_beginners__business_owners__and_developers__Free_monthly_meetups_covering_WordPress_topics__networking__and_skill-building_

5. Visit the About page directly at `/about/`, enter the password, then return to the front page. The section should now render its content normally.
6. Set the page's visibility back to Public. The front page should render the excerpt and the "Read more" link exactly as before this change.
WordPress_Sunshine_Coast_–_Community_group_for_WordPress_users_of_all_levels_—_beginners__business_owners__and_developers__Free_monthly_meetups_covering_WordPress_topics__networking__and_skill-building_

7. Confirm the two existing states still behave: with no About page at all the organizer sees the "Create" button, and with it in draft/private the "Finish the draft" button.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/page-content/render.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/page-content/render.php
@@ -71,6 +71,40 @@ if ( ! $target_page || 'publish' !== $target_page->post_status ) {
 
 $can_edit = current_user_can( 'edit_page', $target_page->ID );
 
+/*
+ * Password protection lives in `get_the_content()`, not in the
+ * `the_content` filter, so rendering the stored content here has to
+ * repeat the check.
+ *
+ * Visitors get nothing rather than the password form: this block embeds one
+ * page inside another, where a prompt for a different page's password would
+ * be confusing in that position. Organizers keep the heading and the edit
+ * link, the way the missing- and draft-page states above do, with a notice
+ * standing in for the content so the front page doesn't look broken to the
+ * person who can change it.
+ */
+if ( post_password_required( $target_page ) ) {
+	if ( ! $can_edit ) {
+		return;
+	}
+
+	$edit_url = get_edit_post_link( $target_page->ID );
+
+	printf(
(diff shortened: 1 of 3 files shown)


mcliwanow (approved):
Code looks good and it tests as described. Thanks!

PR #1980, opened 2026-08-31
Blocks: honour post visibility in the Sessions, Speakers, Sponsors and Organizers blocks

KokkieH:
The Sessions, Speakers, Sponsors and Organizers blocks render their items through `get_all_the_content()` and `get_trimmed_content()` in `mu-plugins/blocks/includes/content.php`. Both read the stored post content directly rather than going through `get_the_content()` / `get_the_excerpt()`, which is where core handles the editor's "Password protected" visibility setting. The blocks therefore treated every published post alike, whatever its visibility. `theme-templates/bootstrap.php` passes a page into the same helper.

This adds the handling, following core's own behaviour rather than inventing something for these blocks:

| Content setting | What the item shows |
| --- | --- |
| `full` | The password form, as `get_the_content()` returns |
| `excerpt` | "There is no excerpt because this is a protected post.", as `get_the_excerpt()` returns |

The post stays in the listing either way, and its title carries core's "Protected:" prefix — `get_the_title()` already applies that, so the views needed no change. A visitor who has entered the password sees the content as before. Submitting the form takes you to the single post, which is core's behaviour: `get_the_password_form()` embeds a hidden `redirect_to` set to the permalink, and `wp-login.php?action=postpass` prefers it over the referer.

The block queries are deliberately unchanged. Password protection guards the body, not the post's existence — an organizer who wants a post out of the listing entirely has "Private" for that — so excluding protected posts from the queries would conflate the two settings and break what people expect these controls to do.

### The offline page is the one exception

`get_offline_page()` now skips a password-protected page, so `get_offline_content()` falls back to its default message. That page is not a listing: the service worker pre-caches it and shows it with no connection, where a password form could not be submitted. Serving a dead form there would be worse than the default message, so this one resolves at the query instead. The helper covers it regardless; the query argument just picks the better of the two outcomes.

### Also here

`organizers/view.php` carried a comment saying organizer posts are not `public` and have no permalinks. That was true when it was written in 968094e7 (2019); the post type became `public` in 3cafe4a9 (2020) and the comment was never updated. It misled a reviewer of this PR, so it is corrected here.

### How to test the changes in this Pull Request

No build is needed — the blocks render from PHP in `source/`, and `build/` holds only compiled JS and CSS.

1. Run the tests:

   ```bash
   ./public_html/wp-content/mu-plugins/vendor/bin/phpunit -c phpunit.xml.dist --group blocks
   ```

2. On a local camp site, open a session at `/wp-admin/edit.php?post_type=wcb_session`, give it some body text you will recognise, and set Status & visibility → Visibility to "Password protected" with a password.
3. Add a page holding the Sessions block, set to list all sessions with the content setting on "Full":

   ```
   <!-- wp:wordcamp/sessions {"mode":"all","content":"full"} /-->
   ```

4. Load that page in a logged-out browser or private window. The session should still be listed, its title reading "Protected: <title>", with the password form where its body would be. The body text itself should not appear anywhere on the page.
Password-test___WordCamp_Seattle_2023

5. Enter the password. You land on the session's own page with its content shown. Go back to the listing — the body now renders inline there too.
6. Clear the `wp-postpass_*` cookie, switch the block's content setting to "Excerpt", and reload logged out. The item should now read "There is no excerpt because this is a protected post." in place of the body. Confirm this with a session that has no manual excerpt, and again with one that has a manual excerpt filled in.
Password-test___WordCamp_Seattle_2023

7. Repeat steps 3–6 for the Speakers, Sponsors and Organizers blocks, e.g. `https://seattle.wordcamp.test/2023/speakers/`.
8. Set the post's visibility back to Public and confirm the listing renders exactly as it did before this branch, in both content settings.
9. Optionally, set a password on the site's Offline page (the one with the `wc_page_offline` meta) and confirm the offline template falls back to its default message rather than showing a form.

### DIFF
--- a/public_html/wp-content/mu-plugins/blocks/includes/content.php
+++ b/public_html/wp-content/mu-plugins/blocks/includes/content.php
@@ -24,8 +24,13 @@ function get_all_the_content( $post ) {
 	/*
 	 * `get_the_content()` holds the password check, and the `the_content`
 	 * filter runs after it, so reading the stored content directly has to
-	 * repeat the check. Return what `get_the_content()` returns here, so a
-	 * listing behaves the way one built on core's loop would.
+	 * repeat the check.
+	 *
+	 * This returns the form ahead of the filters below, where a core loop
+	 * would run them over it. That is deliberate: `wc-post-types` appends
+	 * speaker, slide, video and category markup on `the_content`, none of
+	 * which belongs on a password form. The cost is the paragraph wrapping
+	 * `wpautop()` would have added.
 	 */
 	if ( post_password_required( $post ) ) {
 		return get_the_password_form( $post );
@@ -55,10 +60,14 @@ function get_trimmed_content( $post, $excerpt_length = 55 ) {
 	 * The fallback below reads `post_content` directly, so this needs the
 	 * same check as `get_all_the_content()`. `get_the_excerpt()` applies it
 	 * to both the manual excerpt and the fallback, and stands this sentence
-	 * in their place; the wording matches core's so the two read alike.
+	 * in their place.
+	 *
+	 * Take it from core's catalogue rather than restating it under this
+	 * plugin's domain, so that a listing and the post's own page read alike
+	 * in every locale instead of only in English.
 	 */
 	if ( post_password_required( $post ) ) {
-		return esc_html__( 'There is no excerpt because this is a protected post.', 'wordcamporg' );
+		return esc_html__( 'There is no excerpt because this is a protected post.', 'default' );
 	}
 
 	$post_excerpt = $post->post_excerpt;
--- a/public_html/wp-content/mu-plugins/blocks/includes/content.php
+++ b/public_html/wp-content/mu-plugins/blocks/includes/content.php
@@ -21,6 +21,16 @@ defined( 'WPINC' ) || die();
 function get_all_the_content( $post ) {
 	$post = get_post( $post );
 
+	/*
+	 * `get_the_content()` holds the password check, and the `the_content`
+	 * filter runs after it, so reading the stored content directly has to
+	 * repeat the check. Return what `get_the_content()` returns here, so a
+	 * listing behaves the way one built on core's loop would.
+	 */
+	if ( post_password_required( $post ) ) {
+		return get_the_password_form( $post );
+	}
+
 	$content = wp_kses_post( $post->post_content );
 
 	/** This filter is documented in wp-includes/post-template.php */
@@ -40,6 +50,16 @@ function get_all_the_content( $post ) {
  */
 function get_trimmed_content( $post, $excerpt_length = 55 ) {
 	$post = get_post( $post );
+
+	/*
+	 * The fallback below reads `post_content` directly, so this needs the
+	 * same check as `get_all_the_content()`. `get_the_excerpt()` applies it
+	 * to both the manual excerpt and the fallback, and stands this sentence
+	 * in their place; the wording matches core's so the two read alike.
+	 */
+	if ( post_password_required( $post ) ) {
+		return esc_html__( 'There is no excerpt because this is a protected post.', 'wordcamporg' );
+	}
 
 	$post_excerpt = $post->post_excerpt;
 	if ( ! ( $post_excerpt ) ) {
--- a/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/view.php
+++ b/public_html/wp-content/mu-plugins/blocks/source/blocks/organizers/view.php
@@ -11,7 +11,9 @@ defined( 'WPINC' ) || die();
 /** @var array   $attributes */
 /** @var WP_Post $organizer */
 
-// Note that organizer posts are not 'public', so there are no permalinks.
+// This block doesn't link organizer titles, unlike the sibling blocks.
+// The post type gained permalinks when it became 'public' in 2020; the
+// block was never changed to use them.
 
 setup_postdata( $organizer ); // This is necessary for generating an excerpt from content if the excerpt field is empty.
 ?>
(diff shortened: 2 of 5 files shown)


mcliwanow (approved):
Tests as described and code makes sense too. Thanks for the changes!

public_html/wp-content/mu-plugins/blocks/includes/content.php
```diff
@@ -41,6 +51,16 @@ function get_all_the_content( $post ) {
...
 	$post = get_post( $post );
 
+	/*
+	 * The fallback below reads `post_content` directly, so this needs the
+	 * same check as `get_all_the_content()`. `get_the_excerpt()` applies it
+	 * to both the manual excerpt and the fallback, and stands this sentence
+	 * in their place; the wording matches core's so the two read alike.
+	 */
+	if ( post_password_required( $post ) ) {
+		return esc_html__( 'There is no excerpt because this is a protected post.', 'wordcamporg' );
```
bor0: Small one, and it cuts at the sentence the comment above is about: "the wording matches core's so the two read alike."

It matches in English. In any other locale the two diverge — `get_the_excerpt()` returns this string from core's own catalogue, translated everywhere core is, while the same sentence under `wordcamporg` is a separate string that starts untranslated and stays that way until somebody gets to it on translate.wordpress.org. So a French visitor would read the translated sentence on the session's own page and the English one in the listing, which is the opposite of the intent.

Core's domain gets it for free, and there is precedent for that here — `wporg-groups-frontend/src/blocks/group-settings/render.php:31` does it with "Just another WordPress site":

```php
return esc_html__( 'There is no excerpt because this is a protected post.', 'default' );
```

The trade-off is depending on core keeping the string, which it has since 2.6 and which this PR is already depending on in spirit by copying the wording. `phpcs.xml.dist` sets no `text_domain` property, so the sniff has no opinion either way.

public_html/wp-content/mu-plugins/blocks/includes/content.php
```diff
@@ -21,6 +21,16 @@
...
 	$post = get_post( $post );
 
+	/*
+	 * `get_the_content()` holds the password check, and the `the_content`
+	 * filter runs after it, so reading the stored content directly has to
+	 * repeat the check. Return what `get_the_content()` returns here, so a
+	 * listing behaves the way one built on core's loop would.
+	 */
+	if ( post_password_required( $post ) ) {
+		return get_the_password_form( $post );
```
bor0, line 36: Worth half a sentence in the comment, because the parity here is not quite exact.

Core runs the form through the content filters: `the_content()` calls `get_the_content()` — which returns the form — and then applies `the_content` to the result. This returns the form before the `apply_filters()` below it, so the listing's form is raw where the single post's form has been through `wpautop` and everything else on that hook.

The visible difference is only paragraph wrapping, and skipping the filters is arguably the better default: the plugin hangs four appenders off `the_content` (`wc-post-types.php:73-78`), and a filter appending post-derived material to a password form is a worse failure than a missing `<p>`. So I would keep it as it is — just not describe it as what a core loop would do, since a core loop would filter it.

bor0 (approved):
Ran this locally and checked the behaviour claims against the WordPress in the test environment rather than from memory. It reads right to me: the fix is in the one place both helpers converge, it follows core rather than inventing a third behaviour for these blocks, and leaving the queries alone is the correct call — "password protected" and "private" are two settings and conflating them would take a choice away from organizers.

**Verified.**

- `--group blocks` is 50 tests / 117 assertions green. Reverting just `content.php` and `bootstrap.php` to `production` and rerunning turns 13 of them red, so the new tests are load-bearing rather than decorative — including the four rendered-block cases in both content modes.
- `BASE_REF=origin/production php .github/bin/phpcs-branch.php` clean. The unrelated `Test_QuickBooks_OAuth_State` failure in a full run is pre-existing on `production`; this branch does not touch it.
- Core 7.1 in the container backs the description: `get_the_password_form()` really does embed `redirect_to` as the permalink (`post-template.php:1807-1809`), `get_the_excerpt()` returns that exact sentence (`:424`), and the field id is `pwbox-{ID}`, so several protected items in one listing each get their own form rather than colliding.
- The organizers comment correction is right. `wcb_organizer` registers `'public' => true` (`wc-post-types.php:1663`); the post type registered `'public' => false` a few lines down is `wcb_volunteer` (`:1702`), which is probably where the original confusion came from.
- Nothing else in these blocks reads the body. The Schedule block's internal REST pull asks for `id`, `link`, `slug`, `title`, the session meta and the terms — no content or excerpt — so a protected session reaches the schedule JS as a title only, already carrying core's prefix.
- No block-level caching in any of the four controllers, so the output of these helpers — which now varies with the `wp-postpass_` cookie — isn't stored and replayed to a different visitor from inside the app.
- The offline-page change has exactly one consumer path (`get_offline_page()` → `get_offline_content()` → the template and `add_offline_template_cachebuster()`), so `has_password => false` cannot surprise anything else, and the cachebuster keeps working because it hashes whatever `get_offline_content()` returns. Reasoning about it at the query rather than the helper is right for the reason you give: a form the visitor cannot submit is worse than the default message.

Two notes inline, neither blocking.

**One thing for after this, not for this PR.** The single-session view has the other half of the same shape. `wc-post-types` hooks `the_content` to append the speaker list, slides link, video link and session categories (`wc-post-types.php:73-78`), none of those callbacks consults `post_password_required()`, and core's `the_content()` applies the filters to the password form it just built. So on a site above the id gates that isn't using block templates, an organizer who password-protects a session gets the body withheld and then "View Session Video" rendered underneath the form. Unchanged by this branch either way, and the body is still safe — but it is the same question one level down, and an organizer who has just protected a session would reasonably expect the link to go with it.

PR #1981, opened 2026-09-01
Budgets: ask the same question of the file as of the request it joins

bor0:
## Summary

Follow-on to #1975, which brought the Files metabox's field into line with the rule the metabox itself applies -- a file already on another post keeps the post it has. That covered the file's parent but not the file, so an unattached upload could still be moved onto a request by someone with no claim on it.

- `is_file_attachable()` now takes `edit_post` on the file as well. `post_parent` is the field `privacy.php` reads to decide who sees a payment file, so writing it takes the capability any other route to that field takes -- which runs #1975's `map_meta_cap` guard for payment files and ordinary post capabilities for everything else. Same answer the Media Library would give the same person, rather than a second rule.
- `attach_existing_files()` asks `edit_post` about the request it's writing to. Both callers already sit behind `post_edit_is_actionable()`, which asks the same thing, so nothing changes today -- it's a public helper that writes to whatever request it's handed, and shouldn't depend on a caller having asked.
- `save_request()` no longer builds a stub post out of `$_POST['original_post_status']` to decide whether the request is still editable. See below.

An organizer attaching a co-organizer's unattached receipt still works -- that's a supported way to file a request, and `privacy.php` shows the file to both of them. Covered by a test.

## How to review this PR

`includes/wordcamp-budgets.php` first, then `includes/reimbursement-request.php`.

- **Why `edit_post` and not "own uploads only".** The stricter rule reads well but breaks the shared-upload workflow: one organizer uploads the receipts, another files the request. `privacy.php` explicitly supports that (the requester sees files on their own request), so the picker would keep offering files the server then silently dropped. `edit_post` is the rule that already governs `post_parent` everywhere else.
- **The `original_post_status` stub is safe to drop, and this was measured, not assumed.** The comment on it said it existed because the stored status is already the new one by the time `save_request()` runs, so `draft` → submitted would be judged on the status it moved to and the fields entered with it dropped. That is no longer what decides it: `post_edit_is_actionable()` asks `edit_post` first, and this post type's `map_meta_cap` callback adds `manage_network` to that for a request past `draft`/`auto-draft`/`wcb-incomplete` when `action` isn't `edit`. So a save that would have needed the stub returns before reaching it. Confirmed on the test stack: at `save_post` after a `draft` → `wcb-pending-approval` transition, `post_edit_is_actionable()` is already `false`. What reaches the check is a request still in the editable statuses, which reads the same either way.
- **Unrelated observation, not addressed here.** The above also means editing fields *and* submitting in one click drops those edits, since the handler bails before writing them. That predates this branch and isn't touched by it — happy to look at it separately if it's worth a ticket.
- `tests/test-reimbursement-request.php` is new: `save_request()` had no coverage, and the status boundary is what these three tests pin.

## Test plan

- [x] phpunit — `WordCamp Payments` 38/38, `WordCamp Budgets Dashboard` 54/54, `WordCamp Post Types` 52/52
- [x] Each new guard verified to fail without its fix (reverted the check, watched the test go red, restored)
- [x] phpcs — no new violations on any changed file (error/warning counts identical to the pre-change baseline on the two `includes/` files; both test files clean)
- [ ] npm run build — n/a, no JS changed
- [x] Manual browser pass on the local stack (below)

### Manual test steps

On `seattle.wordcamp.test`, with an unattached PDF owned by an Editor standing in for a co-organizer's upload.

1. **Author-role volunteer, low-privilege actor.** Logged in as an Author and opened `post-new.php?post_type=wcb_reimbursement`. Reachable, as expected — the post types gate creation on `publish_posts`.
2. **The attack shape.** Set `wcb_existing_files_to_attach` to the Editor's file id on their own draft and submitted the form directly (the hidden input is populated in the browser, so this is the realistic shape). The draft saved — status went `auto-draft` → `Draft` — and the file was refused: `post_parent` still `0` in the DB, "You haven't uploaded any files yet" on the screen.
3. **Negative control.** Removed the new check, repeated step 2 unchanged: the file was taken (`post_parent` → the volunteer's draft) and `victim-auth.pdf` was listed on their own request, directly under the "Files uploaded by other users are hidden to protect privacy" note. Restored the check, reset `post_parent`, repeated again — refused.
4. **Editor, no regression.** Logged in as an Editor and attached both their own unattached file and a second Editor's unattached file to a new request. Both attached, both listed in the metabox.
5. **The `original_post_status` change.** Same save also set Name of Payer; confirmed `_wcbrr_name_of_payer` persisted, so dropping the stub doesn't cost the legitimate draft save. Then moved the request to `wcb-approved` and resubmitted with `original_post_status` forced back to `draft` and a rewritten payer name: nothing was written, meta unchanged. Verified the same POST behaves identically on an unmodified checkout, so the refusal is pre-existing and not introduced here.
6. Fixtures (users, attachments, requests) deleted afterwards.

https://claude.ai/code/session_01HGUDuVhzdUeLnMNxTW7z6T

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
+++ b/public_html/wp-content/plugins/wordcamp-payments/includes/reimbursement-request.php
@@ -549,19 +549,17 @@ function save_request( $post_id, $post ) {
 	validate_and_save_notes( $post, $_POST['wcbrr_new_note'] );
 
 	/*
-	 * We need to determine if the user is allowed to modify the request -- in terms of this plugin's post_status
-	 * restrictions, not in terms of current_user_can( 'edit_post', N ) -- but at this point in the execution
-	 * the status has already changed from the original one to the new one, so user_can_edit_request() would often
-	 * return an incorrect result, because it would be evaluating the new status, when it should use the old one.
-	 * That would result in all the meta fields the user entered being ignored when going from `draft` to
-	 * `submitted`, `info_requested` to `submitted`, etc.
+	 * The status this reads is the stored one, which by now is the status the save just wrote. That used to be
+	 * worked around with a stub post built from `$_POST['original_post_status']`, so the check saw the status the
+	 * request had before the save -- otherwise a `draft` being submitted would be judged as `wcb-pending-approval`
+	 * and the fields entered alongside it dropped.
 	 *
-	 * To avoid that, we create a stub WP_Post with the original post status, and give that to
-	 * user_can_edit_request() instead.
+	 * That's no longer what decides it. `post_edit_is_actionable()` above asks `edit_post`, which this post type's
+	 * `map_meta_cap` callback refuses for a request past the editable statuses, so a save that would have needed
+	 * the stub has already returned. What's left here is a request still in those statuses, which reads the same
+	 * either way -- and reading the real post keeps the decision out of the request body.
 	 */
-	$original_post = new WP_Post( (object) array( 'post_status' => $_POST['original_post_status'] ) );
-
-	if ( user_can_edit_request( $original_post ) ) {
+	if ( user_can_edit_request( $post ) ) {
 		$text_fields = array( 'name_of_payer', 'currency', 'reason', 'reason_other' );
 		validate_and_save_text_fields( $post_id, $text_fields, $_POST );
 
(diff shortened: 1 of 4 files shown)


mcliwanow (approved):
LGTM. Ran it locally.

- Suites match: Payments 38/38, Budgets Dashboard 54/54, Post Types 52/52.
- Reverted each guard on its own and watched its test go red: the file edit_post check, the request edit_post check, the original_post_status stub.
- Seattle: author attaching an editor's unattached file refused (post_parent stays 0), editor attaching own + a co-organizer's file both go through.

Stub removal holds up. post_edit_is_actionable() bails before it on anything past the editable statuses. 🙂

PR #1983, opened 2026-09-01
Event List: encode event values before building the block's markup

bor0:
## Summary

- `wporg-events-2023`'s event list block builds markup as a string and assigns
  it to `innerHTML`. The `escapeHtml()` helper it routed values through
  returned its input unchanged, so titles and locations reached the DOM as
  written. Values now go through `@wordpress/escape-html`, which skips
  already-encoded entities — `wporg_events` rows carry plenty of them — and
  distinguishes element-text from attribute context.
- Adds `escapeUrl()` for the `href`. Encoding says nothing about a URL's
  scheme, so it's the wrong tool there. No base is passed to `URL`, so empty,
  malformed and protocol-relative values fail the parse instead of resolving
  against the current page.
- `esc_attr()` -> `esc_url()` on two links fed by organiser-supplied meta: the
  WordCamp Counts report and the meetup admin's clickable link. `esc_attr()`
  quotes without constraining the scheme.

`block.json` is deliberately untouched — its `viewScript` handles are
positional and `index.php` hardcodes the index. The new dependency is
externalised by wp-scripts into `view.asset.php` instead.

## Test plan

- [x] phpunit — `WordCamp Reports` 18/18, `WordCamp Post Types` 52/52
- [x] phpcs — no new errors in either PHP file (86 and 7 pre-existing, unchanged)
- [x] `npm run build` / `npm run lint:js` — clean
- [x] DOM pass on the built bundle, with the pre-change bundle as a control

### Manual test steps

1. Build the theme and load a page carrying the event list block.
2. Give an event row a title containing markup, a title containing a quote,
   and a `javascript:` URL. All three render as text / an empty `href`, and
   every row still renders.
3. Give a row an already-encoded title and location (`&#8211;`, `&amp;` — the
   dev fixture has thousands). They render as the characters they stand for,
   not as literal entities.
4. Give a row an empty, malformed, and protocol-relative URL. Each yields an
   empty `href` rather than a link.

Verified in a real browser and in a DOM harness. The harness was run against
the pre-change bundle first as a control, to confirm the assertions actually
discriminate — two did not (an image is never fetched and inline handlers
don't run there), so only the structural assertions are relied on.

### DIFF
--- a/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-admin.php
+++ b/public_html/wp-content/plugins/wcpt/wcpt-meetup/class-meetup-admin.php
@@ -186,7 +186,7 @@ if ( ! class_exists( 'Meetup_Admin' ) ) :
 		 */
 		protected function print_clickable_link( $link ) {
 			?>
-		<a href="<?php echo esc_attr( $link ); ?>" target="_blank">
+		<a href="<?php echo esc_url( $link ); ?>" target="_blank">
 			<?php echo esc_html( $link ); ?>
 		</a>
 			<?php
--- a/public_html/wp-content/plugins/wordcamp-reports/views/html/wordcamp-counts.php
+++ b/public_html/wp-content/plugins/wordcamp-reports/views/html/wordcamp-counts.php
@@ -201,7 +201,7 @@ $first_time_legend = '<span class="description small">Y / N / ?</span>';
 
 		<?php foreach ( $data['wordcamps'] as $event ) : ?>
 			<tr>
-				<td><a href="<?php echo esc_attr( $event['info']['URL'] ); ?>"><?php echo esc_html( $event['info']['Name'] ); ?></a></td>
+				<td><a href="<?php echo esc_url( $event['info']['URL'] ); ?>"><?php echo esc_html( $event['info']['Name'] ); ?></a></td>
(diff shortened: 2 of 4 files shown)


mcliwanow (approved):
LGTM. Checked both behavior changes against the theme's dev fixture (38,716 rows) rather than by eye.

Entities land correctly now, `&#8211;` and `&amp;` render as `–` and `&`. Ten rows change their visible text and all ten are improvements: titles wrapped in angle brackets used to get eaten by the parser.

Dropping the base costs nothing, all 38,716 URLs are absolute so none get rejected. 259 come back normalized, 258 of those a trailing slash on a bare origin plus one lowercased hostname, so no destination moves.

Wiring is right too: block.json untouched so the handle stays `-2`, `wp-escape-html` lands in `view.asset.php`, bundle references `wp.escapeHtml` rather than inlining it. Build and `lint:js` clean, didn't re-run phpcs.

Follow-ups, none blocking:

- `event-rsvp/view.js:475,478,501` still builds `href`/`src` with `escAttr()` while its own `render.php:247` uses `esc_url()`. Happy to take it.
- `getEventDateTime` calls `toISOString()` on a date that can be invalid, one bad timestamp and the block sits on "Loading events..." forever. Pre-existing.
- `google-map/src/components/list.js:36` renders the same `url` with no scheme check, needs an upstream issue.

PR #1986, opened 2026-09-01
Budgets: judge a request's editability on the status it had before the save

bor0:
## Summary

Budget requests submitted straight from the editor arrive at Central empty — no sponsor, no currency, no amount, no description. Reported for sponsor invoices in #team-orion; vendor payments and reimbursements have the same defect.

- Submitting moves a request out of the statuses its requester may edit, and the row already carries the new status by the time `save_post` runs. The handlers that write the entered values start with `post_edit_is_actionable()` → `edit_post`, so on the very save that submits, the requester is refused and nothing is stored.
- Reachable since #1966: `get_map_meta_cap_post()` now prefers `$args[0]` over the global `$post`. On an edit-form save that global is the object `post.php` read *before* `edit_post()` updated the row, so it still said `draft` and the check passed by accident.
- Fix: record the status each budget request held before the save that's running, and have the three CPTs' `map_meta_cap` callbacks and `user_can_edit_request()` read that instead of the stored status.
- Nothing is loosened: a request already past the editable statuses before the save reads the same either way. Only the transition itself changes.

Nothing was deleted — the values were never written — so invoices already submitted blank have to be redrafted.

Paired with @mcliwanow on this.

## How to review this PR

Read `wordcamp-budgets.php` first — `remember_status_before_save()`, `forget_status_before_save()`, `get_status_for_edit_check()` — then the four call sites.

**Where the reading is taken from.** `post_updated` would hand over the old post directly, but it and `wp_after_insert_post` fire for every post on every site in the network, and only these three post types have any use for them. The reading comes instead from each CPT's own `wp_insert_post_data` callback, past the post-type guard those already carry: that hook runs before the row is written, so the stored status is still the one the request had when the save began. No new always-on hook. The cleanup is attached from there too and detaches itself when the last entry goes, so a request that saves anything else never carries it either.

**How narrow the window is.** The record is honoured only while `doing_action( 'save_post' )`, which is where all four consumers sit. `wp_insert_post()` skips `wp_after_insert_post` when a caller passes `$fire_after_hooks = false`, and an entry left behind by one of those must not be able to widen anything — so the window does not depend on the cleanup having run.

**Why it can't be pushed around.** The value is read off the row with `get_post_status()`, never taken from the caller, so it is always a status the request genuinely held moments earlier. There is nothing an attacker can put here that the request did not already have. It is keyed by post ID, so it says nothing about any other record.

**Other notes.**
- `user_can_edit_request()` needed the same treatment: `save_request()` calls it after `post_edit_is_actionable()` with the post whose status is already the new one, so fixing only the capability layer would still have dropped the reimbursement text fields.
- `daff0185` documented the current behaviour as intentional ("a save that would have needed the stub has already returned"). That comment describes the bug; this replaces the stub's job with something that doesn't read the request body.

## Test plan

- [x] phpunit — 75/75 `WordCamp Budgets Dashboard`, 40/40 `WordCamp Payments`; full suite 1251 tests, only 3 pre-existing `Test_Groups_Blocks` failures (confirmed present on `production` too)
- [x] New tests fail against `origin/production` (7 failures + 2 errors) and pass here
- [x] The access-control assertions below hold on `production` *and* here, and fail against the pre-#1966 helper
- [x] phpcs clean on the new test file; no new findings on any changed line
- [x] Manual browser pass, both roles, before and after — re-run against the final wiring

### Automated coverage

`wordcamp-payments-network/tests/test-submit-transition.php`, 21 cases.

```bash
docker exec wordcamporg-phpunit_wp-1 bash -lc \
  'cd /app && ./public_html/wp-content/mu-plugins/vendor/bin/phpunit \
     -c phpunit.xml.dist --filter Test_Submit_Transition'
```

Each case drives the CPT's real submit path — the request body first, then the same array handed to `wp_update_post()`, the way `edit_post()` does — with the CPT's own status filter attached, so the recording under test is the one that runs in production.

**What has to start working**

| Test | What it pins down |
| --- | --- |
| `test_submitting_a_sponsor_invoice_stores_its_fields` | **The end-to-end one.** Real `save_invoice()` handler, real nonces, `send-invoice` set, `$GLOBALS['post']` staged the way `post.php` leaves it. Asserts the invoice lands at `wcbsi_submitted` **and** that all five `_wcbsi_*` values are stored. The reported bug, in one test. |
| `test_save_handler_may_act_while_submitting` (×3 CPTs) | The gate a `save_post` handler actually sees, asserted from inside a `save_post` callback — not after `wp_update_post()` returns, where the record has already been dropped. |

**What must not start working** — these pass on `production` too, so they are guarding, not demonstrating:

| Test | What it pins down |
| --- | --- |
| `test_editor_is_refused_a_non_draft_request_in_every_argument_form` (×3) | The access-control property #1966 restored: an Editor holding `edit_others_posts` and no network capability is refused `edit_post` **and** `delete_post` on a non-draft request — as an int, a numeric string and a `WP_Post`, with a foreign global `$post` set, in the `wp_ajax_upload_attachment()` request context where that distinction is actually read. Fails on all three post types against the pre-#1966 helper. |
| `test_submitting_one_request_does_not_open_another` (×3) | The record is keyed by post: submitting your own draft leaves the verdict on somebody else's approved request unchanged. |
| `test_save_handler_may_not_act_on_an_already_submitted_request` (×3) | A requester may submit, but may not then edit what they submitted. |
| `test_save_handler_may_not_act_on_an_approved_request` (×3) | Nor touch one approved for payout. |
| `test_recorded_status_is_only_honoured_inside_a_save` | Suppresses the cleanup the way a `$fire_after_hooks = false` caller does, then asserts the request still reads as submitted and `edit_post` is still refused. |
| `test_network_admin_may_act_on_a_submitted_request` (×3) | Central can still correct a submitted request. |
| `test_status_for_edit_check_falls_back_to_the_stored_status` | Outside a save it's just the stored status. |

To watch the first group fail, run the suite with `includes/` checked out from `production`:

```bash
git checkout origin/production -- public_html/wp-content/plugins/wordcamp-payments/includes/
# re-run the phpunit command above  ->  7 failures, 2 errors
git checkout HEAD -- public_html/wp-content/plugins/wordcamp-payments/includes/
```

### Manual test steps

Local docker stack. Two actors, exercised separately.

**Setup** — the Community dropdown reads a network option that is empty on a fresh local install, so the form can't be completed without seeding it:

```bash
docker exec wordcamporg-wordcamp.test-1 bash -lc \
  'wp --path=/usr/src/public_html/mu site option update wordcamp_qbo_classes \
     --format=json --allow-root <<< "{\"42\":\"Seattle, WA\"}"'
```

`seattle.wordcamp.test/2023` already ships a complete sponsor (`ACME Corp`, post 109) and `editorrole` — a subsite administrator who is **not** a network admin, which is the role that reproduces this. `admin` is the super admin.

1. **As `editorrole`, before the fix** (`git checkout origin/production -- .../includes/`) — Budget → Sponsor Invoices → Add New. Set sponsor `ACME Corp`, community `Seattle, WA`, description, currency `USD`, amount. Click *Send Invoice*. → Invoice stored as `wcbsi_submitted` with **zero** `_wcbsi_*` meta. Reopening it shows every field blank under "Status: Submitted".
2. **As `admin`**, `wordcamp.test/wp-admin/network/admin.php?page=sponsor-invoices-dashboard` → the invoice reads "Auto Draft / Error: The item does not have a currency", which is the reported symptom.
3. **As `editorrole`, after the fix** — identical flow, new invoice. → `wcbsi_submitted` with sponsor, class, currency, description and amount all stored. Central shows it complete alongside the broken one from step 1.
4. **Restriction still holds** — still as `editorrole`, on the invoice submitted in step 3, changed the amount and description in the DOM and posted the edit form directly (the Update button is correctly absent). → Refused; meta unchanged.
5. **Normal editing still works** — as `editorrole`, edited an existing *draft* invoice's amount and saved. → "Post updated", new value stored.

Steps 3–5 were re-run against the final wiring after the recording moved off `post_updated`, since that is a different code path from the one the first pass exercised.

Every end state was checked against the database, not just the UI:

```bash
docker exec wordcamporg-wordcamp.test-1 bash -lc \
  'wp --path=/usr/src/public_html/mu --url=https://seattle.wordcamp.test/2023/ \
     post meta list <ID> --keys=_wcbsi_sponsor_id,_wcbsi_qbo_class_id,_wcbsi_description,_wcbsi_currency,_wcbsi_amount --allow-root'
```

Fixtures deleted afterwards (invoices, the `wordcamp_qbo_classes` option, and the `wc_wordcamp_payments_index` rows, which the post deletion already cleaned up).

Vendor payments and reimbursements were **not** exercised in the browser — they're covered by the automated tests only.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-payments-network/tests/test-submit-transition.php
+++ b/public_html/wp-content/plugins/wordcamp-payments-network/tests/test-submit-transition.php
@@ -1,0 +1,356 @@
+<?php
+
+namespace WordCamp\Budgets_Dashboard\Tests;
+
+use WP_UnitTestCase, WordCamp_Budgets;
+
+defined( 'WPINC' ) || die();
+
+/**
+ * What a budget request's `save_post` handler is allowed to do as the request is submitted.
+ *
+ * Submitting moves the request out of the statuses its requester may edit, and the row carries the new status
+ * by the time `save_post` runs. The editability checks have to keep reading the status the request held before
+ * that save, or the handler that writes the fields entered alongside the submission is refused and the request
+ * arrives empty.
+ *
+ * @group budgets-dashboard
+ */
+class Test_Submit_Transition extends WP_UnitTestCase {
+	/**
+	 * A subsite administrator: authors requests, but is not a network admin.
+	 *
+	 * @var int
+	 */
+	protected static $requester_id;
+
+	/**
+	 * The status each budget CPT moves to when its requester submits it.
+	 *
+	 * @var array<string, string>
+	 */
+	const SUBMITTED_STATUSES = array(
+		'wcb_sponsor_invoice' => 'wcbsi_submitted',
+		'wcp_payment_request' => 'wcb-pending-approval',
+		'wcb_reimbursement'   => 'wcb-pending-approval',
+	);
+
+	/**
+	 * Create the shared user.
+	 *
+	 * @param \WP_UnitTest_Factory $factory
+	 */
+	public static function wpSetUpBeforeClass( $factory ) {
+		self::$requester_id = $factory->user->create( array( 'role' => 'administrator' ) );
+	}
+
+	/**
+	 * The `save_post` handlers run nonce checks that `wp_die()` outside a real request, so keep them off while
+	 * the fixtures are built. Tests that want one attach it themselves.
+	 */
+	public function set_up() {
+		parent::set_up();
+
+		$_POST    = array();
+		$_REQUEST = array();
+
+		wp_set_current_user( self::$requester_id );
+
+		$this->detach_save_handlers();
+		$this->detach_status_filters();
+	}
+
+	/**
+	 * Re-attach the `save_post` handlers detached in `set_up()`.
+	 */
+	public function tear_down() {
+		$this->attach_save_handlers();
+		$this->attach_status_filters();
+
+		unset( $GLOBALS['post'] );
+
+		parent::tear_down();
+	}
+
+	/**
+	 * Detach every budget CPT's `save_post` handler.
+	 */
+	protected function detach_save_handlers() {
+		remove_action( 'save_post', 'WordCamp\Budgets\Sponsor_Invoices\save_invoice', 10 );
+		remove_action( 'save_post', 'WordCamp\Budgets\Reimbursement_Requests\save_request', 10 );
+		remove_action( 'save_post', array( $GLOBALS['wcp_payment_request'], 'save_payment' ), 10 );
+	}
+
+	/**
+	 * Re-attach every budget CPT's `save_post` handler.
+	 */
+	protected function attach_save_handlers() {
+		add_action( 'save_post', 'WordCamp\Budgets\Sponsor_Invoices\save_invoice', 10, 2 );
+		add_action( 'save_post', 'WordCamp\Budgets\Reimbursement_Requests\save_request', 10, 2 );
+		add_action( 'save_post', array( $GLOBALS['wcp_payment_request'], 'save_payment' ), 10, 2 );
+	}
+
+	/**
+	 * Detach the `wp_insert_post_data` filters that normalise each CPT's status.
+	 *
+	 * They read the request body to decide the status, so a test that stores a status directly has to set it
+	 * aside. The end-to-end test attaches them again.
+	 */
+	protected function detach_status_filters() {
+		remove_filter( 'wp_insert_post_data', 'WordCamp\Budgets\Sponsor_Invoices\set_invoice_status', 10 );
+		remove_filter( 'wp_insert_post_data', 'WordCamp\Budgets\Reimbursement_Requests\set_request_status', 10 );
+		remove_filter( 'wp_insert_post_data', array( $GLOBALS['wcp_payment_request'], 'wp_insert_post_data' ), 10 );
+	}
+
+	/**
+	 * Re-attach the `wp_insert_post_data` filters detached above.
+	 */
+	protected function attach_status_filters() {
+		add_filter( 'wp_insert_post_data', 'WordCamp\Budgets\Sponsor_Invoices\set_invoice_status', 10, 2 );
+		add_filter( 'wp_insert_post_data', 'WordCamp\Budgets\Reimbursement_Requests\set_request_status', 10, 2 );
+		add_filter( 'wp_insert_post_data', array( $GLOBALS['wcp_payment_request'], 'wp_insert_post_data' ), 10, 2 );
+	}
+
+	/**
+	 * Create a stored draft request of the given type, authored by the requester.
+	 *
+	 * @param string $post_type
+	 *
+	 * @return int
+	 */
+	protected function create_draft( $post_type ) {
+		return self::factory()->post->create( array(
+			'post_type'   => $post_type,
+			'post_status' => 'draft',
+			'post_author' => self::$requester_id,
+		) );
+	}
+
+	/**
+	 * Put the request into the state `wp-admin/post.php` leaves before `edit_post()` updates the row.
+	 *
+	 * @param int $post_id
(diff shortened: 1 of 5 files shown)


mcliwanow (approved):
Approving. I traced the same regression independently this morning and landed on the same chain, `user_can_edit_request()` second gate included. (I got as far as a `pre_post_update` variant of the recorder before this PR appeared. The shape here is tighter: no network-wide hooks, and a recorded status that only answers inside `save_post`.)

What I verified beyond reading the diff:

- Hook order against the WP copy in `.docker/test_suite`: `wp_insert_post_data` runs before the row write, `save_post` after it, `wp_after_insert_post` last. The recording point genuinely reads the old status, and every consumer sits inside the window `doing_action( 'save_post' )` bounds.
- The recorded value comes off the row of the post being saved, so nothing in the request body can plant a status the post didn't hold. It can widen the answer only across the transition this save performed, and only for this post.
- The negatives in `test-submit-transition.php` are real: submitted and approved requests stay refused, the argument-form control keeps the pre-#1966 helper from coming back, and the end-to-end invoice test asserts the exact symptom from the report.

Two notes, neither blocking:

1. `save_post_{$post_type}` fires before plain `save_post`, so a future handler on the typed hook would read the stored status and drop its fields again. Nothing hooks it today. Worth a sentence in the `get_status_for_edit_check()` docblock.
2. `WCP_Payment_Request::wp_insert_post_data()` records behind only its post-type check, while the other two record behind `post_edit_is_actionable()`, so payment requests also record on trash and bulk paths. The entries are inert there.

PHP test jobs were still pending when I checked; assuming green before merge.

bor0:
Thanks — both notes checked out against the code, and I've acted on both rather than documenting them. `0fe6b5e1`.

**1. The typed hook.** Confirmed: core fires `save_post_{$post_type}` at `post.php:5259` and plain `save_post` at `5270`, and `doing_action()` is just `doing_filter()`, which matches the exact hook name — so the window genuinely did not cover the typed form. I wrote the assertion first and it failed on all three post types, which made it worth closing rather than describing. `get_status_for_edit_check()` now goes through `is_inside_a_save()`, which answers for both forms, and `test_a_typed_save_post_handler_sees_the_same_answer` covers all three. It fails without the change.

**2. The asymmetric recording.** Also confirmed — `WCP_Payment_Request::wp_insert_post_data()` sat behind only the post-type check, so payment requests recorded on trash, bulk and autosave paths where sponsor invoices and reimbursements did not. Your read that the entries are inert there is right: the value is the status the row genuinely held, so on those paths it answers exactly as the stored status would. I've put it behind `post_edit_is_actionable()` anyway, so all three ask the same question and there's no path-dependent difference to reason about later.

On the pending jobs: green now — `PHP Tests (8.3)`, `(8.4)`, `(8.5)`, JavaScript and All. Locally, 78 in `WordCamp Budgets Dashboard` and 40 in `WordCamp Payments`; full suite is 1254 with the same three pre-existing `Test_Groups_Blocks` failures that are on `production`.

Two things worth recording since you traced this independently:

- I ran the reported request end to end against the branch on a local stack, with a control. Same request against the pre-#1966 helper reproduces it in full; against this branch it is refused at the capability check with nothing created. The submit window is a per-request static and does not survive into a second request — a request of the actor's own, asked immediately afterwards in a separate request, is refused once submitted, while their draft is still accepted.
- All three types are now exercised in a browser, not just sponsor invoices. Vendor payment and reimbursement submits both advance the status and keep the values entered alongside them; the reimbursement one goes through `validate_and_save_text_fields()` behind `user_can_edit_request()`, which is the second gate you mentioned.

One thing this deliberately does not fix: `required_fields_complete()` validates only the keys present in the submission, so a request body that omits them entirely still promotes a blank invoice. Pre-existing and independent of this chain — a separate PR if we want it.

PR #1991, opened 2026-09-02
Post Types: honour post visibility in the single session and speaker views

KokkieH:
Follow-up to #1980, from [this review comment](https://github.com/WordPress/wordcamp.org/pull/1980#pullrequestreview-5067383317).

#1980 handled the listing blocks. The single post view has the same shape one level down. `wc-post-types` hooks six callbacks onto `the_content` that add material to single session and speaker posts — the speaker list, slides link, video link and session categories on a session, and the avatar and session list on a speaker. Core applies the `the_content` filters to the password form it substitutes for a protected post's body, and none of the six consulted the post's visibility, so they added their markup around the form.

All six already shared the same guard:

```php
if ( ! $this->is_single_cpt_post( 'wcb_session' ) || site_supports_block_templates() ) {
	return $content;
}
```

This moves that into `should_add_to_content()` next to `is_single_cpt_post()`, with the visibility check alongside the two conditions it already had, and routes all six through it. Consolidating rather than repeating the check six times means the next callback added here inherits it instead of having to remember it.

Nothing changes for a post with no password: `post_password_required()` returns false immediately, and the four tests covering the ordinary case pass with and without this change.

`add_nofollow_to_sponsor_links` is left alone — it only rewrites links already present in the content it is handed, so it adds nothing to a password form.

### Also here

`mu-plugins/tests/test-blocks-content.php` from #1980 set the `wp-postpass_*` cookie and cleared it inline, so a test failing part way through left the cookie set for later tests. It is cleared in `tear_down()` now. This surfaced while writing the tests below: a leaked cookie made one of them pass for the wrong reason.

### How to test the changes in this Pull Request

1. Run the tests:

   ```bash
   ./public_html/wp-content/mu-plugins/vendor/bin/phpunit -c phpunit.xml.dist --testsuite "WordCamp Post Types"
   ```

2. These callbacks only run on sites using a classic theme, and only above the site-id gates each one checks, so a local camp site needs both. Activate a classic theme, then drop this in `public_html/wp-content/mu-plugins/` to open the gates:

   ```php
   <?php
   add_filter( 'wcpt_session_post_speaker_info_min_site_id', '__return_zero' );
   add_filter( 'wcpt_session_post_slides_info_min_site_id',  '__return_zero' );
   add_filter( 'wcpt_session_post_video_info_min_site_id',   '__return_zero' );
   add_filter( 'wcpt_speaker_post_avatar_min_site_id',       '__return_zero' );
   add_filter( 'wcpt_speaker_post_session_info_min_site_id', '__return_zero' );
   ```

3. Edit a session at `/wp-admin/edit.php?post_type=wcb_session`: give it body text, assign a speaker, a session category, a slides URL, and a video URL. The video field only accepts a `wordpress.tv` host — anything else is stored empty.
4. View the session's own page logged out. The body, the speaker list, the slides link, the video link and the categories should all be there. This is the baseline; confirm it before step 5, because the rest of the check is meaningless if these were never rendering.
Protected_session___WordCamp_Seattle_2023

5. Set Status & visibility → Visibility to "Password protected", give it a password, and reload logged out. The page should show the password form on its own — no speaker list, no slides link, no video link, no categories.
Protected_session___WordCamp_Seattle_2023

6. Enter the password. The body and all four additions should come back.
7. Repeat on a speaker post, which has its own two: the avatar above the content and the session list below it.

| Public | Protected |
| --- | --- |
| Awesome_Author___WordCamp_Seattle_2023 | Awesome_Author___WordCamp_Seattle_2023 |

8. Set visibility back to Public and confirm both pages render exactly as they did in step 4.
9. Remove the mu-plugin file from step 2.

### DIFF
--- a/public_html/wp-content/mu-plugins/tests/test-blocks-content.php
+++ b/public_html/wp-content/mu-plugins/tests/test-blocks-content.php
@@ -67,6 +67,11 @@ class Test_Blocks_Content extends WP_UnitTestCase {
 	public function tear_down() {
 		unset( $GLOBALS['post'] );
 
+		// Cleared here rather than inline, so a test that fails part way
+		// through cannot leave a later one thinking the visitor holds the
+		// password.
+		unset( $_COOKIE[ 'wp-postpass_' . COOKIEHASH ] );
+
 		set_current_screen( 'dashboard' );
 
 		parent::tear_down();
@@ -166,8 +171,6 @@ class Test_Blocks_Content extends WP_UnitTestCase {
 
 		$content = get_all_the_content( $post_id );
 		$excerpt = get_trimmed_content( $post_id );
-
-		unset( $_COOKIE[ 'wp-postpass_' . COOKIEHASH ] );
 
 		$this->assertStringContainsString( self::PROTECTED_BODY, $content );
 		$this->assertStringContainsString( self::PROTECTED_BODY, $excerpt );
--- a/public_html/wp-content/plugins/wc-post-types/tests/test-content-appenders.php
+++ b/public_html/wp-content/plugins/wc-post-types/tests/test-content-appenders.php
@@ -37,6 +37,13 @@ class Test_Content_Appenders extends WP_UnitTestCase {
 
 		$this->plugin = new WordCamp_Post_Types_Plugin();
 
+		// `is_single_cpt_post()` reads the post type's own query var, which a
+		// plain permalink carries directly. Set the structure here rather than
+		// inheriting whatever an earlier test left behind -- a pretty one needs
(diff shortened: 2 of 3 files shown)


mcliwanow (approved):
Ran it locally on seattle.wordcamp.test with Twenty Ten and the gates from the description. Public pages render everything, protected pages logged out render only the password form, and the postpass cookie brings it all back. Same fetch with `wc-post-types.php` from `production` puts the speaker list, both links and the categories under the form, so the bug is real and this closes it.

`should_add_to_content()` is the right place for it imo.

Approving 👍

PR #1992, opened 2026-09-02
Sponsors: store agreement attachments with the private status

KokkieH:
Makes sponsor agreement attachments private, rather than letting them inherit their sponsor's status, and gives them a random suffix in the name as `wordcamp-payments` does for its own files. A sponsor's other uploads — its logo, images in its content — are left alone.

An agreement uploaded from the Sponsor screen is named and made private in the request that creates it, so nothing has to be moved afterwards. One chosen from the existing Media Library is handled when the sponsor is saved, which does rename the file it already had.

Adds a one-time `wp wc-sponsor-agreements` migration for agreements attached before this, to be removed once it has been run. It sets the status and records that the file still needs renaming; renaming is being handled separately.

## Testing

On an event site, e.g. `https://seattle.wordcamp.test/2023/`, as an admin.

1. **Sponsors → Add New**, use **Attach Signed Agreement** to upload a PDF. Publish.
2. Reload. **View Agreement** opens the file, and its name has a 16-character suffix.
3. `wp --url=<site> eval 'echo get_post( <ATTACHMENT_ID> )->post_status;'` → `private`.
4. Repeat with a **JPG or PNG** as the agreement. Same result — this is the path that depends on the browser sending the field, so it's the one worth checking after any change to `wcb-spon.js`.
5. On the same sponsor, set a **featured image** and add an image to the content. Both stay `inherit` under the names they were uploaded with.
6. Upload a PDF through **Media → Add New**, then attach it with **Attach Signed Agreement** and press Update. It becomes `private` and is renamed on save.
7. The sponsor is unchanged: it renders on the front end, and `/wp-json/wp/v2/sponsors/<SPONSOR_ID>` still reports `"status":"publish"`.
8. **Media → Library** still lists and opens every agreement above for an admin or editor.

The migration, for agreements attached before this:

9. `wp wc-sponsor-agreements scan --verbose` reports which sites have any.
10. `wp --url=<site> wc-sponsor-agreements backfill --dry-run`, then without `--dry-run`. Each is reported `Private: yes`, a second `scan` reports none left, and **View Agreement** still opens the file. The files keep their names — they are recorded for the separate rename pass with `_wcorg_sponsor_agreement_needs_rename`.

    One site per process, so across the network:
    ```
    wp wc-sponsor-agreements scan | while read -r site; do
        wp --url="$site" wc-sponsor-agreements backfill
    done
    ```

Automated: `phpunit --testsuite "WordCamp MU Plugins" --group sponsor-agreements` (45 tests). The full `WordCamp MU Plugins`, `WordCamp Post Types` and `WordCamp Payments` suites pass, apart from failures that also occur on `production` — which ones varies by environment.

### DIFF
--- a/public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
+++ b/public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
@@ -15,7 +15,8 @@
 namespace WordCamp\Sponsor_Agreements\Backfill;
 
 use WP_CLI, WP_CLI_Command;
-use cli\Table;
+
+use function WP_CLI\Utils\format_items;
 
 use function WordCamp\Sponsor_Agreements\add_csprn_to_filename;
 use function WordCamp\Sponsor_Agreements\make_agreement_private;
@@ -74,7 +75,13 @@ function get_public_agreement_ids( $blog_id = null ) {
  * Nothing links to an agreement by URL -- both plugins hold an attachment ID and resolve it through
  * `wp_get_attachment_url()` -- so the new name reaches every reader.
  *
- * The generated sizes move with it. They sit beside the original under names derived from it.
+ * Every file Core derives from the upload moves with it, under one new base name:
+ *
+ * - the generated sizes, which for a PDF are renderings of the first page and for an image are copies of
+ *   it at up to 2048px;
+ * - the full-size original, when Core scaled the upload down. `_wp_attached_file` then points at the
+ *   `-scaled` copy while `original_image` holds the name the sizes are derived from, so that -- not the
+ *   attached file -- is what the new base is built from.
  *
  * Only ever called for the site the process was started on, because `ms_upload_constants()` defines
  * `UPLOADS` for whichever site that was -- see `backfill()`.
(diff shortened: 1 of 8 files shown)


mcliwanow:
I tried the backfill against a 3000x3000 JPEG agreement (a phone photo of a signed page is probably the most common case here). WP stores that as `foo-scaled.jpg`, keeps the original as `original_image = foo.jpg`, and names the sizes `foo-300x300.jpg` ... `foo-2048x2048.jpg`. `rename_agreement_file()` takes `$old_base` from the attached file, so it's `foo-scaled`, none of the sizes match it, and `original_image` is never looked at. The command says `Renamed: yes`, but only the `-scaled` file moved. The original and all seven sizes still have their old names, and the 1024/1536/2048 ones are perfectly readable copies of the contract.

PDFs and images under 2560px are fine, their generated files start with the same base name.

I suggest using `original_image` as the base when it's set, renaming that file too, and swapping the leading base in every size. `test_an_existing_agreement_is_renamed()` uses a fake PDF with no sizes, so a big image fixture would catch this. Worth fixing before the run, the command is gone afterwards.

KokkieH:
@mcliwanow Image-size files should now be handled correctly by the backfill. That's a good catch!

bor0 (commented):
Focused pass over the CLI, as asked.

The shape is right. The direct-table `scan` that avoids `switch_to_blog()` is a good call, the reasoning about `ms_upload_constants()` matches what the code actually does, and `rename_agreement_file()` handles the genuinely awkward parts carefully -- scaled originals, two sizes naming one file, the `-` separator that tells `photo-150x150.jpg` from `photobooth-150x150.jpg` -- each with a test pinning it.

Everything below is in the command layer rather than in the helpers, and I don't think that's a coincidence: `Command::scan()` and `Command::backfill()` are the only things in this file with no test coverage, while every function they call has some. The ordering issue at L398-399 would have surfaced the moment anyone wrote a test that runs `backfill` twice.

Ranked:

1. **L398-399 -- the rename half isn't resumable.** Setting the status first removes the attachment from the only query that can ever find it again. This contradicts "Safe to re-run" in the docblock and in the PR description. The one I'd call blocking.
2. **L359 -- `--skip-rename` makes that permanent** for every agreement on a site, and reports success while doing it.
3. **L310 -- a failed query reads as a clean site,** so `scan` can under-report and nothing in the output says so.
4. **L206 -- a crash mid-loop leaves disk and DB out of step,** on a command whose own docblock cites memory pressure on the sandbox as a known problem.
5. **L108 -- `file missing` conflates an absent file with a misconfigured run,** which is the exact mistake the one-site-per-process rule exists to prevent.
6. **L393 -- `--dry-run` doesn't enumerate anything,** and no run names the site it acted on.

None of this is wrong on the happy path. It's the recovery and reporting paths, which for a one-shot migration across the network are the ones that decide whether a bad run is noticed and fixable. I'd want a couple of tests over `backfill()` itself -- run twice, and run with a rename that fails -- before this goes near the sandbox.

public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
```diff
@@ -0,0 +1,489 @@
...
+			'left_behind' => 0,
+		);
+
+		foreach ( $agreement_ids as $agreement_id ) {
+			if ( $dry_run ) {
+				$migrated = true;
+				$outcome  = $untouched;
+			} else {
+				$migrated = make_agreement_private( $agreement_id );
+				$outcome  = $skip_rename ? $untouched : rename_agreement_file( $agreement_id );
```
bor0: The status change lands before the rename, and `get_public_agreement_ids()` only returns `post_status = 'inherit'` rows -- so once `make_agreement_private()` has succeeded, no later `backfill` can see this attachment again.

That makes the rename half a single-attempt operation, which the docblock above ("Safe to re-run: a site with nothing outstanding does one query and stops") and the PR description ("Idempotent and resumable") both promise it isn't. Concretely: a run reports `1 moved, 3 left`, the operator fixes whatever caused it, re-runs, and gets `No sponsor agreements left to migrate.` The three files keep the names they had, and nothing is left that can pick them up.

Swapping the two lines fixes it without any extra state -- rename first, set the status only once the rename came back clean. A crash between the two then leaves the row `inherit`, so the next run redoes both; worst case is a second suffix on an already-renamed file, which is harmless and visible.

public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
```diff
@@ -0,0 +1,489 @@
...
+		"SELECT DISTINCT agreement.ID
+		FROM {$prefix}posts AS agreement
+		INNER JOIN {$prefix}postmeta AS sponsor_meta
+			ON CAST( sponsor_meta.meta_value AS UNSIGNED ) = agreement.ID
+		INNER JOIN {$prefix}posts AS sponsor
+			ON sponsor.ID = sponsor_meta.post_id
+		WHERE sponsor_meta.meta_key IN ( $key_placeholders )
+		AND sponsor.post_type IN ( $type_placeholders )
+		AND agreement.post_type = 'attachment'
+		AND agreement.post_status = 'inherit'",
```
bor0, line 64: The other half of the resumability point at L398. `inherit` is the only thing that makes an attachment visible to this migration, and `backfill()` clears it on the first pass regardless of what happened on disk afterwards.

If the ordering in `backfill()` stays as it is, this query needs a second arm -- attachments already carrying `_wcorg_sponsor_agreement` whose file still has no suffix -- otherwise the command gets exactly one attempt per attachment and no way to tell, later, which attempts fell short.

public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
```diff
@@ -0,0 +1,489 @@
...
+	 * the wrong directory and silently rename nothing. Use `scan` to find the sites that need this.
+	 *
+	 * Safe to re-run: a site with nothing outstanding does one query and stops.
+	 *
+	 * ## OPTIONS
+	 *
+	 * [--dry-run]
+	 * : Report what would change, without changing it.
+	 *
+	 * [--skip-rename]
```
bor0: Worth reconsidering while the ordering at L398 stands: `--skip-rename` sets the status on every agreement on the site and touches no file, which by the query at L64 permanently forecloses the rename for all of them -- and the run ends on `%d agreements processed.`, i.e. reports success.

If it's here for a staged rollout, inverting it (`--status-only` being the arm that *can't* run first) would keep the option safe. Otherwise `--dry-run` already covers the look-before-you-leap case and this could just go.

public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
```diff
@@ -0,0 +1,489 @@
...
+ * }
+ */
+function rename_agreement_file( $attachment_id ) {
+	$outcome       = array(
+		'renamed'     => 0,
+		'left_behind' => 0,
+	);
+	$attached_path = get_attached_file( $attachment_id );
+
+	if ( ! $attached_path || ! is_file( $attached_path ) ) {
```
bor0: This early return yields `renamed: 0, left_behind: 0`, which `describe_rename()` renders as `file missing`. Accurate for a genuinely absent file -- but it is also exactly what the operator sees when the process is pointed at the wrong uploads directory, i.e. the `ms_upload_constants()` / `UPLOADS` trap the docblock warns about. In that case every row reads `Private: yes` / `file missing`, the summary is a plain success, and nothing distinguishes it from a site whose files were already gone.

Two cheap separations: carry `$attached_path` out in the outcome so the report can print what it looked for, and have `backfill()` assert up front that `wp_upload_dir()['basedir']` belongs to the site it is about to migrate. The whole one-site-per-process rule currently rests on a comment; four lines would make it rest on the code.

public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
```diff
@@ -0,0 +1,489 @@
...
+	 */
+	public function scan( $args, $assoc_args ) {
+		global $wpdb;
+
+		$verbose = isset( $assoc_args['verbose'] );
+		$sites   = $wpdb->get_results( "SELECT blog_id, domain, path FROM {$wpdb->blogs} ORDER BY blog_id" );
+		$found   = 0;
+
+		// A site whose tables have gone is a broken row, not a reason to stop part-way through a network.
+		$wpdb->suppress_errors( true );
```
bor0: `get_col()` returns `array()` when the query fails, so with errors suppressed a broken site is indistinguishable from a site with nothing to migrate. It isn't just skipped -- it is counted as clean in the `%d of %d sites have agreements to migrate.` summary, and its URL never reaches the pipe that drives `backfill`.

On a network this size that's the failure mode least likely to be noticed. Reading `$wpdb->last_error` after each call and reporting the sites that errored on STDERR would keep the "a broken row isn't a reason to stop" behaviour while making the gap visible.

Minor, same call: `suppress_errors()` returns the previous value. Capturing it and restoring that at L334, rather than forcing `false`, keeps this from changing state it didn't own.

public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
```diff
@@ -0,0 +1,489 @@
...
+
+		$metadata['sizes'][ $size ]['file'] = $move( $details['file'] );
+	}
+
+	if ( ! empty( $metadata['file'] ) ) {
+		$path_prefix      = str_contains( $metadata['file'], '/' ) ? dirname( $metadata['file'] ) . '/' : '';
+		$metadata['file'] = $path_prefix . $new_attached;
+	}
+
+	if ( $metadata ) {
```
bor0: The metadata and `_wp_attached_file` are written once, after every `@rename()` has already happened. If the process dies inside the loop -- and the `scan` docblock cites real memory pressure on the production sandbox -- the files sit on disk under their new names while the DB still points at the old ones. The attachment stops resolving, the files are orphaned, and the mapping between them exists nowhere.

Moving these two writes inside `$move()`, or flushing after each successful rename, narrows the window to a single file. If that's more churn than you want in a file that's being deleted anyway, the deployment notes should at least carry a recovery procedure, because reconstructing the pairing afterwards means matching files to attachments by hand.

public_html/wp-content/mu-plugins/wp-cli-commands/backfill-sponsor-agreements.php
```diff
@@ -0,0 +1,489 @@
...
+		$format        = $assoc_args['format'] ?? 'table';
+		$agreement_ids = get_public_agreement_ids();
+		$results       = array();
+		$unfinished    = array();
+		$untouched     = array(
+			'renamed'     => 0,
+			'left_behind' => 0,
+		);
+
+		foreach ( $agreement_ids as $agreement_id ) {
```
bor0, line 214: Two things that would help on a run this wide:

- `--dry-run` reports `pending` and nothing else. For a command that renames files across the network, a dry run that doesn't name the files it intends to move (attached file, `original_image`, each size) or the base it would give them is hard to sign off on -- and `rename_agreement_file()` already computes all of that before it touches anything.
- Nothing in the output names the site being migrated. `--url` is easy to forget, and forgetting it quietly targets the main site instead. Printing `home_url()` above the table would make each run self-documenting, which matters when that output is the only record the operator keeps.

Also worth a `WP_CLI::confirm()` on the non-dry-run path with the usual `--yes` escape, given what the command does.

bor0:
**Follow-up: checked on a sandbox, and the ordering at L398-399 has a concrete trigger**

I wanted to know whether `@rename()` would behave when this is run from a sandbox — sandboxes share the database, but I wasn't sure they share the filesystem. So I checked against a real event site. The answer turns out to be the opposite of what I assumed, and it lands squarely on the ordering at L398-399.

### What I found

- `blogs.dir` is an NFS export shared with production. The sandbox resolves attachments to files that genuinely exist: `is_file()` is true, with real sizes and mtimes. The premise "a sandbox can't see the files" is wrong for event sites.
- The upload directories are owned by `nobody:nogroup`, mode `0755`, and WP-CLI runs as a different uid. `is_writable()` comes back `no`.

### Why that combination is the bad one

Because `is_file()` succeeds, `rename_agreement_file()` does *not* take the early return at L108. It runs all the way to `@rename()`, which fails on permissions and increments `left_behind`. By that point `make_agreement_private()` has already committed `post_status` to the shared production database.

So one `backfill` run from a sandbox would set the status on every agreement it found, rename nothing, and — because `get_public_agreement_ids()` filters on `post_status = 'inherit'` (L64) — put those rows permanently beyond the reach of any later run, from any host, with any permissions. The `%d of %d agreements are unfinished` warning does fire, but by then it is reporting something that can no longer be acted on.

`scan` and `--dry-run` are both safe: neither writes to the database or to disk.

### What I'd change

Both of these are extensions of comments already on this review:

1. **Reverse L398-399** — rename first, set the status only once the rename came back clean. On its own that turns the scenario above into a no-op which can simply be retried from the right host.
2. **Add a preflight to `backfill()`** that refuses to start when it can't do the job. One correction to what I wrote at L108: the check needs to be on `dirname( get_attached_file( $id ) )` — the year/month directory the rename actually happens in — not on `wp_upload_dir()['basedir']`. Those are different directories and can differ in permissions.

With both, this exact run becomes a `WP_CLI::error()` on a clear message rather than two half-migrated rows.

### One more thing worth confirming before the real run

Central is the main site of the main network, so `_wp_upload_dir()` keeps it on `wp-content/uploads` rather than `blogs.dir`. That is a different path, and not necessarily the same mount or the same ownership. Since the `mes_sponsor_agreement` files live there, it is worth checking central separately rather than assuming the event-site result carries over.

That leaves one operational question for the deployment notes: which user do `wc-*` commands run as against production, and does it have write access to `blogs.dir`? That determines where this can be run at all — and it seems worth answering before the migration rather than during it.

KokkieH:
Latest commit narrows the migration to setting the attachment status. The file renaming has been split out and will be handled separately, so this can be deployed and the status backfill run while that's still being worked out.

@bor0 that removes most of what your review covered — the ordering, `--skip-rename`, the crash window between disk and database, and the uploads-directory question all go with it, since the command now reads and writes the database only. The two that still applied are fixed: `scan` reports the sites it couldn't read instead of counting them clean, and each run names the site it acted on.

bor0:
Took another pass over the branch, now at `331b51e9`.

Ran the suites locally:

- `phpunit --testsuite "WordCamp MU Plugins" --group sponsor-agreements` - 45/45 pass.
- Full `WordCamp MU Plugins` suite - 417 tests, one failure (`Test_QuickBooks_OAuth_State::test_authorize_url_carries_the_stored_state`). It fails the same way on `production`, so it is not from this branch. Thanks for dropping the test name from the description, that is better than either of us guessing which one the next person will hit.

Also spot-checked that each fix in the last commit has a test behind it: reverting any one of the three on its own fails exactly one test. It does.

My notes on `rename_agreement_file()` are addressed. I have a couple of small follow-on points about the deferred file work, left with the team in the usual place along with the rest of the review, since they are easier to discuss with the full context to hand. Neither is a change to this branch.

Nothing blocking from me.

bor0 (approved):
Approving at `331b51e9`.

My notes on `rename_agreement_file()` are all addressed, and each fix has a test that fails when that fix alone is reverted. I checked that rather than taking it: reverting any one of the three fails exactly one test.

Suites on the branch: 45/45 in `--group sponsor-agreements`, and 417 in the full `WordCamp MU Plugins` suite with only the QuickBooks failure that fails the same way on `production`. CI is green on 8.3, 8.4 and 8.5.

I have two small follow-on points about the deferred file work, left with the team in the usual place. Both are notes for whoever picks that up rather than changes to this branch, so they are not holding anything here.

Worth waiting on @mcliwanow before merging, since they raised the scaled-image case that led to the rewrite and had the last word on the approach.

mcliwanow (approved):
The scaled-image case is handled and tested, and the upload path on `wp_unique_filename` / `wp_insert_attachment_data` / `add_attachment` is what I had in mind. Checked against core trunk that `wp_ajax_upload_attachment()` runs the nonce, `upload_files` and `edit_post` checks before any of those fire. Approving.

Issue #600, opened 2017-12-04
Feed shows posts while in 'Coming soon' mode


ghost:
Imported from https://meta.trac.wordpress.org/ticket/3309 
Created by @ePascalC:

Belfast has 'coming soon Mode ON' (see https://2018.belfast.wordcamp.org/) as it's about to happen in May next year, but the feed on https://2018.belfast.wordcamp.org/feed/ is suddenly displaying posts thanking the sponsors...

Feed should probably not show anything as long as 'coming soon' mode is ON ?

ghost:
[Comment](https://meta.trac.wordpress.org/ticket/3309#comment:1) by @ePascalC:

@coreymckrill referred to this change just performed, but I don't see why it would have an impact: https://meta.trac.wordpress.org/changeset/6226

ghost:
[Comment](https://meta.trac.wordpress.org/ticket/3309#comment:2) by [@RMarks](https://profiles.wordpress.org/RMarks/):

Although feeds were discussed in https://meta.trac.wordpress.org/ticket/2934, it wasn't addressed.  Just these two:

In [\[5801\]](http://meta.trac.wordpress.org/changeset/5801):
WordCamp Coming Soon: Disable JP OpenGraph when Coming Soon enabled

In [\[6226\]](http://meta.trac.wordpress.org/changeset/6226):
WordCamp Coming Soon: Disable anonymous REST requests in Coming Soon mode

ghost:
[Comment](https://meta.trac.wordpress.org/ticket/3309#comment:3) by [@RMarks](https://profiles.wordpress.org/RMarks/):

I just read the comments in [Slack](https://wordpress.slack.com/archives/C08M59V3P/p1512404518000391).  I know when working on https://meta.trac.wordpress.org/ticket/2934, feeds were revealing content prior to either of my patches being applied.

ghost:
[Comment](https://meta.trac.wordpress.org/ticket/3309#comment:4) by @ePascalC:

[@RMarks](https://profiles.wordpress.org/RMarks/) ok, thanks for the update.

I suppose something like this could somewhere be added when going into Coming Soon mode:

```
add_action( 'init', 'coming_soon_remove_feeds', 99 );
function coming_soon_remove_feeds() {
	// Disable all feeds
	foreach( array( _, '_rdf', '_rss', '_rss2', '_atom', '_rss2_comments', '_atom_comments' ) as $feed ) {
[...]

PR #546, opened 2020-08-06
Add notice about visibility in feeds to post edit screen when site is in Coming Soon mode

timiwahalahti:
Modifies the existing notice shown when editing posts when Coming Soon mode is enabled to warn about posts being visible on RSS and WordPress.org profile feeds. Also adds the same notice to Gutenberg block editor.

Fixes #540

Props nao, @ePascalC, @CdrMarks 

### How to test the changes in this Pull Request:

1. Switch Coming Soon mode on for some WordCamp site
2. Go to create a new post with the classic editor
3. Go to create a new post with the block editor

PR #1994, opened 2026-09-02
Coming Soon: Stop feeds from serving posts while active

chrisdavidmiles:
## Summary

Coming Soon mode hides a WordCamp site's page, REST API, Open Graph tags, subscriber email,
and Publicize. This change does the same thing for feeds (`/feed/`, `/comments/feed/`, and the
RSS/Atom/RDF variants). Refs #600, [Meta Trac Ticket 3309](https://meta.trac.wordpress.org/ticket/3309).

## Technical details

The Coming Soon plugin hides the site by swapping the page for the placeholder on the
`template_include` filter via `override_theme_template()`. But feeds never reach that filter.
For a feed request, core dispatches the feed in `wp-includes/template-loader.php` and returns
before `template_include` runs. The REST API had the same gap and got its own lock.

This change fixes it for feeds using the `template_redirect` hook, which *does* fire for feed
requests (it's where we already disable Open Graph). A new `disable_feeds()` method there:

- Runs all `is_feed()` requests through `wp_die( 403 )`, including RSS2, RSS, Atom, RDF, and the
  comments feed.
- Removes core `feed_links` / `feed_links_extra` discovery tags so the placeholder stops
  advertising the now-blocked feeds.
- It's gated on the plugin's existing `override_theme_template` flag, so it matches everything
  else. If you're working on the site as an organizer, you can see and test everything as though
  the site is launched. If you're signed out, you get the anonymous view and everything is
  hidden, including the feed advertisements.

## How to test the changes in this Pull Request:

1. On a WordCamp site with at least one published post, turn on Coming Soon mode.
2. As a logged-out visitor, open a feed in your browser: `https://<camp>.wordcamp.org/feed/`
   (also try `/comments/feed/`). **Before:** the published posts are there in full. **After:**
   a `403`, and the `<link rel="alternate">` feed tags are gone from the Coming Soon page's
   source (`view-source:` the front page).
3. Confirm organizers are unaffected: logged in as an editor, open the feed. It still loads
   normally.
4. Launch the site (Coming Soon off) and confirm feeds and their discovery tags behave normally
   again.

Automated coverage: `tests/test-feed-lockdown.php` (modeled on `test-rest-lockdown.php`) covers
the 403, the tag removal, the launched-site case, and a logged-in editor keeping feed access.
Full suite green on multisite (17 tests, 30 assertions). PHPCS clean on changed lines.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
+++ b/public_html/wp-content/plugins/wordcamp-coming-soon-page/classes/wordcamp-coming-soon-page.php
@@ -13,6 +13,7 @@ class WordCamp_Coming_Soon_Page {
 		add_action( 'wp_head',                    array( $this, 'render_dynamic_styles'           )        );
 		add_filter( 'template_include',           array( $this, 'override_theme_template'         )        );
 		add_action( 'template_redirect',          array( $this, 'disable_jetpacks_open_graph'     )        );
+		add_action( 'template_redirect',          array( $this, 'disable_feeds' ) );
 		add_filter( 'rest_request_before_callbacks', array( $this, 'disable_rest_endpoints'       ), 99, 3 );
 		// Again after the callbacks, since `before_callbacks` only sets a response and a
 		// later filter on the same request can replace it. `PHP_INT_MAX` so nothing runs after it.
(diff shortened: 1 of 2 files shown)


eidolonnight:
I would not deploy this suggested fix as it creates a false sense of security. As noted in #546 these WordCamp posts appear in user profiles (possibly elsewhere too) which are all too easily crawled. This requires either a comprehensive fix or the notification solution shared in the aforementioned issue.

bor0:
Agreed that posts remain reachable via w.org profile activity — that's out of scope here, so I'd suggest dropping the "Fixes #600" and leaving the issue open with the remaining vectors listed.

On the incremental point: Coming Soon is already a series of per-vector locks (REST, OG, Publicize, subscriber email, robots.txt, structured data). `robots.php` skips coming-soon sites entirely, so crawlers are told nothing — yet `/feed/` still serves full posts to a direct request. This closes that gap rather than making a new promise.

#546's notice would pair well with it, though its copy needs updating since the RSS half stops being true.

chrisdavidmiles:
> suggest dropping the "Fixes https://github.com/WordPress/wordcamp.org/issues/600"

Done

PR #2000, opened 2026-09-04
Speaker Feedback: display submitted answers as plain text

bor0:
Feedback answers are attendee-submitted free-form text, so they are now displayed as such, and the feedback
view is appended after core's shortcode pass instead of before it.

## Testing steps

1. On a camp site, open a session that is inside its feedback window and submit feedback containing bracketed
   text, e.g. `It was great, I liked the [demo] part.`
2. Approve it under *Tickets → Feedback*. The moderation list shows the answer unchanged.
3. As the speaker or an organizer, open the session's `/feedback/` page. The answer renders as literal text,
   brackets included, and the feedback form and rating stars are unaffected.
4. `./public_html/wp-content/mu-plugins/vendor/bin/phpunit -c phpunit.xml.dist --testsuite "WordCamp Speaker Feedback"`
   → 68 tests, 189 assertions, all passing.

### DIFF
--- a/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/view.php
+++ b/public_html/wp-content/plugins/wordcamp-speaker-feedback/includes/view.php
@@ -19,7 +19,8 @@ defined( 'WPINC' ) || die();
 
 const SPEAKER_VIEWED_KEY = 'sft-speaker-viewed-feedback';
 
-add_filter( 'the_content', __NAMESPACE__ . '\render' );
+// The appended view is not post content, so it runs after core's shortcode pass at 11.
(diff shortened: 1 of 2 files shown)


mcliwanow (approved):
tests well, thanks for the changes

bor0:
For anyone arriving from the `strip_shortcodes()` discussion: this branch does not strip. `sanitize_answer_for_display()` encodes the `[` and `]` delimiters, so a nested `[[caption caption="B"]x[/caption]]` is inert along with everything else — encoding cannot manufacture a shortcode the way a removal pass can. The filter also moves to priority 12 so the appended view lands after core’s shortcode pass rather than in front of it.

PR #2001, opened 2026-09-04
Forms to Drafts, Blocks, Post Types: keep submitted titles as plain text

bor0:
Titles submitted through the public forms are free-form text, so they are now stored and rendered as such,
the way the submitted bodies already were. The listing blocks and the `wc-post-types` content appenders
render a title as text too, which also covers titles already in the database.

## Testing steps

1. On a camp site, submit the Call for Sponsors form with a Company Name containing bracketed text, e.g.
   `Acme [demo] Co`, then publish the drafted sponsor.
2. Add the Sponsors block to a page and list that sponsor. The name reads `Acme [demo] Co`, and the
   permalink still works.
3. Repeat through Call for Speakers (`Name`, `Topic Title`) and Call for Volunteers (`Name`) — the
   Speakers and Sessions blocks, and a classic-theme camp's session/speaker pages, all read the same.
4. `./public_html/wp-content/mu-plugins/vendor/bin/phpunit -c phpunit.xml.dist`
   → 1382 tests, 3550 assertions. The three `Test_Groups_Blocks` password-protected-page failures are
   present on `production` too and are unrelated to this branch.

### DIFF
--- a/public_html/wp-content/mu-plugins/3-helpers-misc.php
+++ b/public_html/wp-content/mu-plugins/3-helpers-misc.php
@@ -430,6 +430,34 @@ function wcorg_json_encode_attr_i18n( $raw_value ) {
 }
 
 /**
+ * Encode the shortcode delimiters in submitted text.
+ *
+ * Submitted free-text is stored and later rendered inside post content, where core parses shortcodes.
+ * Encoding the `[` and `]` delimiters keeps that text as literal characters, so a submission is shown
+ * as written rather than run. The result reads as `[` and `]` to a human.
+ *
+ * Applying this twice is a no-op: `&#91;` and `&#93;` contain no delimiter of their own. The result is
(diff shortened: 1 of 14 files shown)


mcliwanow (approved):
Thanks for the changes, code looks good and tests well.

bor0:
For anyone arriving from the `strip_shortcodes()` discussion: this branch does not strip. Everything goes through one shared `wcorg_escape_shortcodes()` helper that encodes the `[` and `]` delimiters, so a nested `[[caption caption="B"]x[/caption]]` is inert along with everything else — encoding cannot manufacture a shortcode the way a removal pass can, and applying it twice is a no-op.

It is applied on both sides: on the way in at `wordcamp-forms-to-drafts`, and again where the blocks and the `wc-post-types` appenders concatenate a title into post content, which is what covers titles already in the database.

PR #2007, opened 2026-09-07
Groups: show an error when publishing a stale event draft

### LINKED ISSUE #1978: Create event modal silently blocks publishing a draft whose date has passed
### Describe the bug

Continuing a draft whose date has already passed, from the Create event modal, cannot be published and gives no feedback. The modal's date input carries min set to today (`MINIMUM_EVENT_DATE`, localised from `inc/modal.php`), so the browser's native form validation blocks submit. The only signal is focus jumping to the Date field with a tooltip that disappears immediately; no request is sent and no notice is shown. The server already returns a clear message for this case (wporg_groups_past_event_date, "Event date cannot be in the past."), but the client never lets the request reach it.

### To reproduce

1. As an organiser, open a group that has an autosaved draft dated before today.                                                                                                                            
2. Sidebar + Create event.
3. In Continue from a draft, pick that draft.
4. Click Create event. Nothing happens; the Date field takes focus. The network tab shows no publish request.

### Expected behavior

Either a visible notice ("Event date cannot be in the past. Pick a new date.") or the date pre-filled with the next valid date when a stale draft loads.

### WordCamp

Groups network, any group site.

 - Device: MacBook Pro
 - OS: macOS
 - Browser: Chrome
 - Version: Latest

av3nger:
The Create event modal blocked a continued draft with a past date and gave no feedback. The Date input carries `min` set to today, so the browser refused the submit before any request left the page.

New events still can't be given a past date. When a loaded draft already has one, the publish request now reaches the server and the modal shows its "Event date cannot be in the past." error. The organiser fixes the date and submits again.

Fixes #1978

Props vanyukov

1. Open "+ Create event", fill in a title, start time and duration, and type a date before today into the Date field. Wait for "Draft saved at", then close the modal.
2. Open "+ Create event" again and pick the draft under "Continue from a draft".
3. Click "Create event". The modal should show "Event date cannot be in the past." and the draft should stay a draft.
4. Change the date to a future date and submit again.
5. The existing draft is published, with no duplicate event.
6. Start a new event and confirm the Date field still refuses a date before today.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/event-form.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/event-form.js
@@ -95,6 +95,9 @@ function EventForm(
 	const [ saving, setSaving ] = useState( false );
 	const [ error, setError ] = useState( '' );
 
+	// The browser-side minimum date only applies to a brand-new event; the server validates a loaded post's date.
+	const [ isExistingPost, setIsExistingPost ] = useState( false );
+
 	/**
 	 * Description is intentionally NOT in form state - the inline block editor
 	 * owns it. We grab the current value via `descriptionRef` only at
@@ -148,6 +151,7 @@ function EventForm(
 					return;
 				}
 				setVenues( res.venues || [] );
+				setIsExistingPost( !! res.is_editing );
 				setInitialDescription( res.fields.description || '' );
 				setFeaturedImage( {
 					id: res.fields.featured_image_id || 0,
@@ -280,7 +284,7 @@ function EventForm(
 					label={ __( 'Date', 'wporg-groups-frontend' ) }
 					type="date"
 					value={ form.date }
-					min={ isEdit ? undefined : MINIMUM_EVENT_DATE }
+					min={ isExistingPost ? undefined : MINIMUM_EVENT_DATE }
 					onChange={ ( v ) => updateField( 'date', v ) }
 					required
 					__nextHasNoMarginBottom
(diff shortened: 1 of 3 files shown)


bor0 (approved):
Verified locally, with a control run. This fixes #1978.

The strongest evidence is your own e2e spec. I ran it against both branches:

- **this PR:** passes in 12.8s
- **`production`, same spec:** fails at exactly `await expect( date ).not.toHaveAttribute( 'min' )`

So the test genuinely pins the behaviour it claims to, and the bug it describes is real.

Checked the mechanism rather than assuming it: `is_editing` is a real field on the form-data response (`rest.php:797`), computed as `$event_id > 0 && Event::POST_TYPE === get_post_type( $event_id )`. A continued draft is a `gatherpress_event` post, so it resolves true regardless of post status, and `event-form.js:144` does pass the draft id through as `event_id`. The chain holds.

Also confirmed `isEdit` is still used elsewhere in the component (line 358), so swapping this one call site doesn't leave a dangling prop.

Security angle, since this relaxes a client-side constraint: the server is still authoritative. Your new `test_publish_draft_rejects_past_date` proves a past-dated draft is rejected with `wporg_groups_past_event_date` **and** stays a draft, so the only thing being relaxed is the browser's pre-flight check, which was never the protection.

Checks: phpcs clean, Groups suite green, jest 41 green, build clean.

Nice touch asserting `input.validity.rangeUnderflow` instead of the `min` attribute on the fresh-form half — the comment explaining why (a broken localisation renders `min=""`, which passes an attribute check while enforcing nothing) is exactly the kind of thing that saves the next person an afternoon.

PR #2009, opened 2026-09-08
Groups: e2e for event-form edit mode and the settings-tab Speakers field

### LINKED ISSUE #2005: Groups: e2e gap on event-form edit mode and the Speakers field
#1935 consolidated the modal and the settings Events tab onto one shared `EventForm`, but the new e2e only covers create/publish. Two paths moved in that refactor and nothing exercises them:

- **Edit mode.** Reopening a published event on either surface. A regression that resends recurrence on edit, or applies the create-only `min` date constraint, would pass today.
- **Speakers.** Settings-only, passed through `children` rather than the shared state (`events-tab.js:110`). `_event_speakers` is written by a second request after the event save, so a regression that clears it is invisible.

Suggested: one settings-tab test that selects a speaker, saves, reopens the event, and asserts the speaker and the datetimes survived.

Raised by Copilot on #1935; deferred there rather than blocking the merge.

av3nger:
#1935 consolidated the modal and the settings Events tab into a shared `EventForm`, and its e2e only covered create and publish. This adds the settings-tab test suggested in #2005: select a speaker, publish, reopen the event with "Edit this event", save, and check that `_event_speakers` and the start datetime persist.

The spec also checks that the date input carries the minimum on the create form and not in edit mode, so the create-only date constraint has an assertion on each side. A REST unit test covers the other regression named in review, that a published event's recurrence comes back as `locked`. Only the unlocked cases had a test before.

The modal's edit mode is not covered here. Its edit button only renders for users who cannot manage group settings, so it needs an author-tier account and an event that account owns. The date `min`, the recurrence gate and the locked notice all live in the shared `EventForm`, so the settings-tab test exercises them for both surfaces

Fixes #2005, See #1935

Props gedex

### Screenshots

Test-only change.

1. The e2e workflow for this PR runs `tests/e2e/event-form-edit.spec.js` and passes.
2. The unit-test workflow passes `test_event_form_locks_recurrence_for_published_event` in the WordCamp MU Plugins suite.
3. To see the spec fail, make `event-form.js` always pass `MINIMUM_EVENT_DATE` as the date input's `min`, or make the settings tab save an empty `_event_speakers` list. Either turns the e2e run red.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/test-rest.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/test-rest.php
@@ -817,4 +817,21 @@ class Test_Groups_REST extends Groups_TestCase {
 
 		$this->assertSame( $stored_before, (string) get_post_field( 'post_title', $event_id, 'raw' ) );
 	}
+
+	/**
+	 * The form marks a published event's recurrence as locked. The frontend relies on this flag to leave
+	 * recurrence out of the edit payload.
+	 */
+	public function test_event_form_locks_recurrence_for_published_event(): void {
+		$editor_id = self::factory()->user->create( array( 'role' => 'editor' ) );
+		wp_set_current_user( $editor_id );
+
+		$event_id = create_event( $this->event_request( $this->base_event_params() ) )->get_data()['id'];
+
+		$load = new WP_REST_Request( 'GET', '/wporg-groups/v1/event-form-data' );
+		$load->set_param( 'event_id', $event_id );
+		$recurrence = get_event_form_data( $load )->get_data()['fields']['recurrence'];
+
+		$this->assertTrue( $recurrence['locked'] );
+	}
 }
--- a/tests/e2e/event-form-edit.spec.js
+++ b/tests/e2e/event-form-edit.spec.js
@@ -1,0 +1,94 @@
+const { test, expect } = require( '@playwright/test' );
+const { login } = require( './utils/login' );
+
+/**
+ * Edit mode of the shared event form, through the group-settings Events tab.
+ * The tab is the only surface with the Speakers field, and the only edit
+ * surface an editor-tier user sees: the group page's modal edit button is
+ * rendered for users who cannot manage group settings, so it needs its own
+ * author-tier account and event and is not covered here.
+ *
+ * Requires the `organiser2` (`password`) editor-tier test user, and
+ * `organiser1` to exist as a group member to pick as the speaker.
+ */
+test.describe( 'event form edit mode', () => {
+	// See event-form.spec.js for why the date sits this far out.
+	const DATE = '2099-03-04';
+	const START = '12:00';
+	const SPEAKER = 'organiser1';
+
+	/**
+	 * Reads the event's meta through REST rather than the rendered page.
+	 *
+	 * @param {import('@playwright/test').Page} page
+	 * @param {string}                          slug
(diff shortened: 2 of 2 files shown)


public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/test-rest.php
```diff
@@ -817,4 +817,21 @@ public function test_event_form_data_title_is_decoded(): void {
 
 		$this->assertSame( $stored_before, (string) get_post_field( 'post_title', $event_id, 'raw' ) );
 	}
+
+	/**
+	 * The form marks a published event's recurrence as locked. The frontend relies on this flag to leave
+	 * recurrence out of the edit payload.
+	 */
+	public function test_event_form_locks_recurrence_for_published_event(): void {
```
gedex, line 825:
```
docker compose -f docker-compose.phpunit.yml exec phpunit_wp phpunit --testsuite 'WordCamp Groups' --filter Test_Groups_REST

.............................                                     29 / 29 (100%)

Time: 00:01.240, Memory: 77.00 MB

OK (29 tests, 89 assertions)
```

👍🏾

public_html/wp-content/mu-plugins/wporg-groups-frontend/tests/test-rest.php
```diff
@@ -817,4 +817,21 @@ public function test_event_form_data_title_is_decoded(): void {
...
+		$editor_id = self::factory()->user->create( array( 'role' => 'editor' ) );
+		wp_set_current_user( $editor_id );
+
+		$event_id = create_event( $this->event_request( $this->base_event_params() ) )->get_data()['id'];
+
+		$load = new WP_REST_Request( 'GET', '/wporg-groups/v1/event-form-data' );
+		$load->set_param( 'event_id', $event_id );
+		$recurrence = get_event_form_data( $load )->get_data()['fields']['recurrence'];
+
+		$this->assertTrue( $recurrence['locked'] );
```
gedex, line 835: this test case looks good

tests/e2e/event-form-edit.spec.js
```diff
@@ -0,0 +1,94 @@
...
+ * Edit mode of the shared event form, through the group-settings Events tab.
+ * The tab is the only surface with the Speakers field, and the only edit
+ * surface an editor-tier user sees: the group page's modal edit button is
+ * rendered for users who cannot manage group settings, so it needs its own
+ * author-tier account and event and is not covered here.
+ *
+ * Requires the `organiser2` (`password`) editor-tier test user, and
+ * `organiser1` to exist as a group member to pick as the speaker.
+ */
+test.describe( 'event form edit mode', () => {
```
gedex, line 14:
```
npx playwright test tests/e2e/event-form-edit.spec.js --headed

Running 1 test using 1 worker

  ✓  1 ….js:34:2 › event form edit mode › keeps the speakers and drops the create-only date minimum when editing (5.7s)

  1 passed (6.7s)
```

works as expected 👍🏾 

Claude claimed following:

> this logs in as `organiser2`, which `event-form.spec.js:73` also uses. The workflow comment (`e2e-tests.yml:209-212`) sets out one editor account per test, so this quietly makes `organiser2` the exception.

i don't see it may overlap with 2 workers and checks in pr are passed so all good

gedex:
thanks for the pr!

gedex (approved):

PR #2012, opened 2026-09-10
Groups: bucket recurring occurrences by their own date in event archives

### LINKED ISSUE #2011: Recurring series moves to Past once its first occurrence ends
### Describe the bug

Once the first occurrence of a recurring GatherPress series has ended, the series disappears from the group's Upcoming view. Every remaining occurrence shows up under Past instead, dated weeks or months into the future.

The archive queries filter on the wrong date. GatherPress builds the upcoming/past `WHERE` comparison with `$wpdb->prepare( ' AND %i.%i >= %s', ... )`. The `%i` placeholder wraps both identifiers in backticks. `Query::clauses()` in `mu-plugins/gatherpress-recurring-events/includes/class-query.php` then tries to swap GatherPress's date columns for `COALESCE(occurrence date, master date)` with a plain `str_replace()` on the unquoted form. GatherPress builds `ORDER BY` by string concatenation, so that part matches. The `WHERE` never does. The result is a query that sorts by occurrence date but still filters by the series master row, which holds the first occurrence's date.

While the first occurrence is still in the future, the series passes the Upcoming filter and the occurrence join turns it into one card per occurrence, so the page looks right. The moment the first occurrence ends, the master end date drops below "now". Upcoming returns nothing for the series and Past returns every occurrence row, future ones included.

### To reproduce

1. On a group site, publish a recurring event whose first occurrence is already in the past. A weekly series that started last week will do.
2. Open the group's front page or its `/event/` archive. Upcoming is the default view.
3. The series is missing, even though its next occurrence is days away.
4. Open `/event/?event_time=past`.
5. Every remaining occurrence of the series is listed as past.

### Expected behavior

Upcoming lists each future occurrence of the series in date order. Past lists only the occurrences that have finished.

### WordCamp

https://events.wordpress.org/group/gatherpress/ shows no events. The next occurrence is this Friday.
https://events.wordpress.org/group/gatherpress/event/?event_time=past lists the future occurrences as past.

 - Device: any
 - OS: any
 - Browser: any
 - Version: n/a

av3nger:
Rewrite GatherPress's archive date comparison so it accepts the backtick-quoted identifiers that `$wpdb->prepare( '%i.%i' )` emits, not only the unquoted form. Until now the extension's `str_replace()` matched `ORDER BY` but never `WHERE`, so archives filtered on the series master date. Root cause and symptoms are in the linked issue.

Also adds a PHPUnit test that publishes a series with two finished and two future occurrences and asserts each archive bucket returns exactly those two. It fails on `production` and passes with the fix.

Fixes #2011

Props vanyukov

### Screenshots

None, server-side query change only.

1. On a group site, publish a weekly recurring event whose first occurrence starts now and ends in a few minutes, with a count of four or more.
2. After the first occurrence ends, open the group's `/event/` archive. Upcoming should list the remaining occurrences in date order and `?event_time=past` only the finished one. On `production` the series is missing from Upcoming and Past lists every future occurrence.
3. Run `vendor/bin/phpunit -c phpunit.xml.dist --testsuite "GatherPress Recurring Events"`.

### DIFF
--- a/public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
+++ b/public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
@@ -28,6 +28,13 @@ defined( 'WPINC' ) || die();
  */
 final class Test_GatherPress_Recurring_Events extends WP_UnitTestCase {
 
+	/** Leave the screen global as we found it for whatever runs next. */
+	public function tear_down() {
+		unset( $GLOBALS['current_screen'] );
+
+		parent::tear_down();
+	}
+
 	/** Weekly expansion keeps local wall time across DST. */
 	public function test_weekly_recurrence_preserves_wall_time_across_dst(): void {
 		$start = new DateTimeImmutable( '2026-10-26 18:00:00', new DateTimeZone( 'America/Los_Angeles' ) );
@@ -478,6 +485,13 @@ final class Test_GatherPress_Recurring_Events extends WP_UnitTestCase {
 	 * occurrence was listed under Past.
 	 */
 	public function test_archive_queries_use_occurrence_dates(): void {
+		/*
+		 * `wordcamp-remote-css/tests/bootstrap.php` defines `WP_ADMIN` for the
+		 * whole run, so `is_admin()` is true unless a screen says otherwise, and
+		 * `Query::clauses()` exempts wp-admin. This is a front-end archive query.
+		 */
+		set_current_screen( 'front' );
+
 		$post_id = self::factory()->post->create(
 			array(
 				'post_type'   => 'gatherpress_event',
--- a/public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
+++ b/public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
@@ -18,6 +18,7 @@ use WordPressdotorg\GatherPress_Recurring_Events\Occurrences;
 use WordPressdotorg\GatherPress_Recurring_Events\Plugin;
 use WordPressdotorg\GatherPress_Recurring_Events\Rule;
 use WP_Comment_Query;
+use WP_Query;
 use WP_UnitTestCase;
 
 defined( 'WPINC' ) || die();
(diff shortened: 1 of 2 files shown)


public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
```diff
@@ -467,6 +475,81 @@ public function test_rule_from_post_normalizes_stored_monthly_weekday(): void {
...
+				'post_type'               => 'gatherpress_event',
+				'post__in'                => array( $post_id ),
+				'orderby'                 => 'datetime',
+				'order'                   => 'DESC',
+				'gatherpress_event_query' => 'past',
+				'include_unfinished'      => 0,
+			)
+		);
+
+		$this->assertStringContainsString( 'COALESCE(gpre_occ_query.datetime_end_gmt', $upcoming->request );
```
gedex, line 548: Why only check the WHERE half?

The regex has to handle two spellings of the column name, and this only proves one. Break the other and the test still passes, but the archive sorts by the wrong date.

gedex (commented):
Checked the pattern against the real clauses: both the backticked WHERE and the unquoted ORDER BY rewrite, the occurrences table and the non-`_gmt` columns stay untouched. The filter runs after GatherPress has built its clauses, so there's always something to rewrite. The test's dates land in the same buckets no matter what time of day it runs.

public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
gedex:
```suggestion
		$this->assertStringContainsString( 'COALESCE(gpre_occ_query.datetime_end_gmt', $upcoming->request );
		$this->assertStringContainsString( 'COALESCE(gpre_occ_query.datetime_start_gmt', $upcoming->request );
```

gedex:
looks good to me. tested and works as described. thanks!

public_html/wp-content/mu-plugins/gatherpress-recurring-events/tests/test-gatherpress-recurring-events.php
gedex: i'll just apply this given it only ensuring robust test case

gedex (approved):

PR #2013, opened 2026-09-10
CampTix: keep a refunded attendee refunded when a later payment result arrives

mcliwanow:
## Why

`payment_result()` sets every attendee on an order to whatever the gateway reports. The cancelled branch checks the attendee's current status first (#1717) and lists `refund` among the statuses it refuses to downgrade. The completed and pending branches don't check at all, so a repeated payment result for an order that was since refunded puts the attendees back to `publish`. Gateways keep reporting the session as paid after a refund (Stripe records the refund on the charge, not the session), so the Stripe return URL, the timed-out-order recovery and a repeated PayPal IPN all reach that branch with a completed result.

## What changed

- A completed, pending or failed result leaves an attendee in `refund` alone, transaction record included, and logs `Refusing to change a refunded attendee` against it. Draft, pending and cancelled attendees still publish on a completed result, so a late webhook still recovers a cancelled order as before.
- The refund-all batch wrote `tix_refund_transaction_id` and its details to the first attendee's row for every related attendee on the order. They now go on the related attendee.
- Loading the Stripe return URL again after a refund now renders the tickets page (status unchanged) instead of redirecting to the ticket.
- Tests: refunded attendee on a completed, pending and failed result, a two-seat refunded order, draft/pending/cancel → publish, and the refund-all batch on a two-seat order.

## Testing

- `Test_CampTix_Plugin`: 4 failures on `production`, all passing on the branch.
- Local camp with Stripe mocked at the HTTP layer, `curl` only, same session against `production` and the branch: pay, refund, load the return URL again. `production` republishes the attendee (and for a two-seat order both seats, on top of a second order that took them meanwhile: 4 published on a capacity of 2). The branch leaves them in `refund` (2 published, 2 refunded). Pay, force `cancel`, then return: publishes on both. Pay, refund, then a return that reports a canceled intent: `production` moves the row to `failed`, the branch leaves it `refund`. Refund-all on a two-seat order: both rows refunded on both, and on the branch the second row carries the refund transaction id too.
- phpcs is clean on the touched lines.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -6749,6 +6749,16 @@ class CampTix_Plugin {
 
 			$old_post_status = $attendee->post_status;
 
+			// A refund reverses the charge, so a later completed, pending or failed result for
+			// the same order leaves the attendee and its transaction record untouched. Gateways
+			// keep reporting the session as paid after a refund (Stripe records the refund on the
+			// charge, not the session), and a gateway error on a repeated return would otherwise
+			// drop the refund marker.
+			if ( 'refund' === $attendee->post_status && in_array( $result, array( self::PAYMENT_STATUS_COMPLETED, self::PAYMENT_STATUS_PENDING, self::PAYMENT_STATUS_FAILED ), true ) ) {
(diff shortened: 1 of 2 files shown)


KokkieH (approved):
Tests as described.

Issue #1301, opened 2024-04-29
Ticket quantity does not reduce until paid


haqadn:
**Describe the bug**
The ticket quantity is not reduced until the payment is complete. This makes it possible to oversell tickets.

**To reproduce**

1. Create a paid ticket with quantity 1.
2. As a visitor, fill out the information for this ticket and go to the payments page.
3. From a different tab, as a visitor again, fill out the information for the ticket and go to the payment page again.
4. Pay for both tickets.
5. You should now have sold two tickets even though only one was available.

**Expected behavior**

The ticket should have said unavailable as soon as the first form was submitted, and the the second ticket purchase should have stopped in two places:
1. Before accessing the check-out form.
2. After submitting and being redirected to the checkout-form.

**Screenshots / Screencasts**

None

**WordCamp**

This was experienced on WordCamp Sylhet 2024 (and 2023). But, it's not specific to any WordCamps.

**System (please complete the following information):**
N/A

Issue #46, opened 2019-03-25
Catch rejected promises 


iandunn:
Originally discussed at https://github.com/WordPress/wordcamp.org/pull/33#discussion_r267527566

```
iandunn
[MIGHT] I'm getting an Unhandled promise rejection error here. It's probably just because I haven't finished setting up the API endpoint the way that the block expects, but the underlying issue is that it feels like a code smell that we're not catching rejections.

https://stackoverflow.com/questions/40500490/what-is-an-unhandled-promise-rejection

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

It seems equivalent to doing a try without a catch in PHP. Should we add catch everywhere we use promises?

@coreymckrill , @vedanshujain

----

coreymckrill
In the Sessions block I added catch in the SessionsSelect component, because that's where the promises ultimately get resolved. Looks like I neglected to carry that back to the Speakers block...

----

vedanshujain
@iandunn you will probably get this when request fails, I am not sure if we can do anything inside catch statement which will fix it and is not complicated. So user would have to reload their page in case this happens.
Please feel free to add a new issue for our standard behavior in case any network call fails across any WordCamp block(maybe display a notification on top of editor?)
```

coreymckrill:
If we move to using data stores (#69), we might not need to worry about catching rejected promises anymore...

coreymckrill:
I think this is irrelevant now that #69 has been merged. We're not directly handling promises from `apiFetch` anymore.

PR #2016, opened 2026-09-11
CampTix: count checkouts in progress against ticket, coupon and reservation quantities

mcliwanow:
### Why

A paid checkout creates the attendee as a draft and only publishes it once payment is confirmed. Every quantity check runs through `get_purchased_tickets_count()` and `get_used_coupons_count()`, and both counted `publish` and `pending` only, so the ticket count didn't drop while a purchase was in progress and the seat stayed on offer to everyone else. Two people buying the last ticket at the same time both finished, and it sold twice. That is #1301, and it is what happened at WordCamp Sylhet in April 2024 with ordinary buyers.

Two commits: the first counts checkouts in progress, the second makes the count-and-claim one step so simultaneous checkouts can't both take the last seat.

### What changed

**Count checkouts in progress**

- `get_purchased_tickets_count()` and `get_used_coupons_count()` count `draft` attendees. The "remaining" number on the ticket form, the Tickets and Coupons admin columns, and a reservation's "Used" column now include checkouts in progress. `cancel`, `failed`, `timeout` and `refund` still don't count.
- The gateway re-checks the order right before charging, and by then the buyer's own drafts exist. So the counting functions take an optional list of attendee IDs to exclude, and `verify_order()` passes the order's own attendees. Without it, buying the last seat would fail: the buyer's own draft would count against them and empty their order.
- An abandoned checkout keeps its seat until `review_timeout_payments()` moves it to `timeout` at 24 hours, matching the Stripe Checkout session lifetime. That sweep moves from the daily cron to the ten-minute one, so the seat comes back at 24 hours rather than anywhere up to 48.

**Sell the last seat once**

- `form_checkout()` takes a named database lock (`GET_LOCK`) per ticket and coupon in the order, in ID order, while it counts and writes its drafts, and releases them before any gateway call. A second checkout for the same rows waits until the first has claimed its seats. It's a mutex, not a transaction: statements stay autocommit, so the count after the lock is current and nothing needs rolling back.
- While the lock is held the counting queries read live rather than from the query cache, because the order is counted once before checkout and again under the lock. Everywhere else (ticket form, admin columns, reports) they stay cached as before.
- Nothing renders while the lock is held. A checkout that can't get the lock within ten seconds, or whose draft can't be written, is asked to try again; partially written drafts are removed so they never hold seats.
- On production's HyperDB a `GET_LOCK` is treated as a read and would land on a replica where no one else is waiting, so the request is pinned to the primary first. Local `wpdb` has no such method and uses one connection.

Visible effect for buyers: during a rush, checkouts for the same ticket are processed one behind another (tens of milliseconds each), and the odd buyer who loses the last seat is told to try again instead of being sold a ticket that isn't there.

Known follow-up, tracked separately and to land close behind: a buyer who abandons a paid checkout without using the gateway's cancel link has their own draft hold their single-use coupon or quantity-1 reservation until the 24-hour timeout, where before this change they could simply retry. There is no self-service way out in those 24 hours: the attendee status control on the edit screen needs `manage_sites`, so an organiser can't release a stuck draft either, and each case is a Central ticket until the follow-up lands.

### Testing

`Test_CampTix_Admin` and `Test_CampTix_Plugin`, the branch's test files against `production`'s `camptix.php` and against the branch:

| | `production` | branch |
|---|---|---|
| the counting and lock tests | fail | pass |

New tests cover a draft at each of the three checks, the four finished statuses, the 24-hour sweep and its schedule, an order not blocking its own re-check, the locks taken in ID order and released, the count reading live only under the lock, and a draft that can't be written aborting the order and removing the ones already written. Five revenue-report tests skip on both for lack of `intl`.

Parallel checkouts against a quantity-limited ticket, coupon and reservation, on a standalone install with the gateway mocked: `production` sells past the limit, the branch sells exactly the limit.

### DIFF
--- a/public_html/wp-content/plugins/camptix/camptix.php
+++ b/public_html/wp-content/plugins/camptix/camptix.php
@@ -219,8 +219,9 @@ class CampTix_Plugin {
 
 		add_action( 'tix_scheduled_every_ten_minutes', array( $this, 'send_emails_batch' ) );
 		add_action( 'tix_scheduled_every_ten_minutes', array( $this, 'process_refund_all' ) );
+		// A draft attendee holds its seat until it times out, so sweep often enough that an abandoned checkout releases it at 24 hours, not up to 48.
+		add_action( 'tix_scheduled_every_ten_minutes', array( $this, 'review_timeout_payments' ) );
 
-		add_action( 'tix_scheduled_daily', array( $this, 'review_timeout_payments' ) );
 		add_action( 'tix_scheduled_daily', array( $this, 'cron_camptix_stats_ticket_validation' ) );
 
 		if ( ! wp_next_scheduled( 'tix_scheduled_every_ten_minutes' ) )
@@ -3189,12 +3190,12 @@ class CampTix_Plugin {
 	/**
 	 * Returns true, if a reservation is valid, and can be used to purchase a ticket.
 	 */
-	function is_reservation_valid_for_use( $token ) {
+	function is_reservation_valid_for_use( $token, $exclude_attendee_ids = array() ) {
 		$reservation = $this->get_reservation( $token );
 		if ( ! $reservation )
 			return false;
 
-		$count = $this->get_purchased_tickets_count( $reservation['ticket_id'], $reservation['token'] );
+		$count = $this->get_purchased_tickets_count( $reservation['ticket_id'], $reservation['token'], $exclude_attendee_ids );
 		if ( $count < $reservation['quantity'] )
 			return true;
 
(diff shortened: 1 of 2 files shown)


bor0 (approved):
Approving the rebuilt branch. The named-lock rewrite is the right primitive: statements stay autocommit so each count reads the latest rows, which removes the isolation and timeout save/restore, the level mapping and the transaction half of the shutdown guard. Acquisition is ID-ordered, the locks are released before the gateway call, nothing renders while held, and the partial-insert cleanup correctly replaces the rollback.

Two one-liners I'd like in the commit that deploys:

1. Move `register_shutdown_function()` above the acquisition loop. As written it only runs after every lock is taken, so a fatal partway through leaves the earlier locks held with nothing registered to release them.
2. Check the `RELEASE_LOCK` return value and log when it isn't `1`. It returns 0 when another session holds the lock and NULL when it doesn't exist, and a release that didn't release is currently silent.

Also worth adding to the sandbox checks: confirm every `RELEASE_LOCK` comes back `1`, and that connections are non-persistent. Same host is not the same connection, and that distinction matters more here than it did for the transaction version.

Detail and the note on what ships without #46 are on the tracking issue.

PR #2043, opened 2026-09-11
Groups: constrain featured image aspect ratio on pages to 16/9

### LINKED ISSUE #2034: Groups: constrain the featured image height on pages
A tall featured image on a page such as About renders at full height and dominates the screen. `page.html` renders the featured image with no aspect ratio, while the front page and single-event templates both pin it to 16/9. Reproduced locally; a one-attribute fix brings it in line.

Props @supernovia ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9890)).

wppoland:
Fixes #2034.

### Description
On pages such as About, tall featured images rendered without an aspect ratio, causing them to render at full height and dominate the viewport.

This update sets `"aspectRatio": "16/9"` on the `core/post-featured-image` block in `page.html`, bringing it in line with the group site's `front-page.html`, `single.html`, and `single-event.html` templates.


bor0 (approved):
Verified locally. Set a deliberately tall 400x2000 featured image on the About page and measured the rendered size:

- Before: 1160x**5800** px, `aspect-ratio: auto 400 / 2000`
- After: 1160x**653** px, `aspect-ratio: 16 / 9`, `object-fit: cover`

That matches `front-page.html` and `single-event.html`, so all three templates now agree. Groups test suite green, no phpcs change. Merging.

PR #2044, opened 2026-09-11
Groups: Make online event link optional for online events

### LINKED ISSUE #2029: Groups: online events should not require a link
Marking an event as online makes the link field mandatory, but some groups share the joining link privately closer to the date. The field is a `required` URL input, so there is no way to publish without one. Reproduced locally.

Props @mauteri ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9883)).

wppoland:
Fixes #2029

### Changes
- Removes `required` validation in REST API (`inc/rest.php`) when creating or updating an online event with an empty online event link.
- Removes `required` attribute on the `online_event_link` text field in the event form component (`event-form.js`).
- Adds a unit test in `tests/test-rest.php` verifying that online events can be created without a link.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/inc/rest.php
@@ -1044,13 +1044,6 @@ function persist_event( int $event_id, WP_REST_Request $request ) {
 	if ( $fields['time_start'] === $fields['time_end'] ) {
 		return new WP_Error( 'wporg_groups_bad_time_range', 'End time must be after start time.', array( 'status' => 400 ) );
 	}
-	if ( $fields['is_online'] && '' === $fields['online_event_link'] ) {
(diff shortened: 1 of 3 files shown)


bor0 (approved):
Verified locally end to end.

Created an online event with the link field left empty: it saves, `gatherpress_online_event_link` is correctly absent rather than empty-string, the `online-event` venue term is still applied, and the front end renders "Online event" as a plain span with GatherPress's "Link available for attendees only." tooltip rather than a broken link.

Checked the security angle too, since this removes a validation rule: sanitisation is untouched. `online_event_link` still has `esc_url_raw` as its `sanitize_callback` in the arg schema, and `sync_online_event_link()` applies `esc_url_raw` again and deletes the meta on the empty path. Nothing new is trusted.

Groups suite 378/378, phpcs unchanged from baseline. Merging.

PR #2045, opened 2026-09-11
Groups: Allow clearing events search to restore full list

### LINKED ISSUE #2024: Groups: clearing the events search does not restore the full list
Clicking the search field's native clear button empties the input but does not re-submit, so the filtered list stays on screen. Submitting the now-empty field triggers the browser's "Please fill out this field" validation instead of resetting, because the input is marked `required`. Reproduced locally.

Props @superfurri ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9886)).

wppoland:
Fixes #2024

### Changes
- Removes `required` attribute from the search input on the events archive (`gatherpress-groups-tweaks.php`), preventing the browser from blocking submission when the input is emptied.
- Automatically resets and navigates to the events archive link when the browser's native clear button `(x)` is clicked or an empty search form is submitted.
- Reverts the forced `event_time=all` search default back to `upcoming` if an empty search query is submitted, restoring the full default list of upcoming events.
- Adds test cases for the search block filter and empty search parameter behavior.

### DIFF
--- a/public_html/wp-content/mu-plugins/groups/gatherpress-groups-tweaks.php
+++ b/public_html/wp-content/mu-plugins/groups/gatherpress-groups-tweaks.php
@@ -226,7 +226,8 @@ add_action( 'init', __NAMESPACE__ . '\register_event_speakers_meta' );
 /**
  * Rewrite the search block form action on the events archive to submit
  * to the archive URL instead of the default search URL, so search results
- * stay scoped to events.
+ * stay scoped to events. Also remove the required attribute and handle clearing
+ * the search field to restore the full list of upcoming events.
  */
 add_filter(
 	'render_block_core/search',
@@ -242,6 +243,17 @@ add_filter(
 			$content
 		);
 
+		// Remove required attribute safely and mark the events search form for the clear script.
+		$processor = new \WP_HTML_Tag_Processor( $content );
+		while ( $processor->next_tag() ) {
+			if ( 'FORM' === $processor->get_tag() ) {
+				$processor->set_attribute( 'data-events-search-form', '1' );
+			} elseif ( 'INPUT' === $processor->get_tag() ) {
+				$processor->remove_attribute( 'required' );
+			}
+		}
+		$content = $processor->get_updated_html();
+
 		// Add hidden field to default to "all" time when searching.
 		$content = str_replace(
 			'</form>',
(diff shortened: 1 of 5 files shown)


bor0 (changes requested):
The behaviour is right. I verified both halves locally: clicking the native clear button now navigates back to the clean archive and restores the full list (167 events), and submitting an empty field no longer triggers "Please fill out this field". Groups suite 380/380, phpcs unchanged from baseline.

I'd still like three things tightened before this merges.

**1. Inline `<script>` injected from a render filter.** The handler is appended as a raw `<script>` string inside `render_block_core/search`. That bypasses `wp_enqueue_script`, the build pipeline, and the eslint job in `linter.yml`, and it is emitted **twice** on the archive because two search blocks render there. Please move it to a real script file enqueued on the archive, or attach it with `wp_add_inline_script` against a registered handle. As written it also binds to every `.wp-block-search` on the page rather than just the events form, which is harmless today only because both forms happen to point at the archive.

**2. The `required` removal is a regex over raw HTML.**

```php
$content = preg_replace( '/\s+required(?:="[^"]*"|\b)/i', '', $content );
```

That strips the literal word "required" anywhere in the block markup preceded by whitespace, not just the attribute, so something like `aria-label="Login required"` would silently become `aria-label="Login"`. `WP_HTML_Tag_Processor` does this safely:

```php
$p = new WP_HTML_Tag_Processor( $content );
if ( $p->next_tag( array( 'tag_name' => 'INPUT' ) ) ) {
    $p->remove_attribute( 'required' );
}
$content = $p->get_updated_html();
```

The matching assertion `assertStringNotContainsString( 'required', $output )` is brittle for the same reason and would break on any future copy containing that word.

**3. The empty-search normalisation is copy-pasted four times** across `gatherpress-groups-tweaks.php` (twice), `functions.php`, and `archive-events-empty.php`. One shared helper would stop these drifting apart.

None of this is a correctness problem today, it's about the shape the code leaves behind.

wppoland:
Thank you for the detailed feedback @bor0! All three points have been addressed in the latest commit:

1. **Standalone script**: Extracted the search clear logic into a dedicated script (`public_html/wp-content/themes/groups-site/assets/js/events-search.js`), enqueued via `wp_enqueue_script` on `is_post_type_archive( 'gatherpress_event' )`. It specifically targets forms with `data-events-search-form="1"` rather than all search forms on the page.
2. **Safe `WP_HTML_Tag_Processor`**: Replaced regex with `WP_HTML_Tag_Processor` to mark the form with `data-events-search-form` and safely remove the `required` attribute from the input. Updated test assertions to inspect the attribute directly via `WP_HTML_Tag_Processor` as well as verifying that no `<script>` tag is output.
3. **Centralized normalization**: Added `normalize_event_time_filter()` in `gatherpress-groups-tweaks.php` to consolidate the empty search check across the archive query filter, body class generator in `functions.php`, and `archive-events-empty.php`. Added unit tests covering empty queries, whitespace, text searches, and non-all time filters.

All 383 tests in the WordCamp Groups suite pass, and phpcs is clean.

bor0 (approved):
All three points addressed, and the behaviour still works. Approving.

- The inline `<script>` is gone, replaced by `assets/js/events-search.js` enqueued properly with `filemtime()` versioning. Confirmed the rendered archive now has zero inline handler copies and one `<script src=...>`.
- `WP_HTML_Tag_Processor::remove_attribute( 'required' )` replaces the regex, so the word "required" can no longer be stripped out of unrelated attribute values.
- `normalize_event_time_filter()` gives the empty-search logic one home.

Re-tested both behaviours: clearing the field navigates back to the clean archive and restores the full list, and submitting an empty field no longer raises the browser validation tooltip. Groups suite 383/383, phpcs unchanged from baseline.

One leftover, non-blocking: the query-vars filter at `gatherpress-groups-tweaks.php:387` still has the old inline copy rather than calling `normalize_event_time_filter()`, which is defined about a hundred lines above it in the same file. That's the call site that actually shapes the query, so it's the one most worth routing through the helper.

bor0:
Resolved the conflict with `production` by merging it in (follow-up commit, not a rebase, so your history stays intact).

The clash was only in `test-gatherpress-groups-tweaks.php`, where #2050 added its RSVP comment-query tests at the same point in the file that this branch adds its search-clear and `normalize_event_time_filter` tests. Both are purely additive, so both are kept.

Verified after merging: Groups suite **393/393**, phpcs unchanged from baseline, and both features working side by side on a local build — an empty search restores the full list (1 event → 166), while #2050's guard still returns 0 comments for the discussion query and 1 for an explicit RSVP query. Now mergeable.

PR #2046, opened 2026-09-11
Groups: Restrict allowed blocks in event description editor

### LINKED ISSUE #2042: Groups: description editor accepts blocks the event form cannot round-trip
The description editor has no allowed-block restriction, so blocks such as pullquote, gallery and cover can be inserted. On save they are written twice, and when the event is reopened for editing they are filtered out of the form entirely while still rendering on the public page, leaving no way to see, edit or remove them. Found while verifying the call-for-testing reports rather than reported by a tester. Reproduced locally.

Found during verification of the [call for testing](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9873).

wppoland:
Fixes #2042

### Changes
- Restricts `allowedBlockTypes` in the event description `BlockEditorProvider` settings to only the blocks supported by `DESCRIPTION_BLOCK_NAMES` (`core/paragraph`, `core/heading`, `core/list`, `core/list-item`, `core/image`, `core/quote`, `core/separator`, `core/code`, `core/preformatted`, `core/group`, `core/columns`, `core/column`).
- Filters parsed initial blocks on mount to drop unsupported blocks before rendering.
- Adds test coverage in `tests-js/description-editor.test.js`.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/constants.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/constants.js
@@ -1,0 +1,18 @@
+/**
+ * Allowed core blocks for the event description prose.
+ * Kept in sync with DESCRIPTION_BLOCK_NAMES in inc/defaults.php.
+ */
+export const ALLOWED_BLOCK_TYPES = [
+	'core/paragraph',
+	'core/heading',
+	'core/list',
+	'core/list-item',
+	'core/image',
+	'core/quote',
+	'core/separator',
+	'core/code',
+	'core/preformatted',
+	'core/group',
+	'core/columns',
+	'core/column',
+];
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/description-editor.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/description-editor.js
@@ -21,6 +21,9 @@ import {
 import { useDispatch } from '@wordpress/data';
 import { registerCoreBlocks } from '@wordpress/block-library';
 import { createBlock, parse, serialize } from '@wordpress/blocks';
+import { ALLOWED_BLOCK_TYPES } from './constants';
+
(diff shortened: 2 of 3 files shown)


bor0 (approved):
Verified locally and this does what it says. `/pullquote` no longer offers anything in the slash inserter, while allowed blocks like `/heading` still insert normally. That closes the round-trip hole in #2042: blocks the form can't display on re-open can no longer be added in the first place. Jest green, build clean.

Approving on the substance. One thing worth addressing, and one process note.

**Worth addressing:** the block list now exists twice, as `ALLOWED_BLOCK_TYPES` in `constants.js` and `DESCRIPTION_BLOCK_NAMES` in `inc/defaults.php`, kept together only by a comment. These will drift, and when they do the failure is silent: a block insertable in the editor but stripped on save, which is the exact class of bug this PR is fixing. Passing the PHP constant through to JS (localised alongside the other modal data) would make that impossible.

**Process note:** I've merged #2047, which rewrites the same file, so this now needs a rebase before it can go in. Flagging because #2046, #2047, #2048 and #2049 all edit `description-editor.js` and each one conflicts the others.

Also, as on the sibling PRs, please leave the unrelated em dash to hyphen comment rewrites out of the diff.

bor0:
Resolved the conflict with #2047 by merging `production` in (follow-up commit rather than a rebase, so your branch history stays intact).

The clash was only in `ensureCoreBlocksRegistered()`, where #2047 added `registerCoreFormatTypes()` to the same guard this branch touches. Kept both: core formats register alongside core blocks, and `allowedBlockTypes` still restricts the inserter.

Verified the two features coexist locally: `getFormatTypes()` returns the full set (bold/italic/link/...), and the block inserter now offers only Paragraph, Image and Heading, with no Pullquote or Gallery. Groups suite 379/379, jest 43/43, build clean. Full CI green including Playwright.

Merging. The `constants.js` / `DESCRIPTION_BLOCK_NAMES` duplication from my review is still worth a follow-up.

PR #2047, opened 2026-09-11
Groups: Enable inline formatting in description editor

wppoland:
Fixes #2017

### Changes
- Imports and calls `registerCoreFormatTypes` from `@wordpress/format-library` in `description-editor.js`, registering inline format types (bold, italic, link, etc.) in the event description editor toolbar.
- Enqueues `wp-format-library` stylesheet alongside `wp-components` and `wp-block-editor` in `enqueue_supplementary_assets()` in `inc/modal.php`.
- Adds test coverage in `tests/test-event-manage-block.php`.

PR #2048, opened 2026-09-11
Groups: Add undo/redo history stack and shortcut support in description editor

### LINKED ISSUE #2018: Groups: undo (Cmd/Ctrl+Z) does nothing in the description editor
Typing in the event description and pressing Cmd/Ctrl+Z has no effect. The editor mounts a bare `BlockEditorProvider` with no `ShortcutProvider` and no undo stack, so there is no history to step back through. Reproduced locally.

Props @mauteri ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9883)).

wppoland:
### Description
This PR resolves #2018 by adding undo and redo support to the event description block editor.

Previously, `DescriptionEditor` mounted a bare `BlockEditorProvider` with no `ShortcutProvider` and no undo/redo stack, so pressing `Cmd/Ctrl+Z` had no effect.

Changes in this PR:
- Wrap `BlockEditorProvider` in `ShortcutProvider` from `@wordpress/keyboard-shortcuts` and register `core/editor/undo` and `core/editor/redo`.
- Implement a dedicated history manager (`description-editor-history.js`) that distinguishes between transient keystrokes (`onInput`) and committed block changes (`onChange`), maintaining `past` and `future` stacks.
- Handle keyboard shortcuts (`Cmd/Ctrl+Z` for undo, `Cmd/Ctrl+Shift+Z` / `Ctrl+Y` for redo), reverting uncommitted typing bursts or stepping through committed states.
- Deep clone block snapshots to avoid state mutation.
- Add unit tests covering undo/redo history management (`description-editor-history.test.js`).

Fixes #2018

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/description-editor-history.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/description-editor-history.js
@@ -1,0 +1,201 @@
+/**
+ * History manager for the event description block editor.
+ *
+ * Implements an undo/redo stack that distinguishes between transient
+ * keystrokes (onInput) and committed block changes (onChange).
+ *
+ * @package WordCamp\Groups\Frontend
+ */
+
+export const MAX_HISTORY_LENGTH = 50;
+
+/**
+ * Deep clone an arbitrary value (objects, arrays, primitives).
+ *
+ * @param {*} value The value to deep clone.
+ * @return {*} Cloned value.
+ */
+function deepClone( value ) {
+	if ( null === value || typeof value !== 'object' ) {
+		return value;
+	}
+	if ( Array.isArray( value ) ) {
+		return value.map( deepClone );
+	}
+	const copy = {};
+	for ( const key of Object.keys( value ) ) {
+		copy[ key ] = deepClone( value[ key ] );
+	}
+	return copy;
+}
+
+/**
+ * Deep clone an array of blocks while preserving clientIds and attributes.
+ *
+ * @param {Array} blockList Array of block objects to clone.
+ * @return {Array} Cloned array of block objects.
+ */
+export function cloneBlocks( blockList ) {
+	return blockList.map( ( block ) => ( {
+		...block,
+		attributes: deepClone( block.attributes ),
+		innerBlocks: cloneBlocks( block.innerBlocks || [] ),
+	} ) );
+}
+
+/**
+ * Initialize a new history manager state.
+ *
+ * @param {Array} initialBlocks Initial blocks array.
+ * @return {Object} Initialized history manager object.
+ */
+export function createHistoryManager( initialBlocks ) {
+	return {
+		past: [],
+		future: [],
+		present: initialBlocks,
+		lastCommitted: initialBlocks,
+		lastCommittedSerialized: null,
+	};
+}
+
+/**
+ * Record a transient input event (e.g. typing) without committing to past history.
+ *
+ * @param {Object} history   History manager object.
+ * @param {Array}  newBlocks Updated blocks array.
+ */
+export function recordInput( history, newBlocks ) {
+	history.present = newBlocks;
+}
+
+/**
+ * Record a committed change event, saving the previous state into past history.
+ *
+ * @param {Object}   history     History manager object.
+ * @param {Array}    newBlocks   Updated blocks array.
+ * @param {Function} serializeFn Function that serializes a block array to a string.
+ * @param {number}   maxHistory  Maximum number of history entries to retain.
+ */
+export function recordChange( history, newBlocks, serializeFn, maxHistory = MAX_HISTORY_LENGTH ) {
+	const newSerialized = serializeFn( newBlocks );
+	const lastCommittedSerialized =
+		history.lastCommittedSerialized ?? serializeFn( history.lastCommitted );
+
+	if ( newSerialized !== lastCommittedSerialized ) {
+		history.past.push( cloneBlocks( history.lastCommitted ) );
+		if ( history.past.length > maxHistory ) {
+			history.past.shift();
+		}
+		history.future = [];
+		history.lastCommitted = newBlocks;
+		history.lastCommittedSerialized = newSerialized;
+	}
+
+	history.present = newBlocks;
+}
+
+/**
+ * Step back to the previous state in history.
+ *
+ * @param {Object}   history     History manager object.
+ * @param {Function} serializeFn Function that serializes a block array to a string.
+ * @return {Array|null} Restored blocks array, or null if nothing to undo.
+ */
+export function stepUndo( history, serializeFn ) {
+	const lastCommittedSerialized =
+		history.lastCommittedSerialized ?? serializeFn( history.lastCommitted );
+
+	// If there is an uncommitted transient typing burst, undo back to last committed state.
+	if ( history.present !== history.lastCommitted ) {
+		const presentSerialized = serializeFn( history.present );
+		if ( presentSerialized !== lastCommittedSerialized ) {
+			history.future.push( cloneBlocks( history.present ) );
+			const target = cloneBlocks( history.lastCommitted );
+			history.present = target;
+			return target;
+		}
+	}
+
+	if ( history.past.length === 0 ) {
+		return null;
+	}
+
+	const previous = history.past.pop();
+	history.future.push( cloneBlocks( history.lastCommitted ) );
+	history.lastCommitted = previous;
+	history.lastCommittedSerialized = null;
+	history.present = previous;
+	return previous;
+}
+
+/**
+ * Step forward to the next state in redo history.
+ *
+ * @param {Object} history    History manager object.
+ * @param {number} maxHistory Maximum number of history entries to retain.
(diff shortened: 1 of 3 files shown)


bor0 (changes requested):
It works. Verified locally: typing then cmd+Z steps back, repeated cmd+Z walks all the way to empty without closing the modal or losing the form, and cmd+shift+Z redoes correctly. Jest green (51 tests), build picks up `wp-keyboard-shortcuts`.

Some things to tighten before merge.

**1. Untranslated user-facing strings.**

```js
description: 'Undo the last change.',
description: 'Redo the last undone change.',
```

These need `__( ..., 'wporg-groups-frontend' )`. The project has open i18n work (#1964, #1927) and this quietly adds two more untranslated registrations.

**2. Two shortcut mechanisms doing the same job.** The component registers `core/editor/undo` / `core/editor/redo` and uses `useShortcut`, *and* separately implements `handleKeyDown` with its own `navigator.platform` sniffing and key matching. Today only one fires because `useShortcut` calls `preventDefault()` and `handleKeyDown` bails on `event.defaultPrevented`, but that is a subtle ordering dependency to leave in place. Please keep one. If you keep the manual path, `isAppleOS()` from `@wordpress/keycodes` is preferred over the deprecated `navigator.platform`.

**3. Registering into core's namespace on the global store.** `core/editor/undo` and `core/editor/redo` are core's own shortcut names, registered here into the shared `core/keyboard-shortcuts` store and never unregistered on unmount. Please namespace these to this editor (e.g. `wporg-groups/description-undo`) and clean up on unmount.

**4. Minor:** `recordChange` serialises both the incoming blocks and `lastCommitted` on every change event, and `stepUndo` serialises twice more. Fine for a short description, but it is full-document serialisation on a hot path. `cloneBlocks` also only shallow-copies `attributes`, so nested attribute values stay shared by reference between history entries.

Also, as on the sibling PRs, the diff rewrites unrelated em dashes to hyphens in existing comments and drops the `**never**` emphasis in a docblock. Worth leaving untouched lines alone so the diff stays reviewable.

wppoland:
Thank you for the thorough review @bor0! All points have been addressed in the latest commit:

1. **i18n strings**: Wrapped shortcut descriptions in `__( 'Undo the last change.', 'wporg-groups-frontend' )` and `__( 'Redo the last undone change.', 'wporg-groups-frontend' )`.
2. **Unified shortcut mechanism**: Removed the redundant manual `handleKeyDown` event listener and platform sniffing completely, letting `useShortcut` within `ShortcutProvider` handle shortcut interception and cross-platform key combinations.
3. **Namespacing & unregister on unmount**: Namespaced the shortcuts to `wporg-groups/description-undo` and `wporg-groups/description-redo`, and added cleanup in `useEffect` to call `unregisterShortcut` on unmount.
4. **History & serialization optimizations**:
   - `cloneBlocks` now performs a recursive deep clone of block attributes so nested objects and arrays do not share references between history entries.
   - History states now cache `lastCommittedSerialized`, eliminating redundant serialization in `recordChange` and skipping serialization entirely when stepping through committed history states.
5. **Clean diff**: Rebased on latest `upstream/production` and kept untouched comments and docblocks intact.

All 53 JS unit tests and 379 PHPUnit tests pass.

bor0 (changes requested):
The three review points are all properly addressed: strings wrapped in `__()`, shortcuts renamed to `wporg-groups/description-undo` and `-redo` instead of squatting on core's names, `unregisterShortcut` on unmount, and the duplicate `handleKeyDown` with its `navigator.platform` sniffing removed. That's exactly what I asked for.

**But redo is now broken.** I think removing the manual key handler took redo with it.

Repro, on a freshly opened Create event modal:

1. Click into the description, type `GAMMA`
2. Press ⌘Z — text clears, correct
3. Press ⇧⌘Z — **nothing happens**, the editor stays empty

⌘Y, the registered alias, does nothing either. I reproduced this three times, including once with two-second pauses between each step to rule out a timing artifact from my automation. Undo works every time; redo never does.

This worked in the previous revision of the PR, so it's a regression from this round of cleanup.

Two things I checked so you don't have to re-derive them:

- **The shortcuts are registered correctly.** `getShortcutKeyCombination()` returns `{modifier: 'primary', character: 'z'}` for undo and `{modifier: 'primaryShift', character: 'z'}` for redo, with the `⌘Y` alias present. So registration isn't the problem.
- **The history stack itself is fine.** I wrote a throwaway unit test against `description-editor-history.js` covering both redo paths — redoing a transient typing burst that was undone, and the same thing after the editor echoes the undone value back through `onChange`. Both pass. So `stepRedo` is not at fault.

That points at the component wiring for redo specifically rather than the history module. Worth checking whether the `useShortcut` registration for redo is firing at all, since the undo one clearly is.

Everything else here is good, and jest is green at 53 tests — which is part of the problem, since the redo path has unit coverage that passes while the actual keystroke does nothing. A test that drives the shortcut rather than calling `stepRedo()` directly would have caught this.

wppoland:
Thank you for the thorough reproduction and investigation!

The issue was that removing the container-level key handler in favor of relying solely on `useShortcut` caused the contenteditable block context to bypass redo keystrokes (`Shift+Cmd+Z` and `Cmd+Y`), while `Cmd+Z` was being handled by the browser's native DOM undo rather than our editor history stack.

We have addressed this by:
1. Handling keystrokes via `handleEditorKeyDown`, utilizing `isAppleOS()` from `@wordpress/keycodes` to cleanly handle `Cmd+Z` and `Shift+Cmd+Z` on Apple platforms, as well as `Ctrl+Z`, `Ctrl+Shift+Z`, and `Ctrl+Y` on Windows/Linux, calling `event.preventDefault()` so browser native behaviors do not conflict.
2. Removing the duplicate `useShortcut` component so there is only one shortcut execution mechanism in place.
3. Retaining `registerShortcut` and `unregisterShortcut` with translated `__()` strings in the `wporg-groups/` namespace on mount/unmount so the global shortcut store is properly populated.
4. Adding 10 new unit tests in `description-editor-history.test.js` (bringing the suite to 63 passing tests) that directly drive keyboard events across both Apple and non-Apple platforms, including a full keystroke-driven cycle verifying that typing `GAMMA`, pressing `Cmd+Z`, and pressing `Shift+Cmd+Z` accurately clears and restores block content.

bor0 (approved):
Approving. First, a correction: **my previous review was wrong about redo being broken.** That was my testing error, not a bug in your code, and I'm sorry for the detour.

What happened: I was driving the editor with synthetic keystrokes, which never reached your handler. When I checked out `production` as a control — which has no undo feature at all — and pressed ⌘Z in the description editor, the text still cleared. So what I'd been reading as "undo works, redo doesn't" was Chrome's native contenteditable undo the whole time. That method can't distinguish your feature from the browser, and it told me nothing about redo either. Retracting the regression claim.

On the actual changes, all three original points are properly addressed:

- Strings wrapped in `__()`.
- Shortcuts renamed off core's namespace to `wporg-groups/description-undo|redo`, with `unregisterShortcut` on unmount.
- The manual key handling is no longer a loose duplicate — extracting it into a pure `handleEditorKeyDown()` is a better answer than the "pick one" I suggested, because now it's directly testable.

The test coverage that came with it is the strongest part: Mac ⌘Z and ⇧⌘Z in both letter cases, Windows Ctrl+Z and Ctrl+Y, alt-modifier rejection, `defaultPrevented` handling, and a full keystroke-driven undo-redo cycle. 63 tests green, build clean.

Also noting you made ⌘Y deliberately *not* redo on Mac, with a test pinning that. That's correct, and it explains an observation from my last review that I'd wrongly read as a symptom.

One thing worth flagging rather than blocking on: nothing here drives a real browser keystroke, and as I found the hard way, synthetic events don't exercise this path. The repo has Playwright, so an e2e spec covering type → ⌘Z → ⇧⌘Z would be the only way to actually prove this end to end. Failing that, a human pressing ⌘Z in the modal once settles it in seconds. Worth doing before this ships, but it doesn't need to hold up the review.

wppoland:
Thank you @bor0 for the detailed follow-up and the approval! Appreciate you taking the time to investigate and confirm the redo behavior. We will keep the Playwright e2e test suggestion in mind for future test coverage expansions as well.

bor0:
Resolved the conflict with `production` by merging it in (follow-up commit, not a rebase).

The clash was just the import block: #2049 added `@wordpress/hooks` and `@wordpress/media-utils` where this branch adds `@wordpress/i18n`, `@wordpress/keycodes` and `@wordpress/keyboard-shortcuts`. Both groups kept; the media-upload settings and the undo/redo wiring are unrelated.

Verified all four editor PRs now coexist on a local build of the merged branch:

| from | check | result |
|---|---|---|
| #2047 | registered formats | `core/bold`, `core/italic`, `core/link` |
| #2046 | inserter contents | Paragraph, Image, Heading only |
| #2049 | image placeholder | Upload, **Media Library**, Insert from URL |
| #2048 | shortcuts | `description-undo` ⌘Z, `description-redo` ⇧⌘Z |

Build clean with all deps present (`wp-format-library`, `wp-hooks`, `wp-keyboard-shortcuts`, `wp-keycodes`, `wp-media-utils`), jest 63 green. Now mergeable.

PR #2049, opened 2026-09-11
Groups: Pass mediaUpload and MediaUpload to description editor settings

### LINKED ISSUE #2019: Groups: description image block offers no media library or upload
Inserting an image into an event description shows a placeholder reading "Drag and drop an image, upload, or choose from your library", but the only button is **Insert from URL**. `mediaUpload` is not passed in the editor settings, so the promise in the placeholder text cannot be met. Reproduced locally.

Props @patricia70 ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9873)).

wppoland:
### Description
This PR resolves #2019 by providing `mediaUpload` and `MediaUpload` in the `BlockEditorProvider` settings for the event description editor.

Previously, inserting an image block into the event description displayed a placeholder promising "Drag and drop an image, upload, or choose from your library", but only offered an "Insert from URL" button because `mediaUpload` was missing from the editor settings.

Changes in this PR:
- Import `uploadMedia` and `MediaUpload` from `@wordpress/media-utils`.
- Pass `mediaUpload` (delegating to `uploadMedia`) and `MediaUpload` in `BlockEditorProvider` settings.
- Enables the "Upload" button (file picker and REST upload), the "Media Library" modal button, and drag & drop file upload into the image block.

Fixes #2019

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/description-editor.js
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/components/event-form/description-editor.js
@@ -22,6 +22,8 @@ import { useDispatch } from '@wordpress/data';
 import { registerCoreBlocks } from '@wordpress/block-library';
 import { registerCoreFormatTypes } from '@wordpress/format-library';
 import { createBlock, parse, serialize } from '@wordpress/blocks';
+import { addFilter } from '@wordpress/hooks';
+import { uploadMedia, MediaUpload } from '@wordpress/media-utils';
 import { ALLOWED_BLOCK_TYPES } from './constants';
 
 export { ALLOWED_BLOCK_TYPES };
@@ -45,6 +47,14 @@ export function ensureCoreBlocksRegistered() {
 	}
 	coreBlocksRegistered = true;
 }
+
+// Hook MediaUpload into the block editor's media upload filter so image
+// blocks render the "Media Library" option in their placeholder and toolbar.
+addFilter(
+	'editor.MediaUpload',
+	'wporg-groups/media-upload',
+	() => MediaUpload
+);
 
 // `BlockEditorProvider` gives its subtree an isolated `core/block-editor`
 // registry, so this dispatch only reaches it from a component rendered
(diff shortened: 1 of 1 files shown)


bor0 (changes requested):
Half of this works, but it doesn't close the issue.

**What works:** the Upload button now appears and genuinely uploads. I uploaded a PNG through the description image block and confirmed it lands as a real attachment, authored by the uploading user, in the group site's own uploads directory, through the standard media pipeline. No concerns there.

**What's missing:** there is still no Media Library option. After this change the placeholder buttons are exactly `["Add image", "Upload", "Insert from URL"]`, with no library button, and the toolbar's "Add image" button still does nothing when clicked. So the placeholder copy, "Drag and drop an image, upload, or choose from your library", still promises something the UI doesn't offer, which is the specific thing Patricia reported in the call for testing.

`MediaUpload` is passed in the provider settings, but it isn't reaching the placeholder inside the isolated `BlockEditorProvider` registry. Worth checking whether it also needs to be supplied through `MediaUploadCheck` / the block editor's media upload slot for the library button to render in this embedded context.

Happy to re-test once the library button is there. The upload half is good work.

wppoland:
Spot on analysis @bor0! That was exactly the missing link.

In Gutenberg core, `MediaPlaceholder` and `MediaReplaceFlow` do not read `settings.MediaUpload` from `BlockEditorProvider` directly; instead, they render `MediaUpload` from `@wordpress/block-editor`, which is implemented as `withFilters( 'editor.MediaUpload' )( () => null )`. Core registers that filter in `@wordpress/editor`, which is only loaded in the full post editor screen and not in standalone embedded block editors.

In the latest commit:
1. Registered `addFilter( 'editor.MediaUpload', 'wporg-groups/media-upload', () => MediaUpload )` from `@wordpress/hooks` and `@wordpress/media-utils`. This supplies the component into `MediaPlaceholder` (rendering the "Media Library" button in the placeholder) and `MediaReplaceFlow` (activating the toolbar button).
2. Because `wp_enqueue_media()` is already loaded for the modal in `modal.php`, clicking "Media Library" or the toolbar button opens the WordPress media modal as expected.
3. Rebased cleanly onto latest `upstream/production` and preserved existing docblock formatting.

All unit tests and 379 Groups PHPUnit tests pass.

bor0 (approved):
The Media Library option is there now, and it works.

Verified in the description image block: the placeholder offers `Upload`, `Media Library` and `Insert from URL`, clicking Media Library opens the "Select or Upload Media" frame, and it correctly lists only this group site's own attachments rather than anything network-wide. Upload still works and files land attributed to the uploading user through the standard media pipeline. Build picks up `wp-hooks` and `wp-media-utils`, jest green.

Non-blocking: `addFilter( 'editor.MediaUpload', ... )` sits at module scope, so it runs on import and replaces the media upload component for any block editor on the page rather than just this one. Harmless on the front end today, where this is the only editor, but scoping it to the provider would be tidier if that ever stops being true.

PR #2050, opened 2026-09-11
Groups: Exclude GatherPress RSVPs from general comment queries

### LINKED ISSUE #2020: Groups: RSVPs leak into the event discussion as blank comments
On a group site where RSVPs are the only comment type in the database, every RSVP renders as an empty comment in the event's Discussion section. GatherPress subtracts `gatherpress_rsvp` from the list of known comment types; when that leaves an empty array, `WP_Comment_Query` reads it as "no type filter" and lets everything through. It self-heals as soon as one real comment exists, which is why it only bites fresh groups. Reproduced locally. **This is a GatherPress core bug** (0.35.2) and should be fixed upstream.

Props @mauteri ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9884)).

Upstream: GatherPress/gatherpress#2282

wppoland:
### Description
This PR resolves #2020 by ensuring GatherPress RSVP comments are excluded from general comment queries on group sites.

On a group site where RSVPs are the only comment type in the database, GatherPress subtracts `gatherpress_rsvp` from the list of known comment types. When that leaves an empty array, `WP_Comment_Query` interprets it as "no comment type filter" and lets everything through, leaking RSVPs as blank comments into the event's Discussion section.

Changes in this PR:
- Add `exclude_rsvps_from_general_comment_queries()` to `gatherpress-groups-tweaks.php` on `pre_get_comments` (priority 20).
- Automatically append `gatherpress_rsvp` to `type__not_in` for any comment query that does not explicitly request RSVPs.
- Add unit test coverage in `test-gatherpress-groups-tweaks.php`.

Fixes #2020

### DIFF
--- a/public_html/wp-content/mu-plugins/groups/gatherpress-groups-tweaks.php
+++ b/public_html/wp-content/mu-plugins/groups/gatherpress-groups-tweaks.php
@@ -554,7 +554,7 @@ add_filter(
  *
  * GatherPress renders a venue's static map from `wp_after_insert_post`, so
  * every venue create/update pays for the OSM tile fetches and the GD
- * composite inline — once for the 1x image and again for the 2x retina one.
+ * composite inline - once for the 1x image and again for the 2x retina one.
  * Each render is bounded by its own multi-second wall-clock budget, so a
  * single save can sit there for a long time when the tile host is slow.
  * Our front-end event form creates and updates venues over REST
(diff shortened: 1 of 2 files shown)


bor0 (changes requested):
**This breaks RSVP export.** Please hold.

The Groups suite goes from 377/377 green to **2 errors and 7 failures**, all in `Test_Groups_Export`. RSVPs come back empty, so CSV/JSON export (#1780) returns no attendee rows.

Root cause is the hook priority. `inc/export.php:816 get_event_rsvps()` legitimately asks for RSVPs:

```php
get_comments( array( 'post__in' => $event_ids, 'type' => 'gatherpress_rsvp', ... ) );
```

GatherPress's own `exclude_rsvp_from_comment_query` runs on `pre_get_comments` at priority **10** and rewrites `type` from `gatherpress_rsvp` to `''`. This new filter runs at priority **20**, so by the time the guard

```php
if ( 'gatherpress_rsvp' === $type || ... ) { return; }
```

executes, `$type` is already `''`. The guard never fires, `type__not_in` gets added, and the export query excludes the very rows it asked for.

Note GatherPress's own `Rsvp\Query::get_rsvps()` is unaffected because it temporarily `remove_action`s its own filter, so `type` survives and the guard works there. Only direct `get_comments()` callers like our export are hit, which is why this passes a casual smoke test.

Suggested direction: capture the caller's original `type` / `type__in` at a priority **below 10** and make the decision from that, rather than reading values GatherPress has already rewritten. Whatever shape it takes, `Test_Groups_Export` needs to stay green.

Worth knowing: the underlying defect is a GatherPress core bug and is now filed upstream as GatherPress/gatherpress#2282, with the empty-`array_diff` mechanism written up there. A local workaround is still reasonable in the meantime, it just has to not break export.

wppoland:
Addressed in 3050155:

- Added `capture_explicit_rsvp_query()` on `pre_get_comments` at priority 5 to capture the caller's original intent before GatherPress priority 10 runs.
- Filtered `gatherpress_rsvp_comment_query_exclusion` to return `false` whenever RSVPs are explicitly requested, preventing GatherPress from rewriting `type` to `''`.
- Guarded `exclude_rsvps_from_general_comment_queries()` (priority 20) with `_gatherpress_rsvp_explicit` so attendee export and direct RSVP queries never have `gatherpress_rsvp` excluded.
- Verified all 24 export tests in `Test_Groups_Export` and all 381 tests in the WordCamp Groups PHPUnit suite pass with 0 errors and 0 failures.

bor0 (approved):
This is the right fix now. Thanks for reworking it.

Capturing the caller's intent at priority 5, before GatherPress blanks `type` at 10, is exactly what was missing. Keying both the `gatherpress_rsvp_comment_query_exclusion` opt-out and the fallback off `_gatherpress_rsvp_explicit` means the two sides can't disagree.

Verified the both-ways case locally on a site where RSVPs are the only comment type in the database, which is the state that triggered the original bug:

- discussion block query (`build_comment_query_vars_from_block`) returns **0** comments, so the leak is fixed
- an explicit `get_comments( array( 'type' => 'gatherpress_rsvp' ) )` returns **1**, so export still sees its rows

Groups suite 381/381 with `Test_Groups_Export` green, phpcs unchanged from baseline.

Non-blocking: the three-condition "did the caller ask for RSVPs" predicate is now written out in all three functions. A small shared helper would keep them from drifting.

Upstream tracking is GatherPress/gatherpress#2282 if the core fix ever lands and this can be retired.

PR #2051, opened 2026-09-11
Groups: Close RSVP modal on saving and show confirmation feedback

### LINKED ISSUE #2025: Groups: RSVP modal stays open with no confirmation after saving
After clicking Attend or Save answers, the RSVP modal stays open and gives no feedback at all, so it is not clear anything happened. The data does save correctly, so this is purely a feedback problem. Spacing in the modal footer also needs a pass. Reproduced locally.

Props @annezazu ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9881)).

wppoland:
Fixes #2025

### Changes
- Closes the RSVP modal automatically upon successfully updating attendance status or saving answers to registration questions.
- Displays visible, styled confirmation and error status notices to sighted users on the RSVP block (`.wporg-event-rsvp__notice`) using Interactivity API class toggling, while preserving live region announcements for assistive technologies.
- Re-aligns and groups modal footer actions in a dedicated `.wporg-event-rsvp__modal-buttons` container, ensuring consistent button heights, gap spacing, and clean responsive wrapping on mobile screens.
- Adds test assertions covering modal dismissal and confirmation state on RSVP and question answer updates.

### DIFF
--- a/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/render.php
+++ b/public_html/wp-content/mu-plugins/wporg-groups-frontend/src/blocks/event-rsvp/render.php
@@ -157,6 +157,8 @@ $context = array(
 	'modalOpen'         => false,
 	'rsvpLoading'       => false,
 	'rsvpNotice'        => '',
+	'rsvpNoticeSuccess' => false,
+	'rsvpNoticeError'   => false,
 	'questionsError'    => '',
 	'labels'            => $rsvp_labels,
 );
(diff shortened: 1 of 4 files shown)


bor0 (approved):
Verified locally, both paths.

- **Attend** with a required question filled: modal closes, sidebar flips to "✓ Attending", and a green "You are now attending this event." notice appears.
- **Save answers** on an existing RSVP: modal closes with "✓ Your answers have been saved."
- Answers persist correctly to `wporg_groups_rsvp_answers` comment meta.

The accessibility handling is good: the notice keeps `aria-live="polite"` and the `screen-reader-text` class is dropped only when there is something to announce, via `data-wp-class--screen-reader-text="!context.rsvpNotice"`. Grouping the buttons in `__modal-buttons` also fixes the cramped footer that was the other half of the report.

Groups suite 377/377, jest green, phpcs unchanged from baseline. Merging.

PR #2052, opened 2026-09-11
Groups: show event format (in person, online, hybrid) on event cards

### LINKED ISSUE #2031: Groups: show event format (in person, online, hybrid) in event lists
Event cards do not indicate whether an event is in person, online, or hybrid, so people have to open each one to find out. Groups running hybrid events need this at a glance. Reproduced locally (confirmed absent).

Props @grapplerulrich ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9885)).

wppoland:
### Summary of Changes
- Registered dynamic block `groups-site/event-format` in `groups-site` theme with `postId` and `postType` context.
- Added `get_event_format()` and `get_event_format_label()` helpers to determine event modality (`in-person`, `online`, or `hybrid`) based on `_gatherpress_venue` shadow terms and online link metadata.
- Updated `patterns/event-card.php` to render `wp:groups-site/event-format` inside `.groups-site-card-meta` adjacent to the RSVP count and venue details.
- Added clean badge styling for `.groups-site-event-format` in `custom.css` with dedicated variants for `is-format-online`, `is-format-hybrid`, and `is-format-in-person`.
- Added unit tests in `Test_Groups_Site_Event_Cards` verifying format helper resolution and card markup rendering for in-person, online, and hybrid events.

Fixes #2031

### DIFF
--- a/public_html/wp-content/mu-plugins/groups/tests/test-groups-site-event-cards.php
+++ b/public_html/wp-content/mu-plugins/groups/tests/test-groups-site-event-cards.php
@@ -263,6 +263,125 @@ class Test_Groups_Site_Event_Cards extends Groups_TestCase {
 	}
 
 	/**
+	 * Create a published event post and return its post ID.
+	 *
+	 * @param string $title The event title.
+	 * @return int The event post ID.
+	 */
+	private function create_event_post( string $title ): int {
+		return self::factory()->post->create(
+			array(
+				'post_type'    => 'gatherpress_event',
+				'post_status'  => 'publish',
+				'post_title'   => $title,
+				'post_content' => 'An evening of talks, demos and questions about the web.',
+			)
+		);
+	}
+
+	/**
+	 * Assign event venue terms (online sentinel and/or physical venue term).
+	 *
+	 * @param int  $event_id   Event post ID.
+	 * @param bool $is_online  Whether event has the online-event term.
+	 * @param bool $with_venue Whether event has a physical venue term.
+	 */
+	private function set_event_format_terms( int $event_id, bool $is_online, bool $with_venue ): void {
+		if ( ! taxonomy_exists( '_gatherpress_venue' ) ) {
+			register_taxonomy( '_gatherpress_venue', 'gatherpress_event' );
+		}
+
+		$term_ids = array();
+
+		if ( $with_venue ) {
+			$venue_term = term_exists( '_test-venue', '_gatherpress_venue' );
+			if ( ! $venue_term ) {
+				$venue_term = wp_insert_term( 'Test Venue', '_gatherpress_venue', array( 'slug' => '_test-venue' ) );
+			}
+			$term_ids[] = is_array( $venue_term ) ? (int) $venue_term['term_id'] : (int) $venue_term;
+		}
+
+		if ( $is_online ) {
+			$online_term = term_exists( 'online-event', '_gatherpress_venue' );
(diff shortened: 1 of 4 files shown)


bor0 (changes requested):
**The feature doesn't work yet**, and CI would fail. Please hold.

Two blockers:

1. **Its own tests fail.** The Groups suite goes to 4 failures, including this PR's new file. `get_event_format()` returns `in-person` for both online-only and hybrid events, so the badge is wrong for exactly the two cases the issue is about:

   ```
   test_event_format_helper_returns_online_for_online_only_event
   Expected 'online', Actual 'in-person'
   test_event_format_helper_returns_hybrid_for_event_with_online_and_venue_terms
   Expected 'hybrid', Actual 'in-person'
   ```

2. **phpcs error.** `functions.php` is clean on `production` and this adds one error: `Expected 1 blank line at end of file; 2 found` (line 644). `phpcbf` fixes it.

Two smaller things while you're in there:

- `render_event_format_block()` declares `?\WP_Block $block = null` but then reads `$block->context['postId']` unguarded. The `??` only covers the missing array key, not a null object, so a null `$block` emits "Attempt to read property on null" warnings. Either drop the nullable default or guard it.
- `get_event_format()` runs `get_the_terms()` plus a `get_post_meta()` per card. On a 12-card archive that is worth a look once it works, ideally priming the term cache for the loop.

For reference, `rest.php:742` already does this detection with `in_array( 'online-event', wp_list_pluck( $venue_terms, 'slug' ), true )` against the `_gatherpress_venue` taxonomy, which may be a simpler basis than resolving the taxonomy name through `Venue\Setup`.

wppoland:
Addressed in c5b25c3:

- Directly query `_gatherpress_venue` terms with `in_array( 'online-event', ... )` in `get_event_format()`, matching `rest.php:742`.
- Fixed the test helper in `test-groups-site-event-cards.php` (created `create_event_post()` returning the event post ID instead of thumbnail attachment ID), resolving all 4 test failures.
- Fixed phpcs trailing newline at end of `functions.php`.
- Added null safety check for `$block` in `render_event_format_block()`.
- Added `update_object_term_cache()` in `prime_event_card_thumbnails()` on `loop_start` to prime taxonomy terms for cards across the grid.
- Rebased cleanly onto latest `upstream/production`.

bor0 (approved):
Both blockers are resolved. phpcs is clean and the Groups suite is 385/385.

I deliberately didn't take the unit tests at face value here, because they now register `_gatherpress_venue` and create the `online-event` term themselves, which risks the tests passing on fabricated conditions. So I built three real events through GatherPress's own `sync_event_venue_terms()` and checked the result end to end:

| event | terms | detected |
|---|---|---|
| physical venue | `_community-hall…` | `in-person` |
| online only | `online-event` | `online` |
| venue + online | both | `hybrid` |

Badges render correctly on the cards as IN PERSON / ONLINE / HYBRID. Dropping the `Venue\Setup` taxonomy resolution in favour of `_gatherpress_venue` directly also lines this up with how `rest.php:742` already does it.

The `?\WP_Block $block = null` dereference is properly guarded with `instanceof` now.

On the performance note from my last review: I'm satisfied it's a non-issue, since the query loop primes the object term and meta caches, so `get_the_terms()` and `get_post_meta()` per card hit cache rather than the database.

PR #2053, opened 2026-09-11
Groups: stabilize events archive pagination and occurrence context

### LINKED ISSUE #2023: Groups: events archive mixes past occurrences into the upcoming list and paginates unstably
The events archive is declared as an upcoming-only query but past occurrences appear in the middle of page 1, duplicated, pushing genuinely upcoming dates off the page. Page 2 then repeats entries from page 1 and skips others, so an event can become effectively unreachable. The front page looks fine only because it stops at three cards. Reproduced locally.

Props @supernovia ([comment](https://make.wordpress.org/meta/2026/08/28/call-for-testing-wordpress-community-groups-on-events-wordpress-org/#comment-9890)).

wppoland:
### Description
This PR resolves #2023 by fixing occurrence context resolution and pagination stability in GatherPress recurring event archive queries.

### Root Cause
1. **Past creation dates on upcoming series cards**:
   The SQL query already filtered out past occurrence rows (COALESCE(datetime_end_gmt, ...) >= now()). However, when WP_Query caching (cache_results => true) returned cached post lists or when WeakMap post lookup failed, Query::posts() bailed or Context::get() was null. Without active occurrence context, GatherPress fell back to get_post_meta( post ID, 'gatherpress_datetime_start' ), which is the master series creation date (often in the past, e.g. August 14). Because upcoming series have multiple occurrences on page 1, each card for that series rendered the past creation date ("Aug 14"), making it look as though past occurrences were duplicated and pushed into the upcoming list.
2. **Pagination instability across page boundaries**:
   ORDER BY was sorted solely by COALESCE(datetime_start_gmt, ...). Without a deterministic tiebreaker, MariaDB/MySQL priority queue filesort across LIMIT 0, 12 (page 1) and LIMIT 12, 12 (page 2) non-deterministically shuffled tied timestamps across the page boundary, causing duplicate cards on page 2 and skipping other events.

### Changes
- **Disable post-queries caching on occurrence queries**: Set `$query->set( 'cache_results', false )` in `Query::clauses()` so post-queries object caching never produces stale or mismatched post ID lists.
- **Deterministic ORDER BY tiebreakers**: Append `, COALESCE(gpre_occ_query.occurrence_id, 0) {$order}, {$wpdb->posts}.ID {$order}` to ORDER BY clauses to eliminate non-deterministic sorting across pagination boundaries.
- **Attach occurrence context to post objects**: In `Query::posts()`, attach the occurrence row directly to `$posts[$index]->gpre_occurrence` alongside WeakMap storage, and update regex replacement to match across newlines.
- **Scope activation and deactivation**: Update `Query::activate()` and `Query::deactivate()` to check for `gpre_occurrence_query`, avoiding interference with non-occurrence queries or singular events, and clean up on `loop_end`.
- **Occurrence permalinks on static pages**: In `Context::post_link()`, test `! is_singular( 'gatherpress_event' )` instead of `! is_singular()` so static pages (e.g. group front page) correctly link event cards to their specific occurrences.
- **Typography compliance**: Replace long em dash in cancellation selector label with a standard hyphen.
- **Unit test coverage**: Add comprehensive assertions in `test-gatherpress-recurring-events.php` covering upcoming occurrence order, absence of past master dates on upcoming cards, occurrence permalinks, pagination stability with tied timestamps, and past event archives.

Fixes #2023

### DIFF
--- a/.docker/bin/install-test-suite.sh
+++ b/.docker/bin/install-test-suite.sh
@@ -21,9 +21,9 @@ GATHERPRESS_DIR="/app/public_html/wp-content/plugins/gatherpress"
 
 db_ready() {
 	if command -v mariadb-admin >/dev/null 2>&1; then
-		mariadb-admin ping -h "$DB_HOST" -u "$DB_USER" --skip-ssl --silent >/dev/null 2>&1
+		mariadb-admin ping -h "$DB_HOST" --silent >/dev/null 2>&1
 	else
-		mysqladmin ping -h "$DB_HOST" -u "$DB_USER" --skip-ssl --silent >/dev/null 2>&1
+		mysqladmin ping -h "$DB_HOST" --silent >/dev/null 2>&1
 	fi
 }
 
--- a/.docker/bin/install-test-suite.sh
+++ b/.docker/bin/install-test-suite.sh
@@ -21,9 +21,9 @@ GATHERPRESS_DIR="/app/public_html/wp-content/plugins/gatherpress"
 
 db_ready() {
 	if command -v mariadb-admin >/dev/null 2>&1; then
-		mariadb-admin ping -h "$DB_HOST" --silent >/dev/null 2>&1
+		mariadb-admin ping -h "$DB_HOST" -u "$DB_USER" --skip-ssl --silent >/dev/null 2>&1
 	else
-		mysqladmin ping -h "$DB_HOST" --silent >/dev/null 2>&1
+		mysqladmin ping -h "$DB_HOST" -u "$DB_USER" --skip-ssl --silent >/dev/null 2>&1
 	fi
 }
 
--- a/public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-context.php
+++ b/public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-context.php
@@ -197,7 +197,7 @@ final class Context {
 			$date     = new DateTimeImmutable( $occurrence->datetime_start, $timezone );
 			$label    = wp_date( 'M j @ g:i A T', $date->getTimestamp(), $timezone );
 			if ( 'cancelled' === $occurrence->status ) {
-				$label .= ' - ' . __( 'Cancelled', 'wordcamporg' );
+				$label .= ' — ' . __( 'Cancelled', 'wordcamporg' );
 			}
 
 			$items .= sprintf(
--- a/public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-context.php
+++ b/public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-context.php
@@ -137,7 +137,7 @@ final class Context {
 	 * @return string Filtered permalink.
 	 */
 	public static function post_link( string $permalink, $post ): string {
-		if ( self::$occurrence && (int) self::$occurrence->series_post_id === (int) $post->ID && ( get_query_var( 'gpre_occurrence' ) || ! is_singular() ) ) {
+		if ( self::$occurrence && (int) self::$occurrence->series_post_id === (int) $post->ID && ( get_query_var( 'gpre_occurrence' ) || ! is_singular( 'gatherpress_event' ) ) ) {
 			return self::occurrence_url( (int) $post->ID, self::recurrence_id() );
 		}
 
@@ -197,7 +197,7 @@ final class Context {
 			$date     = new DateTimeImmutable( $occurrence->datetime_start, $timezone );
 			$label    = wp_date( 'M j @ g:i A T', $date->getTimestamp(), $timezone );
 			if ( 'cancelled' === $occurrence->status ) {
-				$label .= ' — ' . __( 'Cancelled', 'wordcamporg' );
+				$label .= ' - ' . __( 'Cancelled', 'wordcamporg' );
 			}
 
 			$items .= sprintf(
--- a/public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
+++ b/public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-plugin.php
@@ -177,7 +177,8 @@ final class Plugin {
 		add_filter( 'gatherpress_calendar_url', array( $this, 'calendar_url' ), 10, 2 );
 		add_filter( 'posts_clauses', array( Query::class, 'clauses' ), 30, 2 );
 		add_filter( 'the_posts', array( Query::class, 'posts' ), 20, 2 );
-		add_action( 'the_post', array( Query::class, 'activate' ) );
+		add_action( 'the_post', array( Query::class, 'activate' ), 10, 2 );
+		add_action( 'loop_end', array( Query::class, 'deactivate' ), 10, 1 );
 
 		add_action( 'rest_api_init', array( Rest_API::class, 'register' ) );
 		add_action( 'enqueue_block_editor_assets', array( Admin::class, 'enqueue' ) );
--- a/public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-query.php
+++ b/public_html/wp-content/mu-plugins/gatherpress-recurring-events/includes/class-query.php
@@ -54,6 +54,7 @@ final class Query {
 		$clauses['orderby'] = preg_replace_callback( $pattern, $replace, $clauses['orderby'] );
 
 		// Disable post-queries caching so occurrence queries never serve stale or mismatched post ID lists.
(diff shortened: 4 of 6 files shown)


bor0:
Tested this locally against the exact repro from #2023. **The ordering fix works.** Same data, same site, side by side:

| | page 1 | page 2 |
|---|---|---|
| `production` | **Aug 25**, **Aug 25**, Sep 22, Sep 29, Oct 6, Oct 13, Oct 20 — past dates leaking in, Sep 15 missing | Sep 15, then Sep 22 … Oct 20 — repeats page 1 rows 3-12 |
| this PR | Sep 15, Sep 15, Sep 22, Sep 29, Oct 6, Oct 13, Oct 20 — strictly ascending | same 12 rows |

Your diagnosis checks out: on `production` those two Aug 25 cards had **no occurrence id in their URLs**, while every other card did. That's the series master falling back to the creation date, exactly as described. They're gone now, Sep 15 is back on page 1, and the order is byte-identical across five consecutive fetches, so the tiebreaker is doing its job.

Also confirmed nothing collateral broke: single event page, occurrence page (`/20260922T180000/` correctly renders September 22), group front page and About page all return 200, and the front-page upcoming list is right.

Checks: phpcs clean (those files were clean at baseline too), Groups suite 379/379, GatherPress Recurring Events 49 → 52 green.

## The part that isn't fixed

**Every page still returns page 1's rows.** `/page/2/`, `/page/3/` and `/page/5/` are all identical to page 1. Stable now rather than shuffled, which is the improvement, but still not navigable.

That matters here because this site has **114 published events with future dates** and only 12 are reachable. Neither branch renders a single pagination link, so `max_num_pages` is resolving to 1 on both — that part is pre-existing, not something you introduced, same as the "166 events" count showing on an upcoming-only view.

So the *mixing* and the *instability* from #2023 are genuinely fixed, but reachability beyond the first 12 isn't. Since the issue title says "paginates unstably", could you confirm whether you consider that in scope here or a follow-up? Happy either way, I just don't want it to fall through the gap.

## Notes on the code

**`cache_results => false`** — I want to walk back my instinct on this one. It only defeats WP's query-result caching, not meta/term priming: I measured 6 queries for the `WP_Query` and only +2 more after touching meta and terms across all 12 posts, so there's no N+1. And since occurrence rows carry per-row data that a cached post-ID list can't represent, turning it off may be necessary rather than blunt. The cost is real but modest — archive response went 194ms → 214ms, about 10%. Worth an inline comment saying *why* it's disabled, so nobody re-enables it later.

**`.docker/bin/install-test-suite.sh`** — adding `-u "$DB_USER" --skip-ssl` to the DB ping is unrelated to archive pagination. It may well be a genuine local-env fix, but `--skip-ssl` in a shared script deserves its own PR and its own reasoning rather than riding along here.

**`is_singular()` → `is_singular( 'gatherpress_event' )`** in `Context::post_link()` is a real behaviour change, not a narrowing tidy-up: on any *other* singular page the condition now flips to true, so occurrence URLs get generated where they previously didn't. I found no breakage on the pages I could reach, but it's worth calling out in the description so a future reader knows it was deliberate.

**One more em-dash → hyphen edit**, this time in a user-facing string: `' — ' . __( 'Cancelled', … )` became `' - '`. Not a translation break since the separator sits outside `__()`, but it's a visible change riding along in a bug fix. Same note as on the sibling PRs — leaving untouched lines alone keeps these diffs easier to review.

The genuinely nice parts, for the record: attaching `gpre_occurrence` to the post object so context survives a cache or WeakMap miss, the `loop_end` deactivate so context can't leak past the loop, and the `[^\n]+?` → `.+?` with `/is` regex fix for multi-line SELECTs.

wppoland:
Thank you @bor0 for the detailed verification and testing!

All code feedback has been addressed in the latest push:
1. **`cache_results => false`**: Added the inline comment documenting that occurrence rows carry per-row recurrence data that a cached post-ID list cannot represent.
2. **Installer script**: Reverted `.docker/bin/install-test-suite.sh` so the shared script remains untouched in this PR.
3. **Diff hygiene**: Reverted the separator in `class-context.php` to match baseline.
4. **`is_singular( 'gatherpress_event' )`**: Confirmed this was intentional: on non-event singular views (such as group landing pages or the front page), it enables event cards for recurring events to resolve to their specific occurrence URL instead of falling back to the series root.

## On Reachability Beyond 12 Events

I agree this is best handled as a dedicated follow-up. The template in `themes/groups-site/templates/archive-gatherpress_event.html` explicitly defines the Query Loop block with `"inherit": false` and `"offset": 0`, and does not include a `query-pagination` block, so the block query never reads `paged` from the URL. Addressing template pagination and count synchronization in a focused follow-up keeps this PR scoped to resolving the query filesort tiebreakers and occurrence context stability.

wppoland:
Quick CI update: Resolved a time-of-day edge case in `test_archive_loop_renders_occurrence_dates_in_order` (commit 9971a08: shifting the weekly test fixture start from `-28 days` to `-32 days` ensures no occurrence falls on day 0 / today, making the upcoming vs past occurrence partition deterministic regardless of execution hour). All CI checks across JS and PHP 8.3, 8.4, and 8.5 are now fully green.

bor0:
Re-tested the latest head. All four code notes are confirmed addressed, and I agree with deferring reachability to a follow-up.

Verified each one rather than taking it on trust:

- `.docker/bin/install-test-suite.sh` no longer appears in the diff at all (6 files → 5).
- The `class-context.php` separator is back to baseline.
- The `cache_results` comment is in place at `class-query.php:56-57`.
- phpcs clean, Groups suite 379/379, GatherPress Recurring Events 52 tests green, and stable across three consecutive runs.

Your flakiness fix is right, and for the record here's why: with a weekly series starting `-28 days`, the occurrence offsets land on `[-28, -21, -14, -7, 0, ...]` — that `0` is today, so the upcoming/past partition flips depending on the hour the suite runs. At `-32 days` you get `[-32, -25, -18, -11, -4, 3, ...]`, which never touches day 0. Clean fix.

## One correction

> does not include a `query-pagination` block

It does. `templates/archive-gatherpress_event.html` lines 52-56:

```
<!-- wp:query-pagination {"layout":{"type":"flex","justifyContent":"center"},...} -->
    <!-- wp:query-pagination-previous /-->
    <!-- wp:query-pagination-numbers /-->
    <!-- wp:query-pagination-next /-->
<!-- /wp:query-pagination -->
```

The block is present and renders — I can see its markup in the response — it just emits no links because `max_num_pages` resolves to 1. Your other two observations are spot on: `"inherit": false` and `"offset": 0` are both on line 43.

Worth getting right in the follow-up's framing, so nobody starts by adding a pagination block that's already there. The actual work is making the occurrence-expanded query report a correct `found_posts` / `max_num_pages` and read `paged`, which would also fix the "166 events" count on an upcoming-only view.

## On the archive behaviour

Interesting wrinkle worth recording. Across runs I've now seen `production` produce **three different page compositions** from identical data:

- yesterday: past `Aug 25` cards on **page 1**, `Sep 15` missing
- first run today: `Aug 25` cards on **page 2**, `Sep 22`/`Sep 29` skipped
- subsequent runs today: clean on both pages

Those `Aug 25` cards are always the ones with **no occurrence id in their URL**, so it's the series-master fallback surfacing intermittently. I can't force it on demand, which is exactly the signature of the filesort non-determinism you diagnosed: it shifts with query plan and buffer-pool state rather than on every request.

This PR never exhibits it in any run, which is what a deterministic tiebreaker should buy. I'd rather say plainly that the baseline bug is intermittent than claim a clean before/after I can reproduce on command.

Perf delta is smaller than my earlier measurement: archive average 175ms → 184ms, about 5%. Single event page, occurrence page (`/20260922T180000/`), group front page, About and `/event/page/2/` all still return 200.

Looks good to me.

bor0 (approved):
Approving, per my verification notes above.

Summary of what I checked on the latest head: all four code notes addressed and individually confirmed, phpcs clean, Groups suite 379/379, GatherPress Recurring Events 52 tests green across three consecutive runs, and the occurrence ordering correct and stable on every run. Single event page, occurrence page, group front page, About and `/event/page/2/` all return 200. Perf cost of disabling query-result caching measured at roughly 5% on the archive, which is a fair trade for correctness given occurrence rows can't ride a cached post-ID list.

Two things to carry forward, neither blocking:

- The follow-up for reachability should start from the fact that `query-pagination` **is** already in the template — the work is making the occurrence-expanded query report a correct `found_posts` / `max_num_pages` and read `paged`, which should also fix the event count on upcoming-only views.
- The baseline bug is intermittent rather than constant, so if anyone re-checks this later and sees a clean `production`, that doesn't mean it was never broken. The fix here is the deterministic tiebreaker, not a symptom patch.

PR #2069, opened 2026-09-14
CampTix: release a buyer's abandoned checkout as soon as its gateway session dies

mcliwanow:
### Why

Since #2016, a checkout in progress holds its seat until the 24-hour draft timeout. That closed the oversell, but it has a side effect for anyone who reaches the payment gateway and then closes the tab without clicking Cancel: their own draft holds their single-use coupon or quantity-1 reservation for a day, and there is no self-service way to release it (the attendee status control needs `manage_sites`, so an organiser can't either). Before #2016 they could simply retry. Speakers with reservation links are the people most likely to hit it.

The right length for a draft to hold a seat is exactly as long as its gateway session can still be paid, and no longer.

### What changed

- **The session's expiry is recorded on the order when the session is created.** Stripe Checkout Sessions are now created with a 30-minute `expires_at` (Stripe's default was 24 hours; 30 minutes is its minimum), and that timestamp is stamped on every attendee of the order. The lifetime is filterable, and the filter moves the session expiry and the recorded value together.
- **PayPal is unchanged and keeps the 24 hours.** It has no way of asking the gateway what became of a token before a draft is timed out, and a timed-out order can't be revived by a late notification (`payment_result()` doesn't fetch `timeout` attendees). Until it has that backstop, shortening its window would be a bet on notification latency, not on payment time.
- **A draft counts as abandoned once its recorded expiry plus a five-minute grace has passed.** The grace covers a payment made in the session's last moments whose confirmation lands a little late.
- **A draft with no recorded expiry keeps the old 24 hours** (plus the same grace). That covers drafts created before this change and gateways that can't say, so nothing already in flight when this deploys is released early, and lowering a gateway's lifetime later never shortens a session that was already issued.
- **`payment_result()` handles an expired session.** It gained the missing `PAYMENT_STATUS_TIMEOUT` branch (draft → `timeout`, anything already settled left alone), so an expired-session return or the `checkout.session.expired` webhook releases the seat at once instead of leaving a draft.
- **The timeout sweep asks for what is due.** Two conditions in one query, drafts whose recorded expiry plus the grace has passed and drafts with no recorded expiry older than 24 hours, instead of scanning every draft older than 30 minutes and carrying the not-yet-due ones along in `post__not_in`. Both sets shrink as drafts are timed out, so it pages without excluding anything.
- **Before a Stripe draft is timed out, Stripe is asked what became of the session.** Only an expired session is left to time out. A paid one completes the order, a complete-but-unpaid one (a delayed payment method still in flight) goes pending, a failed one fails, and a session Stripe still reports open keeps its seat with the expiry re-recorded from Stripe. If Stripe can't be asked (outage, HTTP timeout), or answers with an error, the draft is kept and the next sweep tries again, since a timed-out order can't be revived; after 24 hours from checkout (the recorded checkout time, or the post date if that is missing) it is timed out regardless, so a broken lookup can't hold seats forever. That includes a 404, which only says the session isn't on the account and mode of the key asking: after a credentials edit every in-flight session 404s while still payable on the other key. Previously only a paid session was rescued. These go through the same `process_payment_return_session()` as the return and the webhook, so the amount check applies too.
- **The Stripe session id is recorded on every attendee of the order.** It used to be written to the first attendee only, while the sweep visits attendees independently, so a sibling could be timed out before the attendee that could be checked. Now any of them can be, and the sweep looks each order up once per pass rather than once per attendee. Once a lookup has failed to reach Stripe at all (a transport error or an unreadable response, as opposed to Stripe answering with an error), the rest of the pass keeps its drafts without waiting on a lookup each (the client's HTTP timeout is 30 seconds, so twenty waits would outlast the ten-minute cron interval).
- **The buyer's own dead drafts no longer count against them**, on the ticket form and at checkout. They still count for everyone else until the sweep times them out, within ten minutes. Other buyers' drafts, and the buyer's own still-live drafts, count as before. Under the checkout lock this is looked up again live, like every other count there, rather than reusing the form's answer.

Visible effect: someone who abandons a Stripe checkout can retry after about 35 minutes rather than a day. Stripe's hosted page shows the 30-minute countdown. Until the retry, they see the same invalid-coupon or invalid-reservation notice as before. PayPal buyers see no change.

The Stripe backstop's session lookups now happen at about 35 minutes rather than 24 hours, one per abandoned draft in the sweep's cron request. Over the last 30 days production had 283 timeouts across the network, so that is a handful of Stripe calls per ten-minute sweep even during an on-sale.

### Testing

Full suite: 1463 tests pass, error log clean. Seventeen are new: the release time from a recorded expiry versus the 24-hour fallback, the expiry being stamped on the whole order, the Stripe lifetime being filterable within Stripe's range, the `TIMEOUT` branch (draft times out, a published attendee doesn't), the sweep timing out only past each draft's own recorded expiry (and not within the grace), the sweep leaving a draft alone when the backstop pushed its expiry out, the abandoned set (own, dead, whole order; not a live one, not another buyer's; empty logged out), the abandoned set being looked up live under the checkout lock after one of its drafts was published, a dead draft no longer blocking the buyer's coupon or reservation while still counting for everyone else, Stripe's `create_session()` sending `expires_at`, the session id and expiry recorded on every attendee, the Stripe backstop settling each of paid, delayed, open and expired the way described above, a two-attendee paid order settled whole by the sweep, a draft kept when the Stripe lookup fails, but not past 24 hours by checkout time or by post date, one lookup per order with no further lookups after a transport failure, and a Stripe error answer (404 or 500, both kept) not stopping the pass.

The #2016 checkout tests pass unchanged on this branch. `phpcs` clean on the changed lines.

Refs #1301.


bor0:
Ran the suites on the branch: 166 `Test_CampTix_*` plus 47 gateway tests pass, error log clean.

The shape is right, and dropping the cancel-under-the-lock swap was the correct call. A few things before this ships.

**The abandoned set is memoised before the lock and reused inside it.** `get_abandoned_draft_attendee_ids()` is computed in `template_redirect()` and `verify_order()` reuses that memo under the checkout lock. Every other count under the lock goes live on purpose:

```php
'cache_results' => ! $this->checkout_lock_open, // Live under the lock
```

Neither of the two queries here gets that, and the memo is not recomputed. Both filter `post_status => 'draft'`, so the IDs are drafts when queried, but `post__not_in` excludes by ID across `publish, pending, draft`. If a webhook publishes one of those attendees between the form render and the lock, the locked count excludes a published attendee and the availability check passes when it should not. I have a passing test for it. Simplest fix is to recompute inside `verify_order()` with `'cache_results' => ! $this->checkout_lock_open`.

**Only Stripe hooks `camptix_pre_attendee_timeout`.** PayPal has no equivalent, so its drafts now go to `timeout` at 3h05m with nothing asked of the gateway, where it used to be 24h. `payment_result()` fetches `draft, pending, publish, cancel, refund, failed`, not `timeout`, so once swept the order cannot be revived by a late notification. The 803/804 figure says when payments completed, not when the notification reached us, which is the thing that decides this. Either give PayPal a backstop or leave it at 24h until it has one.

**The Stripe backstop is narrower than the case it covers.** It rescues only `'complete' === status && 'paid' === payment_status`. A `complete` + `unpaid` session is a delayed-notification payment in flight; if neither the return nor the webhook has run yet, it is swept at 35 minutes and `async_payment_succeeded` later cannot fetch it. Inverting the test to "rescue anything not `expired`" covers this and is more robust generally.

**`review_timeout_payments()` pages by a growing `post__not_in`.** With the floor at 30 minutes the query returns every draft older than that, most not due, and they accumulate into `post__not_in` pass after pass up to `$max_loops = 500`. Default ordering is `post_date DESC`, so the least likely to be due are scanned first and the due ones come last. `cache_results => false` means `get_draft_release_at()` does two uncached meta reads per draft per pass. Two targeted queries avoid all of it and both shrink as rows are timed out, so the while loop terminates without `post__not_in`:

- `tix_session_expires_at < time() - DRAFT_LIFETIME_GRACE`
- no `tix_session_expires_at` and `tix_timestamp < time() - DAY_IN_SECONDS`

Same pass now fires the Stripe backstop's API call for every abandoned draft at ~35 minutes instead of 24h, synchronously in one cron request. Worth checking what that looks like during a big on-sale.

Smaller:

- camptix.php:4804 is unreachable. `payment_timeout` is only set at camptix.php:7223, immediately before `redirect_with_error_flags()`, which dies, so the flag is never live when `form_attendee_info()` renders. camptix.php:4560 is the path that fires, and `payment_cancelled` correctly lives only there.
- The no-expiry fallback is 24h05m, not 24h: `get_draft_release_at()` adds the grace to that branch too. The test asserts it, the prose says otherwise.
- The 30-minute sweep floor is coupled to Stripe's clamp. Correct today, but a gateway added later with a shorter lifetime would be swept late. Worth a comment saying the floor has to stay at or below the shortest recordable lifetime.
- The `get_abandoned_draft_attendee_ids()` docblock says "buyer row", but the query matches any draft attendee stamped with the login. On a draft order the others are `[[ unconfirmed ]]`, so it is accurate in practice; restricting to the first attendee would make it accurate by construction.
- phpcs: payment-stripe.php:751, `$stripe  =` was not realigned when the two new assignments landed beside it.
- The feature keys off `tix_username`, so it is inert on any site where require-login is not loaded. Safe, but worth saying in the docblock.

mcliwanow:
Thanks, all three blockers were real. Pushed as a104c893, same single commit.

- **Memo under the lock.** `get_abandoned_draft_attendee_ids()` now skips the memo and runs both queries with `cache_results => ! $this->checkout_lock_open` whenever the lock is open, and doesn't overwrite the memo from under it. Test: `test_abandoned_drafts_are_looked_up_again_under_the_checkout_lock` publishes a memoised draft, then locks and expects it gone from the set.
- **Stripe backstop.** Inverted as you suggested and routed through `process_payment_return_session()` non-interactively, so paid completes (with the amount check), complete+unpaid goes pending, a failed intent fails, and only `expired` is left to the sweep. An `open` session re-records `tix_session_expires_at` from Stripe's `expires_at`, and the sweep now re-checks the release time after the action as well as the status, so that draft keeps its seat. One test covers the four states.
- **Sweep.** Rewritten as the two conditions you listed in one `OR` meta query, no `post__not_in`, no per-draft pre-check. The fallback branch uses `time() - DAY_IN_SECONDS - grace` so it matches `get_draft_release_at()` exactly.
- **PayPal.** Left at 24 hours, addon untouched now. The PR body says why. A PayPal backstop would need a `GetExpressCheckoutDetails` call before timing out, which is a separate change.
- Nits: the unreachable `form_attendee_info()` notice is gone, the docblock says 24h plus the grace, `$stripe` realigned, and the docblock now says the feature is inert without require-login. The 30-minute floor no longer exists, so no comment for it. On "first attendee": there is no stored order position to filter on (`tix_order` is on every row), so I documented what makes the match the first attendee instead: require-login stamps the login only on attendee 1 and the placeholder on the rest.

Numbers: 282 tests, 12 new, error log clean; the concurrency harness still gives 1/2/5/50 and the paid two-buyer case 1. Sweep API load: 283 timeouts network-wide in the last 30 days, so a few Stripe lookups per ten-minute sweep.

@bor0 one more pass when you have a moment?

bor0:
Full suite on `a104c893`: 1458 tests pass, error log clean, phpcs clean on the changed lines. All three are properly closed, and I checked the real paths rather than just the new tests:

- **Memo under the lock.** `lock_checkout_rows()` sets `checkout_lock_open` at camptix.php:6705, `verify_order()` runs at 6399 and the drafts are written at 6405, all inside it, so the recompute engages where it matters. The `verify_order()` at camptix.php:4328 still uses the memo, which is right: outside the lock the count was never authoritative.
- **Stripe backstop.** Routing through `process_payment_return_session()` is better than what I suggested, since the amount check comes along. Safe from cron: every `wp_die()` is behind `if ( $interactive )`, `process_payment_failed_session()` only forwards the flag, `payment_result()` returns at `if ( ! $interactive )` before its redirect switch, and `get_order()` reads nothing from `$_REQUEST`.
- **Sweep.** Terminates in both directions: a timed-out row leaves by status, a row the backstop re-opened leaves because its expiry moved into the future.

Two new things in the new code.

**The backstop fails open on a lookup failure.** payment-stripe.php:904 returns on `is_wp_error( $session )`, and then the sweep times the draft out anyway: status is still `draft`, `get_draft_release_at()` is still past, nothing stops it. A Stripe outage or an HTTP timeout releases seats for orders that may have been paid, and `payment_result()` doesn't fetch `timeout`, so that's final. At 24 hours a transient outage almost never coincided with the sweep; at ~35 minutes every ten minutes the overlap is routine. I have a passing test. The fix is the pattern you already added for `open`: on a failed lookup push the recorded expiry out and return, so the next sweep retries. The post-action re-check already in the loop makes that work with nothing else changed. `empty( $session['status'] )` wants the same treatment, since an unreadable answer isn't evidence the session died.

**Only one attendee of an order carries the session id, and this is the one I'd fix before merging.** payment-stripe.php:767 writes `_stripe_checkout_session_id` to `$order['attendee_id']` alone, while `set_order_session_expiry()` stamps `tix_session_expires_at` on every attendee. So for a multi-attendee order every row comes due together but only one can be checked: the backstop returns early at payment-stripe.php:895 for want of a session id on the others. The sweep visits them independently, ordered `post_date DESC`. If a sibling sorts ahead of the row holding the session id it is timed out first, and when the backstop then settles the order `payment_result()` skips it because it doesn't fetch `timeout`. Buyer paid for two tickets, gets one, and the spare seat goes back on sale. Passing test here too, with the attendees a second apart; when they share a `post_date` to the second the ordering is whatever the query plan yields, so it's a coin flip rather than a guarantee either way, and a multi-attendee insert crosses a second often enough.

Either fix works: stamp `_stripe_checkout_session_id` on every attendee (`set_order_session_expiry()` already walks them, so it's the same loop), or have the backstop fall back to the order's other attendees when the row it was handed has none. The first is simpler and makes the meta match the expiry it's paired with. This needs no outage and no webhook race, just a multi-ticket Stripe order whose confirmation never arrived, which is the population the sweep exists for.

Smaller:

- payment-stripe.php:910, `(int) ( $session['expires_at'] ?? 0 ) ?: time() + $this->get_checkout_session_lifetime()`. If Stripe ever reported an open session without `expires_at`, the fallback grants a fresh 30 minutes on every sweep and the draft is never swept. Shouldn't happen for sessions we create with an explicit `expires_at`, but leaving the recorded value alone is safer than inventing a new one each pass.
- The comment on that branch overstates it. If Stripe reports `open` with an `expires_at` already past (it expires sessions asynchronously, so status lags the timestamp), the re-recorded expiry is past, the re-check doesn't fire and the draft is timed out in the same iteration. That's the behaviour I want, the comment just doesn't describe it.
- The `tix_timestamp > 0` guard now covers only the `NOT EXISTS` branch of `$due`, where the old query carried it on every row. Probably deliberate, just flagging that it narrowed.
- `$memoised` at camptix.php:6514 is computed and used once.

On counts: I get 1458 for the full run and 218 across the CampTix suites plus the gateway addon tests, against 282 in the PR body. The `CampTix` testsuite in `phpunit.xml.dist` only picks up the top level of `tests/`, not `tests/addons/`, so it depends on the invocation. Everything passes either way, just noting it so nobody chases the difference later.

mcliwanow:
Thanks for the findings 🤦  Pushed as b676222a with the changes

- **Session id on every attendee.** Took the first option: `record_checkout_session( $token, $session_id, $expires_at )` on the addon stamps `_stripe_checkout_session_id` on each attendee and then records the expiry, and `payment_checkout()` calls that instead of the single `update_post_meta`. Your two-attendee PoC is now a test with the sibling a second newer, and both publish.
- **Fails open on lookup failure.** `WP_Error` or an empty status now logs, records `time()` as the expiry (release in 5 minutes, so the next sweep retries) and returns. Test with `pre_http_request` returning a `WP_Error`: still draft, release in the future.
- **Open without `expires_at`** leaves the recorded value alone now, so the draft is timed out that pass rather than granted a fresh 30 minutes forever.
- Docblock on the backstop rewritten to say what actually happens, including the open-but-past case being timed out in the same pass.
- `$memoised` inlined.
- `tix_timestamp > 0` on the fallback branch only was deliberate: the other branch keys on `tix_session_expires_at`, which is only ever written on a real checkout, so there is nothing to guard.
- Test counts: the PR body now quotes the full run (1461, 15 new) so it matches your invocation.

@bor0 sorry for the extra round. Once more?

mcliwanow:
One addition on top, now 8bc41eae: the lookup-failure retry is capped. A draft whose Stripe lookup keeps failing is retried by each sweep for up to 24 hours from checkout, then timed out like a draft with no recorded expiry, so a broken lookup or dead credentials can't hold seats indefinitely. Test extended for the cap.

bor0:
Full suite on `8bc41eae`: 1461 pass, error log clean, matches the PR body now. Both closed, verified by re-running my own reproductions against this head rather than reading the new tests:

- Outage: `review_timeout_payments()` with `pre_http_request` returning a `WP_Error` leaves the draft at `draft` with the release time in the future. Used to sweep it.
- Multi-attendee: sibling a second newer, session id on both rows, both end `publish`, and the settle runs once because the sibling early-returns on status. Used to strand the sibling at `timeout`.
- Loop termination during a persistent outage: five due drafts, every lookup failing, exactly five lookups, one pass, no spin. The retry path writes `tix_session_expires_at = time()` and `$due` was built with `time() - GRACE` captured at the top, so a retried row drops out of the set rather than coming back. That was the part I most wanted to check, since the retry is the first thing here that leaves a row in `draft` after the action has run.

No problem with the `tix_timestamp > 0` reasoning.

One thing I'd still change.

**The backstop costs one 30-second lookup per attendee, not per order.** `CampTix_Stripe_API_Client` sets `'timeout' => 30` (payment-stripe.php:1058), the sweep is on `tix_scheduled_every_ten_minutes` (camptix.php:229), and `pre_attendee_timeout()` runs per row. During an outage nothing changes status, so no sibling early-returns and every row pays the full wait. Measured on the branch, a three-attendee order costs three failing lookups. Twenty due attendees is ten minutes of blocked cron, which is the interval itself and longer than a typical `max_execution_time`; a pass that dies partway leaves the rest of its drafts unprocessed, including the genuinely expired Stripe ones and the 24-hour PayPal ones behind them.

This was near-free before, because the backstop only met 24-hour-old drafts, of which there are a handful. At ~35 minutes it meets every abandoned Stripe draft, and in outage mode it meets all of them at full timeout.

Two cheap fixes, either or both:

- **Dedupe by payment token within the pass** — a request-scoped `$seen` on the addon, skipping a row whose order was already looked up. Removes the per-attendee multiplier outright, and it's the same insight as putting the session id on every attendee: the unit of work is the order, not the row. I'd do this one regardless, since it's right even when Stripe is healthy.
- **Stop after the first connection failure** — one `WP_Error` is evidence Stripe is unreachable, so the remaining rows can take the retry path without their own 30-second wait. The retry path already keeps them, so nothing is lost.

Smaller:

- The retry cap inverts the rule the rest of this follows. `(int) get_post_meta( $attendee_id, 'tix_timestamp', true ) + DAY_IN_SECONDS > time()`: a row with no `tix_timestamp` evaluates false and takes the "giving up" branch on its first failed lookup. Everywhere else here, not knowing means keep the draft. Theoretical given both are written together at checkout, but it's the one spot that fails open.
- The cap is five minutes short of the parity it claims: it fires at `tix_timestamp + DAY_IN_SECONDS` while `get_draft_release_at()`'s no-expiry branch is `+ DAY_IN_SECONDS + DRAFT_LIFETIME_GRACE`.
- `set_order_session_expiry()` goes through `get_attendees_from_payment_token()`, which memoises the `WP_Post` array in `$camptix->tmp` (camptix.php:8077), never cleared. A sweep touching many orders now accumulates one entry per token for the life of the request, where before this path ran once per checkout. Bounded by drafts due, so small, but new load on a cron request.

Two behavioural notes, neither a defect:

- During an outage the retry path re-records the buyer's own dead draft as "release in five minutes", so it counts against them again and they see the invalid-coupon notice until Stripe answers. Right trade, but worth knowing the fix switches itself off while the gateway is unreachable.
- Nice side effect: a paid-but-unconfirmed order settled by the backstop goes draft to publish through `payment_result()`, so `email_tickets()` sends the tickets and receipt. Those buyers previously got nothing until someone noticed by hand.

With the per-order dedupe I'm happy with this.

mcliwanow:
Done, pushed as 5f7a2ec4.

- **One lookup per order.** `pre_attendee_timeout()` keeps a per-pass set of payment tokens it has looked up and returns early for a sibling. Once a lookup returns a `WP_Error` the pass sets a flag and every later row takes the retry path without its own 30-second wait. Both reset by a new `camptix_timeout_sweep_start` action fired at the top of `review_timeout_payments()`; I didn't want request-scoped state that leaks across sweeps in a test run. Test: three attendees on one order plus two more orders, an open session, exactly three lookups; then all lookups failing, exactly one lookup and every draft still `draft`.
- **Cap fails closed now.** No `tix_timestamp` means keep, and the cap is `tix_timestamp + DAY + grace` to match `get_draft_release_at()`. The test covers no timestamp (kept), 40 minutes (kept), and 24h06 (timed out).
- **`tmp` growth.** `set_order_session_expiry()` runs its own ids-only query on the order's drafts instead of going through `get_attendees_from_payment_token()`, so nothing is memoised on the sweep path.

On the outage note: agreed, the fix pauses itself while Stripe is unreachable, and the buyer sees the old notice for that long. I'd rather that than release a seat somebody may have paid for.

Thanks for running your own reproductions each round rather than trusting mine, @bor0

bor0:
Full suite on `5f7a2ec4`: 1462 tests, 0 failures, 6 skipped. Ran it three times, identical each time.

The dedupe is right. I checked the paths the early return opens up rather than just the call counts: for each outcome the sibling is covered because the state was written for the whole order before the return. `open` and the keep path both go through `set_order_session_expiry()`, which stamps every draft on the token, so the post-action re-check keeps the siblings; `expired` writes nothing and all of them time out; a settled order publishes them all and the siblings early-return on status before reaching the dedupe. No gap. `set_order_session_expiry()` narrowing to drafts is equivalent at all three call sites, and the cap matching `get_draft_release_at()` exactly is right.

Two things, one of them mine from last round.

**`stripe_unreachable` treats a Stripe answer as a failure to reach Stripe.** `send_request()` returns a `WP_Error` for every non-200 (payment-stripe.php:1119, via `handle_error()`), so a structured Stripe error arrives indistinguishable from a connection failure. A 404 `resource_missing` on a stale session id sets the flag, and every later order in the pass is kept without being asked. I have a passing test: one 404 order followed by two whose sessions Stripe reports `expired`, one lookup total, all three still `draft`.

Being fair about the blast radius: the sweep orders `post_date DESC` and `keep_for_next_sweep()` doesn't touch `post_date`, so a stale order stays old and sorts last in its page. Under a hundred due drafts it suppresses nothing, and the rotated-key case fails for every order anyway, so short-circuiting there is exactly right. The damaging shape needs a mix, some orders answering and a 404 one sorting early, which is narrow. My test forces the ordering to show the mechanism.

I'd still fix it, because the distinction it erases is the one the design rests on: "not knowing is not evidence the session died", and a 404 is knowing. The error codes already separate cleanly. Transport failures come from `wp_remote_request` as `http_request_failed`, `camptix_stripe_unexpected_response` is a non-200 nobody could parse (fairly called unreachable), and `camptix_stripe_request_error_%d` means Stripe answered. Gating the flag on the code not being a `camptix_stripe_request_error_` one keeps the behaviour you want and drops the conflation before a third error path gets added on top of it. Same cause, second order: a 404 order is re-asked every sweep for 24 hours for a question whose answer won't change, holding its seat the whole time for a session that provably doesn't exist.

**The retry cap has no floor now, and that one's my fault.** I asked for the no-timestamp case to fail closed and it does, but closed now means forever: `if ( $checked_out && $checked_out + DAY_IN_SECONDS + ... )` skips the give-up branch entirely when `tix_timestamp` is missing, so every sweep re-records the expiry as `time()`. Passing test, three passes with the draft aged ten days between each, still `draft`. The previous commit capped it at 24 hours and I traded that away asking for fail-closed; it needs both. Needs data damage to reach, since the first branch of `$due` has no `tix_timestamp > 0` guard but both metas are written together, though unbounded is a worse failure than either state it replaced. Falling back to the post date when the meta is missing gives the cap a floor without reopening the fail-open case.

Smaller:

- The logs can't tell the two keep paths apart. `keep_for_next_sweep()` writes the same line whether the lookup was attempted and failed or skipped because the flag was already set, and the skip is the one that hides how many orders a single failure affected. One extra word, or a separate line for the skip.
- `reset_timeout_sweep_state()` is tied to a hook rather than to the state's owner. Correct today, since `pre_attendee_timeout()` is only reached through `review_timeout_payments()`, which fires the start action first. But `$stripe_unreachable` on a request-lifetime singleton cleared by a different action is the kind of coupling that survives into any future caller that doesn't fire the hook. Passing a pass identifier through the action and keying the state on it would make it structural rather than a convention.
- `record_checkout_session()` and `set_order_session_expiry()` are now asymmetric: the session id goes on every attendee in any status, the expiry only on drafts. Same set at checkout, but it reads as if it were deliberate.

With the error-code gate and a floor on the cap I'm happy with this.

mcliwanow:
Both done, pushed as 44c29289.

- **Error-code gate.** The unreachable flag is set only when the error code is not a `camptix_stripe_request_error_*` one, so transport failures and unreadable responses short-circuit the pass and a Stripe answer doesn't. And a 404 is now treated as the answer it is: logged, and the draft left to time out, instead of being re-asked for 24 hours. Any other Stripe error keeps the draft as before. Test: a 404 order sorted first, then a 500, then an expired one; three lookups, 404 timed out, 500 kept, expired timed out.
- **Floor on the cap.** Checkout time is `tix_timestamp`, or `get_post_timestamp()` when that is missing, so the cap always has a floor and the no-timestamp case still fails closed. Test: no timestamp and a fresh post date, kept; no timestamp and a post date a day old, timed out. (Found while writing it: `wp_update_post()` resets a draft's date to now whatever you pass, so the tests set the date on the row directly.)
- **Logs.** `keep_for_next_sweep()` takes the reason, so the four keep paths log distinctly: could not reach Stripe, unreachable earlier this sweep so not asking, Stripe answered with an error, Stripe answered without a status.
- **Symmetry.** Both stamps go through a new `get_draft_attendee_ids_from_payment_token()` on the plugin, drafts only, not memoised.
- **Pass identifier.** Left as the hook. One caller today, and a second caller that forgets to fire the start action would also be a second caller of the whole backstop, which I'd rather notice in review than paper over with a key.

@bor0

bor0:
Full suite on `44c29289`: 1463 tests, 0 failures, 6 skipped.

Starting with a correction, because the one problem this round is mine. Last time I wrote that a 404 order was being held "for a session that provably does not exist". That was wrong, and the 404 branch built on it is the one thing I'd change.

**The 404 branch releases a seat on evidence that doesn't support it.** Stripe scopes objects to the account *and mode* of the key used, so `resource_missing` means "not on this account", not "does not exist". The realistic way a 35-minute-old Checkout Session 404s isn't that Stripe forgot it: it's that the key we're asking with is no longer the key that created it. An organiser editing the Stripe credentials, or flipping the sandbox toggle, while checkouts are in flight. In exactly that case the buyer still holds the hosted checkout URL and the session is still payable on the other account.

So this is the one `WP_Error` path that now releases a seat, and the evidence behind it is the weakest of the set. Confirmed on the branch, with the contrast:

```php
// 404: seat released
$this->assertSame( 'timeout', get_post_status( $draft ) );

// every other Stripe error: kept
foreach ( array( 500, 429, 401, 403 ) as $code ) {
    $this->assertSame( 'draft', get_post_status( $draft ), "kept on $code" );
}
```

And the failure is correlated rather than isolated: if the credentials changed, every in-flight session 404s, so the branch releases every abandoned draft on the site in one pass. A credentials change is about the worst moment to be releasing seats in bulk.

I'd put 404 back on the keep path. The cost I complained about, re-asking for 24 hours, is the cost of the rule the rest of this design keeps, and the cap already bounds it. If the repeated lookup is worth avoiding, record on the order that the answer was 404 and skip the lookup on later passes while still keeping the draft to the cap. That gets the efficiency without the release.

Everything else is closed:

- **Error-code gate.** The contrast above is the evidence: 500, 429, 401 and 403 all reach `keep_for_next_sweep()` through the `camptix_stripe_request_error_` branch, and only a transport error or an unreadable response trips `stripe_unreachable`. `handle_error()` adds a single code, so `get_error_code()` is unambiguous. The short-circuit path goes through `keep_for_next_sweep()` too, so the 24-hour cap applies during a long outage rather than holding seats past it.
- **Floor on the cap.** I checked the floor is actually stable, since a floating value would have made it no floor at all: `post_date` on a draft doesn't move on an ordinary update. `wp_update_post( array( 'ID' => $draft, 'post_excerpt' => 'touched' ) )` leaves `get_post_timestamp()` unchanged. What you hit while writing the tests is core overriding a date you pass explicitly for a draft, which is a different thing and doesn't affect the floor.
- **Logs.** Four distinct reasons, including the skip, which was the one that mattered.
- **Symmetry.** Both stamps through `get_draft_attendee_ids_from_payment_token()`, drafts only, not memoised.
- **Pass identifier.** Your reasoning is better than mine. A second caller of the backstop that skips the start action is a thing to catch in review, not to design around.

One small thing: `record_checkout_session()` now runs two queries for the same set of ids, its own and the one inside `set_order_session_expiry()`, where before it ran one and then hit the memo. Nothing next to what checkout already does, but passing the ids through would avoid it.

That 404 branch is the only thing left from me.

mcliwanow:
Good catch, and I should have caught it: the 404 branch was the one `WP_Error` that released a seat, on the weakest evidence of the set. Back on the keep path, pushed as ae0c85bb. I didn't add the "remember the 404 and skip the lookup" variant. One lookup per order per sweep for at most 24 hours is cheap, and it keeps one rule for every error instead of two.

The test now expects the 404 order kept alongside the 500, with the expired one still timed out in the same pass.

`record_checkout_session()` stamps both metas in one loop over one id lookup, so it's back to a single query.

bor0 (approved):
Approving. Six rounds, every finding closed and re-checked against `ae0c85bb` as a regression sweep; full suite 1463 tests green, phpcs clean on all changed lines.

The three residual behaviours are known and deliberate rather than missed: the late-confirmation double-seat bounded by gateway delivery latency, the fix pausing itself while Stripe is unreachable, and PayPal keeping its 24 hours. They belong in the deploy note.

Detail in the comment thread above.

PR #2071, opened 2026-09-14
Groups: repair the RSVP button and featured image in GatherPress's event email

### LINKED ISSUE #2054: Weird styling in RSVP email
Reported by @supernovia 

> Just saw a first “Real” email!  The RSVP button is overlapping in this mail client (Help Scout) (threading screenshot)
> 
> It’s this one:
events.wordpress.org/group/gatherpress/event/a-case-for-community-open-source-events-with-gatherpress

Image

also

> I just noticed once more email issue. On mobile, the image is distorted.

Image

bor0:
## Summary

Fixes #2054. Both faults in the reported email are in GatherPress's `includes/templates/admin/emails/event-email.php`, which this network sends for automatic first-publish notifications and for an organizer's "Message members".

**The RSVP button overlaps what follows it.** It's an `<a>` with `padding: 12px 20px` and no `display`. Padding on an inline box doesn't push the next block out of the way: measured in Chrome, the wrapping `<div>` is **18px** tall while the button paints **41px**, so the blue background prints over the excerpt beneath it — exactly the reporter's Help Scout screenshot. `display: inline-block` makes the padding count, and the wrapper then measures the button's real **42px**. GatherPress's own RSVP-token email already does this to the same kind of button; only this template was missed.

**The featured image stretches on a phone.** It carries `max-width: 100%` plus the `width`/`height` attributes of the full-size file. Narrow the viewport and the width obeys the cap while the height stays where the attribute put it. At a 390px viewport a 1600×900 image rendered **374×900** — a ratio of 0.42 against the source's 1.78. `height: auto` restores it to **374×210**, ratio 1.78.

## Why this is a `wp_mail` filter and not a template edit

GatherPress is gitignored here — installed from a release zip, not vendored — so the template isn't ours to change. It's also rendered from a hardcoded path through `Utility::render_template()`, which bypasses the theme-override chain GatherPress uses for its other templates, and `Rest_Api::send_emails()` has no filter on the body. So the only seam is `wp_mail`, scoped to this template by an HTML comment only it carries.

The edits themselves go through `WP_HTML_Tag_Processor` rather than string replacement, and each is guarded against re-applying, so a body that passes through twice isn't rewritten twice.

**This belongs upstream**, and the tests say so out loud: two canary tests assert the shipped template *still has* the faults being repaired. When GatherPress fixes this, they fail — and the right response is to delete `gatherpress-event-email.php` and its tests, not to update them. I have not opened the GatherPress PR; happy to, if you want it.

## Test plan

- [x] phpunit — `WordCamp Groups` 400/400
- [x] phpcs clean on all three files
- [x] Rendered the real email from the real template for a real event and diffed before/after: exactly two attribute changes, nothing else touched
- [x] Measured both faults in a browser, before and after (numbers above)

Six tests: the two repairs against a genuinely rendered email, idempotency, pass-through for every other email this network sends, and the two canaries.

`tests/bootstrap.php` gains the new file, alongside the other `groups/` mu-plugins it already loads.

### Manual test steps

1. Rendered `event-email.php` through `Utility::render_template()` for a published event with a featured image, captured the body, then ran it through the `wp_mail` filter.
2. **Diff of the two bodies** — one `style` gains `height: auto;` on the ``, one gains `display: inline-block;` on the `<a>`. No other line differs.
3. **Button, at 900px.** Before: wrapper 18px, button 41px, paragraph 5px below the button's painted bottom. After: wrapper 42px, gap 16px.
4. **Image, at 390px.** Before: 374×900 (ratio 0.42). After: 374×210 (ratio 1.78, matching the 1600×900 source).
5. Screenshots of both bodies at 900px and 390px, attached to the conversation rather than here — the local placeholder image is a near-white gradient, so the measurements above are the substantive evidence.

Not exercised: delivery through an actual mail client. The overlap is a layout consequence of an inline box, reproduced and measured in a browser engine; I can't install Help Scout to confirm the fix there.


bor0:
Closing in favour of fixing this upstream, where the template actually lives.

Checking GatherPress trunk changed the picture:

* **Featured image** is already fixed in `develop` by GatherPress/gatherpress#2287, which bounds the image and adds `height: auto`. It has not shipped yet: 0.35.3 is the current release and still has the fault, which is why the report reproduced.
* **RSVP button** is still broken in trunk. Filed as GatherPress/gatherpress#2297 and fixed in GatherPress/gatherpress#2298, with a test that renders the template and asserts the button carries both its padding and a `display`.

So the local `wp_mail` repair this PR proposed would have patched one fault that upstream had already solved and one it hadn't, and would have needed deleting on the next version bump anyway. Not worth carrying.

**events.wordpress.org keeps both faults until GatherPress cuts a release containing #2287 and #2298 and the pinned version here is bumped.** Leaving #2054 open to track that; it isn't fixed by anything merged here.

PR #2083, opened 2026-09-15
Events: apply the event password gate to RSVP roster reads

bor0:
Makes event data visibility follow the event's own password gate across the
occurrence-aware REST routes, the group event REST routes, and the event
blocks. Also scopes the speakers meta auth callback to the target post.

## Testing

Verified locally across password-protected, unlocked, and unprotected events,
through the REST routes and the rendered event page, logged out and as a
member and an organiser, with and without a `preview` query var.

- `phpunit -c phpunit.xml.dist`: 1471 pass, 6 pre-existing skips.
- `phpcs`: no errors on the changed files.

New regression tests cover the gated and ungated paths in each component, and
each was confirmed to fail without its fix.


mcliwanow (approved):
Changes look good. Let's 🚢  it

